PR #140, opened 2022-08-25
fix(network): align hex entity id formatting

alvrs:
Previously we were padding entity ids to a 256 bit string by default to align padded and non-padded
entity id strings. This caused issues in components referencing other entity ids and caused
unnecessary memory usage  for storing 0s. This change aligns entity ids by removing all padding by
default.


Kooshaba (approved):
wasn't able to get it running with mudwar, but in theory i agree

PR #141, opened 2022-08-25
feat(network): add loading state component update stream to SyncWorker

alvrs:
Add a loading state component with id keccak("component.LoadingState") to the sync worker that is
used to broadcast the current loading state to the main thread. This can be used to build a loading
screen in the client by adding a new network layer component with "contractId"
kecccak("component.LoadingState").

BREAKING CHANGE: The loading state component is attached to the entity with id 0x060D (GodID). The
std-client package previously exported a different mudwar specific GodID, which has been replaced
with the 0x060D GodID exported by the network package.


Kooshaba:
tested and working 👍🏼

Kooshaba (approved):

PR #142, opened 2022-08-30
feat(cli): forge bulk upload ecs state script

Kooshaba:
- add `mud bulkupload` cli script


packages/std-client/src/utils.ts
```diff
@@ -56,9 +56,9 @@ export function getGameConfig(
   gameConfigComponent: Component<{ startTime: Type.String; turnLength: Type.String; actionCooldownLength: Type.String }>
 ) {
   const godEntityIndex = world.entityToIndex.get(GodID);
-  if (!godEntityIndex) return null;
+  if (!godEntityIndex) return;
```
alvrs, line 59:
```suggestion
  if (godEntityIndex == null) return;
```
otherwise this fails if god happens to be at index 0

packages/network/src/workers/constants.ts
```diff
@@ -6,4 +6,4 @@ export enum SyncState {
   LIVE,
 }
 
-export const GodID = "0x060D" as EntityID;
+export const GodID = "0x060d" as EntityID;
```
alvrs, line 9: why this change?

alvrs (approved):
lgtm

packages/network/src/workers/constants.ts
Kooshaba: the ID comes back from on-chain lowercase

alvrs (approved):

PR #145, opened 2022-09-14
feat: matching Perlin noise implementations in Solidity and AssemblyScript

alvrs:
- add Perlin noise implementations in Solidity and AssemblyScript.
- add tests to make sure Solidity and AssemblyScript implementations compute the same result and it's proper Perlin noise.


ludns (approved):
incredible code and attention to details. Mr Perlin would be happy.

PR #148, opened 2022-09-15
build: compile and use dist/index.js as entry point when publishing packages

alvrs:
- Previously we were releasing .ts source files in packages on npm. This caused some problems with bundling clients for production.
- This PR adds scripts and configuration to build packages and change the entry point to the built code (dist/index.js) when publishing to npm
- For local development, we're still using src/index.ts as entry point (so changes are propagated to the clients without rebuilding when using the locally linked version of mud in development)


ludns (approved):
lgtm

PR #149, opened 2022-09-15
ci: use forge action instead of manual forge install

alvrs:
- replace manual forge install with forge action
- install ts-node as dev dependency of solecs to fix building docs


ludns (approved):
lgtm

PR #152, opened 2022-09-20
fix: phaser input click$ stream

alvrs:
- Due to a race condition between the `mousedown`/`mouseup` events and Phaser's `pointer.leftButtonDown`, the `click$` stream would not fire on every click. This PR changes the internal logic to use to `mousedown`/`mouseup` as source of truth for the click$ stream.


Kooshaba (approved):
tested and works for me 👍🏼

PR #154, opened 2022-09-20
feat(std-client): add setupContracts

alvrs:
- move `setupContracts` out of individual client codebases and into shared std-client library


Kooshaba (approved):
👍🏼

Kooshaba (approved):
🌡️

PR #156, opened 2022-09-20
fix(network): initial sync

alvrs:
- fix a bug during initial sync: previously we were loading three states during the initial sync (cache state, gap state, live state) and then merged them to get the full initial state. This approach led to an issue when one of the states included a **ComponentValueRemoved** event targeting a value in another state (eg cache state has a component value, gap state applies an event to remove this value, but because the gap state doesn't have the value, the removal event is not actually applied when merging the states). To prevent this issue, this PR changes the initial sync logic to load **events** of the "gap state" and "live state" and apply them to the cache state, such that removal events are considered.

- store cache to indexdb on every block instead of every time interval to avoid storing it before all events of the block have been applied to the local state

### DIFF
--- a/packages/network/src/workers/CacheStore.ts
+++ b/packages/network/src/workers/CacheStore.ts
@@ -58,6 +58,15 @@ export function storeEvent<Cm extends Components>(
   cacheStore.blockNumber = blockNumber - 1;
 }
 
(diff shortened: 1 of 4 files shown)


ludns (approved):
😴 👨‍💻

PR #157, opened 2022-09-22
feat: initial implementation of ecs relay service

authcall:
v1 of ECS relay service for off-chain data relay. Clients can subscribe / unsubscribe to labels and open streams to receive relayed data


alvrs (approved):
great work!! tested and works, integration PRs following

PR #158, opened 2022-09-24
feat(recs): add support for custom type in component

alvrs:
- This PR adds support for custom types in components that don't fit the existing `Types` definitions. This custom type has full typescript support, but on the downside can't be read in javascript, so for example automatic decoders/encoders couldn't be inferred from it.
- First use case of this is the `Action` component defined in the `ActionSystem`, which now allows consumers to define a custom `metadata` type to add more information to actions that might be relevant for rendering.


ludns (approved):
type 🥷

PR #162, opened 2022-09-26
feat(network): add system call stream

Kooshaba:
This PR adds the concept of a `System Call` NetworkEvent to the network package. When configuring the SyncWorker, you can pass in a flag that will add System Calls to the NetworkComponentUpdate events that are normally output.

A System Call is a record of a transaction that results in the change of some Components. The System call output contains the tx hash, the updates that resulted from the tx, and the arguments the System was called with.

This was added to support situations like the following:

Two units are engaging in Combat. When Combat resolves, one has lower health and the other was destroyed.

Normally, we would receive a series of Component updates. Both of their Health went down, and the unit that died had its Position removed (effectively removing it from the game). This is fine until you want to animate this interaction in order to provide context to the user. Why did the Health go down? Why was a unit's Position removed? We have no idea...

Using System Calls, you can subscribe to the specific System that you know triggers these changes. Then you can react to these changes in context specific ways. We now know A) the two combatants from the arguments to the system and B) the changes that resulted from this specific Combat. Therefore we can animate this interaction so the two units swing their swords at each and their health goes down at the proper moment.


packages/network/src/createSyncWorker.ts
```diff
@@ -14,13 +12,13 @@ import { Output } from "./workers/SyncWorker";
...
-  const config$ = new Subject<SyncWorkerConfig<Cm>>();
+export function createSyncWorker() {
+  const config$ = new Subject<SyncWorkerConfig>();
   const worker = new Worker(new URL("./workers/Sync.worker.ts", import.meta.url), { type: "module" });
-  const ecsEvent$ = new Subject<NetworkComponentUpdate<Cm>>();
+  const ecsEvent$ = new Subject<NetworkEvent>();
 
   // Pass in a "config stream", receive a stream of ECS events
-  const subscription = fromWorker<SyncWorkerConfig<Cm>, Output<Cm>>(worker, config$).subscribe(ecsEvent$);
+  const subscription = fromWorker<SyncWorkerConfig, NetworkEvent>(worker, config$).subscribe(ecsEvent$);
```
alvrs, line 21: I see `NetworkEvent` accepts a `Components` generic type - could we keep the generic typing for `createSyncWorker` and pass it to `NetworkEvent`? Makes working with `ecsEvent$` slightly nicer

packages/network/src/workers/SyncWorker.ts
```diff
@@ -108,19 +116,26 @@ export class SyncWorker<Cm extends Components> implements DoWork<SyncWorkerConfi
...
+        // Store cache to indexdb every block
+        if (event.blockNumber > cacheStore.current.blockNumber + 1)
+          saveCacheStoreToIndexDb(indexDbCache, cacheStore.current);
+
+        storeEvent(cacheStore.current, event);
+      }
 
-      storeEvent(cacheStore.current, event);
-      this.output$.next(event as Output<Cm>);
+      this.output$.next(event as NetworkEvent<C>);
```
alvrs, line 138: Do we want to pass system calls to the main thread before the initial sync is done?

packages/network/src/workers/syncUtils.ts
```diff
@@ -356,16 +370,17 @@ export function createFetchWorldEventsInBlockRange(
...
         value: undefined,
         blockNumber,
         lastEventInTx,
         txHash,
-      };
+      } as NetworkComponentUpdate<C>;
 
       if (event.eventKey === "ComponentValueRemoved") {
-        ecsEvents.push(ecsEvent);
+        ecsEvents.push({ ...ecsEvent });
```
alvrs, line 383: why the spread here?

packages/network/src/workers/syncUtils.ts
```diff
@@ -419,3 +435,67 @@ export function createTransformWorldEventsFromStream(decode: ReturnType<typeof c
...
+}
+
+function createBlockCache(provider: JsonRpcProvider) {
+  const blocks: Record<number, Awaited<ReturnType<typeof provider.getBlockWithTransactions>>> = {};
+
+  return {
+    fetchBlock: async (blockNumber: number) => {
+      if (blocks[blockNumber]) return blocks[blockNumber];
+
+      return await provider.getBlockWithTransactions(blockNumber);
```
alvrs, line 497: where do we set the entry in the block cache?

alvrs (dismissed):
Looks great! Really clean and even cleaned up existing code 🔥 Just some minor nits, then ready to merge

packages/network/src/workers/syncUtils.ts
Kooshaba: ah, left over from another change

packages/network/src/workers/syncUtils.ts
Kooshaba: lool never does

alvrs (approved):

PR #163, opened 2022-09-27
feat: basic faucet service implementation

authcall:
v1 of ECS faucet that can be used to drip funds and send any desired ECS tx to the recipient


alvrs (approved):

alvrs (approved):
LFG

alvrs (approved):

PR #165, opened 2022-09-28
feat: increase network performance by reducing unnecessary rpc calls

alvrs:
- send raw transactions instead of via ethers method to avoid unnecessary rpc roundtrips for data we don't need
- process transaction queue concurrently


packages/std-client/src/setup/setupMUDNetwork.ts
```diff
@@ -79,10 +80,17 @@ export async function setupMUDNetwork<C extends ContractComponents, SystemTypes
     signerOrProvider,
   });
 
-  const { txQueue, dispose: disposeTxQueue } = createTxQueue(contracts, network, { devMode: networkConfig.devMode });
+  const gasPriceInput$ = new BehaviorSubject<number>(
+    // If no initial gas price is provided, check the gas price once and add a 30% buffer
+    options?.initialGasPrice || (await signerOrProvider.get().getGasPrice()).toNumber() * 1.3
```
ludns, line 85: probably want to make sure this is not a float; it might make the nodes freak out.

ludns (approved):
lgtm

PR #166, opened 2022-09-29
fix: check event type instead of just value before decoding

authcall:
Fixes the issue where events of type `ComponentValueRemoved` were not correctly decoded when coming from the stream service


alvrs (approved):
Good catch!

PR #167, opened 2022-09-29
feat: add signature verification for all client actions via relay service

authcall:
v2 of relay service with breaking change that requires clients to provide a signature for all actions verifying ownership of address which is their identity on the relay service


ludns:
great PR!
one note re: signature
clients need to sign every message + timestamp; otherwise i can pretend to be you and replay whatever messages i want.
clients and relay will only accept messages if the sig of the hash(message || timestamp) is correct, and if the timestamp is within ~5s of current time

ludns:
+ can we relay sig to client? identity can be recovered from the sig; which means a client can now:
- verify that a message was not spoofed (signed message)
- verify that a message was not replayed (signed timestamp)
- recover the identity of the sender (ecrecover on the sig). right now @alvrs had to encode the message sender's address in the message itself.

ludns (commented):
comment for @alvrs: ECDSA recover is very expensive and here we do it in ethers.js TS native code lol. there might be a lot of messages per second and it could tank our perf.
my proposal is to pipe those to a Worker that does the `verifyMessage`; and then pipe those back to the ECS components on the main thread. we'd have to benchmark, but i can suspect having ~100s of msg per second will start tanking performance.

alvrs:
note: force pushed to rebase on latest main

alvrs:
Did some simple benchmarks and address recovery take around 2ms in the main thread vs around 4ms when using a webworker (due to overhead overhead of data transfer to and from the webworker). But since doing it in a webworker is async and non-blocking, the worst thing that can happen is a backlog of messages to recover (vs a lagging client when doing in the main thread)

packages/services/cmd/ecs-relay/main.go
```diff
@@ -12,6 +12,7 @@ var (
 	port                  = flag.Int("port", 50071, "gRPC Server Port")
 	idleTimeoutTime       = flag.Int("idle-timeout-time", 30, "Time in seconds after which a client connection times out. Defaults to 30s")
 	idleDisconnectIterval = flag.Int("idle-disconnect-interval", 60, "Time in seconds for how oftern to disconnect idle clients. Defaults to 60s")
+	messsageDriftTime     = flag.Int("message-drift-time", 60, "Time in seconds that is acceptable as drift before message is not relayed. Defaults to 5s")
```
ludns, line 15:
```suggestion
	messsageDriftTime     = flag.Int("message-drift-time", 5, "Time in seconds that is acceptable as drift before message is not relayed. Defaults to 5s")
```

ludns (changes requested):

ludns (approved):

alvrs (approved):

PR #172, opened 2022-10-01
feat: add a stream rpc for message push

authcall:
- Open a stream and sent continuous message push requests
- No longer need to ping since every valid relayed message will result in a call to `client.Ping()` thus refreshing the activity timestamp


alvrs:
closing in favor of #172

PR #173, opened 2022-10-03
feat: faucet re-drip after tweet, rpc to get time until next drip

authcall:
- Request drip after verifying tweet without tweeting again on separate endpoint
- Get time until next drip


alvrs (approved):
lgtm

PR #174, opened 2022-10-03
feat: add a stream rpc for message push

alvrs:
- Open a stream and sent continuous message push requests
- No longer need to ping since every valid relayed message will result in a call to client.Ping() thus refreshing the activity timestamp
- Rename checkpoint to snapshot in network packages
- Rename relayer to relay in network packages


authcall (approved):

ludns (approved):
lgtm 🐑 
+1 on diminishing the amount of data over the wire for off-chain tx!

PR #175, opened 2022-10-04
feat: flag for sig verification, rate limiting of push, and check for balance when relaying

authcall:
- Toggle to switch on / off service-side signature verification before relay
- Rate limiting of `/Push` msg relay endpoint per client
- Only allow authenticated clients with balance > 0 to relay via `/Push`

### DIFF
--- a/packages/services/pkg/grpc/relay.go
+++ b/packages/services/pkg/grpc/relay.go
@@ -4,12 +4,14 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"latticexyz/mud/packages/services/pkg/eth"
 	"latticexyz/mud/packages/services/pkg/relay"
 	"latticexyz/mud/packages/services/pkg/utils"
 	pb "latticexyz/mud/packages/services/protobuf/go/ecs-relay"
 	"time"
 
 	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/ethereum/go-ethereum/ethclient"
 	"go.uber.org/zap"
 )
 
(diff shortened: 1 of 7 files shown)


packages/services/pkg/grpc/relay.go
```diff
@@ -289,13 +295,31 @@ func (server *ecsRelayServer) HandlePushRequest(request *pb.PushRequest) error {
...
+	// Get the client object for this identity to make sure it's authenticated.
 	client, err := server.GetClientFromIdentity(identity)
 	if err != nil {
 		return err
 	}
 
-	// Verify that the message is OK to relay.
+	// Check if the authenticated client has a balance. We permit pushes of messages only for
+	// clients which have a non-zero balance.
+	balance, err := eth.GetCurrentBalance(server.ethClient, recoveredAddress)
```
ludns, line 306: little nits from me: if we get 10 messages per second per client with sufficient balance we probably need to cache this and only refresh it every so often (if no balance, maybe refresh every 5 second; and if balance refresh it every 60s? so we have a maximum of 5s of messages dropped if you just received a drip on a new account; and a minute of "free" messages if you use to have fund)

alvrs: agree, otherwise we spam RPC again

alvrs (approved):

packages/services/pkg/grpc/relay.go
authcall: addressed in commit 1d476d5440a722e930eb02c47ddd475aa6487161

PR #182, opened 2022-10-05
expose component functions Set and Remove internally without onlyWriter permissions

chlwys:
Component functions Set and Remove require onlyWriter permissions, which prevent components changing their own values when called by non-systems. Although components should only interact with systems, this limits flexibility when creating 'hybrid components', like ERC721-compatible components.

The only change is adding _set and _remove methods. These are internal and will allow child components to hold some external logic if needed but does not affect anything else

### DIFF
--- a/packages/solecs/src/Component.sol
+++ b/packages/solecs/src/Component.sol
@@ -109,24 +109,7 @@ abstract contract Component is IComponent {
    * @param value Value to set for the given entity.
    */
   function set(uint256 entity, bytes memory value) public onlyWriter {
-    // Store the entity
-    entities.add(entity);
-
-    // Remove the entitiy from the previous reverse mapping if there is one
-    valueToEntities.remove(uint256(keccak256(entityToValue[entity])), entity);
-
-    // Store the entity's value;
-    entityToValue[entity] = value;
-
-    // Add the entity to the new reverse mapping
-    valueToEntities.add(uint256(keccak256(value)), entity);
-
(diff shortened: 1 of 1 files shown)


packages/solecs/src/Component.sol
```diff
@@ -196,4 +162,59 @@ abstract contract Component is IComponent {
   function registerIndexer(address indexer) external onlyWriter {
     indexers.push(IEntityIndexer(indexer));
   }
+
+  /**
+   * Set the given component value for the given entity.
+   * Registers the update in the World contract.
+   * Can only be called by addresses with write access to this component.
```
alvrs, line 169:
```suggestion
   * Can only be called internally (by the component or contracts deriving from it),
   * without requiring explicit write access.
```

packages/solecs/src/Component.sol
```diff
@@ -196,4 +162,59 @@ abstract contract Component is IComponent {
...
+    }
+
+    // Emit global event
+    IWorld(world).registerComponentValueSet(address(this), entity, value);
+  }
+
+  /**
+   * Remove the given entity from this component.
+   * Registers the update in the World contract.
+   * Can only be called by addresses with write access to this component.
```
alvrs, line 197:
```suggestion
   * Can only be called internally (by the component or contracts deriving from it),
   * without requiring explicit write access.
```

alvrs (changes requested):

alvrs (approved):
Thank you!

alvrs (approved):

PR #191, opened 2022-10-10
chore(relay): fix typo

therealbytes:


authcall (commented):
could we consolidate #191 (this), #192 and #193 and push together?

PR #195, opened 2022-10-10
feat(relay): check message data doesn't exceed a configurable limit

therealbytes:
Right now, the data field of a relayed message can have an arbitrary size so it's possible for an attacker to send a giant message. This PR adds a field to the relay server configuration specifying a size limit for message data.


therealbytes:
Should I also add a flag to the configuration to set wether the size limit is enforced, like we do for signature verification and account balance? @authcall

authcall:
> Should I also add a flag to the configuration to set wether the size limit is enforced, like we do for signature verification and account balance? @authcall

yes that sounds good. I think something like `VerifyMessageSize` seems reasonable so that we can use the same kind of check like `if server.config.VerifyMessageSize {`

authcall (approved):

authcall (dismissed):

authcall (approved):

PR #197, opened 2022-10-11
therealbytes/signature-expiration

therealbytes:
Right now, users verify their address in the relay service by signing a static message. This is vulnerable to replay-attacks.

This PR adds an optional ExpirationTime field to the address-verification signature. If provided, the signed message is assumed to be `ecs-relay-service: $ExpirationTime` instead of `ecs-relay-service`.
For the signature to be valid ExpirationTime must be greater than or equal to the current unix time.


packages/network/src/createRelayStream.ts
```diff
@@ -22,7 +22,12 @@ export async function createRelayStream(signer: Signer, url: string, id: string)
   );
 
   // Signature that should be used to prove identity
-  const signature = { signature: await signer.signMessage("ecs-relay-service") };
+  // Set the signature to expire in 5000 ms
+  const signatureExpirationTime: number = Date.now() + 5000;
+  const signature = {
+    signature: await signer.signMessage(`ecs-relay-service: ${signatureExpirationTime}`),
+    expirationTime: signatureExpirationTime,
+  };
```
alvrs, line 30: we'll need to continuously create a new signature somewhere

alvrs, line 30: Right now this only creates a single signature that expires after 5 seconds. We'll need to continuously create new signatures somewhere.

alvrs (changes requested):

authcall:
closing this draft PR for now and will reopen with some comments / re-architecture suggestions once ready

PR #199, opened 2022-10-11
fix(solecs): only allow components to register their own updates, feat(std-client): add support for multiple overrides per component per action

alvrs:
- fix(solecs): only allow components to register their own updates (before it was possible for anyone to call `registerComponentValueSet`)
- feat(std-client): add support for multiple overrides per component per action


ludns (approved):
lgtm

PR #208, opened 2022-10-18
fix: lowercase address for relay

ludns:


alvrs:
for reference: discussed offline and decided to leave checksum active for recovered address from the recover worker (and lowercase it in the consumer instead)

PR #212, opened 2022-10-19
feat: allow specific snapshot chunk ratio

authcall:
Allows client to specify what ratio to use per chunk when chunking a snapshot and sending via a stream


alvrs (dismissed):
lgtm

alvrs (approved):

PR #213, opened 2022-10-20
fix(network): disable browser cache in dev mode

Kooshaba:


packages/network/src/workers/SyncWorker.ts
```diff
@@ -165,7 +166,7 @@ export class SyncWorker<C extends Components> implements DoWork<SyncWorkerConfig
 
     // Load initial state from cache or snapshot service
     this.setLoadingState({ state: SyncState.INITIAL, msg: "Fetching cache block number", percentage: 0 });
-    const cacheBlockNumber = await getIndexDBCacheStoreBlockNumber(indexDbCache);
+    const cacheBlockNumber = !disableCache ? await getIndexDBCacheStoreBlockNumber(indexDbCache) : 0;
```
alvrs, line 169: With this change we still load from cache if no snapshot is available (see https://github.com/latticexyz/mud/pull/213/files#diff-a357062aac9c6239900fad7b7194dbf10291661dd2393b12402bbce0b77d4351R186-R202). Need to prevent the actual cache fetching in line 196 (https://github.com/latticexyz/mud/pull/213/files#diff-a357062aac9c6239900fad7b7194dbf10291661dd2393b12402bbce0b77d4351R197)

Kooshaba: @alvrs just tested locally and this worked. i believe because this line catches it https://github.com/latticexyz/mud/blob/4261e519da871372591136c1ed84d784382eff3e/packages/network/src/workers/SyncWorker.ts#L208

Also looks like `snapshotBlockNumber` is doing the same thing (setting the block number to -1) when snapshot isn't available. am i missing something?

alvrs: I think you're right!

alvrs (approved):

PR #229, opened 2022-11-01
feat(solecs): better inheritance for Component and BareComponent

dk1a:
Allow overriding Component logic (except access control)
And add some code reuse internally so Component inherits BareComponent


alvrs (approved):
Great improvement, thank you!

PR #232, opened 2022-11-01
feat(network,std-client): add support for SystemCall events in default MUD network setup

Kooshaba:
## TODO

Add support when using the Streaming service. Currently only supports events over RPC.

Relevant RFC for Streaming service support: https://www.notion.so/latticexyz/System-Call-Events-in-Stream-Service-b04601c721c644859e890e2cee984c17

### DIFF
--- a/packages/std-client/src/setup/setupMUDNetwork.ts
+++ b/packages/std-client/src/setup/setupMUDNetwork.ts
@@ -13,13 +13,18 @@ import {
   ack,
   Ack,
   InputType,
+  NetworkComponentUpdate,
+  SystemCall,
+  isSystemCallEvent,
 } from "@latticexyz/network";
 import { BehaviorSubject, concatMap, filter, from, map, Observable, Subject, timer } from "rxjs";
 import {
   Component,
   Components,
+  EntityID,
   EntityIndex,
   getComponentEntities,
+  getComponentValue,
   getComponentValueStrict,
   removeComponent,
   Schema,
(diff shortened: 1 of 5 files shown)


packages/std-client/src/setup/setupMUDNetwork.ts
```diff
@@ -194,7 +225,8 @@ function applyNetworkUpdates<C extends Components>(
   components: C,
   ecsEvents$: Observable<NetworkEvent<C>[]>,
   mappings: Mappings<C>,
-  ack$: Subject<Ack>
+  ack$: Subject<Ack>,
+  decodeAndEmitSystemCall: (event: SystemCall<C>) => void
```
alvrs, line 229: let's make this optional for folks without the system call setup

alvrs, line 229:
```suggestion
  decodeAndEmitSystemCall?: (event: SystemCall<C>) => void
```

packages/std-client/src/setup/setupMUDNetwork.ts
```diff
@@ -210,21 +242,24 @@ function applyNetworkUpdates<C extends Components>(
...
+          continue;
+        }
+
+        if (update.value === undefined) {
+          // undefined value means component removed
+          removeComponent(components[componentKey] as Component<Schema>, entityIndex);
+        } else {
+          setComponent(components[componentKey] as Component<Schema>, entityIndex, update.value);
+        }
+      } else if (isSystemCallEvent(update)) {
```
alvrs, line 261:
```suggestion
      } else if (isSystemCallEvent && isSystemCallEvent(update)) {
```

alvrs (approved):

PR #234, opened 2022-11-02
refactor(ecs-browser): removing phaserx deps from ecs-browser

### LINKED ISSUE #231: Remove phaserx dependency from ecs-browser

mirshko:
Removing the shared files from phaserx with locally shared functions inside of the ecs-browser
package inside of shared.ts

fix #231


alvrs (approved):
thank you!!

PR #235, opened 2022-11-02
refactor(ecs-browser): make dev components optional, remove unnecessary dependencies 

mirshko:
- refactor(ecs-browser): removing phaserx deps from ecs-browser
- refactor(entityeditor): remove dep of react-collapse
- refactor(ecs-browser): makes setting of ecs component values optional and cleans up the props


mirshko:
#234 needs to be merged before this.

packages/ecs-browser/src/ComponentEditor.tsx
```diff
@@ -24,11 +24,17 @@ export const ComponentEditor = ({
     <ComponentEditorContainer>
       <ComponentTitle>
         {component.id}
-        <ComponentBrowserButton onClick={() => {
-          removeComponent(component, entity);
+        {setContractComponentValue && (
+          <ComponentBrowserButton
+            onClick={() => {
+              removeComponent(component, entity);
```
alvrs, line 30: Removing local components should be possible even without `setContractComponentValue` function
```suggestion
          <ComponentBrowserButton
            onClick={() => {
              removeComponent(component, entity);
```

packages/ecs-browser/src/ComponentEditor.tsx
```diff
@@ -24,11 +24,17 @@ export const ComponentEditor = ({
...
 
-          if(hasContract(component)) setContractComponentValue(entity, component, {});
-        }}>Remove</ComponentBrowserButton>
+              if (hasContract(component)) setContractComponentValue(entity, component, {});
+            }}
+          >
+            Remove
+          </ComponentBrowserButton>
+        )}
```
alvrs, line 37:
```suggestion
              if (setContractComponentValue && hasContract(component)) setContractComponentValue(entity, component, {});
            }}
          >
            Remove
          </ComponentBrowserButton>
```

packages/ecs-browser/src/ValueEditor.tsx
```diff
@@ -51,6 +51,11 @@ export const ValueEditor = ({
   const persistValue = useCallback(
     (e: Event | React.SyntheticEvent, value: string | null) => {
       e.preventDefault();
+
+      if (!setContractComponentValue) {
+        return;
+      }
+
```
alvrs, line 58: It should still be possible to edit local component values without `setContractComponentValue` function

alvrs (changes requested):
Looks great overall! Only feedback: it should still be possible to edit/remove values from local components if the `setContractComponentValue` function is not provided. Also, since this PR changes touches a bunch of files in `ecs-browser` I'll wait with merging until @kooshaba had a chance to look at it on Monday.

packages/ecs-browser/src/ValueEditor.tsx
mirshko: Won't this purely be client side then? I guess thats fine as well tho!

alvrs: Yes, it's for quickly editing client-only components in dev

packages/ecs-browser/src/Browser.tsx
```diff
@@ -14,26 +14,29 @@ import { Coord } from "./shared";
...
     prototypeComponent?: Component<{ value: Type.StringArray }>;
     nameComponent?: Component<{ value: Type.String }>;
     spawnPrototypeAt?: (prototypeId: EntityID, position: Coord) => void;
     world: World;
   }) => {
+    const devHighlightComponent = defineComponent(world, {
+      value: Type.OptionalNumber,
+    });
+
+    const hoverHighlightComponent = defineComponent(world, {
```
Kooshaba, line 35: I agree that it is cleaner to construct the dev only components inside of the browser itself, but right now we are losing the ability to react (easily) to these component changes during rendering.

In Sky Strife currently the HoverHighlight component is being set as the user's mouse is moving and then that information gets passed into here. I could get the same functionality by grabbing the created Component from this package and hooking it into the existing systems.

I see 2 options:
1. We create the Components outside of the Browser using a helper function (`createBrowserDevComponents` or something) and then pass them in. This is at least easier to adhere to the interface the Browser expects.
2. Create `getDevComponents` helper that gives the game layers access to the Components.

Right now to make it work we would have to grab the components from the World and manually type them. It feels strange because users would have to know these magic components got created and their typings.

Kooshaba (changes requested):
Overall looks good! Left one request though

packages/ecs-browser/src/Browser.tsx
mirshko: Ahh word!

I'll build out option 1, 

I think thats the best option, I didn't think about them being used outside of the Browser component. Good call!

packages/ecs-browser/src/ValueEditor.tsx
Kooshaba: I agree we should remove this to support client-only editing.

alvrs (approved):
lgtm!

PR #239, opened 2022-11-04
fix: allow overriding preset components, replace Owned interface with IERC173, fix IComponent interface

dk1a:
Allow overriding component presets in std-contracts

Standardize IComponent and ISystem under IERC173
(Components are effectively unchanged, Systems now get transferOwnership, and transferOwnership also emits an event)
Move getSchema and registerIndexer into IComponent, fix some comments


packages/std-contracts/src/components/FunctionBareComponent.sol
```diff
@@ -3,7 +3,7 @@ pragma solidity >=0.8.0;
 import "solecs/BareComponent.sol";
 
 struct FunctionSelector {
-  address contr;
+  address target;
```
alvrs, line 6: what's the reason for this change?

alvrs (commented):
Looks good! My only concern are the changes in `FunctionComponent` and `FunctionBareComponent` as they are breaking changes for existing consumers, so i'd like to get a better understanding for why those changes were made

dk1a:
Yeah that shouldn't be there, removed the struct changes

PR #242, opened 2022-11-08
fix(phaserx): scale zoom delta with pinch speed

holic:
`pinchSpeed` wasn't implemented before, now it is!


alvrs (approved):
much better!

alvrs (dismissed):

packages/phaserx/src/createCamera.ts
```diff
@@ -53,6 +59,7 @@ export function createCamera(phaserCamera: Phaser.Cameras.Scene2D.Camera, option
   const wheelSub = wheelStream$
     .pipe(
       sampleTime(10),
+      filter((state) => !state.pinching),
```
holic, line 62: Thinking this should probably before the sample call so we don't drop events? Though it won't be too noticeable since we're sampling at 10ms.

alvrs (approved):

holic:
lgtm!

PR #244, opened 2022-11-11
fix(solecs): restrict write access to Set and MapSet to owner

alvrs:
Fixes a critical security bug in solecs (restricting write access to internal MapSet and Set contracts to the owner of those contracts)

### DIFF
--- a/packages/solecs/src/MapSet.sol
+++ b/packages/solecs/src/MapSet.sol
@@ -5,17 +5,27 @@ pragma solidity >=0.8.0;
  * Key value store with uint256 key and uint256 Set value
  */
 contract MapSet {
+  address private owner;
   mapping(uint256 => uint256[]) private items;
   mapping(uint256 => mapping(uint256 => uint256)) private itemToIndex;
 
-  function add(uint256 setKey, uint256 item) public {
+  constructor() {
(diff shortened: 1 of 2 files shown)


Kooshaba (approved):
Tested on Sky Strife, everything is working correctly 👍🏼

holic:
We should probably have a way for the current owner to update the owner, right?

If so, we might consider inheriting from Ownable in one of the standard set of solidity libraries (OpenZeppelin, etc.)

alvrs:
Agree, waiting for https://github.com/latticexyz/mud/pull/239 before this refactor

ludns:
Can we store ownership in ECS too? no reason to have state that doesn't
conform to our way of storing state. we'll need to have custom getters
client side in case someone builds a frontend for system/component
management using MUD.
I think it's good to expose view functions that implements IOwnable; but
let's avoid saving the state in the systems/components with a custom
storage slot.

ludns:
Or if we don't do that; we can standardise the fact that only components
can write to their map set (which would make sense); and let Components
handle access control themselves with that state stored in ECS.
Mental model here is that all state that could possibly be needed by a CLI
/ a frontend should be in ECS.

PR #250, opened 2022-11-18
feat(solecs): storage access utility

cha0sg0d:
Based on the assumption that `System` contracts in MUD have `IUint256Component components` and `IWorld world` as the first two state variables, we can fetch their addresses directly from storage using assembly `sload`.

As long as this variable invariant holds, as documented [here](https://github.com/latticexyz/mud/blob/927dbc584254031ea563b2c7c9d5441e78468c15/packages/solecs/src/LibStorage.sol#L18), a nice pattern is now possible:

Previously in MUD, any `Library` that wants to read or write from a component needs to be passed the address of the components registry:
```solidity
  function isAdmin(IUint256Component components, address sender) internal view returns (bool) {
    AdminComponent adminComponent = AdminComponent(getAddressById(components, AdminComponentID));
    ...
  }
```

Furthermore, it is tedious from a developer experience to have to call `getAddressById(components, ...)` every time a component needs to be loaded.

Using LibStorage, MUD Libraries and Systems can now do the following:
```solidity
import {LibStorage} from "solecs/LibStorage.sol";
import { AdminComponent , ID as AdminComponentID } from "../components/AdminComponent.sol";

library  LibAdmin {
  function isAdmin(address sender) internal view returns (bool) {
    AdminComponent adminComponent = AdminComponent(LibStorage.comp(AdminComponentID));
    ...
  }
}

```

This is possible because _the library knows the location of the components' address by doing a direct storage read instead of getting passed that address_.

You can reduce boilerplate code even further with two tricks:

1. Alias `LibStorage` to something shorter:
```solidity
import {LibStorage as s} from "solecs/LibStorage.sol";

library  LibAdmin {
  function isAdmin(address sender) internal view returns (bool) {
    AdminComponent adminComponent = AdminComponent(s.comp(AdminComponentID));
    ...
  }
}
```
2. Use a more succinct naming scheme for components:
```solidity
import {LibStorage as s} from "solecs/LibStorage.sol";
import { AdminComponent as Admin , ID as AdminID } from "../components/AdminComponent.sol";

library  LibAdmin {
  function isAdmin(address sender) internal view returns (bool) {
    Admin admin = Admin(s.comp(AdminID));
    ...
  }
}
```

**Using these tricks, we get a `~54` char reduction in code everywhere we want to access a component.**

You can also store the component registry address at the start of a function and use the traditional method to access:

```solidity
import {LibStorage as s} from "solecs/LibStorage.sol";
import { AdminComponent as Admin , ID as AdminID } from "../components/AdminComponent.sol";

library  LibAdmin {
  function isAdmin(address sender) internal view returns (bool) {
    UInt256Components c = s.c();
    Admin admin = Admin(getAddressByID(c, AdminID));
    ...
  }
}
```

### DIFF
--- a/packages/solecs/src/LibStorage.sol
+++ b/packages/solecs/src/LibStorage.sol
@@ -1,0 +1,42 @@
+// SPDX-License-Identifier: Unlicense
+pragma solidity >=0.8.0;
+
+import { IUint256Component } from "./interfaces/IUint256Component.sol";
+import { IWorld } from "./interfaces/IWorld.sol";
+import { getSystemAddressById, getAddressById } from "./utils.sol";
+
+/**
+  This library provides a utility to access the addresses of the components registry
+  and the world directly. 
+
+  It relies on the invariant that the base System class declares components as the first state variable
+  and the world as the second one.
+
+  This enables a direct storage read based on the layout of storage in Solidity
+  https://docs.soliditylang.org/en/latest/internals/layout_in_storage.html#layout-of-state-variables-in-storage
+
+  DO NOT change the order of the base System variables.
+  DO NOT add a variable before the components and world in the base System.
+ */
+
+library LibStorage {
+  function c() internal view returns (IUint256Component components) {
+    assembly {
+      components := sload(0)
+    }
+  }
+
+  function w() internal view returns (IWorld world) {
+    assembly {
+      world := sload(1)
+    }
+  }
+
+  function sys(uint256 systemID) internal view returns (address) {
+    return getSystemAddressById(c(), systemID);
+  }
+
+  function comp(uint256 componentID) internal view returns (address) {
+    return getAddressById(c(), componentID);
+  }
+}


0xhank:
When this method is used within a foundry test, the test breaks -- it reverts when the test attempts to recover the precompile. This may have something to do with the fact that the test suite creates a storage environment that is different from the internal systems and libraries. 

I can still use this method internal to functions, but I need to use raw `getAddressByID` within my testing functions.

image

cha0sg0d:
> When this method is used within a foundry test, the test breaks -- it reverts when the test attempts to recover the precompile. This may be because the test suite creates a storage environment different from internal systems and libraries.
> 
> And since libraries are compiled statically, using this method in libraries also breaks test functions. Tests only work when `LibStorage` is accessed in contracts. I need to use raw `getAddressByID` within my testing functions and libraries for successful testing.
> 
> image

^^ This issue needs to be addressed before merging this PR (@alvrs). The problem is that `MudTest.t.sol` has a different storage layout than the generic `System.sol`. If you call a library directly in a test, you `delegatecall` to that library, which inherits the storage of `MudTest` not `System` and thus LibStorage cannot be used.

To get MudTest to match the storage of a System, we can use the Foundry [store](https://book.getfoundry.sh/cheatcodes/store) Cheatcode to manually set the storage slots, but these needs to be tested thoroughly before use.

alvrs:
> This issue needs to be addressed before merging this PR (@alvrs). The problem is that MudTest.t.sol has a different storage layout than the generic System.sol. If you call a library directly in a test, you delegatecall to that library, which inherits the storage of MudTest not System and thus LibStorage cannot be used.

Right now `MudTest.t.sol` is part of user's codebases, not MUD, so we can't enforce a particular storage layout. We should probably move `MudTest.t.sol` to MUD.

Also I think this library should be part of `@latticexyz/std-client`, not `@latticexyz/solecs` (trying to keep the core solecs surface as small as possible, and this lib is a great example of an optional extension that makes devs live easier, like other things in std-client)

cha0sg0d:
> Right now `MudTest.t.sol` is part of user's codebases, not MUD, so we can't enforce a particular storage layout. We should probably move `MudTest.t.sol` to MUD.
> 
> Also I think this library should be part of `@latticexyz/std-client`, not `@latticexyz/solecs` (trying to keep the core solecs surface as small as possible, and this lib is a great example of an optional extension that makes devs live easier, like other things in std-client)

You mean `std-contracts` right? I can make a new PR with `LibStorage` and `MudTest.t.sol` with the correct storage layout into `std-contracts.

alvrs:
yes sorry std-contracts!

cha0sg0d:
https://github.com/latticexyz/mud/pull/264 Closes #250

PR #251, opened 2022-11-18
feat: add ability to request tx metadata per ecs event

authcall:
Adds new `ecsEventsIncludeTxMetadata` flag to a `ECSStreamBlockBundleRequest` that is used to open a stream. If `ecsEventsIncludeTxMetadata` is set to true, every ECS event will contain a new field with a `TxMetadata` struct that contains information about the transaction that originated that event

```
message TxMetadata {
  string to = 2;
  bytes data = 3;
  uint64 value = 4;
}
```
in typescript (auto generated)
```
export interface TxMetadata {
  to: string;
  data: Uint8Array;
  value: number;
}
```

This is motivated as a replacement for "system call streams"


Kooshaba:
do i need to run the streaming service in a specific way in order to emit the metadata? i'm testing locally and the metadata is blank

Kooshaba:
i layered on this PR that parses the `SystemCall` events out of the stream
https://github.com/latticexyz/mud/pull/254

alvrs (approved):
Good to be merged after #254 is merged into this one

Kooshaba (approved):
🚢

alvrs (approved):

PR #256, opened 2022-11-20
feat(network): add support for external wallets (eg MetaMask)

smallbraingames:
Support for external wallets to mud. 
Allows users to pass in `externalProvider` (which is an ethers.js `ExternalProvider`) to the network config and then use the associated signer for txs.

### DIFF
--- a/packages/network/src/createNetwork.ts
+++ b/packages/network/src/createNetwork.ts
@@ -38,10 +38,10 @@ export async function createNetwork(initialConfig: NetworkConfig) {
   // Get address
   const initialConnectedAddress = config.provider.externalProvider ? await signer.get()?.getAddress() : undefined;
   const connectedAddress = computed(() =>
-    config.privateKey ? new Wallet(config.privateKey).address.toLowerCase() : initialConnectedAddress
+    config.privateKey ? new Wallet(config.privateKey).address.toLowerCase() : initialConnectedAddress?.toLowerCase()
   );
   const connectedAddressChecksummed = computed(() =>
-    config.privateKey ? new Wallet(config.privateKey).address : undefined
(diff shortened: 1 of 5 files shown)


packages/network/src/createNetwork.ts
```diff
@@ -29,14 +29,16 @@ export async function createNetwork(initialConfig: NetworkConfig) {
...
+    if (config.provider.externalProvider) return currentProviders.json.getSigner();
+    const privateKey = config.privateKey;
     if (privateKey && currentProviders) return createSigner(privateKey, currentProviders);
   });
 
   // Get address
+  const initialConnectedAddress = config.provider.externalProvider ? await signer.get()?.getAddress() : undefined;
   const connectedAddress = computed(() =>
-    config.privateKey ? new Wallet(config.privateKey).address.toLowerCase() : undefined
+    config.privateKey ? new Wallet(config.privateKey).address.toLowerCase() : initialConnectedAddress
```
alvrs, line 41: i'm assuming the address we get from `signer.getAddress()` is checksummed? If we we need to normalize to lowercase here and also pass the non-lowercased address in `connectedAddressChecksummed`

alvrs (commented):
Lgtm, just one small nit

packages/network/src/createNetwork.ts
mirshko: Yup it does: https://docs.ethers.io/v5/api/utils/address/#utils-getAddress

chlwys:
hey @alvrs any updates on this? it looks good, want to use it with the main package

smallbraingames:
@chlwys Thank you for the reminder
Was working on making a string lowercase for the past few months but recently had a breakthrough, so this should be good to go now

alvrs (approved):
lgtm 🚢

PR #257, opened 2022-11-21
fix: cross-platform removebytecode for solecs and std-contracts

dk1a:
I've been using mud on windows, saw someone on discord saying it doesn't work, this fixes the only problem I've had to deal with so far. The issues are the bash file loop, and to a lesser extent jq.
I've tested it on ubuntu and windows (+ "\n" in the script is to keep output exactly the same).
I also put the new removebytecode in build instead of types, since it doesn't rly have much to do with typechain, but can change it back if you like

I'm not entirely sure why removebytecode exists, but I guess it's to do with network which uses abi, to avoid unnecessary bloat? Anyways I assumed it's not useless and just removing it wouldn't be best.


packages/solecs/removeBytecode.ts
```diff
@@ -0,0 +1,17 @@
...
+    const abi = JSON.parse(fs.readFileSync(realpath).toString("utf8"));
+    delete abi["bytecode"];
+    delete abi["deployedBytecode"];
+    delete abi["ast"];
+
+    fs.writeFileSync(realpath, JSON.stringify(abi, null, 2) + "\n");
+  }
+}
+
+removeBytecode("abi/");
```
holic, line 17: @alvrs @ludns Is there a package to put shared utils like this so we don't have to duplicate code around? If not, should there be?

dk1a:
Just to note, I imagine cli (and maybe other packages) have more issues, this just fixes solecs and std-contracts

packages/solecs/removeBytecode.ts
ludns: yes `utils` in this repo
`@latticexyz/utils`

dk1a: I had the same thought, but noticed utils wasn't already imported and decided against it. Can move it there then in a bit

dk1a:
Took a while, mostly reading about esm in node and ts-node. Apparently https://nodejs.org/api/esm.html#mandatory-file-extensions is a thing and utils (amongst other packages) isn't a valid module. Which is not fine for directly calling it in node scripts.
So I made utils not a module.
Imports in js files can only be used in a module, so the one js file there is now ts (jest.config.js)

holic:
Ohh, I see what's going on now. I think I misread before.

Another approach is to have `utils` build the node-compatible executable (TS -> JS) and include it in the [`bin` field of `package.json`](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#bin), then use it directly rather than importing the file inside an npm script.

dk1a:
> Ohh, I see what's going on now. I think I misread before.
> 
> Another approach is to have `utils` build the node-compatible executable (TS -> JS) and include it in the [`bin` field of `package.json`](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#bin), then use it directly rather than importing the file inside an npm script.

Made a cli file, although not quite the way you meant it (building it is unnecessary imo)

holic:
Looks great! I defer to @alvrs and @ludns for final review.

dk1a:
Added a tiny fix for network too - semicolons don't work in cmd.exe

alvrs:
Thanks for this PR, this is great! Unfortunately the changes to `utils` cause an error downstream (when locally linking MUD in dev)

CleanShot 2022-11-25 at 14 28 37@2x

To reproduce clone OPCraft, run `yarn link:packages` in mud, then `yarn link:mud` in OPCraft, then `yarn dev` in OPCraft/packages/client

dk1a:
Makes sense, I shouldn't have put a node util into a clientside package, even if it's not used ts will still complain.
I can make a new package, e.g. utils-node. Or make another folder in utils, but that seems easier initially and more problematic later.
@alvrs thoughts?

alvrs:
> I'm not entirely sure why removebytecode exists, but I guess it's to do with network which uses abi, to avoid unnecessary bloat? Anyways I assumed it's not useless and just removing it wouldn't be best.

`removebytecode` was initially added because of some incompatibility between the abi output generated by forge and typechain; if it works with the current versions of forge and typechain we could actually get rid of it

dk1a:
Tested it for solecs, typechain works fine. With bytecode present it generates more stuff, like the deploy method.
But if the problem was typechain working at all, and not the specific methods, then I can make a new much simplified PR to get rid of removebytecode and fix the small problem in network

alvrs:
Yep also just tested, there seems to be an issue with typescript 7 and forge's bytecode, but everything works fine with typechain 8, so i don't think we need removebytecode anymore

PR #258, opened 2022-11-23
style(solecs): rename arguments to args

mirshko:
This is changed as Typechain-exported Contracts do not like the variable name arguments as its a
reserved keyword in strict-mode Classes


alvrs (approved):
thanks!

PR #260, opened 2022-11-23
Fixed generating systemTypes on Windows.

nikoladimitroff:
The issue was that on Windows ${CWD} results in an absolute path which glob doesn't support unless the proper option is passed. The fix just has the CLI use relative paths all the way.

Disclaimer: Can't test ot unix-based systems, but according to the docs of glob in NPM, this should be cross-platform compatible.


alvrs:
Should be fixed by https://github.com/latticexyz/mud/pull/277/files#diff-298dc39417c807aeef6f02a10c644e3c31cef17da467f18815cac1b5670fce8eR48

PR #261, opened 2022-11-23
refactor: recalculate world view on phaser resize, remove window resize handler

holic:
- recalculate world view on phaser resize (otherwise calling `game.scale.resize()` may not load in chunks that were previously out of view)
- remove window resize handler, let phaser do its own resizing and lets userland do its own resize handling

(I'll likely be opening a PR with phaser to improve their resize handling)


holic:
Split out the phaser resizing here: https://github.com/latticexyz/mud/pull/263

Will follow up with a less breaking approach to container resize handling

PR #264, opened 2022-11-25
feat(solecs): add deterministic storage location for systems, add storage access util

cha0sg0d:
# Reasoning

## Current State of MUD

**The non-deterministic location of the `world` and `components` contract addresses creates unnecessary boilerplate code in MUD contracts**

When developing smart contracts in MUD, two common actions include accessing the `components` registry contract and the `world` contract. The addresses of these contracts are manually passed in to the constructor for each System and Component contract in MUD (although Component contracts only use the  `world` address). 

The `components` registry contract is required to read or write any component, and is typically accessed as follows:
```solidity
    EntityComponent entityComponent = EntityComponent(getAddressById(components, EntityComponentID));
```

The `world` contract is most commonly used for getting unique entity IDs:
```solidity
    uint256 spawnEntity = world.getUniqueEntityId();
```

A common practice in MUD development is to put business logic in libraries, so they can be reused between systems. However, moving logic to libraries, combined with the need to access the `world` and `components` contracts, means that _the addresses of these contracts must be passed by value to the libraries that use them._

An example of such behavior looks like this:
```solidity
// LibRoom.sol

  function createRoom(
    IWorld world,
    IUint256Component components,
    Coord memory roomPosition)
    
    EntityComponent entityComponent = EntityComponent(getAddressById(components, EntityComponentID));

    ... 
```
Furthermore, if a function inside `createRoom` needs to access the `components`, that value must be passed in again as well.

```solidity
// LibRoom.sol

  function createRoom(
    IWorld world,
    IUint256Component components,
    Coord memory roomPosition)

    ....
    LibEntities.createSpawn(
      CreateSpawnArgs({
        components: components,
        id: world.getUniqueEntityId(),
        ...
        }))
```

## Proposed Solution

In this PR, we make the locations of the `world` and `components` addresses deterministic in `System` contracts. 

We do this by using [diamond storage](https://dev.to/mudgen/how-diamond-storage-works-90e) to place the addresses of the `world` and `component` contracts at a specific place in the storage of every system.

Now, any function, in a system or a library called by a system that wants to access `world` or `components` _can do so with one storage read instead of threading the addresses through every function._

The new access pattern looks like this:

```solidity
// LibRoom.sol

  function createRoom(
    Coord memory roomPosition)

    EntityComponent entityComponent = EntityComponent(getAddressById(SystemStorage.components(), EntityComponentID));

    ....
    LibEntities.createSpawn(
      CreateSpawnArgs({
// We don't need to pass in `components` anymore because `createSpawn` can access it as needed
        id: SystemStorage.world().getUniqueEntityId(),
        ...
        }))
```

This PR enables MUD contracts (systems, libraries called by systems, and tests) to access the `world` and `components` contracts without passing the addresses around in every function. It also improves code readability because only the relevant arguments (not arguments for storage reads/writes) are passed into library functions

There are additional features being discussed to reduce boilerplate further for accessing components, including aliasing and helper functions, (cc @Flynnsc), but that is for future PRs.

@holic is this what you wanted?


cha0sg0d:
Ok after some introspection, the _only_ difference in storage`MudTest.t.sol` and `System.t.sol` is that `MudTest.t.sol` inherits `DSTest.sol`, which has two boolean variables in the first two bits of slot 0 in storage. This slight difference means that `LibStorage` cannot be used interchangeably with MUD tests that call libraries directly (but it still works for MUD tests that call systems)

To use the `LibStorage` access pattern in testing and deployed contract code, we need to either change the storage layout of `DSTest` or `System`.

My solution is to create ` StdSystem` contract in `std-contracts` that copies the addresses of the `World` contract and `Component registry` at specified locations in storage:

```solidity
  uint256 constant worldID = uint256(keccak256("mud.world"));
  uint256 constant componentsID = uint256(keccak256("mud.components"));
```

If any developer creates systems in the following way:
```solidity
import "std-contracts/systems/StdSystem.sol";

contract MySystem is StdSystem {
// 
}
```

They can use the `LibStorage.sol` utility, which enables access patterns documented [here](https://github.com/latticexyz/mud/pull/250#issue-1455242100)

To ensure that `LibStorage.sol` works for tests as well, just add the following [code](https://github.com/cha0sg0d/mud/blob/ca6ae616d4a5f6a1ffbf5147b00b96250860bbfa/packages/std-contracts/src/test/Sample.t.sol#L22) to your base test contract, likely `MudTest.t.sol`.

This gives the base testing contract the same storage of `World` and `Component` registry as deployed systems.

This closes #250 and moves all logic out of `solecs`.

dk1a:
Apparently we're introducing somewhat different diamond storage at the same time, check out OwnableWritableStorage in #265

dk1a:
Specifically if one were to store world and components the same as I did with owner in OwnableWritableStorage, much of LibStorage wouldn't be necessary, it'd be just helpers for getAddressById

cha0sg0d:
> Specifically if one were to store world and components the same as I did with owner in OwnableStorage, much of LibStorage wouldn't be necessary, it'd be just helpers for getAddressById

Ah yup I see - what do you think is a good way to combine? The only difference between my PR and traditional Diamond storage is that I'm storing addresses of deployed contracts, not structs, so I'm using the `storeAddress` method as part of the `System` constructor to ensure the data gets to the right place.

dk1a:
I meant to just move systems to diamond storage entirely, that would be a very breaking change ofc. See #266

dk1a:
Alternatively, you can just use my pattern to do a non-breaking change in std-contracts

dk1a:
The main idea there is to avoid passing around worldID, componentID. There is only 1 STORAGE_SLOT and only storage lib need know about it

cha0sg0d:
> The main idea there is to avoid passing around worldID, componentID. There is only 1 STORAGE_SLOT and only storage lib need know about it

I just looked at #266 , this makes sense to me. LibStorage doesn't need to exist if we just have `SystemStorage.sol`. If @alvrs concurs, I think #266 can close #264. Perhaps the only benefit of my implementation is that it's entirely in `std-contracts`, but I don't think that's a big deal.

packages/std-contracts/src/libraries/LibStorage.sol
```diff
@@ -0,0 +1,44 @@
+// SPDX-License-Identifier: Unlicense
```
alvrs, line 1: Let's change all licenses to `MIT`

packages/std-contracts/src/systems/SampleSystem.sol
```diff
@@ -0,0 +1,25 @@
+// SPDX-License-Identifier: Unlicense
+pragma solidity >=0.8.0;
+import "./StdSystem.sol";
+import { IWorld } from "solecs/interfaces/IWorld.sol";
+import { LibStorage as s } from "../libraries/LibStorage.sol";
+import { console } from "forge-std/console.sol";
+
+uint256 constant ID = uint256(keccak256("system.sample"));
+
+contract SampleSystem is StdSystem {
```
alvrs, line 10: This only seems to be used in `Sample.t.sol`, should we move it to `src/test`?

alvrs (commented):
Looks good! But will have a look at #266 first as mentioned in the conversation above

dk1a:
@cha0sg0d was a bit busy today, but I kinda wanna hijack this unless you would really prefer to do it urself. A non-breaking version of #266 with a util lib should be easy enough, the main question was whether alvrs was ok with that concept at all, which has been resolved

dk1a:
@cha0sg0d Some things I was thinking of doing differently:

c and w are too short imo, that's personal preference ofc but I just don't like 1 letter vars, and those aren't very often needed either (you'd use sys and comp more I think)

You can call `layout()` in `initSystem` itself, then you don't need using for

I wanted to try and put the storage stuff into System itself. It'd require some testing to make sure it's indeed not a breaking change, not sure yet

cha0sg0d:
@alvrs @dk1a take a look when you get a chance.

Couple of notes:
- I moved all of the `getter` functions (`c()`, `w()`) to `LibStorage` (@dk1a had them in `System.sol`) because we want to use the getters outside of the System contracts.
- Re naming, there are a couple conventions being used that we should align on. Specifically:
```solidity
  struct Layout {
    IUint256Component components;
    IWorld world;
  }

  bytes32 internal constant STORAGE_SLOT = keccak256("solecs.contracts.storage.System");

  function c() public view returns (IUint256Component) {
    return layout().components;
  }

  function w() public view returns (IWorld) {
    return layout().world;
  }

  function sys(uint256 systemID) internal view returns (address) {
    return getSystemAddressById(c(), systemID);
  }

  function comp(uint256 componentID) internal view returns (address) {
    return getAddressById(c(), componentID);
  }
```

Do we want to keep these names short for boilerplate reduction? or better to have the full name and let developers rename the functions as needed.

cha0sg0d:
> @cha0sg0d Some things I was thinking of doing differently:
> 
> c and w are too short imo, that's personal preference ofc but I just don't like 1 letter vars, and those aren't very often needed either (you'd use sys and comp more I think)
> 
> You can call `layout()` in `initSystem` itself, then you don't need using for
> 
> I wanted to try and put the storage stuff into System itself. It'd require some testing to make sure it's indeed not a breaking change, not sure yet

I made the names longer and removed the `using keyword`, thanks!

Why put storage stuff into System itself?

dk1a:
> Why put storage stuff into `StdSystem` itself?

To make it the default and deprecate the old `components`, `world`
(except I'm not entirely sure they'll coexist together as well as I imagine)

dk1a:
I don't mean to ditch the lib and put it all into the system. Rather to use `initSystem` inside the base `System`

cha0sg0d:
Re:
> Oh I meant not StdSystem, but System, so you'd modify solecs, not std-contracts
> To make it the default and deprecate the old components, world
(except I'm not entirely sure they'll coexist together as well as I imagine)

My understanding is that this would require any instance of `components` being replaced with `comp()`, so would be a breaking change. Based @alvrs prev comment, seems like going the no breaking changes route for now makes sense, but if devs like this pattern we could push it to `solecs`.

Re:
> I don't mean to ditch the lib and put it all into the system. Rather to use initSystem inside the base System

I agree that `initSystem` in `LibStorage` is not super clean, but I can elaborate on why I put it there. For test purposes, the test contracts like `SampleTest` need to mirror the storage of the `System`, although they are not 1:1 equivalent. Having `initSystem` in the library enables test contracts to have the same storage, and if the test calls a library directly instead of a system there will be an error unless the storage slots exist in the same place.

wdyt?

dk1a:
> My understanding is that this would require any instance of `components` being replaced with `comp()`, so would be a breaking change. Based @alvrs prev comment, seems like going the no breaking changes route for now makes sense, but if devs like this pattern we could push it to `solecs`.

I jsut explained it poorly, I mean you can put `LibStorage.initSystem(_world, IUint256Component(_components));` into `System`'s contructor, and then you don't need `StdSystem`

> I agree that `initSystem` in `LibStorage` is not super clean, but I can elaborate on why I put it there. For test purposes, the test contracts like `SampleTest` need to mirror the storage of the `System`, although they are not 1:1 equivalent. Having `initSystem` in the library enables test contracts to have the same storage, and if the test calls a library directly instead of a system there will be an error unless the storage slots exist in the same place.

Same here, didn't mean to criticize `initSystem`, just that you can call it in `System`'s constructor.

Although you could also remove `initSystem` and do 
```solidity
  constructor(IWorld _world, address _components) System(_world, _components) {
    LibStorage.layout().world = _world;
    LibStorage.layout().components = _components;
  }
```
but that's a matter of preference, and not important imo

packages/std-contracts/src/libraries/LibStorage.sol
```diff
@@ -0,0 +1,43 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
+
+import { IUint256Component } from "solecs/interfaces/IUint256Component.sol";
+import { IWorld } from "solecs/interfaces/IWorld.sol";
+import { getSystemAddressById, getAddressById } from "solecs/utils.sol";
+
+uint256 constant worldID = uint256(keccak256("mud.world"));
+uint256 constant componentsID = uint256(keccak256("mud.components"));
```
alvrs, line 9: do we still need these?

packages/std-contracts/src/libraries/LibStorage.sol
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * This library will only work if you use `std-contracts/StdSystem.sol` for every system in your contracts
+ */
+
+library LibStorage {
```
alvrs, line 15: Minor naming nit - wdyt about calling this `SystemStorage`? This is a pattern used by `@solidstate` (eg https://github.com/solidstate-network/solidstate-solidity/blob/master/contracts/access/ownable/OwnableStorage.sol) and makes it clearer what storage this library manages. (I can see us adding more different types of storage in the future and probably want to manage them in individual libraries)

packages/std-contracts/src/libraries/LibStorage.sol
```diff
@@ -0,0 +1,43 @@
...
+  bytes32 internal constant STORAGE_SLOT = keccak256("solecs.contracts.storage.System");
+
+  function layout() internal pure returns (Layout storage l) {
+    bytes32 slot = STORAGE_SLOT;
+    assembly {
+      l.slot := slot
+    }
+  }
+
+  function initSystem(IWorld _world, IUint256Component _components) internal {
```
alvrs, line 30: if we rename this library to `SystemStorage`, my preference would be to rename this function to `init`

packages/std-contracts/src/libraries/LibStorage.sol
```diff
@@ -0,0 +1,43 @@
...
+    Layout storage l = layout();
+    l.world = _world;
+    l.components = _components;
+  }
+
+  function comp() public view returns (IUint256Component) {
```
alvrs, line 36: My preference here would be to call the function `components` - i think abbreviation of variable/function names is generally an anti-pattern (https://www.codewithjason.com/variable-name-anti-patterns/); I do see the point of reducing boiler plate, but would rather go in the direction of something @FlynnSC was exploring (autogenerating a library to directly access typed components without the need for manual `getAddressById` and casting)

packages/std-contracts/src/test/Sample.t.sol
```diff
@@ -0,0 +1,36 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
+
+// External
```
alvrs, line 4: What does this comment mean?

packages/std-contracts/src/test/SampleSystem.sol
```diff
@@ -0,0 +1,25 @@
...
+contract SampleSystem is StdSystem {
+  constructor(IWorld _world, address _components) StdSystem(_world, _components) {}
+
+  function execute(bytes memory arguments) public returns (bytes memory) {
+    if (address(s.world()) == address(world)) {
+      console.log("World is correct");
+    }
+    if (address(s.comp()) == address(components)) {
+      console.log("Components are correct");
+    }
```
alvrs, line 19: I think we should change this to use `require` (or `revert`) so the execute function reverts if the world is not correct and the test fails

packages/std-contracts/src/libraries/LibStorage.sol
alvrs: Or `LibSystemStorage` if you prefer. Also why did you remove the `system(uint256 id): address` and `component(uint256 id): address` utils from the library? They seem useful to reduce boilerplate.

packages/std-contracts/src/libraries/LibStorage.sol
cha0sg0d: Removed in [16e2033](https://github.com/latticexyz/mud/pull/264/commits/16e203374934ad3df1115076c0083939d69a1060)

packages/std-contracts/src/libraries/LibStorage.sol
cha0sg0d: Renamed to `SystemStorage` and added `system` and `component` getters in [16e2033](https://github.com/latticexyz/mud/pull/264/commits/16e203374934ad3df1115076c0083939d69a1060)

packages/std-contracts/src/libraries/LibStorage.sol
cha0sg0d: Renamed in [16e2033](https://github.com/latticexyz/mud/pull/264/commits/16e203374934ad3df1115076c0083939d69a1060)

packages/std-contracts/src/libraries/LibStorage.sol
cha0sg0d: renamed in [16e2033](https://github.com/latticexyz/mud/pull/264/commits/16e203374934ad3df1115076c0083939d69a1060)

packages/std-contracts/src/test/Sample.t.sol
cha0sg0d: Removed in [16e2033](https://github.com/latticexyz/mud/pull/264/commits/16e203374934ad3df1115076c0083939d69a1060)

packages/std-contracts/src/test/SampleSystem.sol
cha0sg0d: Changed to revert in [16e2033](https://github.com/latticexyz/mud/pull/264/commits/16e203374934ad3df1115076c0083939d69a1060)

cha0sg0d:
@alvrs should be good to go. What other documentation should be provide so people know they can use the `StdSystem + SystemStorage` pattern? Should we wait till we add Flynn's idea as well?

cha0sg0d:
> @cha0sg0d Wait, are you sure you want to make a new StdSystem instead of calling `SystemStorage.init` in the original System?

As in, touching `solecs`?

dk1a:
yep, my changes were breaking because I removed stuff from `System`. But if you just add `SystemStorage.init` it's not a breaking change

cha0sg0d:
> yep, my changes were breaking because I removed stuff from `System`. But if you just add `SystemStorage.init` it's not a breaking change

I'll add that, and see what @alvrs thinks.

dk1a:
Modifying solecs means you don't need to teach people to use StdSystem, and later making breaking changes to move to diamond storage entirely is easier

cha0sg0d:
> Modifying solecs means you don't need to teach people to use StdSystem, and later making breaking changes to move to diamond storage entirely is easier

Done! Now waiting for @alvrs

dk1a:
@alvrs A little detail that will also be relevant in the future - where do you prefer Storage libs?
I'd put it next to its contract, so same directory as `System`, not in `libraries`
But it is a library, so `libraries` makes sense as well

packages/std-contracts/src/test/Sample.t.sol
```diff
@@ -0,0 +1,34 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
```
alvrs, line 2: Since this test is mostly testing that a system's `SystemStorage` is behaving correctly (and the system can be executed), I'd prefer moving the test into a `System.t.sol` in `solecs/src/test`. All it has to test is that after initializing the system, `SystemStorage.world()/components()` corresponds to `world/world.components()`.

alvrs (commented):
Almost there! I support the decision to put the new `SystemStorage` in the default system. Only downside is systems cost slightly more gas to deploy because `components`/`world` is stored twice, but I think this is a reasonable price to pay to make utilities building on this pattern compatible with the next breaking version of solecs (where we should remove the deprecated `world` / `components` fields). Speaking of which, I think we should add a `@deprecated use SystemStorage.components() instead` comment to the previous variables (line 24, 25 in `System.sol`). Besides that just one small comment about the tests, then we're good to go imo!

alvrs:
> @alvrs A little detail that will also be relevant in the future - where do you prefer Storage libs? I'd put it next to its contract, so same directory as `System`, not in `libraries` But it is a library, so `libraries` makes sense as well
> \- https://github.com/latticexyz/mud/pull/264#issuecomment-1374103021

Good point, I'd follow @solidstate/contracts and put the storage libraries next to the contract too.

packages/std-contracts/src/test/MudTest.t.sol
```diff
@@ -49,5 +50,6 @@ contract MudTest is DSTest {
     alice = utils.getNextUserAddress();
     bob = utils.getNextUserAddress();
     eve = utils.getNextUserAddress();
+    s.init(world, components);
```
cha0sg0d, line 53: @alvrs I'm adding this here so that anyone using MudTest doesn't have to manually call `SystemStorage.init` in each test.

cha0sg0d:
@alvrs Should be good to go now, lmk!

alvrs (dismissed):
Looks good to me! 🚀 Let's wait a couple hours for @holic to take a look before merging though.

packages/std-contracts/src/test/MudTest.t.sol
```diff
@@ -6,6 +6,7 @@ import { IWorld } from "solecs/interfaces/IWorld.sol";
 import { IUint256Component } from "solecs/interfaces/IUint256Component.sol";
 import { Vm } from "forge-std/Vm.sol";
 import { getAddressById } from "solecs/utils.sol";
+import { SystemStorage as s } from "solecs/SystemStorage.sol";
```
holic, line 9: Any reason we can't keep this as `SystemStorage` instead of aliasing to `s`? Feels more readable.

holic (changes requested):
I think I am missing some context on the "why" behind this change. Mind filling in the PR description with the current state of things, limitations, and what these changes unlock? This will help future us when we go digging into why certain changes were made or why the code is shaped in a particular way.

packages/std-contracts/src/test/MudTest.t.sol
cha0sg0d: Handled here: [4859a07](https://github.com/latticexyz/mud/pull/264/commits/4859a0757c204c1b60cea8b97e1cfc4b66d7bc86) !

holic (commented):
> @holic is this what you wanted?

This is _exactly_ what I was looking for, thanks so much for the context! Super clear what this is for. This is a very cool strategy and makes a ton of sense.

Are there any concerns about `SystemStorage.components()` being called within a lib where that lib is being called from not-a-system? Do we need some indication that certain libs are system-only libs and expected to just be used within the context of a System? Or some way for SystemStorage to detect if you're in a system and revert if not?

Does this still work for linked libraries? I assume yes, but just checking.

holic:
Quickly wired this into Emojimon to make sure it works and to see what the diff looks like:
https://github.com/latticexyz/emojimon/commit/74629a87e291b21268969d981a20b46044f4028f

cha0sg0d:
> Are there any concerns about `SystemStorage.components()` being called within a lib where that lib is being called from not-a-system? Do we need some indication that certain libs are system-only libs and expected to just be used within the context of a System? Or some way for SystemStorage to detect if you're in a system and revert if not?

> Does this still work for linked libraries? I assume yes, but just checking.

My understanding is that library's functions is executing with the storage of the contract that called it (inline for embedded `internal` libraries and via `delegatecall` for linked libraries. Libraries have no storage of their own, and this is true for embedded or linked libraries. 

As you said, the only time you would run into trouble is if you call a library that _expects_ to access `SystemStorage` from a contract that does not have the correct addresses loaded at the correct location. I'm not sure what the behavior would be, probably a run-time error.

The only time this happened to me was calling a library directly from a test contract (inherited from MudTest.t.sol). This has been fixed in the PR by initializing SystemStorage in MudTest.

I would say it's not a big thing to worry about, but if you think it's necessary, we could make a `sanityCheck` function in `SystemStorage` that loads the `world` and `components` addresses and calls a function on them.

wdyt?

holic:
That all makes sense to me/is how I'd interpret things working too. I am just thinking about potentially-hard-to-track-down errors when developing and wondering if/how we can detect+revert early if you're using it wrong. A linter could also be a potential solution.

holic (dismissed):
Gonna toss back to @alvrs for any final thoughts + approval.

alvrs:
Thanks for reviewing @holic and thanks for the detailed and elaborate context in https://github.com/latticexyz/mud/pull/264#issue-1464846901 @cha0sg0d! One last thought: could we add some [NatSpec](https://docs.soliditylang.org/en/v0.8.17/natspec-format.html) comments to `SystemStorage`? The comments get parsed and put on the [mud docs website](https://mud.dev/packages/solecs/api/) automatically once this gets merged. And here we could also add a notice for developers that this library expects to be initialized before being used (which is done by default in systems).

alvrs:
Other than that everything looks great and I'm down to merge as soon as we have some natspec comments

cha0sg0d:
> Thanks for reviewing @holic and thanks for the detailed and elaborate context in [#264 (comment)](https://github.com/latticexyz/mud/pull/264#issue-1464846901) @cha0sg0d! One last thought: could we add some [NatSpec](https://docs.soliditylang.org/en/v0.8.17/natspec-format.html) comments to `SystemStorage`? The comments get parsed and put on the [mud docs website](https://mud.dev/packages/solecs/api/) automatically once this gets merged. And here we could also add a notice for developers that this library expects to be initialized before being used (which is done by default in systems).
> 
> Other than that everything looks great and I'm down to merge as soon as we have some NatSpec comments. 🎉

Ok, lmk if this is good!

alvrs (dismissed):
Looks great, amazing work! Thanks for being so patient @cha0sg0d and thanks for the great discussion @dk1a @holic @cha0sg0d

alvrs (approved):

PR #265, opened 2022-11-25
refactor(solecs): use @solidstate Ownable, add OwnableWritable, fix some import formatting

### LINKED ISSUE #245: Use Ownable std implementation for all ownable MUD contracts (Component, System, Set, MapSet))
We should have a way for the current owner to update the owner. We also might consider inheriting from Ownable in one of the standard set of solidity libraries (OpenZeppelin, etc.)

_Originally posted by @holic in https://github.com/latticexyz/mud/issues/244#issuecomment-1311798971_

dk1a:
Closes #245

* Use Ownable from @solidstate/contracts to implement IERC173, also use solidstate for IERC165
* Add IOwnableWritable and OwnableWritable on top of IERC173 and Ownable, to generalize authorization which was previously built into components
* Add Subsystem, which is a System, but uses OwnableWritable instead of Ownable
* Fix some import formatting since I was going through the whole package anyways

Edit: removed paragraph about subsystems since they're moved to #268

### DIFF
--- a/packages/solecs/src/Ownable.sol
+++ b/packages/solecs/src/Ownable.sol
@@ -8,8 +8,6 @@ import { OwnableStorage } from "@solidstate/contracts/access/ownable/OwnableStor
  * IERC173 implementation
  */
 contract Ownable is SolidStateOwnable {
-  using OwnableStorage for OwnableStorage.Layout;
-
   constructor() {
     // Initialize owner (SolidState has no constructors)
     OwnableStorage.layout().owner = msg.sender;
--- a/packages/solecs/src/Ownable.sol
+++ b/packages/solecs/src/Ownable.sol
@@ -1,0 +1,17 @@
+// SPDX-License-Identifier: Unlicense
+pragma solidity >=0.8.0;
+
+import { Ownable as SolidStateOwnable } from "@solidstate/contracts/access/ownable/Ownable.sol";
+import { OwnableStorage } from "@solidstate/contracts/access/ownable/OwnableStorage.sol";
+
+/**
+ * IERC173 implementation
+ */
+contract Ownable is SolidStateOwnable {
+  using OwnableStorage for OwnableStorage.Layout;
+
+  constructor() {
+    // Initialize owner (SolidState has no constructors)
+    OwnableStorage.layout().setOwner(msg.sender);
+  }
+}
--- a/packages/solecs/src/Ownable.sol
+++ b/packages/solecs/src/Ownable.sol
@@ -1,4 +1,4 @@
-// SPDX-License-Identifier: Unlicense
+// SPDX-License-Identifier: MIT
 pragma solidity >=0.8.0;
 
 import { Ownable as SolidStateOwnable } from "@solidstate/contracts/access/ownable/Ownable.sol";
--- a/packages/solecs/src/Ownable.sol
+++ b/packages/solecs/src/Ownable.sol
@@ -12,6 +12,6 @@ contract Ownable is SolidStateOwnable {
 
   constructor() {
     // Initialize owner (SolidState has no constructors)
-    OwnableStorage.layout().setOwner(msg.sender);
+    OwnableStorage.layout().owner = msg.sender;
   }
 }
--- a/packages/solecs/src/OwnableWritable.sol
+++ b/packages/solecs/src/OwnableWritable.sol
@@ -1,4 +1,4 @@
-// SPDX-License-Identifier: Unlicense
+// SPDX-License-Identifier: MIT
 pragma solidity >=0.8.0;
 
 import { IOwnableWritable } from "./interfaces/IOwnableWritable.sol";
--- a/packages/solecs/src/OwnableWritable.sol
+++ b/packages/solecs/src/OwnableWritable.sol
@@ -1,0 +1,45 @@
+// SPDX-License-Identifier: Unlicense
+pragma solidity >=0.8.0;
+
+import { IOwnableWritable } from "./interfaces/IOwnableWritable.sol";
+
+import { Ownable } from "./Ownable.sol";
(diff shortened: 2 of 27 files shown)


cha0sg0d:
cc @Kooshaba, this touches on some of your multiple system contract ideas i think,

alvrs:
1) What's the benefit of Subsystems over Libraries?
2) Could you go into some more detail behind the reasoning of introducing diamond storage (+ benefits, implications)?

Also, let's split up the Ownable refactor and the Subsystem changes into two separate PRs

dk1a:
Removed Subsystem from this PR, will elaborate on it in its own PR.

Diamond storage:

benefits:
- Libraries can directly and easily use diamond storage (something like #264). When storage slots are sequential, inheritance can make them unpredictable or difficult to work with.
- Having it out of namespace removes some ambiguity, like in constructor arguments (but that's just my personal preference).
- Upgradeability, but it is a very minor concern here, usually it's easier to just register new system/component over old one. And owner alone doesn't solve upgradeability issues anyways.

implications:
- Some project clutter, since diamond storage requires an extra file.
- Technically a breaking change (but won't affect most users, directly using _owner wasn't usually necessary, and owner(), transferOwnership() behave the same)
- Using owner directly in the contract will be fiddly, having to import the storage file (but as with the previous point, that should almost never be needed)


The first benefit applies even more so to the other state vars, especially in System: world, components. But that's for another PR (the 2 drafts I made)

dk1a:
Oh also removed the tests, they'll be together with Subsystem since they use it, unless it's rejected changing them back and forth seems unnecessary.

packages/solecs/src/OwnableWritable.sol
```diff
@@ -0,0 +1,45 @@
...
+  /** Whether given operator has write access */
+  function writeAccess(address operator) public view returns (bool) {
+    return OwnableWritableStorage.layout().writeAccess[operator] || operator == owner();
+  }
+
+  /** Revert if caller does not have write access to this component */
+  modifier onlyWriter() {
+    if (!writeAccess(msg.sender)) {
+      revert OwnableWritable__NotWriter();
+    }
```
alvrs, line 24: curious about your preference of `revert` over `require` - could you elaborate?

packages/solecs/src/OwnableWritableStorage.sol
```diff
@@ -0,0 +1,18 @@
+// SPDX-License-Identifier: Unlicense
```
alvrs, line 1: Could we change all the `Unlicense` declarations to `MIT`?

alvrs (commented):
Looking really good, thank you! Just had one minor comment and one question, and need to test locally.

alvrs:
Seems like @solidstate removed `setOwner` from `OwnableStorage` since v0.0.49: https://github.com/solidstate-network/solidstate-solidity/commit/7378a2c9e7ac19c4410bd1a46b9070e703c4e3e4. Should we upgrade to the latest version (0.0.52)?

packages/solecs/src/OwnableWritable.sol
dk1a: it's more about my preference of [custom errors](https://docs.soliditylang.org/en/latest/abi-spec.html#errors) over error strings. Custom errors just don't support require, unfortunately.
I took the specific semantics (Contract__ErrorName) from solidstate.

- Unlike strings, custom errors are guaranteed to be devoid of typos, since you can define them in one place, and use (via imports or inheritance) everywhere
- They allow args (like events), which is the most convenient way to do dynamic errors imo
- They aren't strings, so they should be more gas-efficient
- I just don't like the hardcoded nature of error-strings (this is similar to the 1st point, but about personal preference rather than some objective benefit)

imo it'd be great if mud started leaning towards these event-like errors. But if you ultimately prefer reverts, I can keep that in mind for future PRs

cha0sg0d: Seconding request for use of Custom Errors moving forward!

alvrs: Thanks for elaborating, I agree custom errors are a better approach!

alvrs (approved):
Looks great! We should make sure to mention the requirement to install `@solidstate/contracts` in consuming packages in the merge commit. Technically a breaking change, but not worth waiting for this change until the next breaking version imo.

packages/solecs/src/Ownable.sol
```diff
@@ -0,0 +1,15 @@
...
+import { Ownable as SolidStateOwnable } from "@solidstate/contracts/access/ownable/Ownable.sol";
+import { OwnableStorage } from "@solidstate/contracts/access/ownable/OwnableStorage.sol";
+
+/**
+ * IERC173 implementation
+ */
+contract Ownable is SolidStateOwnable {
+  constructor() {
+    // Initialize owner (SolidState has no constructors)
+    OwnableStorage.layout().owner = msg.sender;
```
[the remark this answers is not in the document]
dk1a: Because it's set in just 1 place and having an explicit method for it seemed unnecessary.
The method also emits an event.
Which is important, because after checking EIP-173, the event should be emitted on creation too.
Since it's merged I'll make a fix PR to use _setOwner, @holic thanks for noticing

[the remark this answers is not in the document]
dk1a: > Because it's set in just 1 place and having an explicit method for it seemed unnecessary.

By that I mean, _setOwner didn't exist in the previous version (I made the PR 2 months ago and recently updated solidstate), and when updating I didn't think to check for a method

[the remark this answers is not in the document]
dk1a: https://github.com/latticexyz/mud/pull/265#issuecomment-1372451237

and this comment is about another `setOwner` which was removed, it didn't emit an event

PR #266, opened 2022-11-26
feat: diamond storage for systems

dk1a:
Draft, mostly untested and doesn't address consequences (like in std-contracts)


alvrs:
Let's consolidate with #264 for now, as discussed in https://github.com/latticexyz/mud/pull/264#issuecomment-1372121908

PR #267, opened 2022-11-26
feat(solecs): add AbstractComponent

dk1a:
Depends on #265

Similar to #266 in terms of diamond storage-related breaking changes
(diamond storage stuff both here and in #266 is motivated by #264 and minor upgradeability concerns)

### DIFF
--- a/packages/solecs/src/AbstractComponent.sol
+++ b/packages/solecs/src/AbstractComponent.sol
@@ -1,0 +1,93 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
+
+import { IWorld } from "./interfaces/IWorld.sol";
+import { IComponent } from "./interfaces/IComponent.sol";
+
+import { OwnableWritable } from "./OwnableWritable.sol";
+
+/**
+ * Components are a key-value store from entity id to component value.
+ * They are registered in the World and register updates to their state in the World.
+ * They have an owner, who can grant write access to more addresses.
+ * (Systems that want to write to a component need to be given write access first.)
+ * Everyone has read access.
+ */
+abstract contract AbstractComponent is IComponent, OwnableWritable {
+  /** Reference to the World contract this component is registered in */
+  address public world;
+
+  /** Public identifier of this component */
+  uint256 public id;
+
+  error AbstractComponent__NotImplemented();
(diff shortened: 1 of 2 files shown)


alvrs:
If we want to move forward with this PR, could you elaborate on the use cases of `AbstractComponent` and `Uint256SetBareComponent`? And I think it would be good to split up the two main changes to reduce the surface of each PR.

dk1a:
I'll keep AbstractComponent here then, set comes later (since it kinda depends on abstract).
Abstract does 2 completely different things, which I should also probably separate:

1. diamond storage for component data. Reasons for that are in #264 #266. Except for components it's less relevant than systems. (I'm thinking this should be ditched from this PR)
2. removes the storage logic, i.e. doesn't have entityToValue or internal _add/_remove defined.

So far the only use case of undefined storage logic is that SetComponent has very custom stuff and doesn't (can't) use entityToValue at all. (which is why the 2 are in 1 PR)

dk1a:
Ok this is just a minimal non-breaking addition of AbstractComponent now.
Btw another benefit of Abstract is BareComponent will be easier to read. But most of that comes from #265, authorization was the big source of clutter

alvrs:
Just merged #265, so this can be rebased on main to reduce the diff

dk1a:
Rebased

alvrs:
One concern I have with encouraging components with custom storage layouts is that it might make it much more complicated to pick up the work on local tx simulation using a local evm (which requires mirroring relevant storage slots of the local evm using the known component state)

dk1a:
I suppose if you hardcode slots in the local evm that would be impossible with AbstractComponent, but I don't really know enough about it to suggest anything.
I certainly wouldn't encourage using AbstractComponent, I only use it for Uint256SetBareComponent because it just wouldn't work any other way (having BareComponent with its default slot unused seems like would just be much worse, for local evm as well).
AbstractComponent could also be used for optimization as I mention in #327, and that may be the use case you want to avoid depending on how you do local evm.
I suppose the main reason I want it in solecs, rather than just plopping SetComponent in my solecslib, is to avoid code repetition and inheritance compatibility issues.

dk1a:
Yep, same as #321

PR #268, opened 2022-11-26
Add subsystem

dk1a:
Depends on #265

> 1. What's the benefit of Subsystems over Libraries?

- Given contract size issues, internal libraries can't always be used. Public libraries can be troublesome during deployment. Whereas Subsystems share OwnableWritable with components, which allows them to have similar authorization logic in deployment.
- Dependency bloat. E.g. SkillSubsystem can have all the mana, effect etc components authorized for it. Then Systems which use SkillSubsystem need to authorize only SkillSubsystem, not all of its components.
- Clarity of responsibility. It can be rather unclear which components a library modifies. This kinda reiterates the 2nd point though.

Naturally subsystems are only needed for projects with many components and systems, and they don't negate the need for libraries either.
The way I see it is rather than ECS this is ECCM - entity component controller model, also with non-component-aware-models, and no actual systems. And in MUD a controller is called system, and model - subsystem. And libraries and non-component-aware-models are both called libraries.
(By non-component-aware I mean libraries that use components, but don't deal with authorization. It could just as well be diamond storage)

### DIFF
--- a/packages/solecs/README.md
+++ b/packages/solecs/README.md
@@ -77,7 +77,7 @@ Everyone has read access.
 For convenience the component implementation in `solecs` also includes a reverse mapping from `keccak256(value)` to set of entities with this value, but this is not strictly required and might change in a future release.
 
 ```solidity
-interface IComponent is IERC173 {
+interface IComponent is IOwnableWritable {
   /** Return the keys and value types of the schema of this component. */
   function getSchema() external pure returns (string[] memory keys, LibTypes.SchemaValue[] memory values);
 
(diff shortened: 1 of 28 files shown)


dk1a:
Subsystems arose from my game, which is close to 20 components now, and I only recently saw the need for them. So I agree that their usefulness is non-obvious, and I still reserve some concerns that maybe my project structure is just flawed.
Also `ERC1155BaseSystem` and `ERC721BaseSystem` in [@dk1a/solecslib](https://github.com/dk1a/solecslib/tree/main/contracts) seem like a good example of subsystems.

alvrs:
I see the use case of making systems modular (allowing to create higher level systems by combining lower level systems) but feel like we shouldn't add a new primitive (subsystems) to solve this issue. I opened a new discussion issue to think about use cases and possible solutions: https://github.com/latticexyz/mud/issues/319

PR #273, opened 2022-11-29
chore: update license declarations to MIT

### LINKED ISSUE #248: Update all license declarations (eg in sol files) to MIT
Some files still use `Unlicense` even though all of MUD is MIT licensed. We should update all license declarations to reflect that.

yoyoismee:
Update license declarations on 68 solidity files.

fixes #248


alvrs (approved):
Looks good, thank you! Just waiting for the CI to run (fixing it rn on #278), then good to be merged

PR #274, opened 2022-11-29
feat: twitter verification / linking behind a flag

authcall:
This essentially just allows you to now call `/Drip` without specifying a username. This same endpoint is the "follow-up" endpoint when using twitter verification. Similarly `/TimeUntilDrip` can be called with or without a Twitter username to return the time until the next drip


authcall:
Additional breaking changes: 
- Changes the drip balance to be computed in ETH instead of wei
- Max cap of faucet computed in ETH instead of wei
- Relay uses ETH instead of wei to compute min balance for message relay, for consistency

Most should not affect the client, but endpoint like `/MinBalanceForPush` on relay has a new format returning the min balance needed for message push in both ETH and wei

PR #276, opened 2022-12-01
fix(cli): mud trace bug for non-local networks

cha0sg0d:
The `cast run` call in `mud trace` was not using the `rpc` argument. It was defaulting to localhost:8545, which breaks for production chains.

This pr fixes that problem.


alvrs (approved):
lgtm, thanks!

alvrs (approved):

PR #283, opened 2022-12-07
feat(std-client): add more granular relationship utilities ⏳

Kooshaba:
the goal is to represent more complex ownership hierarchies. example: DAO -> Overworld Wallet -> Match Player -> Unit

new methods:
- `findComponentInRelationshipChain`: find any entity that has a specified component in the given entities relationship chain. example: find the player that owns this unit.
- `findInRelationshipChain`: find a specified entity in the relationship chain.


packages/std-client/src/utils.ts
```diff
@@ -117,16 +100,61 @@ export function resolveRelationshipChain(
...
+  entity: EntityIndex,
+  searchEntity: EntityIndex,
+  world: World,
+  relationshipComponent: Component<{ value: Type.Entity }, { contractId: string }>
+): EntityIndex | undefined {
+  if (entity === searchEntity) return entity;
+
+  while (hasComponent(relationshipComponent, entity)) {
+    const entityValue = world.entityToIndex.get(getComponentValueStrict(relationshipComponent, entity).value);
+    if (!entityValue) return;
```
alvrs, line 132: same as above, `entityValue` could be 0, so we should do a null check instead
```suggestion
    if (entityValue == null) return;
```

packages/std-client/src/utils.ts
```diff
@@ -117,16 +100,61 @@ export function resolveRelationshipChain(
...
+export function isOwnedByCaller(
+  world: World,
+  ownedByComponent: Component<{ value: Type.Entity }, { contractId: string }>,
+  entity: EntityIndex,
+  playerEntity: EntityIndex
+): boolean {
+  const foundPlayer = findInRelationshipChain(entity, playerEntity, world, ownedByComponent);
+  return foundPlayer === playerEntity;
+}
+
```
alvrs, line 150: this is the usage I meant above; is there another use case for `findInRelationshipChain` that requires the entity index to be returned instead of a boolean?

packages/std-client/src/utils.ts
```diff
@@ -117,16 +99,60 @@ export function resolveRelationshipChain(
   return entity;
 }
 
+export function findEntityWithComponentInRelationshipChain(
```
alvrs, line 102: can we add a short code documentation comment on top of this and `findInRelationshipChain` to explain what it does and how to use it? (then it will show up here: https://mud.dev/packages/std-client/api and be discoverable via the documentation search bar)

packages/std-client/src/utils.ts
```diff
@@ -117,16 +99,60 @@ export function resolveRelationshipChain(
...
+
+export function isOwnedByCaller(
+  world: World,
+  ownedByComponent: Component<{ value: Type.Entity }, { contractId: string }>,
+  entity: EntityIndex,
+  playerEntity: EntityIndex
+): boolean {
+  return findInRelationshipChain(entity, playerEntity, world, ownedByComponent);
+}
+
```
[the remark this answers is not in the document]
Kooshaba: yeah true i've actually defined a custom one in sky strife. honestly a lot of std-client stuff is feeling very user specific at this point

packages/std-client/src/utils.ts
Kooshaba: for sure!

alvrs (approved):
lgtm! Great comments!

PR #286, opened 2022-12-09
fix: separate component registration from components

dk1a:
@alvrs This is a fix for the component registration issue I mentioned in discord, where you can't deploy a new component with an existing ID (RegisterSystem expects the owner of the old component to be the new component, instead of the deployer).

2 and 3 don't have to be in the same PR as 1, but they're very related breaking changes that would go well together imo.

1. Make world.registerComponent call external, like with systems. Remove registerWorld from components.
2. id in Component constructor only existed to support registerWorld, remove it as well. This simplifies the constructor and the component state. Id can still be obtained via `getIdByAddress` when needed.
3. In Component's constructor change `world` from `address` to `IWorld`. I assume it was `address` because some components (like world's registries) had to have world separately initialized, and the constructor could receive `address(0)`. That is no longer the case, and component constructor now treats world exactly the same as system constructor. This also simplifies Component creation (all the places I've seen had to cast IWorld to address).

World.init, std-contracts, tests, LibDeploy.ejs are changed to support the new constructor and registration.
I don't think this breaks any other packages, but I may have missed something.
This is naturally a breaking change for any project that implements or deploys components

### DIFF
--- a/packages/cli/src/contracts/LibDeploy.ejs
+++ b/packages/cli/src/contracts/LibDeploy.ejs
@@ -52,7 +52,8 @@ library LibDeploy {
 <% components.forEach(component => { -%>
 
       console.log("Deploying <%= component %>");
-      comp = new <%= component %>(address(result.world));
+      comp = new <%= component %>(result.world);
+      world.registerComponent(address(comp), <%= component %>ID);
       console.log(address(comp));
 <% });-%>
     } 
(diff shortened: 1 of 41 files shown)


alvrs (commented):
I haven't reviewed in depth yet, but at first glance all the changes look reasonable. Since it's a big breaking change I'd propose postponing for now and then merging it into the upcoming `v2` branch we'll start working on soon.

dk1a:
Sure, closed it for now, will look out for v2 branch

PR #288, opened 2022-12-12
fix: add LoadingState component from SyncWorker

holic:
Without this, your console will see a bunch of logs like this:

image

### DIFF
--- a/packages/std-client/src/setup/setupMUDNetwork.ts
+++ b/packages/std-client/src/setup/setupMUDNetwork.ts
@@ -9,7 +9,7 @@ import {
   InputType,
 } from "@latticexyz/network";
 import { BehaviorSubject, concatMap, from, Subject } from "rxjs";
-import { World } from "@latticexyz/recs";
+import { defineComponent, Type, World } from "@latticexyz/recs";
 import { computed } from "mobx";
 import { keccak256 } from "@latticexyz/utils";
 import { Contract, ContractInterface } from "ethers";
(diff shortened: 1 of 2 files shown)


holic:
Separately: I wonder if it's worth expanding the `Component` interface to include a `name` that is only set at deploy time and can be later looked up, so that we can make thees "unknown component" messages a little more user friendly

dk1a:
> Separately: I wonder if it's worth expanding the `Component` interface to include a `name` that is only set at deploy time and can be later looked up, so that we can make thees "unknown component" messages a little more user friendly.
> 
> Or maybe this is something we can add to the registry and do it for users (e.g. infer the name based on contract name).

I've been thinking about something similar from the contract side recently. As you said the problem with World is it can't know names without some `name` func - it only gets a hashed uint256 and an address.

Instead of changing `Component` you could change registration process itself to have an ID=>name mapping e.g.:
(I'm using this for other hashed entities, not component IDs, just thought I'd share it since it's relevant)

```solidity
contract ReverseHashNameSystem is System {
  constructor(IWorld _world, address _components) System(_world, _components) {}

  function executeTyped(
    string memory name
  ) public returns (uint256) {
    return abi.decode(
      execute(abi.encode(name)),
      (uint256)
    );
  }

  /**
   * @dev Permissionlessly map a hashed entity to its string name.
   * Reverts on invalid UTF-8.
   * Returns the character count of the name.
   */
  function execute(bytes memory args) public override returns (bytes memory) {
    (string memory name) = abi.decode(args, (string));

    ReverseHashNameComponent comp
      = ReverseHashNameComponent(getAddressById(components, ReverseHashNameComponentID));

    // reverts if UTF-8 is invalid
    uint256 charCount = toSlice(name).chars().count();

    uint256 entity = uint256(keccak256(bytes(name)));
    comp.set(entity, name);

    return abi.encode(charCount);
  }
}
```

alvrs (approved):
lgtm!

PR #289, opened 2022-12-12
feat: bump devnode gas limit to 100m

holic:
Lattice/MUD testnet runs with 100m gas limit, so this bumps the local devnode testnet to match so we can do a little more work when e.g. initializing a world

We should also make this configurable via CLI!

Open question around what we want our default to be, or put another way, what chains we want MUD dev to be compatible with and their respective gas limits. Since we haven't done a ton of gas golfing, I assume we prob won't be doing much with MUD on mainnet for now?

### DIFF
--- a/packages/cli/src/commands/devnode.ts
+++ b/packages/cli/src/commands/devnode.ts
@@ -21,7 +21,14 @@ export const handler = async (argv: Arguments<Options>): Promise<void> => {
   console.log("Clearing devnode history");
   const userHomeDir = homedir();
   rmSync(path.join(userHomeDir, ".foundry", "anvil", "tmp"), { recursive: true, force: true });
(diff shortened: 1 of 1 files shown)


alvrs (approved):
lgtm! to your question - mainnet is probably infeasible for now, but rollups on mainnet are feasible (eg https://words3.xyz/, built with MUD, running on Optimism Mainnet)

holic:
> lgtm! to your question - mainnet is probably infeasible for now, but rollups on mainnet are feasible (eg https://words3.xyz/, built with MUD, running on Optimism Mainnet)

I guess what I am wondering is if 100m gas limit is problematic as a default, because people will build things expecting it to deploy fine/within the gas limit (because it's 100m locally) but then have trouble when doing it on a rollup/mainnet.

Went ahead and merged this for now and will revisit soon to make this configurable for e.g. the tutorial (or improve the tutorial gas).

PR #290, opened 2022-12-12
fix: new entities should be included in overrides

holic:
Before this change, creating new entities with overrides on the client (for optimistic rendering) resulted in ~no visible changes.

### DIFF
--- a/packages/recs/src/Component.ts
+++ b/packages/recs/src/Component.ts
@@ -374,7 +374,7 @@ export function overridableComponent<S extends Schema, M extends Metadata, T = u
       if (prop === "removeOverride") return removeOverride;
       if (prop === "values") return valuesProxy;
       if (prop === "update$") return update$;
-      if (prop === "entities") return () => [...overriddenEntityValues.keys(), ...target.entities()];
+      if (prop === "entities") return () => new Set([...overriddenEntityValues.keys(), ...target.entities()]).values();
 
       return Reflect.get(target, prop);
     },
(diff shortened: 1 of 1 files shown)


packages/recs/src/Component.ts
```diff
@@ -374,6 +374,7 @@ export function overridableComponent<S extends Schema, M extends Metadata, T = u
       if (prop === "removeOverride") return removeOverride;
       if (prop === "values") return valuesProxy;
       if (prop === "update$") return update$;
+      if (prop === "entities") return () => [...overriddenEntityValues.keys(), ...target.entities()];
```
alvrs, line 377: maybe we could use `concatIterators(overriddenEntityValues.keys(), target.entities())` here (from `@latticexyz/utils`) to avoid spreading the values intro an array?

holic: This gives type errors elsewhere. We're already using the spread operator above in a similar way. I think this is fine?

Is there something about spread you're worried about?

alvrs: I was worried about the performance overhead (O(n) for spreading vs O(1) for creating a new iterator), but you're raising a valid point - above we have to spread into a set to avoid duplicate entities (`overriddenEntityValues.keys()` might contain keys also included in `target.entities()`), I think we'll have to do the same here

```suggestion
      if (prop === "entities") return () => new Set([...overriddenEntityValues.keys(), ...target.entities()]).values();
```

alvrs (approved):
lgtm!

PR #291, opened 2022-12-12
fix: cli issue with circular dependencies

holic:
Hot reloading would fail when trying to generate a deploy script if you had a circular dependency in Solidity (A.sol imports B.sol and B.sol imports A.sol). Since Solidity compiles circular dependencies fine, I felt like it was probably best to just fix the module resolution here (with a warning) rather than preventing circular deps.


alvrs (approved):
lgtm!

PR #292, opened 2022-12-12
fix(cli): reset LibDeploy.sol using original/cached contents

holic:
Running the dev scripts would delete/overwrite the `LibDeploy.sol` file, which kept later parts of the hot reloading from continuing (e.g. couldn't find deleted file). This may only be visible with linked MUD deps, didn't test with a regular npm install.

### DIFF
--- a/packages/cli/src/utils/codegen.ts
+++ b/packages/cli/src/utils/codegen.ts
@@ -4,7 +4,7 @@ import path from "path";
 
 const contractsDir = path.join(__dirname, "../../src/contracts");
 
-const originalLibDeploy = readFile(path.join(contractsDir, "LibDeploy.sol"));
+const stubLibDeploy = readFile(path.join(contractsDir, "LibDeploy.sol"));
 
 export async function generateLibDeploy(configPath: string, out: string, systems?: string | string[]) {
   // Parse config
@@ -28,5 +28,5 @@ export async function generateLibDeploy(configPath: string, out: string, systems
 }
 
 export async function resetLibDeploy(out: string) {
-  await writeFile(path.join(out, "LibDeploy.sol"), await originalLibDeploy);
+  await writeFile(path.join(out, "LibDeploy.sol"), await stubLibDeploy);
 }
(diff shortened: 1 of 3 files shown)


packages/cli/src/utils/codegen.ts
```diff
@@ -20,11 +22,11 @@ export async function generateLibDeploy(configPath: string, out: string, systems
...
   const libDeployPath = path.join(out, "LibDeploy.sol");
-  writeFileSync(libDeployPath, LibDeploy);
+  await writeFile(libDeployPath, LibDeploy);
 
   return libDeployPath;
 }
 
 export async function resetLibDeploy(out: string) {
-  writeFileSync(path.join(out, "LibDeploy.sol"), readFileSync(path.join(contractsDir, "LibDeploy.sol")));
+  await writeFile(path.join(out, "LibDeploy.sol"), await originalLibDeploy);
```
alvrs, line 31: I'm a bit wary of caching the original LibDeploy and restoring it here - it might lead folks to believe they can change the generated LibDeploy.sol manually and then confuse them about why their changes are not actually picked up. For this reason I'd prefer restoring the "stub version" of LibDeploy (https://github.com/latticexyz/mud/blob/main/packages/cli/src/contracts/LibDeploy.sol) after running the script, which should also make keep imports/hot reload working, but include a fat warning about it being a stub

holic: > For this reason I'd prefer restoring the "stub version" of LibDeploy (https://github.com/latticexyz/mud/blob/main/packages/cli/src/contracts/LibDeploy.sol) 

This _is_ what it's doing.

Maybe you could spin this up locally to see what I am seeing? I started with the mud-template-minimal, linked mud deps locally (including CLI package), and when I went to make any changes to contracts, `LibDeploy.sol` would get overwritten or deleted, thus breaking hot reloading for any future changes.

alvrs: Ohh I see, `originalLibDeploy` actually refers to the stub in the cli package. When reading this yesterday I thought it used the `LibDeploy.sol` file in the user directory and restoring it after deploying.

Could we rename the variable to `stubLibDeploy` or something to make it obvious even to devs reading this late at night? 😄 (This was already unclear before your change, good call to move reading the file outside of the reset function!)

alvrs (approved):
lgtm!

alvrs (approved):

PR #293, opened 2022-12-12
refactor: move worker logs to debug

holic:
Worker logs are really noisy and get in the way of my own app's logs, so this PR moves them into the `debug` library, which namespaces them and makes them opt-in (with the "Verbose" log level and namespace logging toggled with e.g. `localStorage.debug = mud.*`)

image

I'll plan to move more things to this in the future.

### DIFF
--- a/packages/network/package.json
+++ b/packages/network/package.json
@@ -24,6 +24,25 @@
     "build": "rimraf dist && rimraf ./.parcel-cache && yarn prepack && parcel build && tsc --emitDeclarationOnly; yarn postpack",
     "release": "npm publish || echo 'version already published'"
   },
+  "dependencies": {
+    "debug": "^4.3.4"
+  },
+  "peerDependencies": {
+    "@ethersproject/providers": "^5.6.1",
+    "@improbable-eng/grpc-web": "^0.15.0",
+    "@latticexyz/recs": "^1.28.1",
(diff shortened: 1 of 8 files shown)


packages/network/package.json
```diff
@@ -24,6 +24,25 @@
...
+    "@latticexyz/utils": "^1.28.1",
+    "@protobuf-ts/grpcweb-transport": "^2.7.0",
+    "async-mutex": "^0.3.2",
+    "ethers": "^5.7.1",
+    "mobx": "^6.5.0",
+    "nice-grpc-web": "^2.0.0",
+    "observable-webworker": "^4.0.1",
+    "rxjs": "^7.5.5",
+    "threads": "^1.7.0"
+  },
```
holic, line 45: These haven't changed, just moving them up above `devDependencies` for clarity

packages/network/src/workers/debug.ts
```diff
@@ -0,0 +1,3 @@
+import { debug as parentDebug } from "../debug";
+
+export const debug = parentDebug.extend("workers");
```
holic, line 3: this is fairly granular and maybe a little overkill but I figured it'd be a better starting place to expand on

alvrs (approved):
much better!

PR #294, opened 2022-12-12
feat(react): add react package

holic:
Adds official React MUD package! This will replace the React usages sprinkled into other packages (I'll do a clean up in a breaking follow up).


packages/mud-react/package.json
```diff
@@ -0,0 +1,72 @@
+{
+  "name": "@latticexyz/mud-react",
```
holic, line 2: wasn't sure what to call this! `@latticexyz/react` felt incorrect and I think what I actually want is something like `@mud/react`

packages/std-client/src/hooks.ts
```diff
@@ -3,6 +3,8 @@ import { ObservableSet } from "mobx";
 import { useEffect, useMemo, useState } from "react";
 import { filter, Observable } from "rxjs";
 
+// Deprecated! Use hooks exported from @latticexyz/react package instead.
+
```
alvrs, line 7:
```suggestion
// @deprecated Use hooks exported from @latticexyz/react package instead.
```

Using the official tag to integrate with vscode and others (vscode highlights usages of deprecated functions by striking them through)

packages/react/src/useDeprecatedComputedValue.ts
```diff
@@ -0,0 +1,15 @@
+import { IComputedValue } from "mobx";
+import { useEffect, useState } from "react";
+
+// TODO: migrate the rest of the mobx stuff to rxjs
```
alvrs, line 4: Opened an issue to track this, let's add a link to the code comment: #339

packages/react/src/useComponentValue.ts
```diff
@@ -0,0 +1,33 @@
...
+  entityIndex: EntityIndex | undefined,
+  component: Component<S, Metadata, undefined>
+): ComponentValue<S> | undefined;
+
+export function useComponentValue<S extends Schema>(
+  entityIndex: EntityIndex | undefined,
+  component: Component<S, Metadata, undefined>,
+  defaultValue?: ComponentValue<S>
+) {
+  const entities = useEntityQuery([Has(component)]);
```
alvrs, line 20: would this hook cause a react rerender every time any value in the component changes or only if the value of the entity specified in the `entityIndex` param changes?

alvrs (commented):
Looking good! 

Let's add a short `README.md` - this show up on npm's package overview page when releasing this package (like here: https://www.npmjs.com/package/@latticexyz/solecs)

packages/react/src/useComponentValue.ts
holic: Ohh this is a great point. It'd currently reevaluate for anything matching the query. I may rewrite this to not depend on `useEntityQuery` and be more mindful of rendering only when the single entity value changes.

packages/react/README.md
```diff
@@ -0,0 +1,20 @@
+# MUD React
+
+React hooks (and more) for building MUD clients.
+
+### useComponentValue
+
+Returns the value of the component for the entity, and triggers a re-render as the component is added/removed/updated.
+
+```
+const position = useComponentValue(entity, Position);
```
holic, line 10: While building Emojimon, I was thinking about ECS in terms of entity-first, where "entity" is often the thing to key off of, and components are types of values on an entity.

However, I noticed our internal APIs (`setComponent`, `removeComponent`), etc. all have components first, then entities. I wonder if this is because we internally model a component as a mapping of entities <> values.

How do you feel about the position of the arguments here?

holic:
Quickly wired this into Emojimon to make sure it works and to see what the diff looks like:
https://github.com/latticexyz/emojimon/commit/74629a87e291b21268969d981a20b46044f4028f

packages/react/README.md
alvrs: Yes that's mostly the reason - when thinking of components as tables and entities as rows in that table, it felt more natural to first specify which table to modify and then which row in that table. In general I don't feel strongly about the order of arguments, can totally see how putting entity first is more intuitive when thinking about entities as objects and components as parameters on that object. My main argument for changing the order here to also put component first would be consistency with existing functions.

packages/react/package.json
```diff
@@ -0,0 +1,55 @@
...
+    "@rollup/plugin-node-resolve": "^13.1.3",
+    "@rollup/plugin-typescript": "^8.3.1",
+    "@testing-library/react-hooks": "^8.0.1",
+    "@types/react": "^18.0.12",
+    "eslint-plugin-react": "^7.31.11",
+    "eslint-plugin-react-hooks": "^4.6.0",
+    "jest": "^29.3.1",
+    "mobx": "^6.4.2",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0",
```
alvrs, line 35: is `react-dom` also required as peer dependency?

packages/react/src/useComponentValue.test.ts
```diff
@@ -0,0 +1,88 @@
...
+    });
+    expect(result.current).toEqual({ x: 0, y: 0 });
+
+    act(() => {
+      removeComponent(Position, entity);
+    });
+    expect(result.current).toBe(undefined);
+  });
+
+  it("should re-render only when Position changes for entity", () => {
```
alvrs, line 44: love these tests!! 🔥

packages/react/src/useComponentValue.ts
```diff
@@ -0,0 +1,48 @@
...
+    setValue(entityIndex != null ? getComponentValue(component, entityIndex) : undefined);
+    if (entityIndex == null) return;
+
+    const queryResult = defineQuery([Has(component)], { runOnInit: false });
+    const subscription = queryResult.update$.subscribe((update) => {
+      if (update.component === component && update.entity === entityIndex) {
+        const [nextValue, prevValue] = update.value;
+        // TODO: fix types in `defineQuery` so we don't have to cast this
+        setValue(nextValue as ComponentValue<S>);
+      }
```
alvrs, line 42: We can use the `isComponentUpdate` type guard to fix the type
```suggestion
      if (isComponentUpdate(update, component) && update.entity === entityIndex) {
        const [nextValue] = update.value;
        setValue(nextValue);
      }
```

We could also try some typescript magic in `defineQuery` to return a component update typed to any of the components passed in via the fragments

packages/react/tsconfig.json
```diff
@@ -0,0 +1,102 @@
...
+    // "noImplicitOverride": true,                       /* Ensure overriding members in derived classes are marked with an override modifier. */
+    // "noPropertyAccessFromIndexSignature": true,       /* Enforces using indexed accessors for keys declared using an indexed type */
+    // "allowUnusedLabels": true,                        /* Disable error reporting for unused labels. */
+    // "allowUnreachableCode": true,                     /* Disable error reporting for unreachable code. */
+
+    /* Completeness */
+    // "skipDefaultLibCheck": true,                      /* Skip type checking .d.ts files that are included with TypeScript. */
+    "skipLibCheck": true /* Skip type checking all .d.ts files. */
+  },
+  "exclude": ["dist", "**/*.test.ts", "packages"]
```
alvrs, line 101: ooc why do we need to exclude `packages` here?

alvrs (commented):
Looks great! Just a some minor comments and a minor preference for changing the order of parameters for consistency.

alvrs:
Should we add you as the codeowner for this package to auto-request reviews from? https://github.com/latticexyz/mud/blob/main/CODEOWNERS (Can also set the codeowner for `ecs-browser` to kooshaba while at it)

packages/react/README.md
holic: Chatted with @alvrs some more and here's what I got out of the convo:

Data modeling with MUD can be thought of in one of two ways
- Entity-first, where the entity is the focus and data (components/values) are "attached" to an entity. I was thinking about it like this, where my "get component value" operation is more like a "get this value of this key from the entity object"
- Component/Table-first, where it's more like a relational database and the entity ID is a primary key. In the future, we may want more complicated keys (tuples) that would expand past just an entity ID. So asking for a component value is more like "get the value in the component table for this primary key aka entity ID".

Sounds like we want to guide folks towards the latter mental model, esp. as we push further into SQL-like capability.

Will update these hooks!

packages/react/package.json
holic: nope! will remove

packages/react/tsconfig.json
holic: it's not needed, this was just copy pasta

alvrs (approved):
Amazing! 🚢

PR #296, opened 2022-12-13
docs: add intro to MUD

alvrs:


guides/introduction.md
```diff
@@ -0,0 +1,177 @@
...
+
+  function getValue(uint256 entity) external view returns (uint256);
+
+  function getEntitiesWithValue(uint256 value) external view returns (uint256[] memory);
+}
+
+```
+
+Components don't contain any logic besides the logic required to store and load component values.
+All application logic lives in systems, which don't contain any persistent state but work with state stored in components.
```
MarcoWorms, line 88:
```suggestion
All application logic lives in systems with no persistent state, and they work with state stored in components.
```

guides/introduction.md
```diff
@@ -0,0 +1,177 @@
...
+
+Note how multiple systems can share the same state and how there is no requirement for the systems and components to be deployed by the same developer.
+
+At the center of every MUD application is the World contract.
+The World is a global namespace, in which components and systems are registered with a fixed id.
+The World contract is permissionless - anyone can register additional components and systems.
+This is the most notable difference between this architecture and existing ones like the [Diamond pattern](https://eips.ethereum.org/EIPS/eip-2535).
+It is possible because state is encapsulated in separate components contracts with individual write access.
+
+When a component value is changed, the update is registered on the World contract (via `registerComponentValueSet`).
```
MarcoWorms, line 101:
```suggestion
When a component value changes, the update is registered on the World contract (via `registerComponentValueSet`).
```

guides/introduction.md
```diff
@@ -0,0 +1,177 @@
...
+This causes the World contract to emit an event (`ComponentValueSet`), which can be used by clients and indexers to mirror the entire World state.
```
MarcoWorms, line 102:
```suggestion
This causes the World contract to emit an event (`ComponentValueSet`), which clients and indexers can use to mirror the entire World state.
```

guides/introduction.md
```diff
@@ -0,0 +1,177 @@
...
+    },
+  }),
+};
+
+const network = await setupMUDNetwork<typeof components, SystemTypes>(config, world, components, SystemAbis);
+```
+
+Note: the on-chain component is always considered the ground truth.
+Setting a value on-chain will also update the local component, but setting a local value won't automatically update the on-chain value.
+Instead, the client should call an on-chain system, which modifies the on-chain component and thereby triggers an update of the local component.
```
MarcoWorms, line 151:
```suggestion
Instead, the client should call an on-chain system, which modifies the on-chain component and triggers an update of the local component.
```

guides/introduction.md
```diff
@@ -0,0 +1,177 @@
...
+
+## Further reading
+
+MUD contains a few other packages besides the ones mentioned in this overview (SOLECS, Network, Services, RECS):
+
+- `cli`: The MUD command line interface contains scripts to make the life of developers easier when working with MUD - scaffolding new projects, managing the development environment, contract hot reloading, contract deployment, type generation, etc.
+- `ecs-browser`: A data explorer to visualize and edit on-chain and local component values - made possible by storing state in a standardized and introspectable way.
+- `std-contracts`/`std-client`: MUD standard libraries for contract and client development.
+
+Now that you have a basic understanding of the core of MUD and the principles behind it, let's dive into a MUD project.
```
MarcoWorms, line 177:
```suggestion
Now that you have a basic understanding of the core of MUD and its principles, let's dive into a MUD project.
```

guides/introduction.md
```diff
@@ -0,0 +1,177 @@
+# Introduction
```
Kooshaba, line 1: it could be cool to include a gif of your increment system with hot reloading here (im a fan of gifs as the first thing people see)

PR #297, opened 2022-12-15
fix(phaserx): update tween references after phaser bump

holic:


packages/phaserx/src/utils/tween.ts
```diff
@@ -31,10 +31,10 @@ export async function tween(config: TweenBuilderConfig, options?: { keepExisting
...
 
 export function removeAllTweens(gameObject: Phaser.GameObjects.GameObject) {
-  // Access some hidden tween manager properties because tweenManager.getTweensOf does not return tweens in _add
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  const tweenManager = gameObject.scene.tweens as any;
-  if (!tweenManager) return;
-  const allTweens = [...tweenManager._add, ...tweenManager._pending, ...tweenManager._active] as Phaser.Tweens.Tween[];
-  for (const tween of allTweens) if (tween.hasTarget(gameObject)) tween.stop();
+  const tweenManager = gameObject.scene.tweens;
+  for (const tween of tweenManager.tweens) {
```
holic, line 35: image

Not sure if this includes all the tweens (add, pending, active) but I assume it would given the description?

Kooshaba:
this fixed the tween errors, but I'm getting visual artifacts on the background image now. it's not the end of the world and I will ask in phaser discord if anyone else has gotten this.

Kooshaba:
seems to be related to zoom levels. i should just do the zoom hacks that ember does and this might go away.

Kooshaba (approved):

packages/phaserx/src/utils/tween.ts
alvrs: the previous implementation was there due to a very obscure and hard to reproduce bug in Ember, where creatures were flying over the map (while they where not supposed to). Our assumption was that the game objects still had active tweens when they where reused for another entity (as part of the object pool), and executed the previous stale tween before the next one. This seemed to have been fixed by manually removing stopping tweens in all sub arrays (`_add`, `_pending`, `_active`).

It seems like these internal subarrays don't exist anymore in the tween manager, so I'd assume they're all included in that tween array.

Down to merge this change, just putting this context here in case the flying creatures appear again.

Also, maybe we don't need this custom logic anymore and could use `getTweensOf` from the tween manager? https://github.com/photonstorm/phaser/blob/v3.60.0-beta.10/src/tweens/TweenManager.js#L855

alvrs (approved):

PR #299, opened 2022-12-15
fix(std-client): allow default components to be passed in to setupMUDNetwork

alvrs:
* before #288 users had to manually define the `LoadingState` component. Since #288 `LoadingState` is created by default in `setupMUDNetwork`, overriding any user-defined `LoadingState` component
* this caused previous clients who had a custom `LoadingState` component and used it for a loading screen to get stuck in that loading screen because their component wasn't updated anymore
* With this PRs components passed in by users get priority over default components (eg if a component with contract id `component.LoadingState` is passed in, it will be used as `LoadingState` component instead of creating a new one)

### DIFF
--- a/packages/std-client/src/setup/setupMUDNetwork.ts
+++ b/packages/std-client/src/setup/setupMUDNetwork.ts
@@ -176,7 +176,17 @@ function findOrDefineComponent<Cs extends ContractComponents, C extends Contract
   components: Cs,
   component: C
 ): C {
-  return (
-    (Object.values(components).find((c) => c.metadata.contractId === component.metadata.contractId) as C) ?? component
-  );
+  const existingComponent = Object.values(components).find(
+    (c) => c.metadata.contractId === component.metadata.contractId
+  ) as C;
+
+  if (existingComponent) {
+    console.warn(
+      "Component with contract id",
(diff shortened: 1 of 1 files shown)


packages/std-client/src/setup/setupMUDNetwork.ts
```diff
@@ -156,3 +165,18 @@ export async function setupMUDNetwork<C extends ContractComponents, SystemTypes
...
+ * @returns component if it exists in components object, otherwise the component passed in
+ */
+function findOrDefineComponent<Cs extends ContractComponents, C extends ContractComponent>(
+  components: Cs,
+  component: C
+): C {
+  return (
+    (Object.values(components).find((c) => c.metadata.contractId === component.metadata.contractId) as C) ?? component
+  );
+}
```
alvrs, line 182: this approach is needed to enforce the component types (ie `SystemRegistry` needs to have the `SystemRegistry` type, a general `C[keyof C]` is not strict enough)

holic (dismissed):
Any thoughts on adding a `console.warn` when we detect a component is already defined but we want to include the built-in one?

Feels like an anti-pattern to have these maybe defined twice?

holic (approved):
🙏

PR #300, opened 2022-12-16
fix(cli): better logs, more resilience, better gas price mgmt

alvrs:
* add better logs to `deploy` command (forwarding forge errors)
* add manual gas price check (forge was using a way too high gas price when deploying to the lattice testnet, so the balance requirements where way too high)
* fix a bug where LibDeploy would be overridden during deploy and not reset afterwards (missing `await`)

### DIFF
--- a/packages/cli/src/contracts/LibDeployStub.sol
+++ b/packages/cli/src/contracts/LibDeployStub.sol
@@ -1,0 +1,23 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
+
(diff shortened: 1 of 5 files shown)


packages/cli/src/contracts/LibDeployStub.sol
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: MIT
```
alvrs, line 1: Moving the "source of truth" of the LibDeploy.sol stub to its own file to avoid overriding it in during the deploy process.

Kooshaba (approved):
🚢

PR #301, opened 2022-12-19
feat(cli): add support for subsystems in writeAccess to LibDeploy

dk1a:
Depends on #265. Technically doesn't depend on #268, they can be merged in any order.
Allows subsystems to be used in writeAccess of deploy.json
Backwards-compatible


alvrs:
Closing for now until we agreed on a solution to the root issue of making systems more modular (discussion in #319), see https://github.com/latticexyz/mud/pull/268#issuecomment-1372252545

PR #309, opened 2023-01-02
feat(cli): allow arbitrary project structure for LibDeploy and system types

dk1a:
#311 would help with highlighting for Vm, but isn't required.

Removes the need for systems to be in `src/systems` or components in `src/components`.
`src` is taken from foundry.toml, names are taken from deploy.json, and then relative paths are built accordingly.
The new requirement is to just not have 2 identically named files (in different folders) for 1 system/component.
This should work fine with most (probably all) existing project structures, but also allows changing them.

The main things that changed are:
- `generateLibDeploy` and its dependents (like codegen-libdeploy, deploy-contracts, deploy, test)
- `generateSystemTypes` and its dependents (types, system-types, deploy-contracts, deploy)
- test's `handler` and `generateAndDeploy` generate LibDeploy in `test` dir from forgeConfig (which by default is `test`, not `src/test`. Not really a breaking change, but a change. Is it bad?)
- `deploy` copies `Deploy.sol` script to `test` dir along with LibDeploy.

TODO for a followup:
- `hsr.ts`: tracking uses `system` folder only, so only works with the old structure; default root should come from getForgeConfig instead of a hardcoded argument
- `trace.ts`: hardcoded `/src/components/` etc


dk1a:
I'll be testing it for a while longer, but I removed dependencies, it seems to work, and should be ready for a general review

dk1a:
added support for initializer libs,
improved Deploy.sol generation and made it work better with tests

[another emojimon branch](https://github.com/dk1a/emojimon/tree/18d12df918685416a023fb3b3cb49b22bf04ca19) to test that it works with current projects (gotta link it locally). Basically the only things that would change are local LibDeploy+Deploy

dk1a:
Closing for now, parts of it will be in a separate PR, the rest will PR to v2

PR #311, opened 2023-01-03
build: update forge-std dependency

dk1a:
While working on #309 I found the very old (and disjointed across .vscode's settings, global node_modules, and cli's node_modules) forge-std to be an issue, so this updates and syncs it.

There're some minimal cli changes mostly to use Script instead of DSTest, but do tell if you want that in a separate PR.


dk1a:
rebased to incorporate #357, and updated forge-std to latest (`StdInvariant` just got added to `Test` base)

alvrs (approved):
lgtm

PR #312, opened 2023-01-03
fix(cli): do not copy System test ABIs during build 🧱

Kooshaba:
i thought this was fixed at some point, but it's still happening for me. lmk if there is a better way to fix this.


alvrs (approved):
lgtm 🚢

PR #313, opened 2023-01-03
feat(cli): add deploy option to specify whether dev flag should be appended to client url

Kooshaba:
this became very annoying once we switched to using actual gas fees on the testnet. defaults to dev=true.


alvrs (approved):
lgtm 🚢

alvrs (approved):

PR #321, opened 2023-01-05
feat(std-contracts): add Uint256SetBareComponent

dk1a:
Follow-up to #267

SetComponent is useful for some many-to-many relationships
(they can also be solved via just more components, but that can get very inconvenient and complicated very fast)

Examples of how I use it:
[LearnedSkillsComponent](https://github.com/dk1a/wanderer-cycle/blob/8b4cdc32d17e711a3f60162a59e23057e1a64510/packages/contracts/src/skill/LearnedSkillsComponent.sol)
[GuiseSkillsComponent](https://github.com/dk1a/wanderer-cycle/blob/8b4cdc32d17e711a3f60162a59e23057e1a64510/packages/contracts/src/guise/GuiseSkillsComponent.sol)
[EquipmentSlotAllowedComponent](https://github.com/dk1a/wanderer-cycle/blob/8b4cdc32d17e711a3f60162a59e23057e1a64510/packages/contracts/src/equipment/EquipmentSlotAllowedComponent.sol)
[AffixAvailabilityComponent](https://github.com/dk1a/wanderer-cycle/blob/8b4cdc32d17e711a3f60162a59e23057e1a64510/packages/contracts/src/affix/AffixAvailabilityComponent.sol)

With skills for example, a set component is handy to avoid instantiating unique skill entities for each character

### DIFF
--- a/packages/solecs/README.md
+++ b/packages/solecs/README.md
@@ -77,7 +77,7 @@ Everyone has read access.
 For convenience the component implementation in `solecs` also includes a reverse mapping from `keccak256(value)` to set of entities with this value, but this is not strictly required and might change in a future release.
 
 ```solidity
-interface IComponent is IERC173 {
+interface IComponent is IOwnableWritable {
   /** Return the keys and value types of the schema of this component. */
   function getSchema() external pure returns (string[] memory keys, LibTypes.SchemaValue[] memory values);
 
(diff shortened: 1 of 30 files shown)


dk1a:
Also a question: is it good that I make a separate `MapSetExtended`, or would it be better to extend the original `MapSet`?

dk1a:
> Same question here as in [#267 (comment)](https://github.com/latticexyz/mud/pull/267#issuecomment-1428760436) - are you fine with closing this for now to focus on patches in v1 and move new features to v2? (This one could be an on-chain indexer in v2)

Yep, I'm very much in favor of focusing on v2, left these in case you want them in v1 basically

PR #328, opened 2023-01-06
build(services): add explicit typescript dependencies

dk1a:
This is a fix for the root cause of a confusing issue with vite (although since it can be hard to reproduce I'm not completely sure this PR solves it).
It's currently addressed via e.g. exclude/include in https://github.com/latticexyz/mud-template-react/blob/main/packages/client/vite.config.ts , or installing nice-grpc (and maybe other packages) can do the trick
Also it doesn't show up for locally linked mud

The issue seems to stem from a deeper issue with ts-proto itself, see https://github.com/stephenh/ts-proto/issues/147

### DIFF
--- a/packages/services/package.json
+++ b/packages/services/package.json
@@ -20,5 +20,10 @@
     "rimraf": "^3.0.2",
     "ts-proto": "^1.126.1"
   },
-  "gitHead": "218f56893d268b0c5157a3e4c603b859e287a343"
+  "gitHead": "218f56893d268b0c5157a3e4c603b859e287a343",
+  "dependencies": {
+    "long": "^5.2.1",
+    "nice-grpc-common": "^2.0.0",
+    "protobufjs": "^7.1.2"
+  }
 }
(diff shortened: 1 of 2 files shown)


alvrs:
> This is a fix for the root cause of a confusing issue with vite 
> \- https://github.com/latticexyz/mud/pull/328#issue-1522895599

Could you add some more details to the issue that this is trying to solve?

dk1a:
ts-proto doesn't use peerDependencies for imports it generates (which is the root issue).
So one must manually look at imports generated in `protobuf/ts` and add dependencies for them (which is what this PR does).
Otherwise packages that import `services`, like `network` (and all the clients that rely on it), will have imports without explicitly specified deps.
So vite, which can be strict with that, will throw an error when it sees an import of a non-existent package.
The error doesn't always happen cause the deps are probably somewhere in the tree, so if u're lucky and the bundler isn't strict it compiles fine.
Even if vite didn't throw errors occasionally, explicitly specifying deps is just good practice anyways.

alvrs (approved):
lgtm

PR #330, opened 2023-01-07
feat(std-contracts): built-in hasValue() for string components

JirAcheron:
Just a bit of a verbose check that seems to be getting used somewhat
frequently on our end. Feel free to ignore and close if there was a good
reason not to include this.


alvrs:
This is a reasonable addition; the only reason we didn't include this is to keep the interface of all components in sync (we could of course add this method to all components too). I'll still close this PR for now as we're moving fast on v2 (which changes the way components are done), and are focussing on bug fixes for v1 (while moving new features to v2).

PR #338, opened 2023-01-10
fix(solecs): use _setOwner in Ownable constructor

dk1a:
https://github.com/latticexyz/mud/pull/265#discussion_r1065996811

Fixes compliance with EIP-173 and makes the code look a bit better


alvrs (approved):
thanks @holic for noticing and thanks @dk1a for rapidly fixing! 🚢

PR #350, opened 2023-01-18
feat(guides): system call stream guide first draft

Kooshaba:


guides/animating.md
```diff
@@ -0,0 +1,111 @@
+---
```
alvrs, line 1: I wonder if we should create a new subfolder `game development` inside of `guides` and put this in there. Right now `mud.dev` doesn't really mention game development, but mostly positions MUD as a framework for complex Ethereum applications. (With games being a prime example of a complex eth app.) In that `game development` section, we can later also add an introduction into how to use MUD for game development (ie if we move towards a more general "table" based approach for vanilla MUD, we could explain how to use those tables for ECS and how ECS is a useful mental model for games).

alvrs (approved):
This will be very useful! Approving this, with the sidenote that I think we should move it into `guides/game development` (can be auto-expanded on the website too)

Kooshaba:
@alvrs moved into a new Game Dev section

alvrs (approved):

PR #351, opened 2023-01-18
fix(network): throw errors from txQueue calls

### LINKED ISSUE #315: Gas estimation errors do not propagate up to client
When a transaction, called in the client by
```
systems["system.xxx"].executeTyped(...)
```
cause a revert, an error is thrown:

Screenshot 2022-12-16 at 16 56 29

https://github.com/latticexyz/mud/blob/915506d7e7ca0b5a68afb646388bb9d4bb689879/packages/network/src/createTxQueue.ts#L192

What I would like is to be able to catch this error on the app-level.

holic:
closes #315

### DIFF
--- a/packages/network/src/createTxQueue.ts
+++ b/packages/network/src/createTxQueue.ts
@@ -35,7 +35,7 @@ export function createTxQueue<C extends Contracts>(
       gasLimit: BigNumberish
     ) => Promise<{ hash: string; wait: () => Promise<TransactionReceipt> }>;
     estimateGas: () => BigNumberish | Promise<BigNumberish>;
-    cancel: () => void;
+    cancel: (error: any) => void;
     stateMutability?: string;
   }>();
   const submissionMutex = new Mutex();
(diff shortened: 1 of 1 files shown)


packages/network/src/createTxQueue.ts
```diff
@@ -190,7 +190,9 @@ export function createTxQueue<C extends Contracts>(
         gasLimit = await txRequest.estimateGas();
       } catch (e) {
         console.error("[TXQueue] GAS ESTIMATION ERROR", e);
-        return txRequest.cancel();
+        throw e;
+      } finally {
+        txRequest.cancel();
```
Kooshaba, line 195: this cancels every tx

holic: ahh I see, so it makes sure that if one gas estimate fails, the whole queue is canceled?

holic: https://github.com/latticexyz/mud/blob/f18c39831f5d5ab0186529eb2cfaee4991570e6f/packages/network/src/createTxQueue.ts#L155

I see now that this just throws a generic error. Maybe I should rip out `txRequest.cancel` and just let the real error bubble up.

holic: > this cancels every tx

I see what you mean now 🙈

packages/network/src/createTxQueue.ts
```diff
@@ -219,6 +224,8 @@ export function createTxQueue<C extends Contracts>(
         // Nonce handeling
         if (shouldIncreaseNonce) incNonce();
         if (shouldResetNonce) await resetNonce();
+        // Bubble up error
+        if (error) txRequest.cancel(error);
```
holic, line 228: This still doesn't help with tx errors if it passes the gas estimation and returns a plain "transaction failed" error message:
image

alvrs (approved):
lgtm 🚢

PR #352, opened 2023-01-19
feat(store): add foundation for v2 data modelling

alvrs:
**TODO**
- [x] Streamline naming in Storage and Memory libraries (probably just use load and store instead of read and write?)
- [x] Move all internal functions to a separate StoreCoreInternal library to discourage using it
- [x] Change Storage library functions to make it clearer which argument is offset and which is length
- [x] Turn all storage pointer to uint256 for consistency (uint256 is better than bytes32 because it's easier to do arithmetic on)
- [x] change `bytes.concat` to `abi.encodePacked` where possible
- [x] Change all manual gas logs in tests to `// !gasreport` comments to use the `mud gas-report` cli command

## Follow ups
- [ ] Combine or further split `Bytes`, `Buffer`, `Memory` - they have some overlap and it’s not entirely clear where the boundaries are right now

**Low hanging fruits for gas improvements**

- [ ] Add `pushToDynamicField` and `popFromDynamicField` functions to store core lib to avoid writing entire array when adding a single element (eg `CallbackArray`)
- [ ] `Storage.write` and `Storage.read`
- [ ] Decoding values returned by store core library using as little memory copies as possible
- [ ] All low level libraries (`Bytes`, `Buffer`, `Memory`, `Storage`, `SchemaLib`, `PackedCounterLib`)
- [ ] Maybe: use abi.encodePacked for passing data into StoreCore? We'd need to construct the `PackedCounter` in StoreCore, but gas implications would be interesting. Maybe we can even use abi.encodePacked for data coming out of StoreCore, and have utils to decode them into non-packed with a given schema.


packages/solecs/v2/tables/Vec2Table.sol
```diff
@@ -0,0 +1,90 @@
...
+import { StoreSwitch } from "../StoreSwitch.sol";
+import { StoreCore } from "../StoreCore.sol";
+import { SchemaType } from "../Types.sol";
+import { Bytes } from "../Bytes.sol";
+
+// -- User defined schema and id --
+
+bytes32 constant id = keccak256("mud.store.table.vec2");
+
+struct Schema {
```
holic, line 15: would this need to be called `Schema` to be picked up by autogen or could this be `Vec2`? thinking that we'll probably be passing these kinds of data around and having everything named `Schema` would be annoying

packages/solecs/v2/tables/Vec2Table.sol
```diff
@@ -0,0 +1,90 @@
...
+  uint32 x;
+  uint32 y;
+}
```
holic, line 18: what would a scalar look like in this model? would we always have to pass around a single-value struct and call `.value` on it (like we do on the client)? is there a more ergonomic approach?

dk1a: I think it's entirely possibly to make a special case for scalars, which would be entirely autogen's responsibility.
`Schema` is more of a convenience thing for users - StoreCore doesn't care about it.

packages/solecs/v2/tables/Vec2Table.sol
alvrs: Agree, it should a named struct for each table instead of `Schema` everywhere!

dk1a: For autogen I was imagining a sol-type-file/ts/json which is the config with all the schemas, from which all the .sol files are spawned

packages/solecs/v2/tables/Vector2Table.sol
```diff
@@ -0,0 +1,93 @@
...
+import { console } from "forge-std/console.sol";
+import { IStore } from "../IStore.sol";
+import { StoreSwitch } from "../StoreSwitch.sol";
+import { StoreCore } from "../StoreCore.sol";
+import { SchemaType } from "../Types.sol";
+import { Bytes } from "../Bytes.sol";
+
+// -- User defined schema and id --
+
+bytes32 constant id = keccak256("mud.store.table.vector3");
```
holic, line 13:
```suggestion
bytes32 constant id = keccak256("mud.store.table.vector2");
```

alvrs:
very much WIP - don't look

dk1a:
a tiny improvement suggestion:
`abi.encodePacked(x, y);` vs `bytes.concat(bytes4(x), bytes4(y));` etc - shorter and no need for typecasts

packages/store/src/StoreCore.sol
```diff
@@ -0,0 +1,524 @@
...
+  event MudStoreDeleteRecord(bytes32 table, bytes32[] key);
+
+  bytes32 internal constant SLOT = keccak256("mud.store");
+  bytes32 internal constant SCHEMA_TABLE = keccak256("mud.store.table.schema");
+
+  error StoreCore_TableAlreadyExists(bytes32 table);
+  error StoreCore_TableNotFound(bytes32 table);
+  error StoreCore_NotImplemented();
+  error StoreCore_InvalidDataLength(uint256 expected, uint256 received);
+  error StoreCore_NoDynamicField();
```
holic, line 34: I noticed some of our error names use `_` and recently others do not. We should prob converge on some pattern.

I also wonder about similarly prefixing errors with `Mud`. Seems like that's the direction folks are going with ERC721, etc. errors because it helps with overlapping names.

That said, I'm now seeing `event MudStore...(...)` in the context of `StoreCore` and wondering again about no prefix vs. prefixing everything internal. 🙈  Can't decide!

packages/store/src/IStore.sol
```diff
@@ -0,0 +1,74 @@
...
+  // Set partial data at schema index
+  function setField(
+    bytes32 table,
+    bytes32[] memory key,
+    uint8 schemaIndex,
+    bytes memory data
+  ) external;
+
+  // Register hooks to be called when a record or field is set or deleted
+  function registerHooks(bytes32 table, IStoreHook hooks) external;
```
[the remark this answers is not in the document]
holic: second argument still needs `hooks` -> `hook` (I should have suggested a diff)

alvrs:
Alright, not 100% done yet (see TODOs in PR description), but ready to have a closer look @dk1a @holic @ludns @authcall @Kooshaba. I know this is a massive PR, but let's try to get it to a point where we're happy with it as a foundation, and then iterate further in separate follow-up PRs (which makes it easier to work on follow up tasks in parallel). Also, once this is merged into `v2`, we can see gas impact of all follow up PRs via the diff in the `gas-report.txt` file.

packages/store/src/IStore.sol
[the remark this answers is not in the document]
alvrs: ahh good catch

alvrs:
merging into `v2` branch now, so we can start working on followups in separate PRs

packages/store/src/Buffer.sol
```diff
@@ -0,0 +1,368 @@
...
+    uint256 _ptr = ptr(self);
+    uint128 _dataLength = uint128(data.length);
+    uint128 _length = length(self);
+    uint128 _newLength = _length + uint128(data.length);
+
+    // Update the current buffer length
+    _setLengthUnchecked(self, _newLength);
+
+    // Copy over given data to the buffer
+    assembly {
```
dk1a, line 201: You copy in 32 byte chunks, whereas the memory isn't word-aligned, so for if e.g. `_dataLength == 8`, memory corruption can happen.
[A verbose but pretty efficient alternative](https://gist.github.com/dk1a/6944af1557268e86b6d9d7ad801a93f9), for long bytes it can even costs less gas

packages/store/src/Buffer.sol
```diff
@@ -0,0 +1,368 @@
...
+pragma solidity >=0.8.0;
+
+// Acknowledgements:
+// Based on @dk1a's Slice.sol library (https://github.com/dk1a/solidity-stringutils/blob/main/src/Slice.sol)
+// Inspired by @ethier's DynamicBuffer.sol library (https://github.com/divergencetech/ethier/blob/main/contracts/utils/DynamicBuffer.sol)
+
+// Optimized for allocating a fixed size buffer and writing to it / reading from it.
+
+// First 16 bytes are the pointer to the data, followed by 16 bytes of data length.
+type Buffer is uint256;
```
dk1a, line 11: I wonder if 10/10/10 split for ptr/length/capacity would be better, then slicing can be done in-place without copying bytes (since you don't need a new length field in-memory)
[LinkedListLib](https://github.com/brockelmore/memmove/blob/master/src/LinkedList.sol) has a 3-way bitpacking example, although it's not very relevant

dk1a: But also, could you get rid of capacity entirely? it seems like an intermediate internal concept, you really need it only for checking overflow. If optimization converges on appendUnchecked anyways, capacity might be useless?

packages/store/src/Buffer.sol
```diff
@@ -0,0 +1,368 @@
...
+
+  /**
+   * @dev copies the buffer contents to a new memory location and lays it out like a memory array with the given size per element (in bytes)
+   * @return arrayPtr the pointer to the start of the array, needs to be casted to the expected type using assembly
+   */
+  function toArray(
+    Buffer self,
+    uint256 elementSize,
+    bool leftAligned
+  ) internal pure returns (uint256 arrayPtr) {
```
dk1a, line 250: wdyt about using `bytes32[] memory array` instead of `uint256 arrayPtr`? Cast will also need to be changed, but otherwise everything including gas (just tested) stays the same, and we avoid using pointers more than necessary.
It'd also be clearer that `toArray` can't make `bytes[]`/`string[]` arrays

packages/store/src/Bytes.sol
```diff
@@ -0,0 +1,705 @@
...
+        outputPtr := add(outputPtr, _bytesPerElement) // Go to next output slot
+      } {
+        mstore(outputPtr, shl(shiftBits, mload(inputPtr))) // Store the value in minimal bytes
+      }
+    }
+  }
+
+  // Needs unique name to avoid conflict with `from(uint32)`
+  function fromUint8(uint8 input) internal pure returns (bytes memory output) {
+    return abi.encodePacked(input);
```
dk1a, line 128: seems easier to just use `abi.encodePacked` instead of having 96 converter methods; I guess this is a leftover after `bytes.concat`?

packages/store/src/Bytes.sol
```diff
@@ -0,0 +1,705 @@
...
+    for (uint256 i = 0; i < output.length; i++) {
+      output[i] = SchemaType(uint8(input[i]));
+    }
+    return output;
+  }
+
+  /**
+   * Converts a tightly packed uint32 array into a regular uint32 array.
+   */
+  function toUint32Array(bytes memory input) internal pure returns (uint32[] memory output) {
```
dk1a, line 211: If `Buffer` becomes 10/10/10 or ditches capacity, these methods can operate on `Buffer` directly

packages/store/src/Bytes.sol
```diff
@@ -0,0 +1,705 @@
...
+   ************************************************************************/
+
+  function equals(bytes memory a, bytes memory b) internal pure returns (bool) {
+    if (a.length != b.length) {
+      return false;
+    }
+    return keccak256(a) == keccak256(b);
+  }
+
+  function setLengthInPlace(bytes memory input, uint256 length) internal pure returns (bytes memory) {
```
dk1a, line 242:
```suggestion
  function setLengthInPlace(bytes memory input, uint256 length) internal pure returns (bytes memory) {
    assert(length <= input.length);
```
changes cost from 16 to 82

packages/store/src/Bytes.sol
```diff
@@ -0,0 +1,705 @@
...
+    for (uint256 i = 0; i < lengths.length; ) {
+      chunks[i] = slice(data, sum, lengths[i]);
+      unchecked {
+        sum += lengths[i];
+        i += 1;
+      }
+    }
+    return chunks;
+  }
+}
```
dk1a, line 705:
```suggestion
```
unused

packages/store/src/PackedCounter.sol
```diff
@@ -0,0 +1,140 @@
...
+  ) internal pure returns (PackedCounter) {
+    bytes32 rawPackedCounter = PackedCounter.unwrap(packedCounter);
+
+    // Get current lengths (total and at index)
+    uint256 accumulator = total(packedCounter);
+    uint256 currentValueAtIndex = atIndex(packedCounter, uint8(index));
+
+    // Compute the difference and update the total value
+    int256 lengthDiff = int256(newValueAtIndex) - int256(currentValueAtIndex);
+    accumulator = uint256(int256(accumulator) + lengthDiff);
```
dk1a, line 124:
```suggestion
    // Update the total value according to the difference
    if (newValueAtIndex >= currentValueAtIndex) {
      accumulator += newValueAtIndex - currentValueAtIndex;
    } else {
      accumulator -= currentValueAtIndex - newValueAtIndex;
    }
```
I'd prefer uints when possible, and this saves some gas: 884 -> 830

packages/store/src/Schema.sol
```diff
@@ -0,0 +1,228 @@
...
+      }
+    }
+
+    // Require max 14 dynamic fields
+    uint8 dynamicFields = uint8(_schema.length) - staticFields;
+    if (dynamicFields > 14) revert SchemaLib_InvalidLength(dynamicFields);
+
+    // Store total static length, and number of static and dynamic fields
+    schema = Bytes.setBytes1(schema, 0, bytes1(bytes2(length))); // upper length byte
+    schema = Bytes.setBytes1(schema, 1, bytes1(uint8(length))); // lower length byte
```
dk1a, line 63:
```suggestion
    schema = Bytes.setBytes2(schema, 0, bytes2(length)); // 2 length bytes
```

packages/store/src/StoreCore.sol
```diff
@@ -0,0 +1,559 @@
...
+      dynamicDataLocation = StoreCoreInternal._getDynamicDataLocation(table, key, i);
+      Storage.store({ storagePointer: dynamicDataLocation, offset: 0, data: new bytes(encodedLengths.atIndex(i)) });
+      unchecked {
+        i++;
+      }
+    }
+
+    // Delete dynamic data length
+    uint256 dynamicDataLengthLocation = StoreCoreInternal._getDynamicDataLengthLocation(table, key);
+    Storage.store({ storagePointer: dynamicDataLengthLocation, data: bytes32(0) });
```
dk1a, line 220: Deleting dynamic data seems unnecessary, since you delete its length; a new record that sets the new length would be responsible for initializing the elements

packages/store/src/StoreCore.sol
```diff
@@ -0,0 +1,559 @@
...
+
+  /************************************************************************
+   *
+   *    GET DATA
+   *
+   ************************************************************************/
+
+  /**
+   * Get full static record for the given table and key tuple (loading schema from storage)
+   */
```
dk1a, line 234:
```suggestion
  /**
   * Get full data (static and dynamic) for the given table and key tuple, loading schema from storage
   */
```

packages/store/src/StoreCore.sol
```diff
@@ -0,0 +1,559 @@
...
+    // Load the dynamic data length if there are dynamic fields
+    PackedCounter dynamicDataLength;
+    uint256 numDynamicFields = schema.numDynamicFields();
+    if (numDynamicFields > 0) {
+      dynamicDataLength = StoreCoreInternal._loadEncodedDynamicDataLength(table, key);
+      outputLength += 32 + dynamicDataLength.total(); // encoded length + data
+    }
+
+    // Allocate a buffer for the full data (static and dynamic)
+    Buffer buffer = Buffer_.allocate(uint128(outputLength));
```
dk1a, line 263: Coming back to [the Buffer comment](https://github.com/latticexyz/mud/pull/352/files#r1096575924) - here it could be something like:
```solidity
bytes memory result = new bytes(outputLength);
Buffer buffer = Buffer_.fromBytes(result);
```
I guess I'm suggesting a `Slice` at this point, not `Buffer`

packages/store/src/StoreCore.sol
```diff
@@ -0,0 +1,559 @@
...
+  function _getStaticData(bytes32 table, bytes32[] memory key) internal view returns (bytes memory) {
+    Schema schema = _getSchema(table);
+    return _getStaticData(table, key, schema);
+  }
+
+  /************************************************************************
+   *
+   *    GET DATA
+   *
+   ************************************************************************/
```
dk1a, line 400:
```suggestion
  /************************************************************************
   *
   *    GET DATA
   *
   ************************************************************************/

  /**
   * Get full static record for the given table and key tuple (loading schema from storage)
   */
  function _getStaticData(bytes32 table, bytes32[] memory key) internal view returns (bytes memory) {
    Schema schema = _getSchema(table);
    return _getStaticData(table, key, schema);
  }
```

packages/store/src/StoreCore.sol
```diff
@@ -0,0 +1,559 @@
...
+   */
+  function _getStaticDataLocation(bytes32 table, bytes32[] memory key) internal pure returns (uint256) {
+    return uint256(keccak256(abi.encode(SLOT, table, key)));
+  }
+
+  /**
+   * Get storage offset for the given schema and (static length) index
+   */
+  function _getStaticDataOffset(Schema schema, uint8 schemaIndex) internal pure returns (uint256) {
+    uint256 offset = 0; // skip length
```
dk1a, line 475: the comment seems confusing, like it used to skip something but now doesn't

packages/store/src/System.sol
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
+
+// TODO: Use existing ERC2771 implementations
```
dk1a, line 4: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/metatx/ERC2771Context.sol

packages/store/src/Types.sol
```diff
@@ -0,0 +1,84 @@
...
+  Bytes24Array,
+  String,
+  Address,
+  AddressArray
+}
+
+/**
+ * Get the length of the data for the given schema type
+ * (Because Solidity doesn't support constant arrays, we need to use a function)
+ * TODO: add more types and make it more efficient (avoid linear search)
```
dk1a, line 23: avoiding both linear and binary searches is probably a good idea, something like https://github.com/dk1a/mud/blob/c4cae23ac0284d6d475b683ad145daaf748e910e/packages/solecs/v2/SchemaType.sol#L225

packages/store/src/Utils.sol
```diff
@@ -0,0 +1,27 @@
...
+   * @dev Left-aligned bit mask (e.g. for partial mload/mstore).
+   * For length >= 32 returns type(uint256).max
+   *
+   * length 0:   0x000000...000000
+   * length 1:   0xff0000...000000
+   * length 2:   0xffff00...000000
+   * ...
+   * length 30:  0xffffff...ff0000
+   * length 31:  0xffffff...ffff00
+   * length 32+: 0xffffff...ffffff
```
dk1a, line 20: Since you changed it from bytes to bits the comments are confusing.
But it should probably be inlined into `_loadPartialWord` and `_storePartialWord` and combined with the other bit manipulation for efficiency anyways

dk1a (commented):
IStore looks good; though I'm not sure how useful MudStoreDeleteRecord is, since it's basically like MudStoreSetRecord with all zeros. There doesn't seem to be a built-in `has` like with components (which makes sense, since this is generic storage and you usually don't waste a slot on an existence flag)

To summarize about `Buffer` - imo it currently combines the roles of a static buffer and slice. I'd explore not storing capacity, refactoring it to a `Slice`, and pairing it with `bytes` where necessary (and `bytes` length can be reduced if that's required).
Tried to avoid talking about gas for the most part.

Should I wait for merge before moving to gas optimizations and codegen?

packages/store/src/Buffer.sol
alvrs: Good point! Mind changing this in a follow up PR?

packages/store/src/Buffer.sol
alvrs: Yeah I agree - the whole point of this library is to make encoding/decoding data in `StoreCore` and the typed libs as cheap as possible, and I think it can be optimized to the use cases we have there instead of trying to be as general as possible. Since in the use cases we have there we always know the capacity in advance and just use the buffer as an intermediate to append to (or slice from), I think we can get rid of the explicit capacity. Mind changing this in a follow up?

packages/store/src/Buffer.sol
alvrs: agree with this change!

packages/store/src/Bytes.sol
alvrs: Agree!

packages/store/src/Bytes.sol
alvrs: This started out as a purely internal function, but it seems not to be used anywhere anymore. Will remove.

packages/store/src/StoreCore.sol
alvrs: Agree with this!

packages/store/src/StoreCore.sol
alvrs: indeed a leftover! Removing

packages/store/src/System.sol
alvrs: yeah true, will remove the comment. For future reference - we don't need a trusted forwarder check here, because systems don't contain own storage, but instead act on the caller's storage (either via delegatecall or via callbacks, see `StoreSwitch`), so any caller can be "trusted".

packages/store/src/Utils.sol
alvrs: True, removed the comments for now, but agree with inlining it later

alvrs:
First of all, thanks a lot for your review!

> though I'm not sure how useful MudStoreDeleteRecord is, since it's basically like MudStoreSetRecord with all zeros. There doesn't seem to be a built-in has like with components (which makes sense, since this is generic storage and you usually don't waste a slot on an existence flag)

There is a subtle difference: with support for callbacks via the `IStoreHook`, it's possible to create an "on-chain indexer" to store whether "an entity has a component", because the hooks for `setRecord` and `deleteRecord` are different. To make off-chain indexers consistent with this, they have to distinguish between `MudStoreSetRecord` and `MudStoreDeleteRecord`.

> To summarize about Buffer - imo it currently combines the roles of a static buffer and slice. I'd explore not storing capacity, refactoring it to a Slice, and pairing it with bytes where necessary (and bytes length can be reduced if that's required).

I agree with this, commented above!

> Should I wait for merge before moving to gas optimizations and codegen?

I just pushed fixes for the comments marked as resolved, and think we should work on the unresolved ones in follow up PRs.
Happy to merge into `v2` tomorrow if you agree, so you can get started on optimization and autogen.

dk1a:
> There is a subtle difference: with support for callbacks via the IStoreHook, it's possible to create an "on-chain indexer" to store whether "an entity has a component", because the hooks for setRecord and deleteRecord are different. To make off-chain indexers consistent with this, they have to distinguish between MudStoreSetRecord and MudStoreDeleteRecord.

Right, makes sense. Later I'd been thinking it might have to do with indexers, hadn't thought about hooks.

> Mind changing this in a follow up PR?

Sure, especially with Buffer comments I was expecting to do the changes myself, it overlaps with gas stuff too.

> Happy to merge into v2 tomorrow if you agree, so you can get started on optimization and autogen.

lgtm

dk1a (approved):

PR #353, opened 2023-01-23
docs: add current directory since 'docker build' requires PATH as arg…

b5710546232:
Dear mud team,
I would like to help adding document
since docker build need PATH as augment 
example
```sh
docker build .
```

Therefore I would like to add . as current directory in the example docuement

Thank you :)


authcall (approved):
nice catch thanks for this!

PR #354, opened 2023-01-25
fix(cli): use --prune-history in devnode to avoid filling disk

holic:
Temporary workaround for https://github.com/foundry-rs/foundry/issues/3623

This may impact the ability to trace anvil txs, but that seems better than filling someone's disk after leaving anvil running.

(This change is motivated by the fact that this happened to me a few times in the last week, waking up to none of my other OSX apps working because my disk was _completely_ full, due to 1.5TB of anvil logs.)


holic:
This seems to cause other issues: https://github.com/foundry-rs/foundry/pull/3488#issuecomment-1404112469

holic:
The latest version of anvil now has better log/history pruning. Use `foundryup` to upgrade!

PR #357, opened 2023-01-26
chore: upgrade github deps to ones with package.json for npm install

holic:
npm is more strict about installing deps from github and expects a package.json file to exist. Otherwise you get errors like we're seeing reported from our users:

image

This upgrades each dep to the earliest version with a `package.json` file (and switches memmove to @dk1a's fork that has `package.json` present until https://github.com/brockelmore/memmove/pull/1 is merged).


alvrs (approved):
lgtm 🚢

PR #359, opened 2023-01-30
docs: system call tweaks, dev improvements

holic:
- adds command (`yarn docs:dev`) for running docs locally
- changes default port to one that doesn't conflict with MacOS
- small (opinionated) tweaks to system call stream article (mostly changing casing)

### DIFF
--- a/guides/game_development/animations_from_system_calls.md
+++ b/guides/game_development/animations_from_system_calls.md
@@ -1,16 +1,12 @@
----
-order: 100
----
+# Animations from system calls
 
-# Animations in Response to System Calls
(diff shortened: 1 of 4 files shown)


guides/game_development/animations_from_system_calls.md
```diff
@@ -1,16 +1,12 @@
----
-order: 100
----
+# Animations from system calls
```
holic, line 1: Simplified the title, updated the file path (url slug) to match, and changed ~most things to use sentence case rather than capitalized headings

guides/game_development/index.yml
```diff
@@ -1,3 +1,2 @@
 expanded: false
 order: 1
-icon: play
```
holic, line 3: this looked a little noisy in the sidebar so I got rid of it for now

alvrs (approved):

PR #360, opened 2023-01-31
docs: emojimon tutorial

holic:
image

TODO
- [ ] better image sizing
- [ ] add more images
- [ ] update mud
- [ ] is there a better option for "collapsed" code in code blocks? currently using `…` but I don't love it
- [ ] refactor using new react hooks
- [ ] move `moveTo` and `moveBy` helpers to some sort of `api` export rather than within hooks
  - [ ] change `useMovement` to be just `useKeyboardMovement`
  - [ ] same with join game
- [ ] move solidity `SingletonID` into initial commit
  - [ ] PR to mud to do this ~everywhere?
- [ ] move loading screen into initial commit?
  - [ ] start project with black bg and white text?
- [ ] use new init lib pattern instead of init system?
- [ ] step 5: update to `import { BoolComponent } ...` instead of importing everything
- [ ] step 8: check that we can use this encoding/decoding approach (using string only in the schema type)
- [ ] step 8+: update terrain rendering to use grid instead of absolute/relative position? though we'll need position context for the encounter animation, so maybe it's fine
- [ ] step 7.2: wrap coord in join optimistic rendering?
- [ ] step 9: obstruction check in join optimistic rendering?
- [ ] use tailwind-animate for appear-in transitions?


alvrs (approved):
This is really amazing! Great job! 🎉🚀

Issue #310, opened 2023-01-02
recs entity values are passed by reference


0xhank:
In TS, if I use `getComponentValue` when the value is a typescript object, it returns a reference to the component value. This means I can update the component values without using `setComponent`. This creates issues when I want to compare the old update and the new update in a `createUpdateSystem` function, because the before and after update values are identical. To me it makes sense to pass the component by value, this way the only way to alter the object is through setComponent.

For example (please ignore terrible code):
```
  const selectedActions = getComponentValue(SelectedActions, ship) || {
    actionTypes: [ActionType.None, ActionType.None],
    specialEntities: ["0" as EntityID, "0" as EntityID],
  };

    if (index == -1) {
      const unusedSlot = selectedActions.actionTypes.indexOf(ActionType.None);
      if (unusedSlot == -1) return;
      selectedActions.actionTypes[unusedSlot] = action; // <- this line actually updates this slot in selectedActions.
      selectedActions.specialEntities[unusedSlot] = cannonEntity;
    } else {
      selectedActions.actionTypes[index] = ActionType.None;
      selectedActions.specialEntities[index] = "0" as EntityID;
    }
    setComponent(SelectedActions, ship, {  // <- This component update will return identical before/after values because the component was already updated above.
      actionTypes: selectedActions.actionTypes,
      specialEntities: selectedActions.specialEntities,
    });
```

alvrs:
The top-level component value object is a copy, but the second-level values (ie the `actionType` array in your example) are passed by reference https://github.com/latticexyz/mud/blob/ccb78462af8ecb2b39fc597382fdc02e736ab85d/packages/recs/src/Component.ts#L148-L163

I agree this can lead to annoying bugs - we could change this behavior by adding another step in line 159 to unwrap and rewrap array-type component values.

0xhank:
my current workaround is using `structuredClone,` which was added in node v17: 

```const actions = structuredClone(selectedActions);```

not sure if you want to add this yourself for backwards compatibility concerns.

alvrs:
the array elements should be primitive types right? Then `value[key] = [...val]` should work too (with a check like `if(ArrayTypes.includes(component.schema[key]))` before)

holic:
another option is to deep freeze the objects so we can continue to pass by reference but avoid mutations

R-Morpheus:
> another option is to deep freeze the objects so we can continue to pass by reference but avoid mutations

@holic  what do you think about this auxiliary function in this file?

[...]

PR #361, opened 2023-01-31
feat(cli): add initialization libs to deploy

dk1a:
deploy.json already has "init" for systems.
But systems are contracts. Scripts are best run through libraries.

Example:
An expensive `addItem` function, which adds a game item prototype with all its properties. Say it costs 1mil gas per item. I have an initializer which adds 100 of those. 100mil gas exceeds the limit.
- If the 100 adds are in a system contract, it will revert. And making a bunch of wrapper init systems is a really bad idea.
- If the 100 adds are in a library, `forge script` can at least split it into several transactions. But also the real contract calls (component sets) use only a small portion of all that gas.

(this is a simplified edge case, my real use case is unfinished and more convoluted, but similar enough, e.g. [InitEquipmentAffixSystem](https://github.com/dk1a/wanderer-cycle/blob/d0892364000075b2afeca610a6883b8d0445acb5/packages/contracts/src/init/InitEquipmentAffixSystem.sol) uses way more than the transaction gas limit)

I tested that it minimally works via a slightly modified [emojimon](https://github.com/dk1a/emojimon), if you wanna just clone it.

Some concerns:
- `libs` may not be the best name, open to suggestions
- The whole idea of modifying `LibDeploy` may be a bit too hardcodey. Maybe having a user-controllable `Deploy.sol` would be best. But that's way beyond the scope of this simple PR. (#309 kinda goes in that direction a bit, making scripts more local)

### DIFF
--- a/packages/cli/src/contracts/LibDeploy.ejs
+++ b/packages/cli/src/contracts/LibDeploy.ejs
@@ -25,6 +25,13 @@ import { <%= component %>, ID as <%= component %>ID } from "components/<%- compo
 import { <%= system.name %>, ID as <%= system.name %>ID } from "systems/<%- system.name %>.sol";
 <% }); -%>
 
+<% if (libs.length > 0) { -%>
+// Init Libraries (requires 'libraries=...' remapping in project's remappings.txt)
+<% libs.forEach(({name}) => { -%>
+import { <%= name %> } from "libraries/<%- name %>.sol";
+<% }); -%>
+<% } -%>
+
 struct DeployResult {
   IWorld world;
   address deployer;
@@ -100,5 +107,9 @@ library LibDeploy {
 <% } -%>
     console.log(address(system));
 <% });-%>
+
+<% libs.forEach(({name, func}) => { -%>
+    if(init) <%= name -%>.<%= func ?? 'init' -%>(world);
+<% }); -%>
   }
 }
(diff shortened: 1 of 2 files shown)


holic:
Love this idea.

I wonder if it's worth adding a `scripts` directory (and follow the foundry `*.s.sol` naming scheme) to keep the initialization scripts/logic separate from the long-running logic (libs).

dk1a:
> I wonder if it's worth adding a `scripts` directory (and follow the foundry `*.s.sol` naming scheme) to keep the initialization scripts/logic separate from the long-running logic (libs).

`scripts` is kinda like `src`, `test`, `out`, it should exist in the root (according to the defaults of foundry book, and my preference).
Something like `Deploy.sol` is a script (#311 is related), and should probably be `Deploy.s.sol`.

But I would prefer libraries that are used by scripts to still be inside `src`. They can be however kept in a separate dir from the other libs. This is mostly because #310 needs libs in `src`, but also allows them to be in arbitrary subfolders (for example `src/init` for init libs).

alvrs (approved):
Great addition!

holic:
Can we add docs for this? Maybe here: https://mud.dev/guides/getting_started/#deployment

dk1a:
> Just so I understand, this is meant to replace the current pattern of `InitSystem`s, right?

Yep

> Can we add docs for this? Maybe here: https://mud.dev/guides/getting_started/#deployment

I'll make a PR

PR #364, opened 2023-02-02
fix(cli): rename libs to initializers, always use init as func

dk1a:
For #361

### DIFF
--- a/guides/getting_started.md
+++ b/guides/getting_started.md
@@ -131,6 +131,9 @@ It contains the list of components and systems to deploy, and which systems need
 This file is the source of truth for MUD's deployment tools.
 If a component or system is not included, it is not deployed, even if it is placed in the `components` or `systems` directory.
 
+`initializers` are optional libraries. After components and systems are deployed, `init` function of each initializer will be called with `IWorld` as the sole argument, e.g. `LibInitData.init(world)`.
+Initializers should be in the `libraries` directory.
+
 **Example**: `deploy.json`
 
 ```json
@@ -141,7 +144,8 @@ If a component or system is not included, it is not deployed, even if it is plac
       "name": "IncrementSystem",
       "writeAccess": ["CounterComponent"]
     }
-  ]
+  ],
+  "initializers": ["LibInitData"]
 }
 ```
 
--- a/guides/getting_started.md
+++ b/guides/getting_started.md
@@ -131,7 +131,7 @@ It contains the list of components and systems to deploy, and which systems need
 This file is the source of truth for MUD's deployment tools.
 If a component or system is not included, it is not deployed, even if it is placed in the `components` or `systems` directory.
 
-`initializers` are optional libraries. After components and systems are deployed, `init` function of each initializer will be called with `IWorld` as the sole argument, e.g. `LibInitData.init(world)`.
+The `initializers` field in the `deploy.json` file allows the specification of libraries that implement a function named `init` which takes a single argument `IWorld`. This `init` function is executed after the deployment of the world, components, and systems, as part of the deployment process (e.g. `LibInitData.init(world)`).
 Initializers should be in the `libraries` directory.
 
 **Example**: `deploy.json`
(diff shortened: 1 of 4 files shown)


guides/getting_started.md
```diff
@@ -131,6 +131,9 @@ It contains the list of components and systems to deploy, and which systems need
 This file is the source of truth for MUD's deployment tools.
 If a component or system is not included, it is not deployed, even if it is placed in the `components` or `systems` directory.
 
+`libs` are optional. After components and systems are deployed, libs will be called with `IWorld` as the sole argument, e.g. `LibInitData.init(world)`.
+Libs should be in the `libraries` directory.
```
holic, line 135: We should probably add more context to answer the "why" behind this: what is this useful for?

guides/getting_started.md
```diff
@@ -141,7 +144,8 @@ If a component or system is not included, it is not deployed, even if it is plac
       "name": "IncrementSystem",
       "writeAccess": ["CounterComponent"]
     }
-  ]
+  ],
+  "libs": [{ "name": "LibInitData", "func": "init" }]
```
holic, line 148: It'd be nice to change `libs` to `libraries` to match the directory name (and the verbosity of `components` and `systems`) above. And maybe `func` -> `function` or just ~always default to `init` and make it not configurable?

Another thought is an `initializers` key to make it clear what they're for:
```json
{
  "initializers": ["LibInitData"]
}
```

holic (changes requested):

dk1a:
I like `initializers`, changed it up, and with the new name it seems self-descriptive too

holic (approved):
👏  Looks great, thank you!

My only other curiosity now is in terms of `mud dev` and "hot reloading" of contracts, when do initializers run? Once only on initial world deploy? On every change to the initializer lib?

dk1a:
> My only other curiosity now is in terms of mud dev and "hot reloading" of contracts, when do initializers run? Once only on initial world deploy? On every change to the initializer lib?

Good point, I use hsr too rarely to have thought of this. I made initializers respect the `init` flag, but it's kinda useless, especially for hot reloading. I think I should move inits from `deploySystems` into `deploy` and use `_reuseComponents`, so initializers run only if components are redeployed

holic:
cc @alvrs for input

dk1a:
Oh, and to be more specific, hsr always reuses components afaik, so `deploy-contracts --watch` will ignore any changes to components and initializers, and only update the system code

dk1a:
> Oh, and to be more specific, hsr always reuses components afaik, so deploy-contracts --watch will ignore any changes to components and initializers, and only update the system code

After some testing turns out `deploy-contracts --watch` always had `reuseComponents: false`, which means it just didn't work? Since components can never be redeployed (#286).
I fixed it and everything seems to work as I expect and think it should (as in, `deploy-contracts --watch` deploys everything and runs initializers once, and on any change only redeploys the systems).

@alvrs I may be missing something, check out the changes to see if they make sense and the bug really existed

guides/getting_started.md
```diff
@@ -131,6 +131,9 @@ It contains the list of components and systems to deploy, and which systems need
 This file is the source of truth for MUD's deployment tools.
 If a component or system is not included, it is not deployed, even if it is placed in the `components` or `systems` directory.
 
+`initializers` are optional libraries. After components and systems are deployed, `init` function of each initializer will be called with `IWorld` as the sole argument, e.g. `LibInitData.init(world)`.
```
alvrs, line 134:
```suggestion
The `initializers` field in the `deploy.json` file allows the specification of libraries that implement a function named `init` which takes a single argument `IWorld`. This `init` function is executed after the deployment of the world, components, and systems, as part of the deployment process (e.g. `LibInitData.init(world)`).
```

alvrs:
> After some testing turns out deploy-contracts --watch always had reuseComponents: false, which means it just didn't work?

Good catch! Previously we set `reuseComponents` to `true` when specific systems were provided, but this was recently changed in https://github.com/latticexyz/mud/pull/356/files and the implications for HMR missed. Thanks for fixing!

alvrs (approved):

PR #365, opened 2023-02-02
feat(cli): add gas-report command

alvrs:
Add a `gas-report` command to the CL.

Print the gas report to the console, save it to a file and compare it to a previous gas report if provided.
Requires forge to be installed, and gas test files including `// !gasreport` comments, like this:

```solidity
  contract GasTest is Test {
    function testBuffer() public pure {
      // !gasreport allocate a buffer
      Buffer buffer = Buffer_.allocate(32);
 
      bytes32 value = keccak256("some data");
 
      // !gasreport append 32 bytes to a buffer
      buffer.append(value);
    }
  }
```


packages/cli/src/commands/gas-report.ts
```diff
@@ -0,0 +1,219 @@
...
+  let logs = "";
+  try {
+    logs = (await child).stdout;
+    rmSync(tempFileName);
+  } catch (e: any) {
+    console.log(e.stdout ?? e);
+    console.log(chalk.red("\n-----------\nError while running the gas report (see above)"));
+    rmSync(tempFileName);
+    process.exit();
+  }
```
holic, line 168: With `process.exit()` here, I don't know if this would work, but you might put `rmSync` into `finally`
```ts
try {
 ...
} catch (...) {
  ...
} finally {
  rmSync(tempFileName);
}
```
(unclear if the process exits before it gets here)

holic (approved):
🎉 

Not blocking, but might be nice to have some tests for those regexes to make sure we're parsing, diffing, etc. properly. Happy to ship this as-is though!

alvrs:
> but might be nice to have some tests for those regexes to make sure we're parsing, diffing, etc. properly

Good point! Noting it down as a follow up

packages/cli/src/commands/gas-report.ts
alvrs: yeah I had that initially but then changed it in assumption of `process.exit()` terminating the script immediately without getting to `finally` (didn't test this assumption though)

PR #368, opened 2023-02-03
fix(network): use current block number while waiting for new blocks

dk1a:
Currently network waits for a new block, which means I must use a tickchain locally.
Perhaps it also delays loading for production chains with slowish block times, haven't tested that.

This fix uses the current block number if the rpc returns it before a new block. I may have missed some downstream problems, but it seems like a harmless change


alvrs (approved):
Lgtm!

PR #370, opened 2023-02-06
fix(solecs): make OwnableWritable overridable

Boffee:
Make it possible to customize component write access.


dk1a:
Thanks! Could you also add `override` to the functions (since they override the interface)?
like
```solidity
function authorizeWriter(address writer) public virtual override onlyOwner {
```

alvrs (approved):
lgtm, just waiting for CI to pass.

PR #371, opened 2023-02-06
feat(cli): allow initializers to utilize SystemStorage

dk1a:
Followup to #364
only thought of this after starting to use initializers - like with tests, they could use SystemStorage, it just needs to be initialized somewhere


alvrs (approved):
lgtm

PR #378, opened 2023-02-07
feat(store): add storage optimizations

dk1a:
Gonna to try to make self-contained PRs.
Tried to make it readable. Not necessarily the final version of storage, but it's already noticably better gas-wise.
This is also the 1st step to removing Buffer.
Should I include gas-report.txt changes in PRs?

(leftMask is back to bytes, otherwise I'd need even more unchecked blocks)

### DIFF
--- a/packages/store/src/Storage.sol
+++ b/packages/store/src/Storage.sol
@@ -7,13 +7,20 @@ import { Bytes } from "./Bytes.sol";
 import { Memory } from "./Memory.sol";
 import "./Buffer.sol";
 
+/**
+ * @dev This implementation is very gas efficient.
+ * TODO Probably not fully optimized.
+ * Full assmebly may be a bit more efficient than unchecked blocks, but would be even less readable.
+ */
 library Storage {
   function store(uint256 storagePointer, bytes memory data) internal {
     store(storagePointer, 0, data);
   }
 
   function store(uint256 storagePointer, bytes32 data) internal {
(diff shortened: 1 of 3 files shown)


holic:
> Should I include gas-report.txt changes in PRs?

You can! But there's also a GitHub Action that will generate these (I just approved it for this PR, so it should publish the new report shortly.)

holic:
@alvrs  looks like the gas report doesn't work on forks (which makes sense, not sure it'd have permission to commit to the fork anyway)
```
fatal: invalid reference: dk1a/v2-optimize-storage
Error: Invalid status code: 128
```

packages/store/src/Storage.sol
```diff
@@ -88,107 +138,114 @@ library Storage {
...
-        );
-
-        // Else, just load the word
-      } else {
-        buffer.appendUnchecked(_loadWord(storagePointer + i), 32);
-      }
-    }
-  }
+    uint256 bufferLength = buffer.length();
+    load(storagePointer, length, offset, buffer.ptr() + bufferLength);
```
alvrs, line 162: why are we adding the buffer's length to the buffer's pointer to get the memory pointer?

alvrs (commented):
Looks good! Mind opening this as a new PR from a branch within the repo? Then the gh action should world. (We'll change the action to not try to commit but just check if the gas report is up to date soon too.)

PR #380, opened 2023-02-08
docs: emojimon tutorial (part 2)

holic:
A continuation of #360  (we merged early in prep for a workshop)

Follow ups noted in #384

### DIFF
--- a/tutorials/emojimon/step-10/step-10.7.md
+++ b/tutorials/emojimon/step-10/step-10.7.md
@@ -1,3 +1,98 @@
 # 10.7. Run away
 
-TODO
+Alongside our "throw" button, let's build a "run away" button to flee the encounter.
+
+## Flee system
+
+Just like the throw system, we'll add a flee system. To keep it simple, we won't use any randomness with this - when you want to flee, you can. Maybe you'll want to add a chance of "you can't run away"?
+
+```sol packages/contracts/src/systems/EncounterFleeSystem.sol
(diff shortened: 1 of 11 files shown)


holic:
Going to call this _good enough_. I need to move on to something else for a bit! Will come back to this and address TODOs (easier to iterate now that _something_ is up there).

tutorials/emojimon/step-10/step-10.7.md
```diff
@@ -1,3 +1,98 @@
...
+    },
+    {
+      "name": "EncounterThrowSystem",
+```
+
+## Add run button
+
+Since our flee system always allows you to run away, we don't need to listen for system call updates to determine the outcome. We can just call the system and wait for the transaction to complete before updating the toast.
+
+Because the encounter screen is shown only when you're an encounter, you'll see that it will automatically disappear when you run away. This is the nice thing about MUD and declarative, responsive UI!
```
alvrs, line 54:
```suggestion
Because the encounter screen is shown only when you're in an encounter, you'll see that it will automatically disappear when you run away. This is the nice thing about MUD and declarative, responsive UI!
```

alvrs (approved):
This is great! Should we move the remaining follow up todos into a follow up issue and mark it `good first issue`? Then folks who want to contribute can pick a couple and fix them.

Good to merge this!

alvrs (approved):

PR #382, opened 2023-02-08
test(store): add script to simulate store events on local chain for development purposes

alvrs:
Run `yarn script` in `packags/store` to deploy a new store on a local chain and set a record and a field on it to emit the `MudStoreSetRecord` and `MudStoreSetRecord` events. Requires a local node with gas price 0 to be running at `http://localhost:8545` (eg via `anvil --block-base-fee-per-gas 0`)

### DIFF
--- a/packages/store/script/Store.s.sol
+++ b/packages/store/script/Store.s.sol
@@ -1,0 +1,58 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
+
+import "forge-std/Script.sol";
+import { StoreView } from "../src/StoreView.sol";
+import { StoreCore } from "../src/StoreCore.sol";
+import { Schema, SchemaLib } from "../src/Schema.sol";
+import { SchemaType } from "../src/Types.sol";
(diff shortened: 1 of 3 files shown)


packages/store/script/Store.s.sol
```diff
@@ -0,0 +1,58 @@
...
+  ) public override {
+    StoreCore.setField(table, key, schemaIndex, data);
+  }
+}
+
+contract StoreScript is Script {
+  function run() public {
+    vm.startBroadcast();
+
+    uint256 blockNumber = block.number;
```
holic, line 37: I've found that this isn't the actual block number that these txs are executed on. Forge seems to run this script in its entirety to grab all the txs that need executing, then executes each of them in a batch, so this will often be behind by a couple blocks.

alvrs: that's good to know!

alvrs:
Don't think this is needed anymore since we have https://github.com/latticexyz/v2sandbox

PR #383, opened 2023-02-08
fix(store): emit MudStoreSetRecord event when registering a new schema, register SchemaTable's schema

alvrs:
We weren't emitting a `MudStoreSetRecord` event when storing to the schema table, because the schema table doesn't use the usual `setRecord` path of setting data (because that one requires the table to have a schema, so it's a circular dependency). 

Also, we weren't registering the SchemaTable's schema in the SchemaTable. Not too important, but good for completeness.

### DIFF
--- a/packages/store/src/StoreCore.sol
+++ b/packages/store/src/StoreCore.sol
@@ -309,10 +309,6 @@ library StoreCoreInternal {
   bytes32 internal constant SLOT = keccak256("mud.store");
   bytes32 internal constant SCHEMA_TABLE = keccak256("mud.store.table.schema");
 
-  event MudStoreSetRecord(bytes32 table, bytes32[] key, bytes data);
-
-  error StoreCore_InvalidDataLength(uint256 expected, uint256 received);
-
   /************************************************************************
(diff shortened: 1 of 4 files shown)


packages/store/src/StoreCore.sol
```diff
@@ -308,6 +309,8 @@ library StoreCoreInternal {
   bytes32 internal constant SLOT = keccak256("mud.store");
   bytes32 internal constant SCHEMA_TABLE = keccak256("mud.store.table.schema");
 
+  event MudStoreSetRecord(bytes32 table, bytes32[] key, bytes data);
+
   error StoreCore_InvalidDataLength(uint256 expected, uint256 received);
```
dk1a, line 314: I think you could use StoreCore's events/errors instead of declaring them again

dk1a (approved):
lgtm, just a small nit about duplicate declarations

packages/store/src/StoreCore.sol
alvrs: good point!

PR #385, opened 2023-02-09
feat(std-contracts): decode null getRawValue

Boffee:
Don't revert on null getRawValue response. Retrieving data from components should behave similarly to reading from mappings. If revert is intended, getValue should revert with custom a error.


alvrs (commented):
I generally agree with this change, but unfortunately it's a breaking change (consumers might expect the current behaviour of revert if the value doesn't exist), so we'd have to release a new major version to integrate this. Since we're currently working on a `v2` version of MUD that changes the way components are modelled, this will soon become obsolete. We could however throw a custom error if the value doesn't exist, which wouldn't be a breaking change and improve the developer experience.

alvrs:
Closing this for now, feel free to reopen with a custom error

PR #386, opened 2023-02-09
feat(store): add storage optimizations

dk1a:
Replaced #378

> why are we adding the buffer's length to the buffer's pointer to get the memory pointer?

I no longer use appendUnchecked in Storage. That's a temporary hack so that the new Storage works before I finish the Buffer overhaul

### DIFF
--- a/packages/store/src/Storage.sol
+++ b/packages/store/src/Storage.sol
@@ -8,9 +8,7 @@ import { Memory } from "./Memory.sol";
 import "./Buffer.sol";
 
 /**
- * @dev This implementation is very gas efficient.
- * TODO Probably not fully optimized.
- * Full assmebly may be a bit more efficient than unchecked blocks, but would be even less readable.
+ * TODO Probably not fully optimized
  */
 library Storage {
   function store(uint256 storagePointer, bytes memory data) internal {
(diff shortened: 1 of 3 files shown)


packages/store/src/Storage.sol
```diff
@@ -7,13 +7,20 @@ import { Bytes } from "./Bytes.sol";
 import { Memory } from "./Memory.sol";
 import "./Buffer.sol";
 
+/**
+ * @dev This implementation is very gas efficient.
```
alvrs, line 11: do we need this comment?

packages/store/src/Storage.sol
```diff
@@ -88,107 +138,114 @@ library Storage {
...
-    assembly {
-      data := sload(storagePointer)
-    }
+    // Update the current buffer length
+    buffer._setLengthUnchecked(uint128(bufferLength + length));
   }
 
   /**
-   * @notice Load a partial word from storage into memory
+   * @notice Append raw bytes from storage at the given storagePointer, offset, and length to the given buffer
```
alvrs, line 169:
```suggestion
   * @notice Append raw bytes from storage at the given storagePointer, offset, and length to the given memoryPointer
```
(I know the interface will probably slightly change in a follow up anyways)

packages/store/src/Storage.sol
dk1a: Haha no, I think I just changed an old comment saying something like "not very gas efficient" and moved it from a function to the top of the lib

alvrs (approved):

PR #387, opened 2023-02-09
feat(store): replace Buffer with Slice and reorganize custom coders

dk1a:
Replaces Slice with Buffer, which affects some other files cause they're pretty interdependent.
We're basically doing custom abi coding, so I moved all the encoding/decoding into `abicoder` dir.

There're some TODOs and temporary checks, and optimizations are not finished. I'm thinking of doing push/pop, then autogen, then coming back to do some cleanup.

Some record concerns:
1. `getRecord` doesn't include the 32 bytes for dynamic length if the total dynamic length is 0. That seems like it could be troublesome for autogen later.
2. A particular `setRecord` optimization (of dubious usefulness) that I haven't done yet can be seen in Mixed.sol:
```solidity
bytes memory data = abi.encodePacked(u32, u128, encodedLengths.unwrap(), EncodeArray.encode(a32), bytes(s));
```
this copies around bytes because otherwise I need to deal with custom-packing static data (for dynamic I have a currently unused `encodeTo`).
This optimization can be skipped - using `setRecord` with arrays is a bad idea to begin with.
If we don't want to skip it, it would be much easier if StoreCore's `setRecord` accepted static and dynamic data as 2 separate packed `bytes`.


holic:
> We're basically doing custom abi coding, so I moved all the encoding/decoding into `abicoder` dir.

I wonder if we should move away from "abi coding" as a term to make sure we don't confuse with `abi.encode`.

dk1a:
> I wonder if we should move away from "abi coding" as a term to make sure we don't confuse with `abi.encode`. As I understand it, we're following our own schema (not the ABI) and doing our own encoding/packing of these data?

Yep, we do custom encoding, I agree that the name is misleading, what would you suggest? E.g. `StoreCoder`, `Pack`

The reason I called it abi - it's `StoreCore`'s "abi", and an analogy to `abi.encodePacked`, since we're reimplementing solidity's internal packing for storage. In an ideal world solidity would be extended with something like `abi.encodeForStorage` for the optimal combination of gas, convenience and safety

dk1a:
To clarify more about what's currently called `ABICoder` - it doesn't really depend on custom schemas, or StoreCore, or MUD. It just encodes data in a tightly packed way for storage, exactly like solidity does internally

packages/store/gas-report.txt
```diff
@@ -37,11 +37,12 @@
...
-(test/Slice.t.sol) | Slice (2 bytes) to bytes memory [bytes memory sliceData2 = slice.toBytes()]: 386
-(test/Slice.t.sol) | Slice (32 bytes) to bytes memory [bytes memory sliceData32 = slice.toBytes()]: 382
-(test/Slice.t.sol) | Slice (34 bytes) to bytes memory [bytes memory sliceData34 = slice.toBytes()]: 463
-(test/Slice.t.sol) | Slice (1024 bytes) to bytes memory [bytes memory sliceData1024 = slice.toBytes()]: 750
+(test/Slice.t.sol) | Slice (0 bytes) to bytes memory [bytes memory sliceData0 = slice.toBytes()]: 607
+(test/Slice.t.sol) | Slice (2 bytes) to bytes memory [bytes memory sliceData2 = slice.toBytes()]: 642
+(test/Slice.t.sol) | Slice (32 bytes) to bytes memory [bytes memory sliceData32 = slice.toBytes()]: 510
+(test/Slice.t.sol) | Slice (34 bytes) to bytes memory [bytes memory sliceData34 = slice.toBytes()]: 591
+(test/Slice.t.sol) | Slice (1024 bytes) to bytes memory [bytes memory sliceData1024 = slice.toBytes()]: 4847
+(test/Slice.t.sol) | Slice (1024x1024 bytes) to bytes memory [bytes memory sliceData1024x1024 = slice.toBytes()]: 6605248
```
holic, line 45: 🙈

dk1a: Aside from being curious about memcopy's performance scaling, I actually added this for control.
gas-report was reporting weirdly low values for me on a couple occasions, as if it didn't copy and just called empty funcs

packages/store/src/Slice.sol
```diff
@@ -0,0 +1,107 @@
...
+// First 16 bytes are the pointer to the data, followed by 16 bytes of data length.
+type Slice is uint256;
+
+using SliceInstance for Slice global;
+using DecodeSlice for Slice global;
+
+/**
+ * @title Static functions for Slice
+ */
+library Slice_ {
```
alvrs, line 20: Nit: we moved towards calling these libraries that implement function on a custom type `SliceLib` to make it a bit more explicit what the connection between the library and the type is

packages/store/src/storagecoder/StorageCoder.sol
```diff
@@ -0,0 +1,108 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
+
+import { Slice, Slice_ } from "../Slice.sol";
+import { SchemaType } from "../Types.sol";
+
+library StorageCoder {
```
alvrs, line 7: Naming is hard; but technically this library just does our custom tight packing / encoding / unpacking / decoding, it's not strictly connected to storage. Don't really have a much better proposal either, will think about it more (planning to do a more in-depth review tomorrow)

packages/store/src/Slice.sol
dk1a: Renamed; now it's SliceLib for static, SliceInstance for instance funcs (SliceInstance would never need to be imported)

packages/store/src/storagecoder/StorageCoder.sol
alvrs: I kind of like `TightCoder`, makes the purpose pretty obvious. Also maybe @holic has an idea, he's great with naming.

alvrs:
> A particular setRecord optimization (of dubious usefulness) that I haven't done yet can be seen in Mixed.sol:
>```solidity
>bytes memory data = abi.encodePacked(u32, u128, encodedLengths.unwrap(), EncodeArray.encode(a32), bytes(s));
>```
>this copies around bytes because otherwise I need to deal with custom-packing static data (for dynamic I have a currently unused encodeTo).
>This optimization can be skipped - using setRecord with arrays is a bad idea to begin with.
>If we don't want to skip it, it would be much easier if StoreCore's setRecord accepted static and dynamic data as 2 separate packed bytes.

I'd prefer keeping `setRecord` as the single way to set full records instead of splitting it into partial methods for static and dynamic data, because it reduces the number of data modification events clients/indexers need to handle (we'd need a new event for `StoreSetStaticData` and `StoreSetDynamicData`).

For the optimization - do you know if `abi.encodePacked` puts the resulting data at the free memory pointer? If so, could we just use `abi.encodePacked` for the static data and copy directly to the new free memory pointer with `EncodeArray` (using a slice/memory pointer)? Copying from calldata to memory for every element would still be necessary of course, but at least not multiple copies from the result of `EncodeArray` to the result of `abi.encodePacked`.

packages/store/src/Slice.sol
alvrs: I like that - we should probably adjust `SchemaLib` and `PackedCounterLib` to use the same pattern (right now everything is in the `...Lib`)

dk1a:
> For the optimization - do you know if abi.encodePacked puts the resulting data at the free memory pointer?

I imagine it would have to, but then the optimizer may move things around, so relying on that is very dangerous.

> could we just use abi.encodePacked for the static data and copy directly to the new free memory pointer with EncodeArray (using a slice/memory pointer)?

A good compromise would be to use `abi.encodePacked` for statics and copy them, but then encode dynamics directly to the preallocated memory.
It wouldn't be that hard to do, I just wanted to explore our options.
I'd rather not do it in this PR though.

packages/store/src/storagecoder/StorageCoder.sol
```diff
@@ -0,0 +1,108 @@
...
+  ) internal pure returns (bytes memory data) {
+    uint256 packedLength = array.length * elementSize;
+    data = new bytes(packedLength);
+    encodeTo(array, SliceLib.fromBytes(data), elementSize, leftAligned);
+  }
+
+  /**
+   * @dev Unpacks the slice to a new memory location
+   * and lays it out like a memory array with the given size per element (in bytes)
+   * @return array a generic array, needs to be cast to the expected type using assembly
```
alvrs, line 62: tiny typo
```suggestion
   * @return array a generic array, needs to be casted to the expected type using assembly
```

packages/store/src/storagecoder/DecodeSlice.sol
```diff
@@ -0,0 +1,102 @@
...
+   *
+   *    Slice -> bytes1-32
+   *
+   ************************************************************************/
+
+  function toBytes4(Slice self) internal pure returns (bytes4 output) {
+    return bytes4(self.toBytes32());
+  }
+
+  // `toBytes32` is already defined in `Slice_`
```
alvrs, line 19:
```suggestion
  // `toBytes32` is already defined in `SliceInternal`
```

packages/store/test/storagecoder/StorageCoder.t.sol
```diff
@@ -0,0 +1,42 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
+
+import "forge-std/Test.sol";
+import { SliceLib } from "../../src/Slice.sol";
+import { EncodeArray } from "../../src/storagecoder/EncodeArray.sol";
+
+contract StorageCoderTest is Test {
```
alvrs, line 8: Should we add a few lower level tests for `StorageCoder` itself in addition to the higher level `EncodeArray` / `DecodeSlice`?

alvrs (commented):
looks really good overall!

packages/store/src/Slice.sol
```diff
@@ -0,0 +1,107 @@
...
+  function toBytes(Slice self) internal view returns (bytes memory data) {
+    uint256 ptrSrc = ptr(self);
+    uint256 _length = length(self);
+
+    // Allocate a new bytes array and get the pointer to it
+    data = new bytes(_length);
+    uint256 ptrDest;
+    assembly {
+      ptrDest := add(data, 32)
+    }
```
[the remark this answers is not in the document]
dk1a: I think the annotations are only meaningful in blocks that actually write/read memory, otherwise the compiler assumes it to be memory safe by default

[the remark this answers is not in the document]
dk1a: But regardless of this particular block, I haven't really gotten to annotations yet, I'm sure there're places that could use them

packages/store/src/StoreCore.sol
```diff
@@ -239,41 +238,48 @@ library StoreCore {
...
+    uint256 staticLength = schema.staticDataLength();
+    uint256 outputLength = staticLength;
 
     // Load the dynamic data length if there are dynamic fields
     PackedCounter dynamicDataLength;
     uint256 numDynamicFields = schema.numDynamicFields();
     if (numDynamicFields > 0) {
       dynamicDataLength = StoreCoreInternal._loadEncodedDynamicDataLength(table, key);
-      outputLength += 32 + dynamicDataLength.total(); // encoded length + data
+      // TODO should total output include dynamic data length even if it's 0?
```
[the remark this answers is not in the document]
dk1a: Currently none, this was just something unexpected I noticed. This TODO would be well-resolved by this behaviour just being documented.
I suppose in decoders this could be of some unimportant inconvenience

packages/store/src/StoreCore.sol
```diff
@@ -239,41 +238,48 @@ library StoreCore {
...
+    if (dynamicDataLength.total() == 0) return data;
+    // Advance memoryPointer to the dynamic data section
+    memoryPointer += staticLength;
 
-    // Append the encoded dynamic length to the buffer after the static data
-    buffer.append(dynamicDataLength.unwrap());
+    // Append the encoded dynamic length
+    Memory.store(memoryPointer, dynamicDataLength.unwrap());
+    // Advance memoryPointer by the length of `dynamicDataLength` (1 word)
+    memoryPointer += 0x20;
```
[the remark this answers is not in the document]
dk1a: Currently I'm using hex for pointers specifically. I don't mind changing it, it just felt like a little extra visual indication

packages/store/src/storagecoder/StorageCoder.sol
```diff
@@ -0,0 +1,108 @@
...
+
+import { Slice, SliceLib } from "../Slice.sol";
+import { SchemaType } from "../Types.sol";
+
+library StorageCoder {
+  /**
+   * @dev Copies the array to the location of `packedSlice`,
+   * tightly packing it using the given size per element (in bytes)
+   */
+  function encodeTo(
```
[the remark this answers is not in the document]
dk1a: Renamed it and made it more explicit that I intend to change it later depending on optimization/autogen factors

packages/store/src/storagecoder/StorageCoder.sol
```diff
@@ -0,0 +1,108 @@
...
+    bytes32[] memory array,
+    Slice packedSlice,
+    uint256 elementSize,
+    bool leftAligned // TODO this function is currently not used externally
+  ) private pure {
+    uint256 arrayLength = array.length;
+    uint256 packedPtr = packedSlice.ptr();
```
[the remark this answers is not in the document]
dk1a: https://github.com/latticexyz/mud/pull/387#discussion_r1106473625
I've thought about this actually, but since this function is temporary I'd rather put off optimizing it. Marked it with a clearer TODO

packages/store/test/storagecoder/StorageCoder.t.sol
dk1a: A lot of tests are refactored and added in #396, you should review them there

dk1a:
@alvrs Resolved nits that are changed (or don't seem to need changes)
left some you might want to follow up on

dk1a:
> I kind of like TightCoder, makes the purpose pretty obvious. Also maybe @holic has an idea, he's great with naming.

do you want me to rename it here, or in another PR after we think about the name more?

packages/store/src/StoreCore.sol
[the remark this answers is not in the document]
alvrs: yeah `0x20 = 1 word` is probably common enough to justify hex here

alvrs:
> @alvrs Resolved nits that are changed (or don't seem to need changes)
> left some you might want to follow up on

Changes / comments look good to me, marked them as resolved.

> do you want me to rename it here, or in another PR after we think about the name more?

I think `TightCoder` is already better than `StorageCoder`, so slight preference to rename it now to avoid larger refactoring down the line, but no strong opinion.

packages/store/src/Memory.sol
```diff
@@ -35,8 +35,8 @@ library Memory {
   }
 
   function copy(
-    uint256 ptrDest,
-    uint256 ptrSrc,
+    uint256 toPointer,
+    uint256 fromPointer,
```
alvrs, line 39: Nit - what's the reason for this order? Would have expected `from` before `to`

dk1a: Agree that from,to is more intuitive, changed it.
That was a remnant from memcpy, which was itself based on C's semantics, which has ptrDest first

packages/store/src/storagecoder/StorageCoder.sol
holic: I don't love TightCoder haha

DataPack? DataPacker? BytePacker? BytesPacker? DataCoder? ByteCoder? 🤷‍♂️

holic: how about Serializer?

ludns:
Should we move to the `assembly("memory-safe")` dialect string instead of `/// @solidity memory-safe-assembly`? This comment will be removed in a future version, and it is only useful if we want to support old version of Solidity.
see: https://docs.soliditylang.org/en/v0.8.16/assembly.html#memory-safety

dk1a:
> Should we move to the assembly("memory-safe") dialect string instead of /// @solidity memory-safe-assembly? This comment will be removed in a future version, and it is only useful if we want to support old version of Solidity.
see: https://docs.soliditylang.org/en/v0.8.16/assembly.html#memory-safety

I use the comment because my ide linter really doesn't like `assembly("memory-safe")` (Juan Blanco's solidity for vscode), so I'd rather switch to it when editor support catches up

alvrs (approved):
lgtm!

packages/store/src/storagecoder/StorageCoder.sol
dk1a: > another thought: since this is only a few methods, would it make sense to include them inside another lib (e.g. SliceLib) we already have?

Well, there're only 2 core encode/decode methods, but then there're like 200+ for a plethora of solidity types that use these 2. I'd rather isolate them as much as possible, since it's a ton of code.
They do depend on Slice, but that could actually change, and Slice doesn't depend on them.

PR #388, opened 2023-02-10
refactor(react): improve useEntityQuery usage/rendering

### LINKED ISSUE #379: useEntityQuery hook triggers unexpected rerenders
Reported by @emersonhsieh

### Minimal reproduction case:
https://github.com/emersonhsieh/mud-minimal-test
https://github.com/emersonhsieh/mud-minimal-test/blob/main/packages/client/src/App.tsx

image

holic:
closes #379

Adds better memoizing to `useEntityQuery` arguments so that you don't have to memoize values passed into it (which is currently causing an ~infinite loop if you're not careful).

### DIFF
--- a/packages/react/src/useEntityQuery.ts
+++ b/packages/react/src/useEntityQuery.ts
@@ -1,23 +1,25 @@
-import { defineQuery } from "@latticexyz/recs";
+import { defineQuery, QueryFragment } from "@latticexyz/recs";
 import { useEffect, useMemo, useState } from "react";
+import { useDeepMemo } from "./utils/useDeepMemo";
+import isEqual from "fast-deep-equal";
+import { distinctUntilChanged, map } from "rxjs";
 
-type Args = Parameters<typeof defineQuery>;
-const defaultOptions: NonNullable<Args[1]> = { runOnInit: true };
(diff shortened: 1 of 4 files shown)


packages/react/src/useEntityQuery.ts
```diff
@@ -1,23 +1,25 @@
...
+import { defineQuery, QueryFragment } from "@latticexyz/recs";
 import { useEffect, useMemo, useState } from "react";
+import { useDeepMemo } from "./utils/useDeepMemo";
+import isEqual from "fast-deep-equal";
+import { distinctUntilChanged, map } from "rxjs";
 
-type Args = Parameters<typeof defineQuery>;
-const defaultOptions: NonNullable<Args[1]> = { runOnInit: true };
+// This does a little more rendering than is necessary when arguments change,
+// but at least it's giving correct results now. Will optimize later!
```
alvrs, line 8: Should we open a follow up issue for this?

alvrs (approved):
lgtm!

packages/react/src/useEntityQuery.ts
holic: I am not worried about it at the moment! If we get folks complaining about rerenders, then we can take another pass.

PR #392, opened 2023-02-10
feat(cli): use forge config for paths to src, test, out

dk1a:
Adds forge config getter; and replaces all hardcoded instances of `src`, `test`, `out` with getDir functions.
I also add some constants for `systems` and `components` dirs to the top of the files where they're used.

`inputDir` is removed from systemTypes because it only allows overriding the hardcoded `src`, which now isn't hardcoded (overriding `systems` directory wouldn't work well, and isn't useful for just 1 command).

`src` and `out` changes are seamless, but `test` changes require users to:
- either add `test = src/test` to `foundry.toml`
- or move their `src/test` to `test`

### DIFF
--- a/packages/cli/src/commands/call-system.ts
+++ b/packages/cli/src/commands/call-system.ts
@@ -1,6 +1,8 @@
 import { defaultAbiCoder as abi } from "ethers/lib/utils";
+import path from "path";
 import { Arguments, CommandBuilder } from "yargs";
 import { execLog } from "../utils";
+import { getTestDir } from "../utils/forgeConfig";
 
 type Options = {
   rpc?: string;
@@ -40,13 +42,15 @@ export const builder: CommandBuilder<Options, Options> = (yargs) =>
 export const handler = async (argv: Arguments<Options>): Promise<void> => {
   const { rpc, caller, world, systemId, argTypes, args, calldata, broadcast, callerPrivateKey, debug } = argv;
   const encodedArgs = calldata ?? (argTypes && args && abi.encode(argTypes, args)) ?? "";
+  const testDir = await getTestDir();
+
   await execLog("forge", [
     "script",
     "--fork-url",
     rpc ?? "http://localhost:8545", // default anvil rpc
     "--sig",
     "debug(address,address,string,bytes,bool)",
-    "src/test/utils/Debug.sol", // the cli expects the Debug.sol file at this path
+    path.join(testDir, "utils/Debug.sol"), // the cli expects the Debug.sol file at this path
     caller ?? "0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266", // default anvil deployer
     world,
     systemId || "",
--- a/packages/cli/src/commands/call-system.ts
+++ b/packages/cli/src/commands/call-system.ts
@@ -2,7 +2,7 @@ import { defaultAbiCoder as abi } from "ethers/lib/utils";
 import path from "path";
 import { Arguments, CommandBuilder } from "yargs";
 import { execLog } from "../utils";
-import { getTestDir } from "../utils/forgeConfig";
+import { getTestDirectory } from "../utils/forgeConfig";
 
 type Options = {
   rpc?: string;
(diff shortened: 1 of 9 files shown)


holic:
@alvrs would love to get this in so I can update templates before we release create-mud

packages/cli/src/utils/types.ts
```diff
@@ -5,6 +5,9 @@ import { extractIdFromFile } from "./ids";
 import { rmSync, writeFileSync } from "fs";
 import path from "path";
 import { filterAbi, forgeBuild } from "./build";
+import { getOutDir, getSrcDir } from "./forgeConfig";
+
+const systemsDir = "systems";
```
holic, line 10: I saw this defined in `trace.ts` too. Can we centralize these?

packages/cli/src/commands/system-types.ts
```diff
@@ -2,21 +2,15 @@ import type { Arguments, CommandBuilder } from "yargs";
...
 
 type Options = {
-  inputDir: string;
   outputDir: string;
 };
 
 export const command = "system-types";
 export const desc =
-  "Generates system type file. Note: assumes ABIs of all systems in ./abi and typechain generated types in ./types/ethers-contracts";
+  "Generates system type file. Note: assumes contracts of all systems in <forge src path>/systems folder, ABIs of all systems in ./abi and typechain generated types in ./types/ethers-contracts";
```
holic, line 10: Should we import/use `systemsDir` instead of hardcoding "systems" here?

packages/cli/src/utils/build.ts
```diff
@@ -1,15 +1,17 @@
...
 import { copyFileSync, mkdirSync, readdirSync, rmSync } from "fs";
 import path from "path";
+import { getOutDir } from "./forgeConfig";
 
-export async function forgeBuild(out = "out", options?: { clear?: boolean }) {
+export async function forgeBuild(options?: { clear?: boolean }) {
   if (options?.clear) {
+    const out = await getOutDir();
     console.log("Clearing forge build output directory", out);
     rmSync(out, { recursive: true, force: true });
```
holic, line 10: I think we can use `forge build --force` for this, rather than doing this manually.

image

https://book.getfoundry.sh/reference/forge/forge-build

holic (changes requested):

packages/cli/src/utils/build.ts
holic: Ahh, that's true. Might be worth asking the foundry team about the distinction/best practice. And maybe also an option to `forge build --clean` to clear output dir.

dk1a:
@holic wdyt about the constants.ts?

holic:
> wdyt about the constants.ts?

Good by me! Thanks!

holic (approved):
lgtm!

curious for @alvrs thoughts

alvrs (commented):
Looks good overall, just a couple minor comments

packages/cli/src/utils/build.ts
```diff
@@ -1,15 +1,17 @@
...
   }
 
   console.log("Running forge build");
-  const child = execa("forge", ["build", "-o", out], {
```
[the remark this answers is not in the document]
holic: A change I wanna make soon is generate ABIs separately from the standard forge output with
```
forge build --out abi --extra-output-files abi
```
Then we can `.gitignore` any non `*.abi.json` file so that ABIs can be checked in (and CI jobs to ensure these are up to date).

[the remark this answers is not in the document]
dk1a: > We should check whether that's still the case - if forge's output has changed all good

Tested it on a couple projects, the output is flat, and skipping explicit out doesn't change it

packages/cli/src/commands/call-system.ts
```diff
@@ -1,6 +1,8 @@
 import { defaultAbiCoder as abi } from "ethers/lib/utils";
+import path from "path";
 import { Arguments, CommandBuilder } from "yargs";
 import { execLog } from "../utils";
+import { getTestDir } from "../utils/forgeConfig";
```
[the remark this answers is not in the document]
dk1a: Updated method names

alvrs:
Do we have to rebase this on latest main to include the CI fix?

alvrs (approved):

PR #395, opened 2023-02-11
feat(store): expand schema type

dk1a:
Here I just refactor types, they're used and tested in #396

### DIFF
--- a/packages/store/src/Types.sol
+++ b/packages/store/src/Types.sol
@@ -1,84 +1,238 @@
 // SPDX-License-Identifier: MIT
 pragma solidity >=0.8.0;
 
-// TODO: add remaining types
+using { getStaticByteLength } for SchemaType global;
+
+// WARNING: SchemaType methods use hardcoded enum indexes, review them after any changes to the enum
+// TODO add and implement BYTES_ARRAY, STRING_ARRAY if they are needed (they are special nested arrays)
 enum SchemaType {
-  None, // The first `None` value ends the schema
-  Uint8,
-  Uint16,
-  Uint32,
-  Uint128,
-  Uint256,
-  Bytes4,
-  Bytes32,
-  Uint32Array,
-  Bytes24Array,
-  String,
-  Address,
-  AddressArray
+  UINT8,
+  UINT16,
+  UINT24,
+  UINT32,
+  UINT40,
+  UINT48,
+  UINT56,
+  UINT64,
+  UINT72,
+  UINT80,
+  UINT88,
+  UINT96,
+  UINT104,
+  UINT112,
+  UINT120,
+  UINT128,
+  UINT136,
+  UINT144,
+  UINT152,
+  UINT160,
+  UINT168,
+  UINT176,
+  UINT184,
+  UINT192,
+  UINT200,
+  UINT208,
+  UINT216,
+  UINT224,
+  UINT232,
+  UINT240,
+  UINT248,
+  UINT256,
+  INT8,
+  INT16,
+  INT24,
+  INT32,
+  INT40,
+  INT48,
+  INT56,
+  INT64,
+  INT72,
+  INT80,
+  INT88,
+  INT96,
+  INT104,
+  INT112,
+  INT120,
+  INT128,
+  INT136,
+  INT144,
+  INT152,
+  INT160,
+  INT168,
+  INT176,
+  INT184,
+  INT192,
+  INT200,
+  INT208,
+  INT216,
+  INT224,
+  INT232,
+  INT240,
+  INT248,
+  INT256,
+  BYTES1,
+  BYTES2,
+  BYTES3,
+  BYTES4,
+  BYTES5,
+  BYTES6,
+  BYTES7,
+  BYTES8,
+  BYTES9,
+  BYTES10,
+  BYTES11,
+  BYTES12,
+  BYTES13,
+  BYTES14,
+  BYTES15,
+  BYTES16,
+  BYTES17,
+  BYTES18,
+  BYTES19,
+  BYTES20,
+  BYTES21,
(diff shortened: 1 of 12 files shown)


packages/store/src/Types.sol
```diff
@@ -1,84 +1,238 @@
...
-  Uint128,
-  Uint256,
-  Bytes4,
-  Bytes32,
-  Uint32Array,
-  Bytes24Array,
-  String,
-  Address,
-  AddressArray
+  UINT8,
```
holic, line 9: starting the enum here feels a little dangerous with no null/undefined? should we keep `None` or similar for index 0?

packages/store/src/Types.sol
```diff
@@ -1,84 +1,238 @@
...
-  } else if (schemaType == SchemaType.Uint256 || schemaType == SchemaType.Bytes32) {
-    return 32;
-  } else if (schemaType == SchemaType.Address) {
-    return 20;
-  } else {
-    // Return 0 for all dynamic types
-    return 0;
+  uint256 val = uint8(schemaType);
+
+  if (val < 32) {
```
holic, line 216: does solidity support the use of enums as comparison values? e.g.

```sol
if (val <= SchemaType.UINT256) {
```

packages/store/src/Types.sol
dk1a: I think the downside of `None` is it just adds extra complexity of checking that something isn't `None` in `Store`, and it complicates `getStaticByteLength` a tiny bit

packages/store/src/Types.sol
dk1a: great idea, changed this and the offset subtraction

dk1a: upped get/set gas usage by ~50 though

packages/store/src/Types.sol
alvrs: Since we store `numStaticFields` in the schema itself, we could live without `None` (without `numStaticFields` and no `None` schema type indexers couldn't distinguish between `none` and `uint8`, but with `numStaticFields` they can.)

packages/store/src/Types.sol
```diff
@@ -1,84 +1,238 @@
...
-  } else if (schemaType == SchemaType.Uint128) {
-    return 16;
-  } else if (schemaType == SchemaType.Uint256 || schemaType == SchemaType.Bytes32) {
-    return 32;
-  } else if (schemaType == SchemaType.Address) {
-    return 20;
-  } else {
-    // Return 0 for all dynamic types
-    return 0;
+  uint256 val = uint8(schemaType);
```
alvrs, line 214: how about
```suggestion
  uint256 index = uint8(schemaType);
```
(tiny nit, feel free to ignore if you disagree)

alvrs (approved):
lgtm!

packages/store/src/Types.sol
holic: I guess it depends how often this is called. If it's every get/set, maybe not worth it?

packages/store/src/Types.sol
holic: I don't think I have my head fully wrapped around the store stuff, so long as its easy to otherwise check if something is "unset" then we should be fine here.

packages/store/src/Types.sol
dk1a: Yeah it's called all the time. It basically tells Store what's where in the tightly packed bytes data

packages/store/src/Types.sol
dk1a: I like index, changed it

packages/store/src/Types.sol
dk1a: Reverted, this method seems worth it to prioritize gas.
(It's the reason I suggested in discord to store SchemaType separately, then the method wouldn't be needed at all. But that's just too much complexity, and can be done later if we really want to.)

alvrs (approved):

packages/store/src/Types.sol
```diff
@@ -1,84 +1,238 @@
...
-  } else if (schemaType == SchemaType.Uint256 || schemaType == SchemaType.Bytes32) {
-    return 32;
-  } else if (schemaType == SchemaType.Address) {
-    return 20;
-  } else {
-    // Return 0 for all dynamic types
-    return 0;
+  uint256 index = uint8(schemaType);
+
+  if (index < 32) {
```
alvrs, line 216: just noticed we went back from `enum` to hardcoded values here - what's the gas implications of using enums? I feel like enums are much more readable, but on the other hand the enum probably doesn't change anymore (since it already includes all types), so maybe hardcoded numbers are fine if we have test cases to make sure it's always correct

dk1a: ~50 gas on each set and get, and since there will be a lot of those it will add up

holic: 50 gas in relation to how much gas per set in total? 50 out of 10k gas is different than 50 out of 100k gas

dk1a: ~2500, but it gets worse for multi-field tables. 1% at best, 5% at worst
(all these numbers are very approximate)

holic: imo this optimization is worth it then, as long as we have good test coverage and can reliably detect if anything in this enum changes (adding elements, changing order, etc.)

dk1a: yep, thorough tests are in the next PR (SchemaType by itself isn't meaningful to test, `staticByteLength` is part of the TightCoder suite)

alvrs: agree with this optimization in the context of this being a hot path + the underlying enum is not expected to change + has good test coverage incoming

PR #396, opened 2023-02-11
feat(store): add codegen for all types for tightcoder

dk1a:
Depends on #395
Not sure what to do with autogenerated files, tentatively not gitignoring them for now


packages/store/package.json
```diff
@@ -12,8 +12,9 @@
   "scripts": {
     "prepare": "yarn build && chmod u+x git-install.sh",
     "git:install": "bash git-install.sh",
-    "test": "forge test",
-    "build": "rimraf out && forge build --out out && yarn dist && yarn types",
+    "codegen": "ts-node ./scripts/codegen.ts",
+    "test": "yarn codegen && forge test",
+    "build": "yarn codegen && rimraf out && forge build --out out && yarn dist && yarn types",
```
alvrs, line 17: nit
```suggestion
    "build": "yarn codegen && rimraf out && forge build && yarn dist && yarn types",
```

packages/store/src/tightcoder/DecodeSlice.ejs
```diff
@@ -0,0 +1,81 @@
...
+
+library DecodeSlice {
+  /************************************************************************
+   *
+   *    uint8 - uint256
+   *
+   ************************************************************************/
+<% for (let i = 8; i <= 256; i += 8) { -%>
+
+  function decodeArray_uint<%= i %>(Slice self) internal pure returns (uint<%= i %>[] memory output) {
```
alvrs, line 16: ooc why did you decide for this naming format (`decodeArray_uint256`) instead of something like `decodeUint256Array`? Also, since the library is called `DecodeSlice`, I would have expected the function to be called something like `toUint256Array`, but no strong opinion.

packages/store/src/tightcoder/DecodeSlice.ejs
```diff
@@ -0,0 +1,81 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
+
```
alvrs, line 3: Should we put a comment on the top of the file saying something like "Autogenerated from DecodeSlice.ejs via yarn codegen, do not edit manually"?

alvrs (commented):
Great PR! Just very minor questions/nits. 

Also, do you think this PR would be a good place to rename `TightCoder` into `Serializer` (as proposed by @holic in https://github.com/latticexyz/mud/pull/387#discussion_r1106941413)?

packages/store/src/tightcoder/DecodeSlice.ejs
dk1a: In short I'm going for `<method>_<T>` as way to resemble generics

> why did you decide for this naming format (decodeArray_uint256) instead of something like decodeUint256Array?

The main reason is autogen. After _ comes the exact type name. I'd rather not do more manipulations than necessary since this is already quite complicated.

> Also, since the library is called DecodeSlice, I would have expected the function to be called something like toUint256Array

Similar reason - I want the "method name" to be purely as a prefix, and "type name" purely as a postfix, both separated by `_`.
For `encode`/`decode` - I feel like while writing it I had a harder time remembering which methods did what with `to`/`from`. With decode I had a better analogy to abi.decode etc.

packages/store/src/tightcoder/DecodeSlice.ejs
dk1a: > Also, do you think this PR would be a good place to rename `TightCoder` into `Serializer` (as proposed by @holic in [#387 (comment)](https://github.com/latticexyz/mud/pull/387#discussion_r1106941413))?

I'd rather do such encompassing renamings closer to v2 completion when we have world and autogen merged, to avoid changing WIP branches often. Will give more time to think about the name too

dk1a:
> Also, do you think this PR would be a good place to rename `TightCoder` into `Serializer` (as proposed by @holic in [#387 (comment)](https://github.com/latticexyz/mud/pull/387#discussion_r1106941413))?

I'd rather do such encompassing renamings closer to v2 completion when we have world and autogen merged, to avoid changing WIP branches often. Will give more time to think about the name too

packages/store/src/tightcoder/DecodeSlice.ejs
dk1a: Added one, see if it's ok. I'm not sure we need the comment customized to filename, but mentioning which file shouldn't be edited seems important
(for schema autogen I just write `/* Autogenerated file. Do not edit manually. */` but since here both files are together, it's not very clear)

alvrs (approved):
lgtm!

PR #397, opened 2023-02-13
feat: prepare cli for tablegen, add new schema types to network

dk1a:
Self-contained build pipeline changes, so I thought they should go ahead of the main tablegen PR.
Also new schema ts types are mostly finished, might come in handy if somebody's working on that side of things

About remappings:
```solidity
import { IStore } from "store/IStore.sol";
```
seems way too generic, I thought full paths might be better, like:
```solidity
```
(I'd personally prefer full paths for std-contracts/solecs too)


dk1a:
Refactored it a bit so cli doesn't import network and remains self-contained cjs for now.
Also schema types are exported by cli (that's the simplest way to make them work for now. Making cli into ESM would be one way to fix it; having another cjs package export them would be another, like store)

packages/cli/src/commands/gas-report.ts
```diff
@@ -44,7 +44,7 @@ export const desc = "Create a gas report";
 
 export const builder: CommandBuilder<Options, Options> = (yargs) =>
   yargs.options({
-    path: { type: "array", default: ["Gas.t.sol"], desc: "File containing the gas tests" },
+    path: { type: "array", string: true, default: ["Gas.t.sol"], desc: "File containing the gas tests" },
```
dk1a, line 47: this fixes a tsc error

dk1a:
Closing in favor of #417

PR #398, opened 2023-02-13
refactor(store): rename events

holic:
Nothing else is prefixed, so renaming these for consistency.

### DIFF
--- a/packages/store/src/IStore.sol
+++ b/packages/store/src/IStore.sol
@@ -5,9 +5,9 @@ import { SchemaType } from "./Types.sol";
 import { Schema } from "./Schema.sol";
 
 interface IStore {
-  event MudStoreSetRecord(bytes32 table, bytes32[] key, bytes data);
-  event MudStoreSetField(bytes32 table, bytes32[] key, uint8 schemaIndex, bytes data);
-  event MudStoreDeleteRecord(bytes32 table, bytes32[] key);
+  event StoreSetRecord(bytes32 table, bytes32[] key, bytes data);
+  event StoreSetField(bytes32 table, bytes32[] key, uint8 schemaIndex, bytes data);
+  event StoreDeleteRecord(bytes32 table, bytes32[] key);
 
   function registerSchema(bytes32 table, Schema schema) external;
 
--- a/packages/store/src/StoreCore.sol
+++ b/packages/store/src/StoreCore.sol
@@ -14,9 +14,9 @@ import { IStoreHook } from "./IStore.sol";
 
 library StoreCore {
   // note: the preimage of the tuple of keys used to index is part of the event, so it can be used by indexers
-  event MudStoreSetRecord(bytes32 table, bytes32[] key, bytes data);
-  event MudStoreSetField(bytes32 table, bytes32[] key, uint8 schemaIndex, bytes data);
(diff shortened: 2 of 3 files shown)


packages/store/src/StoreCore.sol
```diff
@@ -14,9 +14,9 @@ import { IStoreHook } from "./IStore.sol";
 
 library StoreCore {
   // note: the preimage of the tuple of keys used to index is part of the event, so it can be used by indexers
-  event MudStoreSetRecord(bytes32 table, bytes32[] key, bytes data);
-  event MudStoreSetField(bytes32 table, bytes32[] key, uint8 schemaIndex, bytes data);
-  event MudStoreDeleteRecord(bytes32 table, bytes32[] key);
+  event StoreSetRecord(bytes32 table, bytes32[] key, bytes data);
+  event StoreSetField(bytes32 table, bytes32[] key, uint8 schemaIndex, bytes data);
+  event StoreDeleteRecord(bytes32 table, bytes32[] key);
```
holic, line 19: It's a bummer we can't import these from `IStore` or the reverse so they can be defined in one place.

packages/store/src/StoreCore.sol
```diff
@@ -332,7 +332,7 @@ library StoreCoreInternal {
     Storage.store({ storagePointer: location, data: schema.unwrap() });
 
     // Emit an event to notify indexers
-    emit StoreCore.MudStoreSetRecord(SCHEMA_TABLE, key, abi.encodePacked(schema.unwrap()));
+    emit StoreCore.StoreSetRecord(SCHEMA_TABLE, key, abi.encodePacked(schema.unwrap()));
```
holic, line 335: Any reason this is `emit StoreCore.StoreSetRecord` instead of `emit StoreSetRecord`?

packages/store/src/IStore.sol
```diff
@@ -5,9 +5,9 @@ import { SchemaType } from "./Types.sol";
 import { Schema } from "./Schema.sol";
 
 interface IStore {
-  event MudStoreSetRecord(bytes32 table, bytes32[] key, bytes data);
-  event MudStoreSetField(bytes32 table, bytes32[] key, uint8 schemaIndex, bytes data);
-  event MudStoreDeleteRecord(bytes32 table, bytes32[] key);
+  event StoreSetRecord(bytes32 table, bytes32[] key, bytes data);
+  event StoreSetField(bytes32 table, bytes32[] key, uint8 schemaIndex, bytes data);
+  event StoreDeleteRecord(bytes32 table, bytes32[] key);
```
holic, line 10: Are these here mostly so we get them as part of the ABI?

Related to my other comment, I wonder if it would make sense to move them into a `library StoreEvents` and emit them as `emit StoreEvents.StoreSetRecord` etc.?

I don't feel that strongly about centralizing them but just trying to make sure we don't end up in a spot where we update one event signature but not the other.

holic: Looks like the events won't be exported in the ABI if used from a library: https://github.com/ethereum/solidity/issues/12573

alvrs: Yep that's the reason we have to declare them twice, it's unfortunate but seems like the only way to emit them in a library but make them part of a contract's ABI

alvrs (approved):
Lgtm!

PR #399, opened 2023-02-13
fix(network): add explicit return type to createFaucetService

dk1a:
Hopefully fixes the bug created (revealed?) by #328
Seems related to https://github.com/microsoft/TypeScript/issues/42873


holic (approved):

dk1a:
My guess about what happened here is #328 introduced a bit of extra complexity into imports/types, and since for faucet they're already very complicated, this broke typescript's return type inference with something like a stack overflow. Explicit return type fixes it because ts doesn't have to compute it, so the complexity goes back down

alvrs (approved):

PR #400, opened 2023-02-14
chore: update discord invite to land in #welcome instead of #team

ludns:
The previous invite link lands in #team and apparently errs for some people. Created a new one (that does not expire) to land in #welcome


ludns:
@alvrs lmk if we can merge this ASAP so we don't have an invalid invite on the doc website

alvrs (approved):

PR #402, opened 2023-02-14
feat(cli): add store config parser and basic schema generator

dk1a:
Very unfinished, has a lot of limitations, but it already works and can be useful in some situations. Expect future PRs to completely change everything about this feature.
See how I use it in `store` for examples


packages/store/mud.config.ts
```diff
@@ -0,0 +1,43 @@
+import { MUDConfig, ContractSchemaType } from "@latticexyz/cli";
+
+const config: MUDConfig = {
+  storePath: "../",
```
alvrs, line 4: what does `storePath` refer to?

packages/store/mud.config.ts
```diff
@@ -0,0 +1,43 @@
+import { MUDConfig, ContractSchemaType } from "@latticexyz/cli";
+
+const config: MUDConfig = {
```
alvrs, line 3: I think this config interface makes sense! We should probably have a `StoreConfig` that includes the table definitions, and then extend the type to a `WorldConfig` when used with the World framework. In the latter we could add a field for `route` in each `Table`, and a separate `systems` top level key to define the systems in a similar fashion.

packages/store/mud.config.ts
dk1a: `import { StoreSwitch } from "../StoreSwitch.sol";`
By default storePath is `@latticexyz/store/src/`, but can be overridden
I should probably rename it, added it quickly just to make imports work

dk1a:
looks like it was a typescript issue. This PR makes cli require a recent typescript version (not sure how recent, but no earlier than 4.0)

packages/cli/package.json
```diff
@@ -70,7 +70,9 @@
     "path": "^0.12.7",
     "solmate": "https://github.com/Rari-Capital/solmate.git#9cf1428245074e39090dceacb0c28b1f684f584c",
     "table": "^6.8.1",
+    "ts-node": "^10.9.1",
```
alvrs, line 73: why is ts-node a dependency instead of a dev dependency now?

packages/cli/src/commands/tablegen.ts
```diff
@@ -0,0 +1,40 @@
...
+import { getSrcDirectory } from "../utils/forgeConfig";
+
+type Options = {
+  configPath?: string;
+};
+
+const commandModule: CommandModule<Options, Options> = {
+  command: "tablegen",
+
+  describe: "TODO write description",
```
alvrs, line 15:
```suggestion
  describe: "Autogenerate MUD Store table libraries based on the config file",
```

packages/cli/src/mud.ts
```diff
@@ -1,8 +1,13 @@
-#!/usr/bin/env node
+#!/usr/bin/env ts-node-esm
```
alvrs, line 1: just for context, what's the reason we have to switch to `ts-node` here? Does this mean consumers of the cli need to have ts-node installed?

packages/cli/src/mud.ts
```diff
@@ -1,8 +1,13 @@
...
 
+// -S ts-node-esm --no-warnings
+// TODO env -S flag may not be well supported, gather feedback
```
alvrs, line 4: what does the flag do?

packages/cli/src/utils/config.ts
```diff
@@ -1,24 +1,149 @@
-// TODO unfinished placeholder
+import { findUpSync } from "find-up";
+import path from "path";
+import { SchemaType } from "@latticexyz/schema-type";
+
+// TODO require may or may not be needed, finish exploring the various configurations
+import { createRequire } from "module";
```
alvrs: What might it be needed for? Do you want to do this exploration in this PR or as a follow up?

packages/cli/src/utils/config.ts
```diff
@@ -1,24 +1,149 @@
...
       };
     };
   };
 }
+
+export async function loadConfig({ configPath }: { configPath?: string }) {
+  configPath = resolveConfigPath(configPath);
+
+  // TODO is this override really necessary?
+  process.env.TS_NODE_COMPILER_OPTIONS = `{ "module": "esnext" }`;
```
alvrs: what's the intention for this override?

packages/store/src/schemas/Mixed.sol
```diff
@@ -12,131 +13,166 @@ import { EncodeArray } from "../tightcoder/EncodeArray.sol";
...
-  function registerSchema(uint256 tableId, IStore store) internal {
-    store.registerSchema(tableId, getSchema());
+  function registerSchema(uint256 _tableId, IStore _store) internal {
+    _store.registerSchema(_tableId, getSchema());
   }
 
   /** Set the table's data */
   function set(
-    uint256 tableId,
+    uint256 _tableId,
```
alvrs, line 46: this PR doesn't include the option to generate "tables" instead of "schemas" yet correct? (table = hardcoding the tableid in the library so it doesn't have to be passed as an argument)

alvrs (commented):
Very impressive! Left a couple comments about the todos (mostly to get more context), but nothing really blocking. As soon as we merge `v2` into `main` (hopefully soon) we should go through the whole codebase, open issues for the TODOs and link the issue in the code comment. In my experience TODO comments are often forgotten if not tracked somewhere outside the code too.

packages/cli/src/utils/config.ts
alvrs: (this is now in `storeConfig.ts`)

packages/cli/src/utils/config.ts
alvrs: (this is now in `storeConfig.ts`)

packages/cli/package.json
dk1a: shebang relies on it. It could be a peerDep, but bad devex. Ultimately it's needed to load the ts config file

packages/cli/src/mud.ts
dk1a: https://github.com/latticexyz/mud/pull/402#discussion_r1117404537 yep. `mud` cli command could be split into 2, so things that don't need config wouldn't need ts-node either. But most things will eventually need config, and 2 cmds sounds bad

packages/cli/src/mud.ts
dk1a: I'm not sure if I need it now, but if there're `Experimental warnings` in console, this flag could remove them. Keeping it in comments to not look for it later

packages/cli/src/utils/config.ts
dk1a: yea I just renamed it. I might do `.cts` and `.ts` support later and might need requires (or might not).
Right now I'm also doing some config reorg so it's easier for us to work on it simultaneously (this require stuff will be segregated into a low level file you wouldn't need to change).
Should take ~hour or less, can merge then.

packages/cli/src/utils/config.ts
dk1a: I think it makes schema-type import in user config work well regardless of user's tsconfig. TODO is about me thoroughly figuring out the exact situations it's needed and possible workarounds

packages/cli/src/mud.ts
alvrs: do you know how much ts node impacts the bundle size? (Old bundle size is 978.82 KB for reference)

packages/store/src/schemas/Mixed.sol
dk1a: no but that's coming soon, today-tomorrow

packages/cli/src/mud.ts
dk1a: Screenshot 2023-02-24 210712

dk1a: looks like surprisingly little

alvrs (approved):

alvrs (approved):

packages/store/src/schemas/Mixed.sol
```diff
@@ -12,131 +13,166 @@ import { EncodeArray } from "../tightcoder/EncodeArray.sol";
...
   }
 
   /** Register the table's schema */
   function registerSchema(uint256 tableId) internal {
     StoreSwitch.registerSchema(tableId, getSchema());
   }
 
-  function registerSchema(uint256 tableId, IStore store) internal {
-    store.registerSchema(tableId, getSchema());
+  function registerSchema(uint256 _tableId, IStore _store) internal {
```
holic, line 40: ooc why the change to underscore-prefixed arguments?

packages/cli/src/utils/errors.ts
```diff
@@ -0,0 +1,54 @@
...
+    message: (...errors: string[]) => `There's one or more errors in your config file:
+
+  ${errors.join("\n  ")}
+
+To learn more about MUD's configuration, please go to [TODO link to docs]`,
+    description: `You have one or more errors in your config file.
+
+Check the error message for details, or [TODO link to docs]`,
+  },
+};
```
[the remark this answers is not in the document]
dk1a: > [wagmi has some really nice errors](https://github.com/wagmi-dev/wagmi/blob/main/packages/core/src/errors.ts) that we can pull inspiration from.

They are indeed very nice, I totally agree that errors can be improved much further.
They weren't a significant feature of this PR, I just wanted to start some improvements going

packages/cli/src/utils/tablegen.ts
```diff
@@ -0,0 +1,60 @@
...
+  return renderedTables;
+}
+
+function getSchemaTypeInfo(schemaType: SchemaType): RenderSchemaType {
+  const staticByteLength = getStaticByteLength(schemaType);
+  const isDynamic = staticByteLength === 0;
+  const typeId = SchemaTypeId[schemaType];
+  return {
+    typeId,
+    typeLoc: isDynamic ? typeId + " memory" : typeId,
```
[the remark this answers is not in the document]
dk1a: typeWithLocation would be even better I think

packages/cli/src/utils/errors.ts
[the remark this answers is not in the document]
holic: yep not at all blocking, but we should come back to this before release

packages/store/src/schemas/Mixed.sol
dk1a: I only allow field/key names that start with `[a-z]`, so they're guaranteed to never clash with underscore-prefixed internal stuff

Issue #393, opened 2023-02-10
Proposal: World framework v2


alvrs:
## Table of contents

- [Abstract](#abstract)
- [Motivation](#motivation)
	- [Issues with the previous approach](#issues-with-the-previous-approach)
	- [Relation to *Store*](#relation-to-store)
- [Design goals](#design-goals)
- [Basic Concepts](#basic-concepts)
	- [Separation of data and logic into tables and systems](#separation-of-data-and-logic-into-tables-and-systems)
	- [Routes](#routes)
	- [Access control](#access-control)
- [Code exploration](#code-exploration)
- [Advanced concepts / extensions](#advanced-concepts--extensions)
	- [Hooks for system calls](#hooks-for-system-calls)
	- [Interface proxy](#interface-proxy)
	- [Register function selectors for root systems](#register-function-selectors-for-root-systems)
	- [Subsystems](#subsystems)
	- [Generic account delegation / approval pattern](#generic-account-delegation--approval-pattern)
	- [Modules](#modules)
	- [Multicall](#multicall)
- [Acknowledgements](#acknowledgements)

## Abstract

This proposal outlines a new version of the World framework, which leverages the *Store* library to move away from individual component contracts to manage state and instead towards managing state in access-controlled tables (backwards compatible to components) in a central World contract. In addition it proposes a new way of interacting with systems via a central entry point on the World contract. This new architecture promises to be more gas-efficient and powerful, enabling a long list of extensions described in the last part.

## Motivation

The goal of the World framework is to facilitate complex and extendable on-chain applications. If used exclusively by a first party developer it can be compared to the [Diamond pattern](https://eips.ethereum.org/EIPS/eip-2535). But the goal of the World framework is allow both first and third party developers to extend existing applications, after they’ve been deployed, to facilitate the creation of [Autonomous Worlds](https://0xparc.org/blog/autonomous-worlds).

### Issues with the previous approach

In the first version of MUD, the World contract served as a central registry for components and systems and as a central source of update events for indexers and clients to reconstruct the component state. State and logic was separated into components and systems, with components being separate contracts (each managing its own access control) and systems being bound to a World (with the World address stored in their storage). This approach led to a couple issues / inconveniences:

- Each system contract can only be part of a single World (unlike in the Diamond pattern where Facets can be part of different diamonds, making it possible to create a new instance of an app by redeploying a single DiamondProxy contract and linking existing Facets to it)
- Unnecessary gas overhead for reading from / writing to components from a system
[...]

PR #405, opened 2023-02-16
feat(world): add World prototype

alvrs:
Prototype for #393


packages/store/src/Store.sol
```diff
@@ -0,0 +1,42 @@
...
+  function getSchema(uint256 table) public view virtual returns (Schema schema) {
+    schema = StoreCore.getSchema(table);
+  }
+
+  // Get full record (including full array, load schema from storage)
+  function getRecord(uint256 table, bytes32[] calldata key) public view virtual returns (bytes memory data) {
+    data = StoreCore.getRecord(table, key);
+  }
+
+  // Get full record (including full array)
```
dk1a, line 23: > including full array

something like `including the full dynamic data` might be better

packages/store/src/StoreView.sol
```diff
@@ -2,74 +2,42 @@
...
 
-  constructor() {
-    StoreCore.initialize();
-  }
-
-  function getSchema(uint256 table) public view virtual returns (Schema schema) {
-    schema = StoreCore.getSchema(table);
-  }
-
   function registerSchema(uint256, Schema) public virtual {
```
dk1a, line 14: I'd add `override` to the relevant `StoreView` methods

packages/world/package.json
```diff
@@ -0,0 +1,51 @@
...
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/latticexyz/mud.git",
+    "directory": "packages/world"
+  },
+  "scripts": {
+    "prepare": "yarn build && chmod u+x git-install.sh",
+    "git:install": "bash git-install.sh",
+    "test": "forge test",
+    "build": "rimraf out && forge build --out out && yarn dist && yarn types",
```
dk1a, line 16: `--out out` doesn't seem necessary anymore

packages/world/remappings.txt
```diff
@@ -0,0 +1,3 @@
+ds-test/=../../node_modules/ds-test/src/
+forge-std/=../../node_modules/forge-std/src/
+store/=../../node_modules/@latticexyz/store/src/
```
dk1a, line 3: we're getting a lot of packages, some with really generic names
wdyt about a more verbose remapping, or even a full path?
(I kinda like how hardhat just doesn't have remappings, makes dependencies more verbose but much easier to understand)

packages/world/src/World.sol
```diff
@@ -0,0 +1,422 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
+
+import { console } from "forge-std/console.sol";
+import { Store, IStoreHook } from "store/Store.sol";
+import { StoreCore } from "store/StoreCore.sol";
+import { Schema } from "store/Schema.sol";
+
+import { OwnerTable, tableId as OwnerTableId } from "./tables/OwnerTable.sol";
```
dk1a, line 9: more relevant for autogen than world: UX would probably benefit from having `tableId` have a non-generic name at declaration in all tables, so you'd just import `OwnerTableId`

packages/world/src/World.sol
```diff
@@ -0,0 +1,422 @@
...
+  /**
+   * Register a new route by extending an existing route
+   */
+  function registerRoute(string calldata baseRoute, string calldata subRoute) public returns (bytes32 routeId) {
+    // Require subroute to be a valid route fragment (start with `/` and don't contain any further `/`)
+    if (!_isSingleLevelRoute(subRoute)) revert RouteInvalid(subRoute);
+
+    // Require base route to exist (with a special check for the root route because it's empty and fails the `has` check)
+    if (!(keccak256(bytes(baseRoute)) == ROOT_ROUTE_ID || RouteTable.has(keccak256(bytes(baseRoute)))))
+      revert RouteInvalid(baseRoute);
```
dk1a, line 54:
```suggestion

    // Require base route to exist (with a special check for the root route because it's empty and fails the `has` check)
    bytes32 baseRouteId = keccak256(bytes(baseRoute));
    if (!(baseRouteId == ROOT_ROUTE_ID || RouteTable.has(baseRouteId)))
      revert RouteInvalid(baseRoute);
```

dk1a, line 54:
```suggestion
    bytes32 baseRouteId = keccak256(bytes(baseRoute));
    if (!(baseRouteId == ROOT_ROUTE_ID || RouteTable.has(baseRouteId)))
      revert RouteInvalid(baseRoute);
```

dk1a (commented):
lgtm overall, just some nits/questions

packages/world/remappings.txt
holic: +1 to verbose/namespaced mapping, and would prefer to keep it tied to the npm namespace, i.e. `@latticexyz/store` for now and `@mud.dev/store` in the future

packages/world/src/schemas/AddressToBytes32.sol
```diff
@@ -0,0 +1,71 @@
...
+import { IStore } from "store/IStore.sol";
+import { StoreSwitch } from "store/StoreSwitch.sol";
+import { StoreCore } from "store/StoreCore.sol";
+import { SchemaType } from "store/Types.sol";
+import { Bytes } from "store/Bytes.sol";
+import { Schema, SchemaLib } from "store/Schema.sol";
+import { PackedCounter, PackedCounterLib } from "store/PackedCounter.sol";
+
+// -- User defined schema and tableId --
+struct AddressToBytes32Schema {
```
holic, line 14: What's the reason for both `Address` and `AddressToBytes32`?

holic (commented):
mostly just questions from me!

packages/store/src/StoreView.sol
```diff
@@ -2,74 +2,42 @@
...
+    bytes32[] calldata,
     uint8,
-    bytes memory
+    bytes calldata
   ) public virtual {
     revert StoreView_NotImplemented();
   }
 
-  function registerHook(uint256, IStoreHook) public virtual {
+  function registerStoreHook(uint256, IStoreHook) public virtual {
```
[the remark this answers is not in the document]
dk1a: since not just arguments, but this whole method isn't usable, I rather think it's better for docs to add something like
```solidity
/**
 * Not implemented in StoreView
 */
```

packages/world/foundry.toml
```diff
@@ -0,0 +1,9 @@
+[profile.default]
+ffi = false
+fuzz_runs = 256
+optimizer = true
+optimizer_runs = 1000000
```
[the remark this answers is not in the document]
dk1a: personally I'd want even more, except that'd impact performance very noticably.
From my experience with [Slice](https://github.com/dk1a/solidity-stringutils/blob/d0a7466de5038da11302e59a747c6697770e49fc/test/Slice.t.sol) a lot of runs can really help with `assume` and weird edge cases

packages/store/src/StoreView.sol
alvrs: which methods to you mean in particular? (`StoreView` just "implements" the `IStore` functions missing from `Store`, but doesn't really override anything)

packages/world/src/World.sol
alvrs: agree! This import wasn't actually used, in the current prototype I have the table libs wrapping around the schema libs to remove the need for explicit handling of the table id, but happy to discuss alternative approaches. How were you thinking of handling it with the autogen?

packages/world/src/World.sol
```diff
@@ -0,0 +1,422 @@
...
+  /************************************************************************
+   *
+   *    REGISTRATION METHODS
+   *
+   ************************************************************************/
+
+  /**
+   * Register a new route by extending an existing route
+   */
+  function registerRoute(string calldata baseRoute, string calldata subRoute) public returns (bytes32 routeId) {
```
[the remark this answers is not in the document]
alvrs: yeah I agree - didn't do the refactor of routeId from `bytes32` to `uint256` yet, but I agree it would be better. Working on it.

packages/store/src/StoreView.sol
dk1a: Yeah that's kinda what I meant - to use override for implemented interface functions. https://github.com/ethereum/solidity/issues/8281 has some discussion on that, and I'm generally in favor of using `overrides` to indicate which functions come from interfaces and not just other contracts. But it's really fine either way, up to you.

packages/world/src/World.sol
dk1a: Since I can generate the whole thing, the separation of schema/table stops being necessary. If you want 2 same-schema tables, autogen handles it for you, and since config is in typescript, you can use variables if you want to reuse schemas.
Do you see a use-case where schema is useful by itself other than reusability?

packages/world/src/World.sol
```diff
@@ -0,0 +1,422 @@
...
+    RouteAccessTable.set({ routeId: baseRouteId, caller: address(system), access: true });
+  }
+
+  /**
+   * Grant access to a given route
+   */
+  function grantAccess(string calldata route, address grantee) public {
+    // Require the caller to own the route
+    bytes32 routeId = keccak256(bytes(route));
+    if (OwnerTable.get(routeId) != msg.sender) revert RouteAccessDenied(route, msg.sender);
```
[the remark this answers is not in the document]
alvrs: A small issue I have with that is usually we need the `routeId` later in the logic, so either we compute the `routeId` twice (once in the main method, once in the helper), or we pass in the `routeId` to the helper directly, which then requires us to grab the `route` pre-image from the `RouteTable` to be able to emit it as part of the error. (In some cases we only have `routeId`, eg `registerSchema`, so we'd need the helper that works with `routeId` anyways.)

The other thing is if we revert in the helper method, it's slightly less clear from looking at the method what the possible outcomes are (because you have to check the helper method first to see that the method can revert with a certain error). Since the check here is pretty simple and readable, I have a slight preference for throwing the error explicitly in the main function.

packages/world/src/World.sol
alvrs: One minor use case I see is if you wanted to create "generic systems" that can be called with table ids as arguments (simple example, an `IncrementSystem` that doesn't hard code its `CounterTable` but receives it as an argument, so it can be used for different tables that fit the schema

dk1a: Yeah I was thinking how to do something along those lines. How about 2 autogen modes - specific table, and generic schema? That way we should keep the benefits, but reduce the file count/complexity

packages/world/foundry.toml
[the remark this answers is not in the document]
alvrs: `optimizer_runs = 1000000` is probably not helpful, this is just some default config I copied from other packages, not sure why it's set so high. Will decrease to `256` for now.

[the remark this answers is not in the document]
alvrs: Okay interestingly I do see some negative gas implications (using `mud gas-report`) if going below 3000 optimizer runs. Will keep it at that for now. (Also didn't really notice a change in build/test time with different optimizer run values)

packages/world/hardhat.config.ts
```diff
@@ -0,0 +1,23 @@
...
+        runs: 200,
+      },
+    },
+  },
+  docgen: {
+    outputDir: "API",
+  },
+};
+
+export default config;
```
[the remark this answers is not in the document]
alvrs: Unfortunately we still need the hardhat setup to use `solidity-docgen` to parse out NatSpec into our markdown docs

packages/world/src/schemas/RouteAccess.sol
```diff
@@ -0,0 +1,86 @@
...
+    StoreSwitch.registerSchema(tableId, getSchema());
+  }
+
+  function registerSchema(uint256 tableId, IStore store) internal {
+    store.registerSchema(tableId, getSchema());
+  }
+
+  /** Set the table's data */
+  function set(
+    uint256 tableId,
```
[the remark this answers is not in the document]
alvrs: agree, would slightly prefer `tableId` i think, but so far we mostly used `table` in `store`. To avoid merge conflicts with @dk1a's store PRs I wouldn't change that in this PR. Also I expect the files inside `schemas` and `tables` to be replaced by auto-generated ones once that's ready

packages/world/src/schemas/RouteAccess.sol
```diff
@@ -0,0 +1,86 @@
...
+    bytes32 routeId,
```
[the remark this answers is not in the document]
alvrs: yep agree, changing that in this PR

packages/world/src/schemas/Address.sol
```diff
@@ -0,0 +1,65 @@
...
+  function get(uint256 tableId, bytes32 key) internal view returns (address) {
+    bytes32[] memory keyTuple = new bytes32[](1);
+    keyTuple[0] = key;
+    bytes memory blob = StoreSwitch.getRecord(tableId, keyTuple);
+    return address(bytes20(blob));
+  }
+
+  function get(
+    uint256 tableId,
+    IStore store,
```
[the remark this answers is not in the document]
alvrs: I agree store, table, key would be more intuitive here. The order is mostly because `tableId` is the argument that is "removed" by the table libraries when they wrap the schema libraries, so it's the first argument of all the functions in the schema library. Other methods except from this one don't have a `store` reference (because it's implicitly either `this` or `msg.sender` via `StoreSwitch`) and it feels weird to remove the first argument in the table libraries for every function except this one where we'd remove the second argument.

packages/world/src/schemas/AddressToBytes32.sol
alvrs: `Address` represents `mapping(bytes32 => address)`, `AddressToBytes32` represents `mapping(address => bytes32)`. `Address` is the underlying schema for the `OwnerTable` (in this case only `address` can be owner of things), `AddressToBytes32` is the underlying schema for the `SystemRouteTable` to store the routeId of a given system. The purpose of the latter is to check whether a system is already registered in a World. Systems (= contracts) can only be registered once per World, because otherwise it would mess up the access mechanism based on addresses.

packages/world/src/World.sol
alvrs: sounds good to me!

packages/store/src/StoreView.sol
alvrs: Oh I see, after reading through the discussion in https://github.com/ethereum/solidity/issues/8281 I'd tend to not doing this, at least not for now - there are good arguments for both sides, but since we don't do it currently with other interfaces in the MUD codebase, I'd aim for consistency in this case.

packages/world/foundry.toml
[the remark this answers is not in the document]
holic: yeah I figured there was prob diminishing returns for tons of runs, but sounds like you found a sweet spot for good gas + fewer runs

packages/world/hardhat.config.ts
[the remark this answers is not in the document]
holic: Oh bummer!

Looks like foundry has a docs command now: https://github.com/foundry-rs/foundry/pull/2701

packages/world/remappings.txt
```diff
@@ -0,0 +1,3 @@
+ds-test/=../../node_modules/ds-test/src/
+forge-std/=../../node_modules/forge-std/src/
+@latticexyz/store/=../../node_modules/@latticexyz/store/src/
```
dk1a, line 3: wdyt about `@latticexyz/store/src/=../../node_modules/@latticexyz/store/src/`?
(or just `@latticexyz/=../../node_modules/@latticexyz/`)
The extra `src/` is only 4 symbols, but this essentially gets rid of remappings.
It could even truly get rid of remappings if using `include_path=../../node_modules` in foundry.toml (except it freaks out in some edge cases with nested deps, but eventually it might work well).

packages/world/index.yml
```diff
@@ -0,0 +1,2 @@
+label: store
```
dk1a, line 1: using world to align `schema-types` package to it. Noticed that this should probably be `world`

packages/world/hardhat.config.ts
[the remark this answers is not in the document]
alvrs: Oh nice, we should look into replacing the hardhat doc gen with the foundry one in the future in all our packages for a leaner setup.

dk1a (approved):
lgtm

packages/store/src/IStore.sol
```diff
@@ -16,20 +16,20 @@ interface IStore {
...
-    bytes32[] memory key,
+    bytes32[] calldata key,
     uint8 schemaIndex,
-    bytes memory data
+    bytes calldata data
   ) external;
 
   // Register hooks to be called when a record or field is set or deleted
-  function registerHook(uint256 table, IStoreHook hooks) external;
+  function registerStoreHook(uint256 table, IStoreHook hooks) external;
```
holic, line 32: this should probably be `IStoreHook hook` (singular) not `IStoreHook hooks`?

packages/store/src/Slice.sol
```diff
@@ -36,16 +36,24 @@ library SliceLib {
     return Slice.wrap((_pointer << 128) | (data.length & MASK_LEN));
   }
 
+  /**
+   * @dev Subslice a bytes array using the given start index until the end of the array (without copying data)
+   */
+  function getSubslice(bytes memory data, uint256 start) internal pure returns (Slice) {
+    return getSubslice(data, start, data.length);
+  }
```
holic, line 44: naming nit: what about just `slice`?

packages/world/tsconfig.json
```diff
@@ -0,0 +1,102 @@
...
+    // "noImplicitOverride": true,                       /* Ensure overriding members in derived classes are marked with an override modifier. */
+    // "noPropertyAccessFromIndexSignature": true,       /* Enforces using indexed accessors for keys declared using an indexed type */
+    // "allowUnusedLabels": true,                        /* Disable error reporting for unused labels. */
+    // "allowUnreachableCode": true,                     /* Disable error reporting for unreachable code. */
+
+    /* Completeness */
+    // "skipDefaultLibCheck": true,                      /* Skip type checking .d.ts files that are included with TypeScript. */
+    // "skipLibCheck": true /* Skip type checking all .d.ts files. */
+  }
+  // "exclude": ["dist", "**/*.spec.ts", "packages"]
```
holic, line 101: we can prob remove all these comments to make the tsconfig easier to read

.vscode/settings.json
```diff
@@ -9,6 +9,7 @@
     "ds-test/=node_modules/ds-test/src/",
     "solecs/=node_modules/@latticexyz/solecs/src/",
     "royalty-registry/=node_modules/@manifoldxzy/royalty-registry/contracts/",
-    "std-contracts/=node_modules/@latticexyz/std-contracts/src/"
+    "std-contracts/=node_modules/@latticexyz/std-contracts/src/",
```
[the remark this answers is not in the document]
dk1a: considering it's a breaking change and not relevant for v2, is it worth it for v1 packages?

packages/store/src/Slice.sol
dk1a: I named it that way pretty much just to avoid confusion with the `Slice` type, and make searching for it easier

.vscode/settings.json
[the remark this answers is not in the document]
holic: we do breaking changes all the time 🙈

PR #412, opened 2023-02-17
feat(network): allow retries inside tx queue

Kooshaba:
i'm looking for some feedback on whether this is reasonable or not.

there are certain situations where we incorrectly estimate gas due to state growth while the transaction is still in flight. i'd like these to automatically retry because it should succeed with the same parameters upon with a new estimate.

there is one thing i don't know: should the nonce increase if there is a gas estimation mistake?

### DIFF
--- a/packages/network/src/createTxQueue.ts
+++ b/packages/network/src/createTxQueue.ts
@@ -227,14 +227,30 @@ export function createTxQueue<C extends Contracts>(
           error &&
           (("code" in error && error.code === "NONCE_EXPIRED") ||
             JSON.stringify(error).includes("transaction already imported"));
+        const incorrectGasEstimation =
+          error &&
+          (JSON.stringify(error).includes("gas required exceeds allowance") ||
+            JSON.stringify(error).includes("intrinsic gas too low"));
         console.log(
           `[TXQueue] TX Sent (error=${!!error}, isMutationError=${!!isNonViewTransaction} incNonce=${!!shouldIncreaseNonce} resetNonce=${!!shouldResetNonce})`
         );
-        // Nonce handeling
-        if (shouldIncreaseNonce) incNonce();
+
         if (shouldResetNonce) await resetNonce();
-        // Bubble up error
-        if (error) txRequest.cancel(error);
+        const canRetry = incorrectGasEstimation || shouldResetNonce;
+
+        if (canRetry) {
+          queue.add(uuid(), {
+            execute: txRequest.execute,
+            estimateGas: txRequest.estimateGas,
+            cancel: txRequest.cancel,
+            stateMutability,
+          });
+        } else {
+          // Bubble up error
+          if (error) txRequest.cancel(error);
+          // Nonce handeling
+          if (shouldIncreaseNonce) incNonce();
+        }
       }
     });
 
--- a/packages/network/src/createTxQueue.ts
+++ b/packages/network/src/createTxQueue.ts
@@ -25,9 +25,10 @@ export function createTxQueue<C extends Contracts>(
   computedContracts: IComputedValue<C> | IObservableValue<C>,
   network: Network,
   gasPrice$: BehaviorSubject<number>,
-  options?: { concurrency?: number; devMode?: boolean }
+  options?: { concurrency?: number; devMode?: boolean; maxRetries?: number }
 ): { txQueue: TxQueue<C>; dispose: () => void; ready: IComputedValue<boolean | undefined> } {
   const { concurrency } = options || {};
+  const maxRetries = options?.maxRetries ?? 3;
 
   const queue = createPriorityQueue<{
     execute: (
@@ -37,6 +38,7 @@ export function createTxQueue<C extends Contracts>(
     estimateGas: () => BigNumberish | Promise<BigNumberish>;
     cancel: (error: any) => void;
     stateMutability?: string;
+    retryCount: number;
   }>();
   const submissionMutex = new Mutex();
   const _nonce = observable.box<number | null>(null);
@@ -166,6 +168,7 @@ export function createTxQueue<C extends Contracts>(
       estimateGas,
       cancel: (error?: any) => reject(error ?? new Error("TX_CANCELLED")),
       stateMutability,
+      retryCount: 0,
     });
 
     // Start processing the queue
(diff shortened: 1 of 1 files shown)


packages/network/src/createTxQueue.ts
```diff
@@ -227,14 +227,30 @@ export function createTxQueue<C extends Contracts>(
...
-        if (error) txRequest.cancel(error);
+        const canRetry = incorrectGasEstimation || shouldResetNonce;
+
+        if (canRetry) {
+          queue.add(uuid(), {
+            execute: txRequest.execute,
+            estimateGas: txRequest.estimateGas,
+            cancel: txRequest.cancel,
+            stateMutability,
+          });
```
holic, line 247: we should probably count the number of retries so this doesn't end up infinitely looping because of bad gas estimate

Kooshaba: ya fair

holic: if txs are queued in order, what happens to all the txs after this one in the queue if this one is retried?

Kooshaba: yeah this would all collapse if you queued a bunch of txs that depend on each other

Kooshaba: but i believe only the first one should retry once, then everything will properly fail in confirmation and not retry

packages/network/src/createTxQueue.ts
```diff
@@ -25,9 +25,10 @@ export function createTxQueue<C extends Contracts>(
   computedContracts: IComputedValue<C> | IObservableValue<C>,
   network: Network,
   gasPrice$: BehaviorSubject<number>,
-  options?: { concurrency?: number; devMode?: boolean }
+  options?: { concurrency?: number; devMode?: boolean; maxRetries?: number }
 ): { txQueue: TxQueue<C>; dispose: () => void; ready: IComputedValue<boolean | undefined> } {
   const { concurrency } = options || {};
+  const maxRetries = options?.maxRetries ?? 3;
```
alvrs, line 31: should we make `0` the default for now to not unexpectedly change then behaviour for other consumers?

packages/network/src/createTxQueue.ts
```diff
@@ -227,14 +230,31 @@ export function createTxQueue<C extends Contracts>(
           error &&
           (("code" in error && error.code === "NONCE_EXPIRED") ||
             JSON.stringify(error).includes("transaction already imported"));
+        const incorrectGasEstimation =
+          error &&
+          (JSON.stringify(error).includes("gas required exceeds allowance") ||
+            JSON.stringify(error).includes("intrinsic gas too low"));
```
alvrs, line 236: do these errors occur in different situations? What are the situations in which each one occurs?

packages/network/src/createTxQueue.ts
```diff
@@ -227,14 +230,31 @@ export function createTxQueue<C extends Contracts>(
...
         console.log(
           `[TXQueue] TX Sent (error=${!!error}, isMutationError=${!!isNonViewTransaction} incNonce=${!!shouldIncreaseNonce} resetNonce=${!!shouldResetNonce})`
         );
-        // Nonce handeling
-        if (shouldIncreaseNonce) incNonce();
+
```
alvrs, line 240: i think if the transaction hit the mempool and wasn't a view function call, we need to increase the nonce in any case. We should verify that though.

packages/network/src/createTxQueue.ts
```diff
@@ -227,14 +230,31 @@ export function createTxQueue<C extends Contracts>(
...
-        // Bubble up error
-        if (error) txRequest.cancel(error);
+        const canRetry = txRequest.retryCount <= maxRetries && (incorrectGasEstimation || shouldResetNonce);
+
+        if (canRetry) {
+          queue.add(uuid(), {
+            execute: txRequest.execute,
+            estimateGas: txRequest.estimateGas,
+            cancel: txRequest.cancel,
+            stateMutability,
```
alvrs, line 249: nit, would this be possible? (haven't tested it)
```suggestion
            ...txRequest
```

alvrs (commented):
if this works it would be a great ux improvement, but we should tripple check and test if it actually all works as expected before merging

packages/network/src/createTxQueue.ts
Kooshaba: i kind of want to change the default behavior for people to add to more of the magic of MUD. do you see a reason why it would be bad or just don't want to surprise people?

Kooshaba: 3 is a little aggressive though... 1 probably makes more sense as a default

alvrs: Mainly to avoid surprising people, in general I think it's a good default behaviour, so also happy with enabling by default. Agree 1 is a better default than 3 though. We should also add a note about this behaviour in the comment above the function, so it can get parsed into the documentation here https://mud.dev/packages/network/api/#createtxqueue

packages/network/src/createTxQueue.ts
alvrs: My main concern here is with handling situations where we need to increase the nonce ([tx was included in a block but failed](https://etherscan.io/tx/0xbf7146c7799b8c1e486260726ad5f383f1e98ee9c726b6a76a7e4d1e57f7c298)) vs those where we don't (tx was declined by ethers or maybe the RPC?). Is there any information in the error that let's us tell for sure if the tx was included in a block? Maybe an abi encoded "Out of Gas" exception or something?

packages/network/src/createTxQueue.ts
Kooshaba: ah yeah this is right

packages/network/src/createTxQueue.ts
ludns: if it's included in a block you get a tx hash back and no error. to check error in a tx in a block you need to fetch the tx details from the tx hash.

Kooshaba: > if it's included in a block you get a tx hash back and no error. to check error in a tx in a block you need to fetch the tx details from the tx hash.

now that i know this, i think this whole PR is misguided. by fetching the tx details we are starting another round trip to the chain. at that point we might as well submit again via other means. i think this whole retrying issue will have to be moved closer to the application layer.

Kooshaba:
i'm closing this after doing more research on how to properly get tx errors. [see comment](https://github.com/latticexyz/mud/pull/412/files/072c0a6b6cff521e9bcf3c67dad37bf9d2fb1eff#r1119295477)

PR #415, opened 2023-02-17
feat: v2 event decoding

holic:
Parses v2 events into schema registrations and data to populate client ECS with it.

TODO
- [x] decode arrays
- [x] use generics in config def so we can carry table types through
- [ ] use generics in decoding to carry schema types through
- [x] figure out key tuples for client (concatenate? json encode?)
- [x] lazily fetch metadata
- [x] update missing key warning only for non-numeric keys
- [ ] add support for key schema
- [x] add peer deps
- [x] ensure no local/linked deps
- [x] remove circular dep https://github.com/latticexyz/mud/pull/415#discussion_r1132935137
- [ ] add tests for long-lived low-level stuff
- [ ] fetch from ECS cache before RPC for schemas/metadata
- [x] figure out how to get things to compile with `mud.config` imports
- [ ] make `parseConfig` more strongly typed and carry through statically defined values
- [ ] flip `StoreConfig` and `StoreUserConfig`, etc. so we default to user-first


packages/network/src/v2/schemas/decodeDynamicField.ts
```diff
@@ -0,0 +1,18 @@
+import { SchemaType } from "@latticexyz/schema-type";
+
+const unsupportedDynamicField = (fieldType: SchemaType): never => {
+  throw new Error(`Unsupported dynamic field type: ${SchemaType[fieldType] ?? fieldType}`);
+};
+
+// TODO: decode per dynamic field type, so we get better return types for each field type
```
alvrs, line 7: Could we create a type like the v1 `ContractSchemaValueTypes` and make the function generic? (`decodeDynamicField<T extends SchemaType>(fieldType: T, data: ArrayBuffer): SchemaTypeToTsType<T>`)

packages/network/src/v2/schemas/decodeData.ts
```diff
@@ -0,0 +1,59 @@
+import { getStaticByteLength } from "@latticexyz/schema-type";
+import { hexToArray } from "@latticexyz/utils";
+import { TableSchema } from "../common";
+import { decodeStaticField } from "./decodeStaticField";
+import { decodeDynamicField } from "./decodeDynamicField";
+
+export const decodeData = (schema: TableSchema, hexData: string): Record<number, any> => {
```
alvrs, line 7: ooc when do you use `const` and when `function` for functions?

packages/network/src/v2/schemas/decodeField.ts
```diff
@@ -0,0 +1,27 @@
...
+    if (index === schemaIndex) {
+      data[schemaIndex] = decodeStaticField(fieldType, bytes, 0);
+    }
+  });
+
+  if (schema.dynamicFields.length > 0) {
+    schema.dynamicFields.forEach((fieldType, i) => {
+      const index = schema.staticFields.length + i;
+      if (index === schemaIndex) {
+        // if (hexData === "0x") console.log("decoding dynamic field");
```
alvrs, line 20:
```suggestion
```

packages/network/src/v2/schemas/decodeStaticField.ts
```diff
@@ -0,0 +1,41 @@
+import { SchemaType } from "@latticexyz/schema-type";
+import { arrayToHex } from "@latticexyz/utils";
+
+const unsupportedStaticField = (fieldType: SchemaType): never => {
+  throw new Error(`Unsupported static field type: ${SchemaType[fieldType] ?? fieldType}`);
+};
+
+// TODO: decode per static field type, so we get better return types for each field type
+export const decodeStaticField = (
```
alvrs, line 9: Same q as with `decodeDynamicField`, could we make this generic?

packages/network/src/v2/schemas/tableMetadata.ts
```diff
@@ -0,0 +1,26 @@
...
+export const metadata: Partial<Record<`${string}:${string}`, TableMetadata>> = {};
+
+// the Contract arguments below assume that they're bound to a provider
+
+export function getMetadata(world: Contract, table: TableId): TableMetadata | undefined {
+  const schemaKey = `${world.address}:${table}` as const;
+  return metadata[schemaKey];
+}
+
+// TODO: make async with optional metadata and lazily fetch from contract?
```
alvrs, line 16: Should we track this TODO in #444?

packages/recs/package.json
```diff
@@ -22,6 +22,9 @@
     "release": "npm publish --access=public"
   },
   "devDependencies": {
+    "@latticexyz/cli": "^1.41.0",
+    "@latticexyz/schema-type": "^1.41.0",
+    "@latticexyz/std-client": "^1.41.0",
```
alvrs, line 27: `recs` is a requirement of `std-client` too, so this creates a circular dependency. Is there a way we can avoid `recs` depending on `std-client`?

packages/recs/src/Component.ts
```diff
@@ -75,7 +75,24 @@ export function setComponent<S extends Schema, T = undefined>(
 ) {
   const prevValue = getComponentValue(component, entity);
   for (const [key, val] of Object.entries(value)) {
-    component.values[key].set(entity, val);
+    if (component.values[key]) {
+      component.values[key].set(entity, val);
+    } else if (!component.metadata?.tableId) {
+      // If component metadata includes the tableId (from `defineStoreComponents`),
+      // we can ignore this value without logging anything.
```
alvrs, line 82: Why is this? Because the update includes additional numeric indices as keys but they are not part of the component?

packages/recs/src/defineStoreComponents.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  const components = Object.fromEntries(
+    Object.entries(config.tables).map(([tableName, table]) => {
+      // TODO: move this to a util to share this logic with cli/autogen
+      const tableId = new TableId(config.namespace, table.fileSelector);
+      // translate v2 schema type to v1 RECS generic type
+      const schema: Schema = Object.fromEntries(
+        Object.entries(table.schema).map(([fieldName, schemaType]) =>
+          // TODO: infer TS types from schema types, map to real key names?
+          [fieldName, Type.T]
```
alvrs, line 20: This relates to the question in `decodeDynamicField` - could we create a type map like v1 `ContractSchemaValueTypes` but mapping from `SchemaType` to recs's `Type` and use it to infer the recs type here, and also give the `components` object strong typing (probably just using `as` while infering the type directly from the config)?

packages/std-client/package.json
```diff
@@ -24,7 +24,9 @@
     "@latticexyz/network": "^1.41.0",
     "@latticexyz/recs": "^1.41.0",
     "@latticexyz/solecs": "^1.41.0",
+    "@latticexyz/store": "^1.41.0",
     "@latticexyz/utils": "^1.41.0",
+    "@latticexyz/world": "^1.41.0",
```
alvrs, line 29: do we need to add these to `peerDependencies` too?

packages/std-client/src/setup/setupMUDNetwork.ts
```diff
@@ -65,7 +68,25 @@ export async function setupMUDNetwork<C extends ContractComponents, SystemTypes
     )
   );
 
+  const storeSchemaTableId = new TableId("mudstore", "schema");
+  const storeSchemaComponent = defineComponent(
+    world,
+    { schema: Type.T },
```
alvrs, line 74: Note: with latest `main` we probably need to add `keyType` here too (and when decoding schemas), see #480

packages/utils/src/v2/TableId.ts
```diff
@@ -0,0 +1,40 @@
...
+    const tableId = new Uint8Array(32);
+    tableId.set(stringToBytes16(namespace), 0);
+    tableId.set(stringToBytes16(name), 16);
+    return arrayToHex(tableId);
+  }
+
+  static fromBytes32(tableId: Uint8Array) {
+    // assumes tableId is a 32-byte hex string, otherwise it left-pads with zeros (for numbers)
+    // this is scary, since zero padding is different depending on the type (bytes types vs number types)
+    // TODO: fix this after we move tableIds to bytes32 instead of uint256
```
alvrs, line 32: Should we add this TODO to #444?

alvrs (commented):
Looking very good already. Had a couple comments, but this should be good to go soon!

packages/network/src/types.ts
```diff
@@ -70,6 +70,10 @@ export type ContractEvent<C extends Contracts> = {
   args: Result;
   txHash: string;
   lastEventInTx: boolean;
+  blockNumber: number;
+  // TODO: make this required, so we can later sort by logIndex when concatenating event types
+  //       would require updating the ECS snapshot, though
+  logIndex?: number;
```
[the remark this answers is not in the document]
holic: if we're doing any sort of caching of raw events, we should def store the block number + log index, as it provides a deterministic order of events (block number isn't enough, and multiple sets of the same record in a block could get reordered and client data corrupted)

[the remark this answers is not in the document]
holic: cc @authcall

packages/network/src/v2/schemas/tableSchemas.ts
```diff
@@ -0,0 +1,49 @@
...
+      existingSchema.then((schema) => {
+        if (schema.rawSchema !== rawSchema) {
+          console.warn("a different schema was already registered for this table", {
+            world: world.address,
+            table,
+            currentSchema: schema,
+            newSchema: rawSchema,
+          });
+        }
+      });
```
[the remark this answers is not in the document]
holic: this is just a developer-friendly warning, and does not need to block the `registerSchema` call

packages/network/src/v2/decodeStoreSetRecord.ts
```diff
@@ -0,0 +1,57 @@
...
+): Promise<ComponentValue> {
+  // registerSchema event
+  if (table.toHexString() === schemaTableId.toHexString()) {
+    const [tableForSchema, ...otherKeys] = keyTuple;
+    if (otherKeys.length) {
+      console.warn(
+        "registerSchema event has more than one value in key tuple, but this method only supports a single key",
+        { table, keyTuple }
+      );
+    }
```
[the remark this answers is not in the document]
holic: If `key` in this case is an entity ID, I wasn't sure of the downstream effects of concatenating etc. the key tuple.

packages/network/src/v2/schemas/decodeData.ts
holic: oof, its inconsistent!

I generally prefer `const` but I've started using `export function` a bit more lately because you can overload type signatures with it, and you can't do that for `const` definitions.

packages/recs/src/Component.ts
holic: Correct!

packages/utils/src/v2/TableId.ts
holic: added https://github.com/latticexyz/mud/issues/484

packages/network/src/v2/decodeStoreSetRecord.ts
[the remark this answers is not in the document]
alvrs: The client only cares about the entity id being a string, so something like `keys.join("/")` shouldn't break anything. It's only an issue if trying to pass the entity ID to a contract that expects a tuple, but that case has to be handled separately anyway.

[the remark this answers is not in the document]
holic: For now I am just going to `JSON.encode(keyTuple) as EntityID`

But for this specific code, these checks are here to make sure our on-chain logic matches client logic in how we use the key for the table name. I need to extract the table name from the key in order to register the schema and metadata (ahead of lazily fetching). So it still makes sense to leave this logic as-is.

packages/cli/src/config/loadStoreConfig.ts
```diff
@@ -7,7 +7,7 @@ export async function loadStoreConfig(configPath?: string) {
   const config = await loadConfig(configPath);
 
   try {
-    return parseStoreConfig(config);
+    return parseStoreConfig(config as any);
```
alvrs, line 10: why is this needed?

packages/network/src/v2/schemas/decodeData.ts
alvrs: I think the `function` definitions are a bit more common in the current codebase, we should align everything at some point for consistency. Not blocking for this PR though.

packages/network/src/v2/schemas/tableMetadata.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  // Skip lazily fetching internal tables
+  if (table.toHexString() === schemaTableId.toHexString() || table.toHexString() === metadataTableId.toHexString()) {
+    return Promise.resolve(undefined);
+  }
+
+  console.log("fetching metadata for table", { table: table.toString(), world: world.address });
+  const metadataPromise = Promise.all([
+    registerSchema(world, metadataTableId),
+    // TODO: figure out how to pass in rawSchema, it was giving me "incorrect length" errors before
```
alvrs, line 59: maybe we have to pad the raw schema string to 32 bytes?

packages/network/src/v2/schemas/tableMetadata.ts
```diff
@@ -0,0 +1,70 @@
...
+    world["getRecord(uint256,bytes32[])"](metadataTableId.toHexString(), [table.toHexString()]),
+  ]).then(([metadataSchema, metadataRecord]) => {
+    // TODO: error if schema or record not found?
```
alvrs, line 62: should we at least log a warning here?

packages/store/package.json
```diff
@@ -13,7 +13,7 @@
     "prepare": "yarn build && chmod u+x git-install.sh",
     "git:install": "bash git-install.sh",
     "codegen": "ts-node ./scripts/codegen.ts",
-    "tablegen": "../cli/dist/mud.js tablegen",
+    "tablegen": "mud tablegen",
```
alvrs, line 16: i'd prefer keeping the local version for this script, so it works in the CI when we do changes to tablegen (without releasing a new version)

packages/world/package.json
```diff
@@ -12,7 +12,7 @@
   "scripts": {
     "prepare": "yarn build && chmod u+x git-install.sh",
     "git:install": "bash git-install.sh",
-    "tablegen": "../cli/dist/mud.js tablegen",
+    "tablegen": "mud tablegen",
```
alvrs, line 15: Same here, see comment above (in store/package.json)

alvrs (commented):
Nothing blocking (except from the comment in package.json of world/store). Are you planning to add all `TODO` comments to some tracker and link the tracker in the code?

Almost there!!

packages/store/package.json
holic: what if, instead, we have CI use a locally linked mud?

packages/cli/src/config/loadStoreConfig.ts
holic: `loadConfig` has a return value of `unknown`, and if I don't case it, I get
image

this is happening now because I changed the argument type in `parseStoreConfig` to help with `defineStoreComponents` (to "expand" the mud config into something reasonable/without tons of shorthand and conditional branching)
image

packages/network/src/v2/schemas/tableMetadata.ts
holic: not sure, but also not too worried since this is a getter and we still have to do both fetch calls anyway

packages/store/package.json
alvrs: That would work too, we just have to set it up. If it "just works" I'm down to do that instead in this PR, but wouldn't lose too much time on it, we can always change it in a follow up

holic: will change it back for easy and add an issue to come back to this

holic: https://github.com/latticexyz/mud/issues/489

packages/world/package.json
holic: https://github.com/latticexyz/mud/issues/489

packages/network/src/v2/ecsEventFromLog.ts
```diff
@@ -19,7 +19,7 @@ export const ecsEventFromLog = async (
   const tableId = TableId.fromBytes32(utils.arrayify(args.table));
   const component = tableId.toString();
   // TODO: revisit key tuple encoding for client
-  const entity = args.key.join(":") as EntityID;
+  const entity = `v2:${args.key.join(":")}` as EntityID;
```
alvrs, line 22: This might break something if I want to use v2 with ECS, and want to refer to another entity in my component

Example: `OwnedBy` component, Bob (0x01) is owned by Alice (0x02), so I have an entry in the OwnedBy table with key `0x01` and value `0x02`. Now I want to do an recs query on the client to get all entities owned by Alice, so I do `getEntitiesWithValue(OwnedBy, {value: Alice})`, but because Alice is normalized to `v2:0x02` it wouldn't find the `0x02` value of key `0x01`.

holic: updated in https://github.com/latticexyz/mud/pull/415/commits/9c13cd7084ed8afb0754a0eefbeb756f9f38e375

alvrs (approved):
Let's go!

PR #416, opened 2023-02-17
refactor(store): change tableId type from bytes32 to uint256

alvrs:
This PR changes the type of `tableId` from `bytes32` to `uint256`.

Reason: `bytes32` clashes with `string memory` in the World interface because sufficiently short strings can be auto-converted to `bytes32`. This causes the following two method signatures to clash:

```solidity
function setRecord(bytes32 tableId, bytes32[] memory key, bytes memory data) external;

function setRecord(string memory tableRoute, bytes32[] memory key, bytes memory data) external;
```


dk1a:
lgtm

dk1a (approved):

PR #421, opened 2023-02-19
docs(emojimon): fix step 4, optimistic rendering for clicks

virtualelena:
I was getting an error in which the emoji didn't render -- the fix is in the "Override position value" in Step 4 of the tutorial. I realized that the Click handler was not calling `moveTo`. By moving it up in scope, and adding to `onClick` for each element in the grid, I was able to get the emoji to appear. 

For the edit on the bottom, I wanted optimistic responsivity for clicking as well as keystrokes. Without this edit, the keystroke and clicking function would be broken.

### DIFF
--- a/tutorials/emojimon/step-4.md
+++ b/tutorials/emojimon/step-4.md
@@ -54,23 +54,30 @@ import { useMUD } from "./MUDContext";
 
 export const GameBoard = () => {
   …
-  useEffect(() => {
-    const moveTo = async (x: number, y: number) => {
-      const positionId = uuid();
-      Position.addOverride(positionId, {
-        entity: playerEntity,
-        value: { x, y },
-      });
+  const moveTo = async (x: number, y: number) => {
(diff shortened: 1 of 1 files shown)


holic:
Good catch, thanks for the changes! I made some tweaks to the tutorial as I was writing it and this one slipped through the cracks.

holic:
Reading through the tutorial again, I realized I had another intent here that wasn't talked about (yet).

Future steps of the tutorial restrict movement to only adjacent spaces, so it makes less sense to have click-to-move (can no longer jump to anywhere on the map). In those later steps, you replace the click-to-move with click-to-join.

I'm planning to rework the initial repo code and some of the steps to put `moveTo` and `moveBy` in a separate part of the app and not have them depend on any React state. Will follow up soon with those changes!

holic:
Closing this, going to address differently in an Emojimon refresh after merging some latest changes.

PR #424, opened 2023-02-20
feat(cli): deprecate create command

holic:
Deprecates `mud create` cli command in favor of `yarn create mud`. This is a "bridge" commit that deprecates the command but still allows it to run (by using `yarn create mud` under the hood). In a future version, we should remove this command all together. Happy to do this sooner if you think that's warranted.

Also deprecated+archived the template repos:
https://github.com/latticexyz/mud-template-minimal
https://github.com/latticexyz/mud-template-react

### DIFF
--- a/packages/cli/src/commands/create.ts
+++ b/packages/cli/src/commands/create.ts
@@ -10,23 +10,21 @@ type Options = {
 export const command = "create <name>";
 export const desc = "Sets up a mud project into <name>. Requires yarn.";
 
+export const deprecated = true;
+
 export const builder: CommandBuilder<Options, Options> = (yargs) =>
   yargs
     .options({
-      template: { type: "string", desc: "Template to be used (available: [minimal])", default: "minimal" },
+      template: { type: "string", desc: "Template to be used (available: [minimal, react])", default: "minimal" },
     })
     .positional("name", { type: "string", default: "mud-app" });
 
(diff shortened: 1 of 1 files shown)


dk1a:
Didn't realize it was essentially deprecated. I can remove it in #417 then, it'll fit in with the other removals and avoid resolving conflicts with this PR

holic:
> Didn't realize it was essentially deprecated. I can remove it in #417 then, it'll fit in with the other removals and avoid resolving conflicts with this PR

Yep, just the `create` command is deprecated.

I would kinda prefer to keep this deprecation separate from the big refactor, just to minimize blast radius/what's changing. Coupling a big refactor and a deprecation is a little scary to me (in case one breaks).

dk1a:
> I would kinda prefer to keep this deprecation separate from the big refactor, just to minimize blast radius/what's changing. Coupling a big refactor and a deprecation is a little scary to me (in case one breaks). Is that okay?

Of course, I meant that your PR goes into main soonish, and my refactor with its removals goes into v2 (#417 isn't needed in v1 as its main purpose is to pave the way for autogen)

holic:
> Of course, I meant that your PR goes into main soonish, and my refactor with its removals goes into v2 (#417 isn't needed in v1 as its main purpose is to pave the way for autogen)

Ohh that sounds perfect. Thank you!

alvrs (approved):
lgtm

PR #426, opened 2023-02-20
feat(cli): log client world URL

holic:
Certain OS/platforms can't open URL automatically, so we should at least log it in the console to make it easier to access.


dk1a:
note to self to duplicate this in #417

alvrs (approved):

PR #429, opened 2023-02-21
feat(schema-type): create schema-type package

dk1a:
Refactoring store would be a separate PR
Pure SchemaType tests are of very minor importance and can be done later (the useful tests are in #396 already - that SchemaType's static length used in encoding/decoding of all types works correctly)


packages/schema-type/.npmignore
```diff
@@ -0,0 +1,6 @@
+*
+
+!dist/**
```
alvrs, line 3: with the current config in `package.json` the `main` field would always be set on the ts version, never the dist version. Do we want that behaviour here? In other ts packages we use this yarn script to change the `main` field to `dist/index.js` before publishing: https://github.com/latticexyz/mud/blob/1f2598cff40cd9f5059b553b9291ffd2c61bacdd/packages/recs/package.json#L19-L20

But it might be that this package is intended to be only used in typescript since `enum` is a typescript feature anyway

packages/schema-type/src/solidity/SchemaType.sol
```diff
@@ -0,0 +1,236 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
+
+using { getStaticByteLength } for SchemaType global;
+
+// WARNING: SchemaType methods use hardcoded enum indexes, review them after any changes to the enum
```
alvrs, line 6: Should we add the same comment in the ts file? (And maybe a comment about always changing all implementations, never just an individual one?)

packages/schema-type/src/typescript/SchemaType.ts
```diff
@@ -0,0 +1,639 @@
...
+  [SchemaType.BYTES32_ARRAY]: "bytes32[]",
+
+  [SchemaType.BOOL_ARRAY]: "bool[]",
+  [SchemaType.ADDRESS_ARRAY]: "address[]",
+
+  [SchemaType.BYTES]: "bytes",
+  [SchemaType.STRING]: "string",
+};
+
+export type ArraySchemaType =
```
alvrs, line 435: Should we add a `DyamicSchemaType` (ArraySchemaType | bytes | string) just to make it more clear that ArraySchemaType doesn't include all dynamic types?

alvrs (commented):
looking great, again just a couple smaller nits / questions

packages/schema-type/.npmignore
dk1a: I actually tried out different things for `main`. Ultimately I kept it as typescript by default since it's flagship feature is an enum, but also include `dist` just in case someone needs `.js` stuff.
I imagine we can try it out like this and change later if necessary, unless you think `.ts` main is bad? I don't have much experience here, and locally it works either way

dk1a:
btw I ditched hardhat since it just doesn't have any docs do generate (no contracts).
But I did figure out how to do [hardhat@esm](https://github.com/NomicFoundation/hardhat/issues/3043) in case we do need it, as this package is a `"type": "module"`

packages/schema-type/src/solidity/SchemaType.sol
dk1a: Added more warnings. I think solidity file gets the core warning to mirror things everywhere, and the other langs should just copy the solidity file

packages/schema-type/.npmignore
alvrs: I agree with this reasoning, let's try it out as is and later change if needed

alvrs (approved):
lgtm

packages/schema-type/.npmignore
holic: the pattern I've seen is building `.js` (for `main`) and `.d.ts` (for `types`):
https://github.com/wagmi-dev/wagmi/blob/main/packages/core/package.json#L27-L28

but that probably only makes sense if there are things in this package useful in JS alone (not TS)

PR #430, opened 2023-02-21
chore: merge main into v2

alvrs:


alvrs:
this would lead to an ugly non-linear history - alternatively we could rebase v2 onto main

dk1a:
> this would lead to an ugly non-linear history - alternatively we could rebase v2 onto main

I don't mind a rebase, I think it's just you, me and @holic using it atm, shouldn't be very disruptive

dk1a (approved):

alvrs:
Just realized main was already merged into v2 in the past, so it's already not a linear history. Will go ahead with a merge commit.

holic:
Ahh oops, yeah the main merge into v2 was my bad - I assumed we would be squash merging v2 sooner than later. We could still probably rebase it in a way that can be cleanly merged into main?

PR #434, opened 2023-02-23
feat: add pushToField to Store and World

dk1a:
Intentionally not adding a new event, but the method's logic accounts for it being added in the future with minimal changes to StoreCore and no changes to World etc (so it could be mostly a network PR)

Notably this allows pushing multiple items at once
No validity checks though (same as setField, they should be added in one separate refactor I think)

`_verifiedTableRouteId`, `_expectRouteAccessDenied` and `_initRouteTableKey`: 
I basically want to reduce code repetition in World but I don't want to complicate this PR by refactoring things unrelated to `pushToStore` yet

setItem and pop could happen in the future, but rn I just need push for proper autogen

### DIFF
--- a/packages/store/test/StoreCore.t.sol
+++ b/packages/store/test/StoreCore.t.sol
@@ -44,6 +44,16 @@ contract StoreCoreTest is Test, StoreView {
     StoreCore.setField(table, key, schemaIndex, data);
   }
 
+  // Expose an external pushToField function for testing purposes of indexers (see testHooks)
+  function pushToField(
+    uint256 table,
+    bytes32[] calldata key,
+    uint8 schemaIndex,
+    bytes calldata dataToPush
+  ) public override {
+    StoreCore.pushToField(table, key, schemaIndex, dataToPush);
+  }
+
   // Expose an external deleteRecord function for testing purposes of indexers (see testHooks)
   function deleteRecord(uint256 table, bytes32[] calldata key) public override {
     StoreCore.deleteRecord(table, key);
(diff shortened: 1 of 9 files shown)


packages/store/test/StoreCore.t.sol
```diff
@@ -503,6 +513,111 @@ contract StoreCoreTest is Test, StoreView {
...
   }
 
+  function testPushToField() public {
+    uint256 table = uint256(keccak256("some.table"));
+
+    {
+      // Register table's schema
+      Schema schema = SchemaLib.encode(SchemaType.UINT256, SchemaType.UINT32_ARRAY, SchemaType.UINT32_ARRAY);
+      StoreCore.registerSchema(table, schema);
+    }
```
holic, line 523: I haven't seen this pattern before. Do the curly braces/block create a new scope? So `schema` is only available inside the scope?

packages/store/src/schemas/AddressArray.sol
```diff
@@ -48,7 +48,6 @@ library AddressArray_ {
 
   /**
    * Push an element to the addresses array
-   * TODO: this is super inefficient right now, need to add support for pushing to arrays to the store core library to avoid reading/writing the entire array
```
holic, line 51: Curious to see if we have gas tests for before/after this change to see how much more efficient this method is compared to the previous.

holic (approved):
lgtm!

packages/store/test/StoreCore.t.sol
dk1a: > Do the curly braces/block create a new scope?

yes, afaik it can also be useful outside of tests to avoid stack limits with many vars

packages/store/src/schemas/AddressArray.sol
dk1a: The real change would be when a separate event is added. Right now there's still a redundant full storage read

dk1a: there should be a gasreport for these changes too though (there're some for push too), wonder why it didn't get added

alvrs:
we need to rebase this one on main now

alvrs (approved):
Looks good overall, just a couple smaller nits

packages/world/src/World.sol
```diff
@@ -326,6 +362,26 @@ contract World is Store {
     StoreCore.deleteRecord(tableRouteId, key);
   }
 
+  /**
+   * Check for access based on `accessRoute`
+   * and return `tableRouteId` for `accessRoute/subRoute`
+   * because access to a route also grants access to all sub routes.
+   */
+  function _verifiedTableRouteId(string calldata accessRoute, string calldata subRoute)
```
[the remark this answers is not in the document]
dk1a: yep I made it to remove code repetition but didn't wanna bloat this PR

packages/world/src/schemas/AddressArray.sol
```diff
@@ -83,6 +81,6 @@ library AddressArraySchemaLib {
 
   function decode(bytes memory blob) internal pure returns (address[] memory addresses) {
     if (blob.length == 0) return new address[](0);
-    SliceLib.getSubslice(blob, 32).decodeArray_address();
+    return SliceLib.getSubslice(blob, 32).decodeArray_address();
```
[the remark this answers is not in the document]
dk1a: I think it broke something for me, which is how I noticed it. Don't remember what though, could have even been some temporary test I used AddressArray for

packages/world/test/World.t.sol
```diff
@@ -111,6 +113,35 @@ contract WorldTest is Test {
...
+      bytes32[] memory keyTuple
+    )
+  {
+    // Register a new route
+    world.registerRoute("", baseRoute);
+
+    // Register a new table
+    tableRouteId = world.registerTable(baseRoute, tableRoute, schema);
+
+    key = keccak256("testKey");
```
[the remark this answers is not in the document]
dk1a: yeah, did this and https://github.com/latticexyz/mud/pull/434#discussion_r1126394460 in 1 go, best of both worlds

packages/world/test/World.t.sol
```diff
@@ -111,6 +113,35 @@ contract WorldTest is Test {
     world = new World();
   }
 
+  // Expect an error when trying to write from an address that doesn't have access
+  function _expectRouteAccessDenied(address caller, string memory route) internal {
+    vm.prank(caller);
+    vm.expectRevert(abi.encodeWithSelector(World.RouteAccessDenied.selector, route, caller));
+  }
+
+  function _initRouteTableKey(
```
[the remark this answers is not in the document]
dk1a: agree in general, and for `_initRouteTableKey` specifically, see https://github.com/latticexyz/mud/pull/434#discussion_r1127164439

for `_expectRouteAccessDenied` I think it's an exception. It's used a lot (14 times) with the same pattern. `caller` is used twice there, so without an internal method higher risk of making a mistake. Also changed it to prank+expectRevert (it was even longer with startPrank+expectRevert+stopPrank, which is yet another mistake risk).
So I refactored it more, to use `_expectRouteAccessDenied` in all 14 cases

alvrs (dismissed):
Looks good! Let's merge this before #467

alvrs:
Looks like we need to resolve the gas-report conflicts with main (easiest way would be to merge main into this and just rerun the gas report)

alvrs (approved):

PR #437, opened 2023-02-24
feat(cli): refactor tablegen and add more advanced features

dk1a:
Pretty much rewrote the whole thing.
StoreUserConfig is a good summary of what's going on.
All tables in `store` are now autogenerated, check them out.
I didn't want to mess with `world` too much, so made just `RouteAccess` autogenerated (it's interesting because it has 2 keys with different types, and uses `storeArgument` flag)

https://github.com/latticexyz/mud/issues/438 has a summary


packages/store/src/tables/Route.sol
```diff
@@ -13,13 +13,16 @@ import { EncodeArray } from "../tightcoder/EncodeArray.sol";
...
+uint256 constant RouteTableId = _tableId;
+
 struct Route {
   address addr;
   bytes4 selector;
   uint8 executionMode;
 }
 
-library Route_ {
+library RouteTable {
```
holic, line 25: I think I missed the naming conversation but did ya'll consider `RouteStore` as a name option?

dk1a: I think it'll be confusing when a table that writes to `Store` is also called `Store`. We already have `World`, which is `Store`. Wouldn't want everything to be a `Store`

holic: My thought is that users will mostly be interacting with these autogen files, not the underlying ones (i.e. Store). But I understand if this creates confusion on our end.

"Table" just feels like an overloaded term. Folks with SQL experience will have preconceived notions about what it does/should do, and folks without SQL experience may be a little confused by it the term "table". We have a chance to shed this here.

dk1a: I like the SQL analogy (I have SQL experience, although I'm obviously not a great example).
We even have hooks/indexers. And a planned postgres integration. In solidity it's just a limited key/value store of course.
But I'm starting to think of how foreign keys would fit in here

packages/cli/src/utils/errors.ts
```diff
@@ -1,54 +1,47 @@
...
-export class CustomError extends Error {
-  constructor(message: string, public readonly parent?: Error) {
-    super(message);
-
-    this.name = this.constructor.name;
-  }
+export class NotInsideProjectError extends Error {
+  name = "NotInsideProjectError";
+  message = "You are not inside a MUD project";
 }
```
holic, line 17: small thing: I would probably add

```ts
export class MudCliError extends Error {}
```
then
```ts
export class NotInsideProjectError extends MudCliError {
  ...
}
```

just so downstream folks can easily catch all CLI errors in one conditional

(I am thinking we might eventually have a centralized `MudError` but a cli-specific one in the meantime is still handy)

holic: That said, I'm not sure how these interact with Zod error classes.

dk1a: > just so downstream folks can easily catch all CLI errors in one conditional

There aren't really supposed to be downstream folks rn, the renderer is cli-only currently, and everything is handled by `logError`.
When I do make it standalone that sounds like a good thing to add though. Zod errors can probably be wrapped to also be identifiable somehow

packages/cli/src/utils/format.ts
```diff
@@ -0,0 +1,33 @@
...
+    let message;
+    if (error instanceof Error) {
+      message = error.message;
+    } else {
+      message = error;
+    }
+    console.log(chalk.yellow(`Error during output formatting: ${message}`));
+    return content;
+  }
+}
```
holic, line 33: Not blocking but we can also call out to `forge fmt` and use the project's local `foundry.toml` (in the same way that we use it for `src` etc. dirs)

```
cat File.sol | forge fmt --raw -
```

dk1a: This method isn't really file-related, it's used at the render stage, so it's not aware of the filesystem at all, and when renderer is standalone it could even be used without files at all.
Also since it's all js using prettier is just more convenient.
I think if people want custom formatting they can do their own `fmt`, this func is just for new mudders who don't have that set up yet, so they don't see super ugly files

packages/store/src/tables/Route.sol
holic: to circle back on this: we all chatted in discord and ended up with removing the suffix altogether on libraries, and adding "data" suffix to structs

packages/cli/src/render-table/index.ts
```diff
@@ -0,0 +1,99 @@
...
+  const renderedTables = [];
+  for (const tableName of Object.keys(config.tables)) {
+    const tableData = config.tables[tableName];
+
+    // struct adds methods to get/set all values at once
+    const withStruct = tableData.dataStruct;
+    // operate on all fields at once; for only 1 field keep them only if struct is also kept
+    const withRecordMethods = withStruct || Object.keys(tableData.schema).length > 1;
+    // field methods can be simply get/set if there's only 1 field and no record methods
+    const noFeldMethodSuffix = !withRecordMethods && Object.keys(tableData.schema).length === 1;
```
alvrs, line 24:
```suggestion
    const noFieldMethodSuffix = !withRecordMethods && Object.keys(tableData.schema).length === 1;
```

packages/cli/src/render-table/index.ts
```diff
@@ -0,0 +1,99 @@
...
+    });
+
+    const fields = Object.keys(tableData.schema).map((name) => {
+      const type = tableData.schema[name];
+      const elementType = SchemaTypeArrayToElement[type];
+      const field: RenderTableField = {
+        ...getSchemaTypeInfo(type),
+        arrayElement: elementType ? getSchemaTypeInfo(elementType) : undefined,
+        name,
+        methodNameSuffix: noFeldMethodSuffix ? "" : `${name[0].toUpperCase()}${name.slice(1)}`,
```
alvrs, line 46:
```suggestion
        methodNameSuffix: noFieldMethodSuffix ? "" : `${name[0].toUpperCase()}${name.slice(1)}`,
```

packages/cli/src/config/loadStoreConfig.ts
```diff
@@ -1,142 +1,100 @@
...
+  .refine((arg) => getStaticByteLength(arg) > 0, "Primary key must not use dynamic SchemaType");
+const PrimaryKeys = z.record(KeyName, PrimaryKey).default({ key: SchemaType.BYTES32 });
+
+const Schema = z
+  .record(ColumnName, z.nativeEnum(SchemaType))
+  .refine((arg) => Object.keys(arg).length > 0, "Table schema may not be empty");
+
+const FullTable = z
+  .object({
+    route: OrdinaryRoute.default("/tables"),
```
alvrs, line 22: looks like this can't be empty correct? I think we should allow registering tables at the root level of a World (eg. `/RootTable`), which would require passing the empty root route here, however this wouldn't work as nicely with using the same route as output path for the generated file

alvrs (commented):
Another very impressive PR! Just a couple smaller nits / questions

packages/cli/src/utils/errors.ts
```diff
@@ -1,54 +1,47 @@
...
+export function logError(error: Error) {
+  if (error instanceof ValidationError) {
+    console.log(chalk.redBright(error.message));
+  } else if (error instanceof ZodError) {
+    // TODO currently this error shouldn't happen, use `fromZodErrorCustom`
+    const validationError = fromZodError(error, {
+      prefixSeparator: "\n- ",
+      issueSeparator: "\n- ",
+    });
+    console.log(chalk.redBright(validationError.message));
```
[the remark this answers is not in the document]
dk1a: > Or should we add a console log saying "If you see this log, please report at ..."?

Strongly considered this actually, but left as-is because this is still WIP and I might end up needing this error or changing this whole method or something

packages/cli/src/config/loadConfig.ts
```diff
@@ -1,28 +1,29 @@
...
-import { ERRORS, MUDError } from "../utils/errors.js";
-
-// TODO require may or may not be needed, finish exploring the various configurations
-import { createRequire } from "module";
-import { fileURLToPath } from "url";
-const require = createRequire(fileURLToPath(import.meta.url));
+import { NotESMConfigError, NotInsideProjectError } from "../utils/errors.js";
 
 // Based on hardhat's config (MIT)
 // https://github.com/NomicFoundation/hardhat/tree/main/packages/hardhat-core
```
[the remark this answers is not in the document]
dk1a: I think 1 small method is still almost verbatim copied, but otherwise not really, will remove it

packages/cli/src/config/loadConfig.ts
```diff
@@ -1,28 +1,29 @@
...
 
-const TS_CONFIG_FILENAME = "mud.config.mts";
+// In order of preference files are checked
+const configFiles = ["mud.config.ts", "mud.config.mts"];
```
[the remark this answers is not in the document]
dk1a: We could, but it may not be trivial and best left for another PR if needed
And I think we should be opinionated and force people into typescript. We already force ESM, might as well go all the way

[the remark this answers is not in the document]
dk1a: Ah right, SchemaType enum. That could work in js but not nearly as nice

packages/cli/src/config/loadStoreConfig.ts
dk1a: Correct, it's the one case that messes up route->folder.
Do we really need root tables?

packages/cli/src/utils/errors.ts
```diff
@@ -1,54 +1,47 @@
...
-
-  private readonly _isMUDError: boolean;
-
-  constructor(errorDescriptor: ErrorDescriptor, messageArguments: string[] = [], parentError?: Error) {
-    super(errorDescriptor.message(...messageArguments), parentError);
-
-    this.errorDescriptor = errorDescriptor;
+export class NotESMConfigError extends Error {
+  name = "NotESMConfigError";
+  message = "MUD config must be an ES module";
```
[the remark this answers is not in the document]
dk1a: `logError` handles that already. I don't think the error itself needs such details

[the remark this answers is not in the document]
alvrs: Makes sense, I missed that `logError` proposes a solution!

packages/cli/src/config/loadStoreConfig.ts
alvrs: I think it might be useful, if only to allow users to minimize route lengths. Just feels like we shouldn't prevent a setup that uses tables at the root. We could even just put the auto-generated file into the "base directory" (right now foundry's `src` config - but we could add a config field to allow overriding the table base directory).

Not blocking for this PR though.

alvrs (approved):

packages/cli/src/config/loadStoreConfig.ts
dk1a: Just allowing root routes is easy actually - replace validateRoute with validateBaseRoute in zod schema.
Having an extra config option for baseDirectory is also pretty trivial, although baseRoute+baseDirectory feels like a bit much.

Btw I don't bundle systems and tables together here. Systems are writers, and for systems that have access to everything (i.e. use delegatecall) I agree root is the best place.

Problems I have with root tables:
- Encourages flat setup. It's intuitive for small projects, but projects that need MUD aren't really small imo. No root simplifies the switch from "/tables" to "/skills", "/items" etc.
- Tables no longer have access groups. If you want to give access to several at once, you'd give root route. You shouldn't give out root access.
- Unlike systems, root tables don't signify a root permission. They don't *have* to exist

Rather than root routes I think it'd be more useful to consider shorthands/config reorgs for setting the same route for multiple tables/systems

My current folders (they would more or less become routes in v2, and this sorta setup is what I'm optimizing for):
Screenshot 2023-02-28 033836

dk1a: On the other hand, if route and folder path are completely separated, it feels like my problems with root tables kinda evaporate. And maybe we should have `route: ""` by default, but `path: "/tables"` by default, and have no significant relationship between the 2

alvrs: > Encourages flat setup. It's intuitive for small projects, but projects that need MUD aren't really small imo. No root simplifies the switch from "/tables" to "/skills", "/items" etc.

I think with v2 there are a lot of good reasons for small projects to use MUD too (eg default compatibility with indexers and network stack), so my preference would be to make all project configurations possible, including simple/flat ones.

I still agree with making more complex / nested project structures like your example easy to achieve too. After thinking about it more I think separating on-chan `route` and filesystem `outputPath` like you proposed makes the most sense, because it allows for any configuration of routes and filesystem structure. This also makes the id computation consistent with systems, where it's always `keccak256(baseRoute/route/SystemName)` with `route` being `""` by default. (Also for systems `route` is independent from the file system path too because they are not auto-generated/)

alvrs: While working on the system deploy script I am realizing it would be most intuitive if `route` allows users to completely override the route, without a default `/SystemName` or `/TableName` appended at the end. So basically if `route` is omitted the default route should be `baseRoute/SystemOrTableName`, while if the route key is present it should be `baseRoute/route`. Otherwise we might prevent some configurations where the route has to be some explicit value that's not the contract name.

dk1a: > While working on the system deploy script I am realizing it would be most intuitive if `route` allows users to completely override the route, without a default `/SystemName` or `/TableName` appended at the end. So basically if `route` is omitted the default route should be `baseRoute/SystemOrTableName`, while if the route key is present it should be `baseRoute/route`. Otherwise we might prevent some configurations where the route has to be some explicit value that's not the contract name.

Sounds good! I can change that in store config in one of the upcoming PRs

PR #448, opened 2023-02-28
feat(store): check in abi, types

holic:
By checking in ABI files, types, etc., they are immediately usable by other dependencies (i.e. network) without a build step. We can use CI to enforce that they're up-to-date.


packages/store/types/IStore.js
```diff
@@ -0,0 +1 @@
+export {};
```
holic, line 1: Although these are empty, they seem to be required to get e.g. next.js' build pipeline working. Otherwise you get errors like
```
packages/app dev: error - ../contracts/types/index.ts
packages/app dev: Module parse failed: Unexpected token (4:12)
packages/app dev: You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
packages/app dev: | /* tslint:disable */
packages/app dev: | /* eslint-disable */
packages/app dev: > import type * as afdSol from "./AFD.sol";
```

packages/store/package.json
```diff
@@ -15,9 +15,11 @@
...
     "tablegen": "../cli/dist/mud.js tablegen",
     "test": "yarn codegen && forge test",
-    "build": "yarn codegen && rimraf out && forge build && yarn dist && yarn types",
-    "dist": "rimraf abi && mkdir abi && cp -rf out/*.sol/*.json abi/ && yarn rimraf abi/*.metadata.json",
-    "types": "rimraf types && typechain --target=ethers-v5 abi/*.json",
+    "abi": "yarn abi:clean && yarn abi:compile && yarn abi:types",
+    "abi:clean": "rm -rf abi types",
+    "abi:compile": "forge build --out abi --force --skip test script",
+    "abi:types": "typechain --target ethers-v5 \"abi/**/*.sol/!(*.abi).json\" --out-dir types && tsc",
+    "build": "yarn codegen && yarn abi",
```
holic, line 22: note to self: we can probably remove the `prepare` step since we check in all the necessary build files

package.json
```diff
@@ -56,6 +56,7 @@
   },
   "scripts": {
     "prepare": "husky install && (forge --version || yarn foundryup) && yarn workspaces run prepare",
+    "build": "yarn workspaces run prepare",
```
alvrs, line 59: feels a bit odd to have the global `build` script mapped to packages' `prepare` script, should we map it to their `build` script instead?

alvrs (commented):
Lgtm, just the one not about the `build` script! Do you have an approximate value of how much faster the setup is now?

package.json
holic: oh yep!

this was a bit of a temporary workaround because (afaict) there's no way to tell yarn to "run script in all workspaces, only if its defined" and I thought `prepare` was what we wanted but turns out everything just runs `yarn build` inside their `prepare`, so will change!

holic:
> Do you have an approximate value of how much faster the setup is now?

Not yet, but it I would guess not a ton for just this package. It'll be more noticeable once I move the rest to this approach.

holic:
interesting, looks like you get different build results when running in the root with `yarn build` vs running inside a package's dir, which is probably why I am seeing mismatch results (i.e. 200 default optimizer runs vs. 3000 configured in each package)

holic:
sounds like this might be unblocked by configuring foundry with `bytecode_hash = "none"`:
https://twitter.com/PaulRBerg/status/1659268482100338702

alvrs:
opened an issue to track (https://github.com/latticexyz/mud/issues/1017), closing this for now

PR #449, opened 2023-02-28
ci: check for gas report changes

### LINKED ISSUE #442: Fix gas-report action
#422 introduced an automatic `gas-report` action to track gas impacts of changes. Initially the action auto-committed the updated gas report, but that caused more problems than it solved. Now ideally the action should run the gas report and fail if the `gas-report.txt` file is not up to date, requiring the dev to manually run the gas report and update the file. However, currently the action always passes. We should fix it by using similar logic as https://github.com/holic/web3-scaffold/blob/main/.github/workflows/contracts.yml#L49

holic:
fixes #442


.github/workflows/gas-report.yml
```diff
@@ -0,0 +1,54 @@
...
+        os: [ubuntu-latest]
+        node-version: [16.x]
+    runs-on: ${{ matrix.os }}
+
+    name: Generate gas report
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          submodules: recursive
```
alvrs, line 25: do we need this? (probably doesn't hurt, just wondering if there is a reason you added it)

alvrs (approved):
lgtm!

.github/workflows/gas-report.yml
holic: this helps if we ever move to true foundry modules

alvrs (approved):

PR #450, opened 2023-02-28
feat(cli): add v2 deployment script

alvrs:
Deployment script for v2 contracts (World, tables, systems).
- Requires a `PRIVATE_KEY` .env variable
- Uses rpc url from `foundry.toml` (profile can be switched with the `--profile` param)
- Uses `mud.config.mts`
- A compatible example repo can be found here: https://github.com/latticexyz/v2sandbox/tree/alvrs/deploy-v2


packages/cli/src/commands/deploy-v2.ts
```diff
@@ -0,0 +1,111 @@
...
+        if (!ethers.utils.isAddress(accessElement) && !systemNames.includes(accessElement)) {
+          console.log(
+            chalk.red(
+              `Invalid access list in "${systemName}": "${accessElement}" is not a valid Ethereum address or name of a local system.`
+            )
+          );
+          process.exit(1);
+        }
+      }
+    }
```
alvrs, line 86: I wonder if most of the logic in this function should be moved into the `zod` validators. I put it here for now because it depends on running `forge build` and then grabbing all `*System.sol` contracts from the file system, which felt like something that doesn't really fit into `zod`. Thoughts @holic @dk1a?

dk1a: I agree it doesn't fit into the world's schema.
But you could add a separate function that accepts `loadWorldConfig`'s output and the results that depend on `forge`. The main benefit there is not so much a zod schema though, but having that function separated from cli

dk1a: Kinda like how my `tablegen` command does just some filesystem writing, and `renderTablesFromConfig` is a separate exported function you can use on its own

dk1a: `renderTablesFromConfig` also doesn't really do any validation (you would need some of course), but I think you don't need to validate world config again. You can just export the parseConfig function as well

alvrs: Another thought I just had is we could pass `existingContractNames` as array into `loadWorldConfig`, so it can be used for validation inside `zod` (checking that the system overrides and excludes only refer to existing contracts). And we could probably use something like `zod`'s `transform` to resolve `accessList` into `accessListSystems` and `accessListAddresses`. Downside is `forge build` would happen before validating the config, but it should be really quick if no contracts change in between the config changes.

packages/cli/src/commands/deploy-v2.ts
```diff
@@ -0,0 +1,111 @@
...
+      }
+      console.error(chalk.red("Error running `forge build`"));
+      process.exit(1);
+    }
+
+    // Get a list of all contracts in output ending with `System`
+    const outDir = await getOutDirectory();
+    const defaultSystemNames = glob
+      // Get all files in outDir ending in `*System.sol  `
+      .sync(`${outDir}/*System.sol`)
```
holic, line 46: should this be `${outDir}/**/*System.sol` to also capture system files in subdirectories?

packages/cli/src/commands/deploy-v2.ts
holic: I defer to @dk1a, who now has more zod experience than me haha

dk1a: > Another thought I just had is we could pass existingContractNames as array into loadWorldConfig, so it can be used for validation inside zod

I haven't really explored your code yet, but it doesn't seem like `existingContractNames` should be a part of the config loader (unless I'm missing something and it's a config parameter).

To summarize: I'm suggesting another little zod schema and a separate method that accepts config+forge stuff, and validates/transforms those lists. And it can assume world config is otherwise validated already

packages/cli/src/commands/deploy-v2.ts
alvrs: It seems like forge flattens the files in the output directory - in my test project my system is at `src/systems/IncrementSystem.sol`, the output is at `out/IncrementSystem.sol`

holic: oh look, I've already reported this 🙈  https://github.com/foundry-rs/foundry/issues/2079

holic:
is [alvrs/sync-store-v2-hack](https://github.com/latticexyz/mud/tree/alvrs/sync-store-v2-hack) the right base branch?

packages/cli/src/config/loadWorldConfig.ts
```diff
@@ -0,0 +1,179 @@
...
+      accessList: SystemAccessList,
+    }),
+  ])
+);
+
+// The parsed world config is the result of parsing the user config
+export const WorldConfig = z.object({
+  baseRoute: BaseRoute.default(""),
+  worldContractName: z.string().default("World"),
+  overrideSystems: z.record(SystemName, SystemConfig).default({}),
```
holic, line 41: What about just `systems`? 

and `exclude` could be an option, e.g.
```
systems: {
  SomeSystemToOverride: {...},
  SomeSystemToExclude: { exclude: true }
}
```

holic (commented):
Are there any artifacts from this deploy process where we know what our system calls are going to be? Like will we know how to make a system call via the world?

packages/cli/src/utils/schema.ts
```diff
@@ -0,0 +1,47 @@
+import { getStaticByteLength, SchemaType } from "@latticexyz/schema-type";
+import { MUDError } from "./errors.js";
+
+/**
+ * Encode a table schema into a bytes32 hex string
+ * Port of `Schema.sol` from `@latticexyz/store`
+ * @param schema The schema to encode SchemaType[]
+ * @returns The encoded schema as a 32 byte hex string
+ */
+export function encodeSchema(schema: SchemaType[]): string {
```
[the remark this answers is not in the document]
dk1a: solidity one uses Bytes extensively, and I wouldn't wanna duplicate that in schema-type.
I think it could be useful to make a `solidity-utils` first, put `Bytes`, `Memory` and maybe some others in there, and then use that in `schema-type` to do schema encoding

packages/cli/src/config/loadWorldConfig.ts
```diff
@@ -0,0 +1,179 @@
...
+   * Contracts named *System files will be deployed by default
+   * as public systems at `baseRoute/ContractName`, unless overridden
+   *
+   * The key is the system name (capitalized).
+   * The value is a SystemConfig object.
+   */
+  overrideSystems?: Record<string, SystemUserConfig>;
+  /** Systems to exclude from automatic deployment */
+  excludeSystems?: string[];
+  /** Script to execute after the deployment is complete */
```
dk1a, line 171: Should probably mention the default

packages/cli/src/utils/foundry.ts
```diff
@@ -0,0 +1,24 @@
+import chalk from "chalk";
+import { execa } from "execa";
+
+async function safeExec(command: string, args: string[]): Promise<string> {
```
dk1a, line 4: this seems like a pretty generic wrapper, maybe it should be in its own file, like a replacement for execLog?

dk1a (commented):
lgtm, just some nits
Using ethers doesn't look particularly easier than LibDeploy, and certainly has its own share of problems. But I think I do prefer to not have a temporary script file if we can avoid it

packages/cli/src/utils/foundry.ts
holic: "safe" is also a misleading term here (since executing arbitrary commands is ~rarely safe)

It looks like `execa` exports `execaSync` which seems to do what you're doing here, but maybe more robust:
https://github.com/sindresorhus/execa#execasyncfile-arguments-options

alvrs:
> is [alvrs/sync-store-v2-hack](https://github.com/latticexyz/mud/tree/alvrs/sync-store-v2-hack) the right base branch?

No, need to rebase on main before merging. But it shows the correct diffs.

packages/cli/src/utils/deploy-v2.ts
```diff
@@ -0,0 +1,280 @@
...
+    //     String(globalNonce.nonce++)
+    //   )
+    // );
+    // return deployedTo;
+
+    const { abi, bytecode } = await getContractData(contractName);
+    try {
+      const factory = new ethers.ContractFactory(abi, bytecode, signer);
+      console.log(chalk.gray(`executing deployment of ${contractName} with nonce ${globalNonce.nonce}`));
+      const deployPromise = factory.deploy({ nonce: globalNonce.nonce++ });
```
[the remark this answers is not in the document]
alvrs: The reason for manual nonce management here is so we can send the transactions without waiting for previous transactions. Without this ethers errored due to duplicate nonces. Unless ethers has an option for this use case that I'm not aware of?

packages/cli/src/config/loadWorldConfig.ts
alvrs: imo `overrideSystems` makes it more obvious that systems not mentioned here are included by default. If it was just called `systems`, I would assume only systems mentioned here to be deployed. In `resovleWorldConfig` we then use the `overrideSystems` and default systems to build the full `systems` key, so the resulting config object only includes the `systems` key (not `overrideSystems` or `excludeSystems`)

alvrs:
> Using ethers doesn't look particularly easier than LibDeploy, and certainly has its own share of problems.

While working on this deploy script I realized that a bunch of things we can do in js would be much harder to realize in a codegenerated solidity script (like registering nested routes, or more advanced features that will come later like "only deploy a contract if it doesn't yet exist on-chain"). But I agree the most important factor is the advantage of not having to generate a temporary file - `LibDeploy.sol not found` has been a very common error for users.

packages/cli/src/config/loadWorldConfig.ts
```diff
@@ -0,0 +1,179 @@
+import { z, ZodError } from "zod";
+import { fromZodErrorCustom, UnrecognizedSystemErrorFactory } from "../utils/errors.js";
+import { BaseRoute, ObjectName } from "./commonSchemas.js";
+import { loadConfig } from "./loadConfig.js";
+import { validateEthereumAddressOrSystemName } from "./validation.js";
+
+// Object names have to start with a capital letter and only contain letters, numbers and underscores
+const SystemName = ObjectName;
+
+// Base routes are ordinary routes with an empty string as a valid value
```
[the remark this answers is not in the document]
alvrs: This went through a bunch of iterations but the final one is very basic and probably doesn't require the comment 👍

packages/cli/src/utils/foundry.ts
alvrs: Yeah not sure why this is called `safeExec`, should definitely be something like `execLog`, and agree with moving it to a general location.

Re `execaSync` - the one here is more of a helper to pipe the `stdout` and throw `stderr` in the main process (plus some default logging), while `execaSync` is just the non-async version of `execa`, so we'd still need the wrapper for setting the `stdio` options I think. Also being async is intended here, because it allows us to execute multiple commands without waiting for the result of the previous one.

alvrs:
> Are there any artifacts from this deploy process where we know what our system calls are going to be? Like will we know how to make a system call via the world?

I think we can create types and utils (generated or even dynamic) for system calls from config + typechain/abitype. That's what I'll work on next, but gonna put it in a follow up PR.

packages/cli/src/utils/deploy-v2.ts
```diff
@@ -0,0 +1,280 @@
...
+      await forge(
+        "script",
+        postDeployScript,
+        "--sig",
+        "run(address)",
+        await contractPromises.World,
+        "--rpc-url",
+        rpc,
+        "--private-key",
+        privateKey,
```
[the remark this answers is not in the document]
alvrs: Unfortunately foundry doesn't seem to pick up the `PRIVATE_KEY` environment variable by default to replace the `--private-key` argument, but only provides a `vm` method to load it in foundry scripts. But since we use ethers to deploy / call contracts it doesn't really matter. I've replaced our `privateKey` cli param with loading the `PRIVATE_KEY` environment variable and never logging it in clear text.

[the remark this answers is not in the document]
alvrs: And also moved to using the rpc url from `foundry.toml` (and adding a `profile` param to our cli, which acts both as selector for the foundry profile, and we can use it to have default logic based on the profile, eg automatically funding the wallet when deploying to the lattice testnet).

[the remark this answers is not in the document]
alvrs: Looks like `FOUNDRY_PROFILE` is ignored in `cast block-number`, maybe also in other cast commands? Reported here: https://github.com/foundry-rs/foundry/issues/4474, will still load the rpc url from the foundry.toml but also pass as an explicit argument to cast for now.

packages/cli/src/utils/deploy-v2.ts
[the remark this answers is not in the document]
holic: this is what I was referring to: https://docs.ethers.org/v5/api/experimental/#experimental-noncemanager

packages/cli/src/utils/deploy-v2.ts
```diff
@@ -0,0 +1,280 @@
...
+  const contractPromises = Object.keys(mudConfig.systems).reduce<Record<string, Promise<string>>>(
+    (acc, systemName) => {
+      acc[systemName] = deployContract(systemName);
+      return acc;
+    },
+    { World: deployContract(worldContractName) }
+  );
+
+  // Create World contract instance from deployed address
+  const WorldContract = new ethers.Contract(await contractPromises.World, WorldABI, signer) as World;
```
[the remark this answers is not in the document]
alvrs: It does break if someone calls their system `World` with the intention of deploying it as a system, but it does allow for users to define their own contract called `World` to use as the world contract with the default config (it's also possible for them to use a contract not called `World` as World by specifying the `worldContractName` config param).

I can add a check for preventing `World` as key in `overrideSystems` to make it more explicit that `World` is not a valid system name though.

alvrs:
Implemented feedback and added a couple cleanups and small features (like auto-retrying transactions if the error is recoverable). Ready for another look. (Commits since your last review: https://github.com/latticexyz/mud/pull/450/files/eabfc00c6a489a2ecb6a7fd4eac1e559fbca80fb..9250d5c466f0ccbdf52c9471de090b8a2a05c899)

packages/cli/src/utils/deploy-v2.ts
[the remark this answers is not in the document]
alvrs: Oh I see, wasn't aware of that! I'd still keep the manual nonce handling for now since it's pretty simple and might help with porting to viem.

dk1a (approved):
lgtm
schema encoder may be better off in its own file, but that can be done together with the other refactoring (after moving solidity Schema)

packages/cli/src/commands/deploy-v2.ts
```diff
@@ -25,19 +28,27 @@ const commandModule: CommandModule<Options, Options> = {
     return yargs.options({
       configPath: { type: "string", desc: "Path to the config file" },
       printConfig: { type: "boolean", desc: "Print the resolved config" },
-      rpc: { type: "string", desc: "RPC endpoint to deploy to", default: "http://127.0.0.1:8545" },
-      privateKey: {
-        type: "string",
-        desc: "Private key of the deployer account",
-        default: "0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80", // Anvil default private key
+      profile: { type: "string", desc: "The foundry profile to use" },
```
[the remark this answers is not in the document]
alvrs: Added this!

packages/cli/src/utils/foundry.ts
```diff
@@ -1,24 +1,99 @@
...
+  return options?.silent
+    ? (
+        await execa("forge", args, {
+          env: { FOUNDRY_PROFILE: options?.profile },
+          stdout: "pipe",
+          stderr: "pipe",
+        })
+      ).stdout
+    : execLog("forge", args, {
+        env: { FOUNDRY_PROFILE: options?.profile },
```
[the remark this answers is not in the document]
alvrs: all environment variables are part of `process.env` (because we use `dotenv` in `mud.ts`), and everything that is part of `process.env` is automatically included in `execa`'s `env`. This means we actually wouldn't need a `--profile` flag in our cli, but I think it's still useful to explicitly know the profile in our cli, so we can 1. log it to the user and 2. add more advanced features later, like auto-funding accounts via faucets if a certain profile is used (which could arguably also be done via a check on the rpc url, but I still think being able to set the profile via a cli param in addition to falling back to the env variable is good ux).

holic (approved):

packages/cli/src/utils/foundry.ts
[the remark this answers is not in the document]
alvrs: Personally I find the ux of using a flag a bit nicer than writing `FOUNDRY_PROFILE=profile` in front of the comment, but this is not a strong opinion. The other thing I was envisioning was adding default options based on the profile arg itself, instead of requiring them in the foundry config. (Eg `--profile lattice-testnet` -> default values for rpc, auto-funding the deployer wallet from the faucet, etc). Using the `foundry.toml` / `FOUNDRY_PROFILE` env variable feels like an unnecessary indirection here. We could separate that use-case from the "foundry profile" arg though. In any case I propose doing this in a follow up.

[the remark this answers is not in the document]
alvrs: Oh one other thing: these are utils, where a consumer might want to explicitly execute one of them with a profile other than the one found in `FOUNDRY_PROFILE`. Defaulting to the `FOUNDRY_PROFILE` happens one level higher in the `command` itself.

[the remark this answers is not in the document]
holic: I'm good with all that! Foundry basically has the same: env vars + matching command args.

PR #451, opened 2023-02-28
feat(cli): reorganize internal structure and add exports for all utilities

dk1a:
this is just internal reorg for our convenience, the commands in deprecated folders aren't yet marked as deprecated because there's nothing to replace them yet. I'm also being greedy with what I put in deprecated folders, some utils can always be moved back if necessary

@alvrs this PR is small in scope but will likely cause significant conflicts with other cli changes, so it's best to deal with this one sooner rather than later

haven't tested with old utils, but this now allows doing `renderTablesFromConfig(await parseStoreConfig(config))` to get raw text output without actually calling cli, and write it to wherever


holic:
very +1 to moving things to deprecated to make it clear what's planning to go away/needs replacing

packages/cli/src/commands/index.ts
```diff
@@ -1,18 +1,19 @@
...
-import codegenLibdeploy from "./codegen-libdeploy.js";
-import deployContracts from "./deploy-contracts.js";
+import bulkupload from "./deprecated/bulkupload.js";
+import callSystem from "./deprecated/call-system.js";
+import codegenLibdeploy from "./deprecated/codegen-libdeploy.js";
+import deployContracts from "./deprecated/deploy-contracts.js";
+import systemTypes from "./deprecated/system-types.js";
+import test from "./deprecated/test.js";
+import trace from "./deprecated/trace.js";
+import types from "./deprecated/types.js";
```
holic, line 10: I don't think I noticed this before but why the explicit `.js` in imports?

dk1a: I'm being compatible with `moduleResolution="nodenext"`. In a sense this could be run in a browser (maybe, eventually)

alvrs (approved):
lgtm!

packages/cli/src/commands/index.ts
dk1a: basically this - https://nodejs.org/api/esm.html#mandatory-file-extensions

PR #459, opened 2023-03-04
feat(cli): add prototypes to tablegen

dk1a:
A prototype is a collection of tables with the same `primaryKeys`
You can use it to set/delete all tables with one method.
Default values are a bit clunky and don't work well with dynamic fields. I'm not yet sure how useful they will be.
Also a helper to get all tableIds, which can be used by a `Prototype` table and `LibPrototype` to dynamically copy prototypes (but these aren't a part of this PR)

also adds minor improvements, in particular `storeArgument` for all relevant methods and a new shorthand `schema: SchemaType....`


alvrs (commented):
Didn't review in-depth yet, but looks good on first sight! Could you add an integration/example branch to `v2sandbox` to illustrate this change?

holic (commented):
(oops, forgot to submit this review)

dk1a:
ready for review

see https://github.com/latticexyz/v2sandbox/tree/dk1a/prototypes for some examples (they're real prototypes from 3 mud games)

v2sandbox doesn't have tests, I just tested that it all compiles (prototypes there are too complicated. It would be useful to add complex tests later though, when prototypes are more stable).
cli itself has some create/destroy tests for simple prototypes.

packages/cli/src/config/parseStoreConfig.ts
```diff
@@ -39,38 +79,93 @@ const TableDataFull = z
...
+  /** Table names used in this prototype, mapped to their options */
+  tables: Record<
+    string,
+    {
+      /**
+       * Default value is either a string for structs, or a record of strings for each key
+       *
+       * The string is rendered in solidity as-is, unescaped and without adding quotes
+       */
+      default?: string | Record<z.input<typeof ColumnName>, string>;
```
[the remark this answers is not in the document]
dk1a: Yeah I'd prefer it to be typed, but that's complicated and would kinda use a lot of v2 networking/schema-type stuff, which I'd rather take/generalize later than reimplement now. It's also not a very important feature and I have some doubts about its usefulness

packages/store/src/Schema.sol
```diff
@@ -188,9 +173,9 @@ library SchemaLib {
     return Schema.unwrap(schema) == bytes32(0);
   }
 
-  function validate(Schema schema) internal pure {
+  function validate(Schema schema, bool allowEmpty) internal pure {
```
dk1a, line 176: this is needed to allow empty keys for singletons

packages/cli/contracts/src/prototypes/Tables23.sol
```diff
@@ -0,0 +1,30 @@
...
+pragma solidity >=0.8.0;
+
+/* Autogenerated file. Do not edit manually. */
+
+import { Table2, Table2Data } from "./../tables/Table2.sol";
+import { Table3 } from "./../tables/Table3.sol";
+import { Enum1 } from "./../Types.sol";
+
+library Tables23 {
+  function create(bytes32 key, Table2Data memory _data_Table2, Enum1 _value_Table3) internal {
```
alvrs, line 11: What do you think about generating these without the `_data` or `_value` prefix? Thinking about UX of using the libraries, especially with a prototypes with lots of value I see people using names arguments like

```solidity
Tables23.create({
  _key: 0x01,
  _data_Table2: table2Data,
  _value_Table3: enum1
});
```

and that would be nicer like

```solidity
Tables23.create({
  _key: 0x01,
  _Table2: table2Data,
  _Table3: enum1
});
```

packages/cli/contracts/src/prototypes/Tables23.sol
```diff
@@ -0,0 +1,30 @@
...
+    Table2.deleteRecord(key);
+
+    Table3.deleteRecord(key);
+  }
+
+  function getTableIds() internal pure returns (uint256[] memory _tableIds) {
+    _tableIds = new uint256[](2);
+
+    _tableIds[0] = uint256(bytes32(abi.encodePacked(bytes16(""), bytes16("Table2"))));
+
```
alvrs, line 27: formatting nit: can we remove the `\n` here and in `create` and `destroy`?

packages/cli/contracts/src/prototypes/Tables23.sol
dk1a: updated it to only resort to field name prefixes for multi-argument tables

dk1a:
closing in favor of #500

PR #460, opened 2023-03-05
feat(react): option for useEntityQuery to re-render on value changes

dk1a:
I have 2 use-cases for this:
1. Getting values from 2+ components using a custom query (e.g. `HasValue` for one, `Has` for the other)
2. Complex chain of methods that use different components to resolve foreign keys. These methods are reused in different hooks with different nesting depths. The only way I see to do this semi-cleanly is to use non-hook getters that call each other while specifying which components they depend on, and then make it all re-render based on any changes to that array of components

1 can maybe be also solved with a more generalized `getComponentValues`
2 would need some deep foreign key support in recs or something, and even then it may not solve everything

### DIFF
--- a/packages/react/src/useEntityQuery.ts
+++ b/packages/react/src/useEntityQuery.ts
@@ -12,11 +12,11 @@ import { distinctUntilChanged, map } from "rxjs";
  * and triggers a re-render as new query results come in.
  *
  * @param fragments Query fragments to match against, executed from left to right.
- * @param options.recomputeOnValueChange False - re-renders only on entity array changes. True (default) - also on component value changes.
+ * @param options.updateOnValueChange False - re-renders only on entity array changes. True (default) - also on component value changes.
  * @returns Set of entities matching the query fragments.
  */
-export function useEntityQuery(fragments: QueryFragment[], options?: { recomputeOnValueChange?: boolean }) {
-  const recomputeOnValueChange = options?.recomputeOnValueChange ?? true;
+export function useEntityQuery(fragments: QueryFragment[], options?: { updateOnValueChange?: boolean }) {
+  const updateOnValueChange = options?.updateOnValueChange ?? true;
 
   const stableFragments = useDeepMemo(fragments);
   const query = useMemo(() => defineQuery(stableFragments, { runOnInit: true }), [stableFragments]);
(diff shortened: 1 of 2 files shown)


packages/react/src/useEntityQuery.ts
```diff
@@ -7,19 +7,29 @@ import { distinctUntilChanged, map } from "rxjs";
...
-export function useEntityQuery(fragments: QueryFragment[]) {
+/**
+ * Returns all matching `EntityIndex`es for a given entity query,
+ * and triggers a re-render as new query results come in.
+ *
+ * @param fragments Query fragments to match against, executed from left to right.
+ * @param recomputeOnValueChange False - re-renders only on entity array changes. True - also on component value changes.
+ * @returns Set of entities matching the query fragments.
+ */
+export function useEntityQuery(fragments: QueryFragment[], recomputeOnValueChange = false) {
```
holic, line 18: Can we make this second argument an object so we can add more options in the future?

holic (approved):
This looks great, thank you!

I wonder if we should default `recomputeOnValueChange: true` even though it would do a lot more rerendering, since most use cases would probably want the entity + its value.

dk1a:
> I wonder if we should default recomputeOnValueChange: true even though it would do a lot more rerendering, since most use cases would probably want the entity + its value.

Changed default to true.
All my current uses (12 calls in 10 different hooks) either need rerender on value changes, or the value doesn't really change so it doesn't matter. I assume other projects would be similar enough

> Or we could create a separate hook where this is the default and values are included as well.

It's much more convenient for me to just call `getComponentValueStrict`, instead of trying to merge arrays of entities and values. And for the nested foreign key stuff I wouldn't want use those values at all.

Having a `getComponentsValues` of some sort where you pass a callback and get typed individual values for all components and then parse them as needed sounds cool, but also harder to make

packages/react/src/useEntityQuery.ts
```diff
@@ -7,19 +7,31 @@ import { distinctUntilChanged, map } from "rxjs";
...
+/**
+ * Returns all matching `EntityIndex`es for a given entity query,
+ * and triggers a re-render as new query results come in.
+ *
+ * @param fragments Query fragments to match against, executed from left to right.
+ * @param options.recomputeOnValueChange False - re-renders only on entity array changes. True (default) - also on component value changes.
+ * @returns Set of entities matching the query fragments.
+ */
+export function useEntityQuery(fragments: QueryFragment[], options?: { recomputeOnValueChange?: boolean }) {
+  const recomputeOnValueChange = options?.recomputeOnValueChange ?? true;
```
holic, line 19: This looks great, thank you!

Not blocking but I'm wondering if something like `updateOnValueChange` would align a little better with existing React terminology. And I'm also wondering if we should invert this and have the option be something like "update on distinct entities" or "ignore component value changes", but that's a mouthful.

holic (approved):

packages/react/src/useEntityQuery.ts
dk1a: update does seem better, changed it

holic (approved):

PR #465, opened 2023-03-07
feat(network): Allow for eip-1559 (type 2) transactions in `network`

0xhank:
Problem was manually setting gasPrice in `createTxQueue` was forcing transactions to be submitted as legacy type, because you can't submit an eip1559 while overriding gasPrice.

This update instead sets the value of `maxPriorityFeePerGas` and `maxFeePerGas` which are the proper ways to set gas price using eip1559 anyway.

### DIFF
--- a/packages/network/src/createTxQueue.ts
+++ b/packages/network/src/createTxQueue.ts
@@ -119,12 +119,13 @@ export function createTxQueue<C extends Contracts>(
 
         // Populate config
         const configOverrides = {
-          gasPrice: gasPrice$.getValue(),
+          maxPriorityFeePerGas: gasPrice$.getValue() * 2,
+          maxFeePerGas: gasPrice$.getValue() * 4,
           ...overrides,
           nonce,
           gasLimit,
         };
-        if (options?.devMode) configOverrides.gasPrice = 0;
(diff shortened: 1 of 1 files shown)


ludns (changes requested):
see comment above

dk1a:
based on my similar experience with lattice testnet using
```solidity
maxPriorityFeePerGas: gasPrice$.getValue() * 2,
maxFeePerGas: gasPrice$.getValue() * 4,
```
was not enough, and I had to hardcode actual numbers instead of `gasPrice$.getValue()`. Haven't looked into the specifics, but there might be something wrong with `gasPrice$`

ludns:
Gas price from the node is incorrect for some reason, and it's generally not recommended to turn gas price into base fee.
I would look at the `deploy-v2` script in the CLI for a robust way to fetch the latest baseFee and  bump it up using ethers [getFeeData](https://docs.ethers.org/v5/api/providers/provider/#Provider-getFeeData)

alvrs:
Closing this for now as it's only relevant for v1 and our focus is on getting v2 stable. Happy to merge a PR to the [`v1`](https://github.com/latticexyz/mud/tree/v1) branch if a contributor wants to pick it up.

PR #466, opened 2023-03-07
feat(cli): add setMetadata to autogen of table libraries

holic:
Pulled out of https://github.com/latticexyz/mud/pull/415

Not sure if this is the _right_ approach but it works/does what I want it to! Happy to make adjustments.


holic:
After making this update to `renderTable`, do I need to run anything to regenerate built-ins?

dk1a:
yeah this is what I was going to do, just didn't prioritize it, looks good.

>After making this update to renderTable, do I need to run anything to regenerate built-ins?

Run tablegen in store and world, and see if all their tests pass. I think committing the table updates in this PR also makes sense

dk1a:
(In a future PR I also start adding tests to cli directly, but that's not relevant here)

holic:
Some tests are now failing but unclear how this is related
```

Failing tests:
Encountered 1 failing test in test/Mixed.t.sol:MixedTest
[FAIL. Reason: StoreCore_InvalidDataLength(0, 197)] setUp() (gas: 0)

Encountered 1 failing test in test/StoreCore.t.sol:StoreCoreTest
[FAIL. Reason: StoreCore_InvalidDataLength(0, 197)] setUp() (gas: 0)

Encountered 1 failing test in test/StoreMetadata.t.sol:StoreMetadataTest
[FAIL. Reason: StoreCore_InvalidDataLength(0, 197)] setUp() (gas: 0)

Encountered 1 failing test in test/StoreSwitch.t.sol:StoreSwitchTest
[FAIL. Reason: Setup failed: 521b3d33:] setUp() (gas: 0)

Encountered 1 failing test in test/Vector2.t.sol:Vector2Test
[FAIL. Reason: StoreCore_InvalidDataLength(0, 197)] setUp() (gas: 0)

Encountered 1 failing test in test/tables/Callbacks.t.sol:CallbacksTest
[FAIL. Reason: StoreCore_InvalidDataLength(0, 197)] setUp() (gas: 0)

Encountered 1 failing test in test/tables/Hooks.t.sol:HooksTest
[FAIL. Reason: StoreCore_InvalidDataLength(0, 197)] setUp() (gas: 0)
```

```
2023-03-07T17:37:08.880946Z ERROR forge::runner: setUp failed reason="521b3d33:" contract=0x34a1d3fff3958843c43ad80f30b94c510645c316

Running 1 test for test/World.t.sol:WorldTest
[FAIL. Reason: Setup failed: 521b3d33:] setUp() (gas: 0)
Test result: FAILED. 0 passed; 1 failed; finished in 1.45ms

Failing tests:
Encountered 1 failing test in test/World.t.sol:WorldTest
[FAIL. Reason: Setup failed: 521b3d33:] setUp() (gas: 0)
```

dk1a:
Maybe somewhere metadata is being set before the metadata table is registered? can look into it a bit later if you want

holic:
> Maybe somewhere metadata is being set before the metadata table is registered? can look into it a bit later if you want

yeah that's exactly what it was, working on a fix now

packages/store/src/StoreCore.sol
```diff
@@ -74,7 +75,11 @@ library StoreCore {
    * Get the schema for the given table
    */
   function getSchema(uint256 table) internal view returns (Schema schema) {
-    return StoreCoreInternal._getSchema(table);
+    schema = StoreCoreInternal._getSchema(table);
+    // A race condition prevents us from doing this check ~everywhere
+    // if (schema.isEmpty()) {
+    //   revert StoreCore_TableNotFound(table);
+    // }
```
holic, line 82: I saw we were doing this check in just one of the read functions (removed below) but not in the others or in any of the writes. I thought this would be a cheap check to do and worth adding everywhere, to make the error messages more useful (was getting `StoreCore_InvalidDataLength` rather than `StoreCore_TableNotFound` in tests).

But, there seems to be a race condition somewhere when registering schemas and/or metadata, where the schema is requested, but not found, therefore we can't write to it. So I think somewhere we're sort of writing data ahead of the schema existing?

packages/store/gas-report.txt
```diff
@@ -14,9 +14,9 @@
 (test/Gas.t.sol) | pass abi encoded bytes to external contract [someContract.doSomethingWithBytes(abiEncoded)]: 6554
 (test/Gas.t.sol) | pass custom encoded bytes to external contract [someContract.doSomethingWithBytes(customEncoded)]: 1381
 (test/Mixed.t.sol) | store Mixed struct in storage (native solidity) [testMixed = mixed]: 92050
-(test/Mixed.t.sol) | register Mixed schema [Mixed.registerSchema()]: 35047
-(test/Mixed.t.sol) | set record in Mixed [Mixed.set({ key: key, u32: 1, u128: 2, a32: a32, s: s })]: 111032
-(test/Mixed.t.sol) | get record from Mixed [MixedData memory mixed = Mixed.get(key)]: 13167
+(test/Mixed.t.sol) | register Mixed schema [Mixed.registerSchema()]: 423749
```
holic, line 17: I'm gonna assume we're okay with this large jump in gas (10x) because its a one-time thing on a world/deploy.

packages/store/src/StoreCore.sol
holic: Just did a bit more testing, and this issue comes from a `getField` call.

holic: image

```

Running 1 test for test/World.t.sol:WorldTest
[FAIL. Reason: Setup failed: Schema not found in getField call] setUp() (gas: 0)
Traces:
  [120595] WorldTest::setUp()
    ├─ [79789] → new <Unknown>@0xFEfC6BAF87cF3684058D62Da40Ff3A795946Ab06
    │   ├─ emit StoreSetRecord(table: 31851884016513291979465943704761379125824557917914706133157551191634376234846, key: [0x466b87090ce2bc34362737dab7b900dfc94a202aee6a6323f6813f7ce59d975e], data: 0x002001005f000000000000000000000000000000000000000000000000000000)
    │   ├─ emit StoreSetRecord(table: 31851884016513291979465943704761379125824557917914706133157551191634376234846, key: [0x15985d03d398bf6f1a658e500e78fbbf35cd784646b3998d4f26a9a9e441a627], data: 0x00000002c5c40000000000000000000000000000000000000000000000000000)
    │   ├─ emit StoreSetRecord(table: 9767772539668730435636382986942077369059354384034778394590643984438732629543, key: [0x15985d03d398bf6f1a658e500e78fbbf35cd784646b3998d4f26a9a9e441a627], data: 0x0000010d000d010000000000000000000000000000000000000000000000000053746f72654d65746164617461000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000040000000000000000000000000000000000000000000000000000000000000008000000000000000000000000000000000000000000000000000000000000000097461626c654e616d6500000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000014616269456e636f6465644669656c644e616d6573000000000000000000000000)
    │   └─ ← 0 bytes of code
    └─ ← "Schema not found in getField call"
```

packages/cli/src/render-table/renderTable.ts
```diff
@@ -64,10 +67,20 @@ ${
   /** Register the table's schema for the specified store */
   function registerSchema(${renderArguments([_typedTableId, "IStore _store"])}) internal {
     _store.registerSchema(_tableId, getSchema());
+
+    (string memory _tableName, string[] memory _fieldNames) = getMetadata();
+    StoreSwitch.setMetadata(_tableId, _tableName, _fieldNames);
```
dk1a, line 72: just noticed that it should be `_store` here, not `StoreSwitch`

packages/store/src/StoreCore.sol
```diff
@@ -39,6 +40,11 @@ library StoreCore {
     // Register other internal tables
     Hooks.registerSchema();
     StoreMetadata.registerSchema();
+    // Then set their table metadata. If this is done before `registerSchema`,
+    // then downstream validation will fail due to `StoreMetadata` and `Hooks`
+    // loosely depending on one another (by way of `StoreCore` internals).
+    Hooks.setMetadata();
+    StoreMetadata.setMetadata();
```
holic, line 47: @dk1a felt cleaner to just separate the two paths rather than reproduce their logic here

It might be worth adding an `initialize` function for each of the autogen libs that does the `registerSchema` and `setMetadata` so that downstream usages can use that. It's just here that we have call them separately because of the intertwined dependency.

dk1a: > @dk1a felt cleaner to just separate the two paths rather than reproduce their logic here (I think you suggested this initially anyway)

yeah this does look better than calling `StoreSwitch` directly

packages/store/src/StoreCore.sol
dk1a: `StoreMetadata.registerSchema();` calls `StoreSwitch.setMetadata(_tableId, _tableName, _fieldNames);`, which tries to call 
`getField` for hooks when setting metadata fields, but the hook table isn't registered yet
And if you do hooks table first the issue is reversed for metadata

alvrs (approved):
Lgtm!

alvrs (approved):

alvrs (approved):

PR #467, opened 2023-03-08
feat(world): simplify access control to namespaces instead of routes

### LINKED ISSUE #456: Proposal: Extending the Filesystem of World
# Introduction
The World has an implicit filesystem described with Routes. In this Filesystem, two types of "files" are stored: Tables and Systems.
Currently, the Filesystem is used to facilitate two features:
- Access control for Store: Systems in the same folder as a Table can write to the Table
- Access control for Systems: Systems in the same folder as a System can call the System

As planned in #393 (World framework), nested routes are used to proxy write access (ie: Systems having access to the parent route of a Table or System in a child route can call / write them). This is however not currently implemented.

I propose three modifications to the Filesystem:
1) Rename subRoute to File
2) Add file types to differentiate between directories (a File that contains other files), Systems, and Tables.
3) Add one additional file type

### 1) Rename subRoute to File and store all File related data into a FileTable
I have found subRoute to be a confusing name; and I propose renaming it to File to match expectations. This change -- combined with adding file types -- turns the current pseudo-Filesystem into something more akin to a traditional filesystem.
An initial FileTable schema could look like such:
* primary key: routeId (`uint256`)
* name: `string`
* parent: routeId (`uint256`)
* content: `uint256` (stores an address for all file type but the TABLE one, where it instead stores the Schema) 
* fileType: `DELEGATECALL`, `CALL`, `TABLE`, `ASSET`
### 2) Add file types to differentiate between directories (a File that contains other files), Systems, and Tables.
Currently, file types are implicit; and different types of resources can have the same full-name (eg: there can be a system at `/mudswap/SwapSystem` and a table at `/mudswap/SwapSystem`, which is IMO very confusing).
I propose creating a unified FileTable which enforces unique File names, even if the Filetype is different.
Enforcing unique names across different file types mean we can drop the file type when operating over files.
The initial set of proposed file types are found in the _Appendix: Description of the file types_
### 3) Add one additional file type
There is a case to be made for non-executable data stored in the filesystem. Additionally, we should pave the way for more file types than the ones we have right now. 
I propose adding one additional file types over the three implicit ones existing (Folder, System, and Table) 
- `ASSET`: a piece of data stored under an address with SSTORE2 that is not executable.
This would allow each World to store static files in the style [EthFS](https://github.com/holic/ethfs).
Systems will be able to write to these files (in practice: deploy a new contract and change the `content` field of the file to the new address). This can be used by clients to load assets from the chain.

### Appendix: Description of the file types
* `FOLDER`: a file that can be used as a baseRoute. created with `createdFolder` (current `registerRoute`)
* `SYSTEM`: the EVM byte code at the corresponding address will be `CALL`ed called from the World (or `DELEGATECALL`ed for root Systems); some context will be appended on the calldata (for now: the original msg.sender) 
* `TABLE`: a Store table, where its Schema can be found in the content column. It doesn’t have an address because it lives on the OS
* `ASSET`: anything else stored via SSTORE2 on-chain.

### LINKED ISSUE #462: Proposal: Simplify World v2 Access Control
## Current Situation / Recap

- In [https://github.com/latticexyz/mud/issues/393](https://github.com/latticexyz/mud/issues/393) / [https://github.com/latticexyz/mud/pull/405](https://github.com/latticexyz/mud/pull/405) we introduced a new access control mechanism for resources in the World (tables, systems, maybe more, see [https://github.com/latticexyz/mud/issues/456](https://github.com/latticexyz/mud/issues/456))
- The access control mechanism is based on routes
    - A single route fragment starts with `/` and doesn’t include further `/`.
    - A route can have arbitrary depth (as many route fragments chained together as you want, eg `/skystrife/combat/AttackSystem`).
    - Access rights are inherited to subroutes: if an address has access to `/skystrife`, it automatically has access to `/skystrife/combat/AttackSystem`.
    - This is achieved by passing two separate parts of the route to any method that requires access control: an `accessRoute`, and a `subRoute`. Access control checks are done on the `accessRoute`, and if the access control passes the `subRoute` is appended to the `accessRoute` to form the full route of the resource to access. (Eg. `accessRoute = /skystrife`, `subRoute = /combat/AttackSystem`)
- This method is relatively gas efficient, because only a single storage access is required to verify the permission of a caller (only checking `accessRoute`) for a nested route of arbitrary depth.
- However, the method is not very developer friendly - a developer has to *know* which parent route their access to a specific resource is based on to be able to specify this part of the route as `accessRoute`.
- Further downsides of this approach:
    - Passing two dynamic length strings as part of every resource access adds the gas overhead of two bytes32 fields that Solidity uses to store the length of the strings.
    - Internally the full route has to be hashed to normalise it to a 32 byte word (eg to make it compatible with *Store,* among other gas efficiency reasons). This is bad for developer experience, because it is not immediately obvious what a route refers to without checking the hash's pre-image from the table that we have to create for this purpose.
    - The whole approach of deeply nested routes and splitting them into `accessRoute` and `subRoute` and thereby inheriting access is relatively hard to onboard / wrap your head around.
    

## Alternative approach 1: Simple Namespaces

- What if instead of allowing arbitrary depth of routes, we limit the depth of routes to 2: `/namespace/file`.
- Access control would be based on the `namespace` by default (= an address with access to the namespace has access to every resource in the namespace), but could also be granted for individual files.
- We would lose support for deeply nested routes at launch, but could always have a separate entry point in the World contract to add support for deeply nested routes using the previous approach, if we realize we need them. So far it doesn’t seem like there are strong examples for situations where deeply nested routes are needed.
- In turn we would get a bunch of advantages:
    - We can limit the size of a namespaces selector to `bytes16`, and the size of a file selector to `bytes15`, so they can fit into a single `bytes32` word when concatenated with a separator character (eg `/`).
        - **More gas efficient:** This means we save on gas for calldata, because instead of 2x `bytes32` for dynamic length strings plus ~1 byte per character in each string, we only require a single 32 byte word for the full concatenated `namespace/file` selector.
        - **Better developer experience for understanding routes:** This also means we can keep the full route to any resource in clear text, without hashing it, as long as the `namespace` and `file` are strings with max length ~16/15 characters respectively. If longer names are required, we could still fall back to hashing them to the required length.
    - **Better developer experience for accessing resources**: When accessing a resource, the caller would provide both the namespace and file selector of the resource they want to access. The access control check could first check access on the namespace (the more common use case), and if it fails check access on the individual `namespace/file` selector. This means the caller doesn’t need to know if their access is based on the namespace of the individual file - the contract would figure it out. In the worst case it would mean two storage accesses (if access is based on the file). As an optimization, we could add a second entry point that can be called with a single  `bytes32 namespace_file` selector, if the caller already knows their access is based on the full file selector, to skip the separate `namespace` check.
    

### Code exploration

- Here is a simple code exploration of what this change would look like in the World contract: [https://github.com/latticexyz/mud/commit/4c4c6e283044031b1aa9fd281d5bb7b4c342e832](https://github.com/latticexyz/mud/commit/4c4c6e283044031b1aa9fd281d5bb7b4c342e832)
    - Very little actually has to change, the required changes make the code simpler (because we don’t need to juggle between “id” and “route pre-image”)
- Here is a little exploration of what calling systems via the World would look like (without wrapper libraries / types)
    
    ```solidity
    // Default - check access first on namespace, then on file if no namespace access:
    world.call("mud", "increment", abi.encodeWithSelector(IncrementSystem.increment.selector, 1)); 
    
    // Optimized - only check access on the full selector if the caller knows they don't have namespace access, or if the system is public
    world.call("mud/increment", abi.encodeWithSelector(IncrementSystem.increment.selector, 2));
    
    // Even simpler if the system only has a single fallback function
    world.call("mud/increment", abi.encode(3));
    ```
    

- Not too relevant for this proposal, but still mentioning here for context:
    - For the first two cases we’d add autogenerated wrapper libraries to allow calling the systems with native types.
    - For the last case we could even go as far as registering a world function selector to be called via the World’s fallback function - this would allow us to easily generate an interface (instead of a library) for the World via which systems can be called with native Solidity types instead of manual `abi.encode`.

---

## Alternative Approach 2: Roles
This came up when discussing the idea with @holic
- What if instead of `namespace`, we implement access control based on `roles`?
- The general approach would be very similar to the `namespace` approach (see above), but instead of `files` being part of exactly one namespace, they could be accessible by different roles. By default all resources by single developer would have access to all the other resources (equivalent to a namespace). But in addition, access to different subsets of the resources could be shared with other roles.
- To have the same gas efficiency as the `namespace` approach, we could store access for each role in a separate entry (instead of eg. an array). As an example: if we want to give `role1` and `role2` access to `systemA`, we’d store `true` in the `AccessTable` at key `role1` and at `role2`, and also store the file data (eg address of a system) in the `SystemTable` at `role1/systemA` and `role2/systemA`. We’d also need to store all roles having access to `systemA` in an array somewhere, to be able to update all references when updating the system. This would increase gas cost for registering new files (one time), but keep the gas cost low for accessing files.
- Unfortunately this approach wouldn’t map as nicely to tables: with namespaces there is a single `bytes32 namespace_table` key that can be used to refer to the table in *Store*, while with roles we’d need another concept of the table id (because `role1/table` and `role2/table` can’t refer to the same table), and therefore an additional storage access for writing/reading from/to tables.
- After thinking about the role approach more, I think it would be slightly harder to wrap your head around than the very simple namespace approach. The role approach would make more fine grained access control to a subset of resources easier though.
- Mostly for the gas efficiency reason I’m personally in favour of the namespace approach, it’s simplicity is a nice bonus.

I’m curious for your thoughts on the general idea of simplifying access control to resources in the World contract, as well as thoughts on the two alternative approaches (or any other alternative approaches that come to mind)

alvrs:
fixes #462 

~WIP, TODOs in this PR~ All done!
- [x] fix failing tests
- [x] merge main into this branch
- [x] remove unused tables
- [x] update deploy CI to handle new World methods
- [x] update tablegen to use namespace/file as table id
- [x] turn ResourceSelector into custom type or at least use "using for bytes32"
- [x] Add ResourceType table, prevent registration of files where files exist (closes #456)

### Looking for feedback on naming:
Right now there are three names: 
* `bytes32 resourceSelector`: identifier for any resource in the World. This includes Systems, Tables, Namespaces.
* `bytes16 namespace`: the upper 16 bytes of the `resourceSelector`, but also implicitly the `resourceSelector` `bytes32(namespace)` (which describes the "Namespace" resource, kind of like the directory)
* `bytes16 file`: the lower 16 bytes of the `resourceSelector`

I wonder whether `file` is the most descriptive name for the lower 16 bytes of the resource selector, and how we can make it more obvious that the `namespace` can also be a resource selector for the Namespace resource if the lower 16 bytes are 0.


packages/world/src/types.sol
```diff
@@ -0,0 +1,11 @@
...
+pragma solidity >=0.8.0;
+
+/* Autogenerated file. Do not edit manually. */
+
+enum Resource {
+  NONE,
+  NAMESPACE,
+  TABLE,
+  SYSTEM
+}
```
alvrs, line 11: User types already being useful!

packages/store/src/Utils.sol
```diff
@@ -6,6 +6,10 @@ library Utils {
     return a / b + (a % b == 0 ? 0 : 1);
   }
 
+  function toString(uint256 input) internal pure returns (string memory) {
+    return string(abi.encodePacked(input));
+  }
```
dk1a, line 11: as a general utility this is a bit misleading, as it's specifically meant for table ids

packages/world/src/World.sol
```diff
@@ -43,163 +58,209 @@ contract World is Store {
...
-
-    // Give caller access to the route
-    RouteAccess.set({ routeId: routeId, caller: msg.sender, value: true });
+    // Give caller access to the new namespace
+    ResourceAccess.set(resourceSelector, msg.sender, true);
   }
 
   /**
-   * Register register a table with given schema at the given route
+   * Register register a table with given schema in the given namespace
```
dk1a, line 80:
```suggestion
   * Register a table with given schema in the given namespace
```

dk1a (commented):
lgtm
should the new tableId be validated onchain for alphanumeric?

packages/world/mud.config.mts
```diff
@@ -24,6 +41,34 @@ const config: StoreUserConfig = {
...
+    //   storeArgument: true, // TODO Add support for store argument in setter function to table autogen
+    //   tableIdArgument: true,
+    // },
+    AddressArray: {
+      // TODO: This table is only used for testing, move it to `test/tables` via the directory config once supported
+      schema: { value: SchemaType.ADDRESS_ARRAY },
+      storeArgument: true,
+      tableIdArgument: true,
+    },
+  },
```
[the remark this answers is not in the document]
dk1a: sure, #459 already fixed this, I can resolve the TODOs after the conflicts too

packages/store/src/Utils.sol
holic: would echo that above! I personally use `toString(uint)` to mean "convert uint to a string of digits" rather than "convert this uint binary to a string binary"

packages/cli/contracts/src/tables/Table1.sol
```diff
@@ -19,7 +19,7 @@ import { PackedCounter, PackedCounterLib } from "@latticexyz/store/src/PackedCou
 // Import user types
 import { Enum1, Enum2 } from "./../types.sol";
 
-uint256 constant _tableId = uint256(keccak256("/Table1"));
+uint256 constant _tableId = uint256(bytes32(abi.encodePacked(bytes16(""), bytes16("Table1"))));
```
holic, line 22: does `abi.encodePacked` leave padding for an empty `bytes16`?

packages/store/src/StoreCore.sol
```diff
@@ -12,16 +12,18 @@ import { Slice } from "./Slice.sol";
...
 
 library StoreCore {
   // note: the preimage of the tuple of keys used to index is part of the event, so it can be used by indexers
   event StoreSetRecord(uint256 table, bytes32[] key, bytes data);
   event StoreSetField(uint256 table, bytes32[] key, uint8 schemaIndex, bytes data);
   event StoreDeleteRecord(uint256 table, bytes32[] key);
 
-  error StoreCore_TableAlreadyExists(uint256 table);
-  error StoreCore_TableNotFound(uint256 table);
+  // Errors include a stringified version of the tableId for easier debugging if cleartext tableIds are used
```
holic, line 23: 😍

packages/cli/contracts/src/tables/Table1.sol
holic: also are we able to make library calls when building these constants? it would be nice to do like
```
uint256 constant _tableId = TableId.from("", "Table1");
```

packages/store/src/Utils.sol
holic: Maybe we combine this with `ResourceSelector`, move it out into a more general purpose thing (i.e. `common` package in the future)?

holic (commented):
This looks great! Only a question + suggestion, nothing particularly blocking.

packages/cli/contracts/src/tables/Table1.sol
alvrs: > does abi.encodePacked leave padding for an empty bytes16?

yes, `bytes16` always takes 16 bytes in `abi.encodePacked`, independent of its content

> also are we able to make library calls when building these constants?

Unfortunately not - we're lucky `abi.encodePacked` is an exception that is considered "constant", we can't even do bitshifts in constants

> and if not, maybe a uint type alias?

Like `type TableId is uint256`? I _think_ we could do that. Can you elaborate on why you'd prefer that over just `uint256`?

packages/store/src/Utils.sol
alvrs: happy to change it to `tableIdToString(uint256)`. I'm a bit hesitant to use `ResourceSelector` here because that's a _World_ concept and i'd like to allow _Store_ to be used with other types of ids that are not ResourceSelectors (maybe someone wants to use a `keccak256` id, or maybe a different type of selector that's not just namespace and file but other fragments)

If the `tableId` is not representable as string, the error log just includes some garbage characters, but I don't think that hurts.

packages/cli/contracts/src/tables/Table1.sol
holic: I guess the type alias wouldn't matter if we can't call methods on that alias. So it's not much better.

alvrs: The slightly annoying thing with a custom `TableId` type is `uint256` can't be implicitly converted to `TableId`, so we'd have to do `TableId.wrap(tableId)` in all consuming libraries that call any Store function. For simplicity my preference would be to keep it `uint256`.

alvrs: Okay think I found a good compromise: with a library `TableId` and `using TableId for uint256` in `StoreCore` we can do `tableId.toString()`, without the overhead of a custom type

packages/store/src/StoreCore.sol
```diff
@@ -326,53 +331,53 @@ library StoreCore {
...
-      return StoreCoreInternal._getDynamicField(table, key, schemaIndex, schema);
+      return StoreCoreInternal._getDynamicField(tableId, key, schemaIndex, schema);
     }
   }
 }
 
 library StoreCoreInternal {
   bytes32 internal constant SLOT = keccak256("mud.store");
-  uint256 internal constant SCHEMA_TABLE = uint256(keccak256("mud.store.table.schema"));
+  uint256 internal constant SCHEMA_TABLE = uint256(bytes32(abi.encodePacked(bytes16(""), bytes16("schema"))));
```
alvrs, line 360: Important node @holic @authcall: this PR changes the ID of the schema table, which is relevant for the network package and MODE

packages/cli/contracts/src/tables/Table1.sol
holic: Would this affect all uints? Wondering if we should be more explicit like castToString.

packages/store/src/StoreCore.sol
holic: Does the empty namespace mean anything special? If not, do we wanna use a “mud” or “mud.internals” namespace for ourselves? Leaves the empty namespace for userland.

alvrs: The schema table is gonna be registered by the deployer of the world, so they'd own the `mud` namespace. Doesn't really matter though, happy to go with a mud specific namespace here

alvrs: updated namespace for all store tables to `mudstore`

packages/cli/contracts/src/tables/Table1.sol
alvrs: this means inside of `StoreCore` you could call the `toString` method on all `uint256`, yes. But as long as we don't do that it should not be a problem. We can also just use `TableId.toString` and remove using for if you think it's an issue.

holic: I've just personally used `toString` on `uint` specifically to translate e.g. `123` into `"123"`. Same is true for a couple of common Solidity libraries: https://github.com/transmissions11/solmate/blob/main/src/utils/LibString.sol#L4-L8

So I'm just a little wary of overlapping naming. I'll personally need to make a mental leap to remind myself that this is more of a cast operation.

holic (approved):

packages/store/src/StoreCore.sol
holic: I like `mudstore` or similar just because errors/reverts will make a little more sense!

PR #468, opened 2023-03-08
ci: validate build artifacts (gas report, table codegen)

holic:
I've noticed a few times where codegen tables were out of sync with the latest branch, so I've extended the previous gas report action to also validate codegen is up to date.


holic:
Gonna wait until https://github.com/latticexyz/mud/pull/470 and merge that in here too

alvrs (approved):

PR #469, opened 2023-03-08
refactor(cli): replace prettier with forge fmt

holic:
Hopefully this gets us more consistent formatting!

I intentionally left all the forge defaults, so you'll see some whitespace changes. I'm fine with this.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -55,8 +55,6 @@
     "nice-grpc-web": "^2.0.1",
     "openurl": "^1.1.1",
     "path": "^0.12.7",
-    "prettier": "^2.8.4",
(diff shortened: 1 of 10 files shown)


holic:
I just noticed that our prettier version for cli (2.8.4) was different than everywhere else in the monorepo (2.6.2) and wondering if some of our inconsistent formatting was due to mismatched versions.

holic:
Looks like what I was able to commit here into this branch is different from what is actually generated, and I think _that_ is the source of our issues. Looks like the tablegen prettier output disagrees with the precommit lint hook (which formats solidity, and with a diff prettier version).

Gonna just align the prettier versions for now.

holic:
See https://github.com/latticexyz/mud/pull/470

PR #470, opened 2023-03-08
style: bump/align prettier versions

holic:
I think this might be the reason we were seeing inconsistent formatting of autogen code!

Autogen was using the CLI's prettier version to do its formatting, then our precommit hook was using another version to reformat before files were committed. And I would guess some of us don't have husky installed/working properly, so we were committing different formattings.


holic:
Gonna wait on this until https://github.com/latticexyz/mud/pull/470 lands and I can merge that in too

alvrs (approved):

PR #471, opened 2023-03-08
fix(cli): avoid fs usage in utils, create deployment output directory if it doesn't exist

alvrs:
Fix to allow usage of cli utils in other non-nodejs packages


dk1a:
lgtm, I similarly just use mkdir

dk1a (approved):

PR #472, opened 2023-03-08
fix(cli): add missing await to tablegen, fix formatting

dk1a:


packages/store/src/tables/Callbacks.sol
```diff
@@ -47,38 +49,34 @@ library Callbacks {
   /** Get value */
   function get(bytes32 key) internal view returns (bytes24[] memory value) {
     bytes32[] memory _primaryKeys = new bytes32[](1);
-
-    _primaryKeys[0] = bytes32(key);
+    _primaryKeys[0] = bytes32((key));
```
holic, line 52: just carrying over discord convo: these extra parens are because our autogen optionally wraps values and, to keep autogen simpler (less inline conditionals), we just leave the parens here.

holic (approved):

PR #475, opened 2023-03-09
fix(cli): add missing await

dk1a:


holic (approved):

packages/cli/src/commands/deprecated/test.ts
```diff
@@ -35,15 +35,18 @@ const commandModule: CommandModule<Options, Options> = {
...
+      await resetLibDeploy(testDir);
       child.kill();
       process.exit();
     });
+
+    // Reset LibDeploy.sol
+    console.log("Reset LibDeploy.sol");
+    await resetLibDeploy(testDir);
+
+    await child;
```
holic, line 49:
```suggestion
    await child;
    
    // Reset LibDeploy.sol
    console.log("Reset LibDeploy.sol");
    await resetLibDeploy(testDir);
```

Does this need to wait until after the tests run before resetting the dir?

packages/cli/src/commands/deprecated/test.ts
```diff
@@ -35,15 +35,18 @@ const commandModule: CommandModule<Options, Options> = {
...
 
-    // Reset LibDeploy.sol
-    console.log("Reset LibDeploy.sol");
-    await resetLibDeploy(testDir);
-
-    process.on("SIGINT", () => {
+    process.on("SIGINT", async () => {
       console.log("\ngracefully shutting down from SIGINT (Crtl-C)");
+      await resetLibDeploy(testDir);
       child.kill();
```
holic, line 41: similarly, wondering if we should kill first, then reset
```suggestion
      child.kill();
      await resetLibDeploy(testDir);
```

packages/cli/src/commands/deprecated/test.ts
dk1a: right, I got confused with the refactor and broke it again haha

holic (approved):

PR #476, opened 2023-03-09
fix(services): fix protobuf import

holic:


dk1a:
check out esModuleInterop in https://github.com/stephenh/ts-proto
I'm not sure it'll fix anything, but it is likely to break something

either way, you should change the flags in makefile rather than manually edit the autogenerated stuff, and rerun `protoc-ts`

holic:
> check out esModuleInterop in https://github.com/stephenh/ts-proto I'm not sure this PR will fix anything, but it is likely to break something
> 
> either way, you should change the flags in makefile rather than manually edit the autogenerated stuff, and rerun `protoc-ts`

Good call! I didn't realize these were auto generated. But also, seems like this workaround is ~common 😭 

https://github.com/stephenh/ts-proto/issues/536#issuecomment-1198154550

alvrs:
Maybe we could add a script to automatically replace those lines until it's fixed in `ts-proto`?

dk1a:
why the star import may be bad:
https://github.com/stephenh/ts-proto/issues/738
https://github.com/vitejs/vite/issues/9018#issuecomment-1179728774

the issue seems to be a macos thing, my guess something to do with tsconfig

holic:
> the issue seems to be a macos thing, my guess something to do with tsconfig

when you say "macos thing", do we think it's an issue at build time? so every browser would see this if the client is built on macos?

PR #478, opened 2023-03-09
chore: add release:yalc script for local release testing

alvrs:
Steps to test a release locally:
1. Run `yarn release:yalc` in mud root
2. Run `npx yalc add @latticexyz/<packageToTest>` in root of the project you want to test with
3. Verify things work
4. Run `npx yalc remove @latticexyz/<packageToTest> && yarn --force` in test project to reset it to the normal state (without yalc package)


package.json
```diff
@@ -65,14 +65,13 @@
     "lint": "eslint . --ext .ts",
     "release": "yarn && lerna publish --no-private --force-publish && yarn retype:updateversion",
     "release:manual": "node scripts/workspaceRun.mjs release",
+    "release:yalc": "(yarn entry:dist && yarn lerna exec npx yalc push); yarn entry:src",
```
holic, line 68: just to clarify, this doesn't actually release anything, just creates a build artifact that _could_ be released or mirrors what would be uploaded to npm?

holic: omg the level of wrapping here is terrifying: `yarn lerna exec npx yalc push`

holic (approved):

package.json
alvrs: > just to clarify, this doesn't actually release anything, just creates a build artifact that could be released or mirrors what would be uploaded to npm?

Correct, it creates the build artifacts and publishes them to a local `yalc` registry

> omg the level of wrapping here is terrifying: yarn lerna exec npx yalc push

🙈

PR #480, opened 2023-03-10
feat: add support for key schemas

alvrs:
* Currently we only store the schema of value columns on-chain, while key columns are implicit by the number of keys emitted in the `StoreSetRecord` event (and always typed `bytes32`).
* This PR adds explicit support for key schemas to `Store`, `World` and `CLI` (tablegen + deploy):
   * Key schemas follow the same format as value schemas and are stored in the Schema table. This changes the schema of the Schema table to `{ valueSchema: Bytes32, keySchema: Bytes32 }` (used to be just `{ schema: Bytes32 }`)


alvrs:
> ooc is the world only doing the minimal necessary to bootstrap? It’s not registering userland schemas and systems etc right? That’s done via deploy script?

Correct, the World only sets up its own internal tables (and soon internal systems)

packages/store/src/StoreCore.sol
```diff
@@ -40,7 +40,11 @@ library StoreCore {
    */
   function initialize() internal {
     // Register internal schema table
-    registerSchema(StoreCoreInternal.SCHEMA_TABLE, SchemaLib.encode(SchemaType.BYTES32));
+    registerSchema(
+      StoreCoreInternal.SCHEMA_TABLE,
+      SchemaLib.encode(SchemaType.BYTES32, SchemaType.BYTES32),
```
dk1a, line 45: I think the first BYTES32 is supposed to be padding for keys now.
If so this took me a while to understand despite knowing how this all works pretty well, a comment would be useful

alvrs: The first `BYTES32` is the `valueSchema`, the second BYTES32 is the `keySchema`. Will add a comment. Also wanna turn the SchemaTable into an autogenerated table at some point, but there might be circular dependencies with the Schema existing...

dk1a: > The first BYTES32 is the valueSchema, the second BYTES32 is the keySchema

right that makes more sense, it's not just padding. What's the UINT256 one then?

alvrs: That's the Schema table's single primary key, which has type UINT256 because it represents a tableId. Added a comment about that too.

PR #482, opened 2023-03-13
feat(world): add support for modules, move logic to RegistrationModule

alvrs:
This PR adds support for "Modules" (see #393 for the original concept).
Modules encapsulate installation logic for coupled pieces of logic/data (see `RegistrationModule` as an example).

This PR also moves out all logic and tables related to "registration" to a `RegistrationModule` and the initialization of a couple core tables to `CoreModule`. This is mainly to reduce the bytecode size of World, which recently passed the contract size limit. With these changes we're well below the limit again.


packages/world/src/modules/core/CoreModule.sol
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * The CoreModule registers internal World tables.
+
+ * Note:
+ * This module is required to be delegatecalled via `World.registerRootSystem`.
+ * because otherwise the registration of tables would require the `registerTable`
+ * function selector to already exist on the World, but it is only
+ * added in `RegistrationModule`, which is installed after `CoreModule`.
+ */
```
holic, line 22: ooc do we have a way to detect if a given module has been installed, aside from looking at each individual thing being registered?

If it's worth the gas, might be cool to have as a way to introspect the world and provide nicer errors (either in client or in Solidity) if something reverts because a module was not installed before another one.

packages/world/src/modules/registration/RegistrationModule.sol
```diff
@@ -0,0 +1,90 @@
...
+ * and all required tables and function selectors in the World.
+
+ * Note:
+ * This module is required to be delegatecalled via `World.registerRootSystem`.
+ * because otherwise the table libraries would try to call methods on the World contract
+ * that are only registered in this module (e.g. `registerTable`).
+ * If the module is delegatecalled, the StoreCore functions are used directly.
+ */
+contract RegistrationModule is IModule, WithMsgSender {
+  error RegistrationModule_OnlyRoot();
```
holic, line 27: this doesn't seem used? and also isn't very self explanatory on its own

packages/world/test/World.t.sol
```diff
@@ -114,7 +117,9 @@ contract WorldTest is Test {
 
   function setUp() public {
     world = IWorld(address(new World()));
-    world.initialize();
+    world.installRootModule(new CoreModule());
+    world.installRootModule(new RegistrationModule());
```
holic, line 121: this is very cool!

holic (approved):
very cool to see this pattern in action! nothing blocking from a review standpoint, mostly questions/thoughts

packages/cli/src/utils/deploy-v2.ts
```diff
@@ -237,6 +237,8 @@ export async function deploy(mudConfig: MUDConfig, deployConfig: DeployConfig):
         throw new MUDError(
           `Error deploying ${contractName}: invalid bytecode. Note that linking of public libraries is not supported yet, make sure none of your libraries use "external" functions.`
         );
+      } else if (error?.message.includes("CreateContractLimit")) {
+        throw new MUDError(`Error deploying ${contractName}: CreateContractLimit exceeded.`);
```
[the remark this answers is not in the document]
alvrs: Agreed! In this case the original error message includes the entire bytecode, so it makes your terminal explode and buries the actual error message, but we could add a CLI flag (sth like `--debug`) to print the full error message too

packages/world/src/WithMsgSender.sol
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
+
+// Similar to https://eips.ethereum.org/EIPS/eip-2771, but any contract can be the trusted forwarder.
+// This should only be used for contracts without own storage, like Systems.
+contract WithMsgSender {
```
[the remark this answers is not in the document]
alvrs: Yep I guess we could call it `Context` too for consistency with OZ! Maybe `WorldContext` (akin to [ERC2771Context](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/metatx/ERC2771Context.sol))?

packages/world/src/modules/core/CoreModule.sol
alvrs: We could add a `Modules` table that stores which modules are installed in a world and a "dependency check" in the `install` method to revert if a required module is not found! One point of friction is it would require modules to have a unique name (at least unique per namespace), but maybe that's worth it for the benefit of better developer experience.

packages/world/src/modules/registration/RegistrationModule.sol
alvrs: good catch! This was used in some iteration step but isn't needed anymore

packages/cli/src/utils/deploy-v2.ts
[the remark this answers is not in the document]
alvrs: (will leave that to a follow PR to not bloat this one too much)

packages/world/src/modules/registration/RegistrationModule.sol
```diff
@@ -0,0 +1,102 @@
...
+  function getName() public pure returns (bytes16) {
+    return REGISTRATION_MODULE_NAME;
+  }
+
+  // The namespace argument is not used because the module is always installed in the root namespace
+  function install(bytes16) public {
+    // Require the CoreModule to be installed in the root namespace
+    if (InstalledModules.get(ROOT_NAMESPACE, CORE_MODULE_NAME).moduleAddress == address(0)) {
+      revert RequiredModuleNotFound(ResourceSelector.from(ROOT_NAMESPACE, CORE_MODULE_NAME).toString());
+    }
```
holic, line 46: we're so close to having a real package system hahaha

packages/cli/src/utils/deploy-v2.ts
```diff
@@ -60,12 +64,20 @@ export async function deploy(mudConfig: MUDConfig, deployConfig: DeployConfig):
       World: worldContractName
         ? deployContractByName(worldContractName)
         : deployContract(WorldABI, WorldBytecode, "World"),
+      CoreModule: deployContract(CoreModuleData.abi, CoreModuleData.bytecode, "CoreModule"),
+      RegistrationModule: deployContract(RegistrationModuleData.abi, RegistrationModuleData.bytecode, "RootModule"),
```
holic, line 68: is this intentionally called `RootModule`?

packages/world/mud.config.mts
```diff
@@ -76,6 +78,20 @@ const config: StoreUserConfig = {
...
+        namespace: SchemaType.BYTES16,
+        mdouleName: SchemaType.BYTES16,
+      },
+      schema: {
+        moduleAddress: SchemaType.ADDRESS,
+      },
+      // TODO: this is a workaround to use `getRecord` instead of `getField` in the autogen library,
+      // to allow using the table before it is registered. This is because `getRecord` passes the schema
+      // to store, while `getField` loads it from storage. Remove this once we have support for passing the
+      // schema in `getField` too.
```
holic, line 92: can we add this to the tracker?

holic (approved):

packages/cli/src/utils/deploy-v2.ts
holic: curious what this would have affected if left uncaught?

alvrs: It's just used for logging, the logs would have said "Deploying RootModule"

Issue #444, opened 2023-02-28
Tracker: v2 gas optimization todos


alvrs:
- [ ] try preallocate for all, encodePacked statics, and direct encode dynamics in 
 `render-table/record.ts` (see https://github.com/latticexyz/mud/pull/422#discussion_r1119458406)

dk1a:
moved 4 tightcoder-specific todos to https://github.com/latticexyz/mud/issues/1132

alvrs:
Moved a couple gas optimization related items to #1133

PR #487, opened 2023-03-14
feat(world): add naive ReverseMappingHook/Module

alvrs:
This PR adds a naive but functional implementation of a `ReverseMapping` on-chain indexer, similar to the `reverseMapping` of `solecs` `Components`. We should revisit the implementation once we added support for `setIndexOfField` (to modify individual values of arrays at a given index) to avoid writing the entire array of keys with a given value every time the value is updated. The other limitation for now is that `ReverseMapping` only supports single keys, no composite keys, just like `solecs`. We can add support for indexing composite keys by adding a version of  `ReverseMapping` with parallel arrays, but we should wait until the single key version is optimized.

Followup tasks / todos are added to #444.

I'm intentionally leaving out CLI support for registering these indexers for now to reduce the scope of this PR, will do that in a separate PR.


packages/store/src/IStore.sol
```diff
@@ -49,7 +49,10 @@ interface IStore {
 interface IStoreHook {
   function onSetRecord(uint256 table, bytes32[] memory key, bytes memory data) external;
 
-  function onSetField(uint256 table, bytes32[] memory key, uint8 schemaIndex, bytes memory data) external;
+  // Split onSetField into pre and post to simplify the implementation of hooks
```
alvrs, line 52: Without this hooks have to reimplement the StoreCore logic to replace a certain piece of bytes in the previous value with the new value. By splitting it up hooks can handle the previous value in `preSetField` and the new value in `postSetField` (see `ReverseMappingHook` as an example). This approach is less gas efficient than letting the hook compute the new value in memory, so we might change this during a gas optimization pass. Will add to #444.

packages/world/src/World.sol
```diff
@@ -43,45 +44,47 @@ contract World is Store, IWorldCore, IErrors {
...
+    // TODO: do we need this?
+    if (InstalledModules.get(module.getName(), keccak256(args)).moduleAddress != address(0)) {
+      revert ModuleAlreadyInstalled(ResourceSelector.from(module.getName()).toString());
     }
 
     Call.withSender({
       msgSender: msg.sender,
       target: address(module),
-      funcSelectorAndArgs: abi.encodeWithSelector(IModule.install.selector, namespace),
+      funcSelectorAndArgs: abi.encodeWithSelector(IModule.install.selector, args),
```
alvrs, line 57: Restricting the arguments passed to a module's install method to `namespace` was too restrictive. None of our current modules even need the `namespace` param, but some (like `IndexModule`) need some configuration to be passed to the `install` method.

packages/world/src/modules/index/ArrayLib.sol
alvrs, line 1: These are a couple utils we need for the current naive implementation of `ReverseMappingHook`. We should optimize the `ReverseMappingHook` once we have support for `setIndexOfField` and then won't need these utils anymore, so I kept them local to `modules/index` for now.

packages/store/src/IStore.sol
```diff
@@ -49,7 +49,10 @@ interface IStore {
...
+  function preSetField(uint256 table, bytes32[] memory key, uint8 schemaIndex, bytes memory data) external;
+
+  function postSetField(uint256 table, bytes32[] memory key, uint8 schemaIndex, bytes memory data) external;
```
holic, line 55: small nit: I kinda like `onBeforeSetField` and `onAfterSetField` for these

packages/store/src/StoreCore.sol
```diff
@@ -169,7 +169,7 @@ library StoreCore {
 
     // Call onSetRecord hooks (before actually modifying the state, so observers have access to the previous state if needed)
     address[] memory hooks = Hooks.get(bytes32(tableId));
-    for (uint256 i = 0; i < hooks.length; i++) {
+    for (uint256 i; i < hooks.length; i++) {
```
holic, line 172: is this a gas saver? if not, I kinda prefer the more explicit version

packages/store/src/StoreCore.sol
```diff
@@ -272,14 +278,20 @@ library StoreCore {
...
+      hook.preSetField(tableId, key, schemaIndex, fullData);
     }
 
     StoreCoreInternal._pushToDynamicField(tableId, key, schema, schemaIndex, dataToPush);
+
+    // Call postSetField hooks (after modifying the state)
+    for (uint256 i; i < hooks.length; i++) {
+      IStoreHook hook = IStoreHook(hooks[i]);
+      hook.postSetField(tableId, key, schemaIndex, fullData);
+    }
```
holic, line 294: not blocking, but thinking aloud: in terms of gas, I wonder if we could do anything to detect if a hook has implemented a given method and skip if not, or if adding a hook should be for a particular operation (e.g. `IStoreOnBeforeSetFieldHook`)

packages/world/mud.config.mts
```diff
@@ -80,8 +80,8 @@ const config: StoreUserConfig = {
     },
     InstalledModules: {
       primaryKeys: {
-        namespace: SchemaType.BYTES16,
-        mdouleName: SchemaType.BYTES16,
+        moduleName: SchemaType.BYTES16,
+        configHash: SchemaType.BYTES32, // Hash of the params passed to the `install` function
```
holic, line 84: nit: what about `argsHash` to align with  other variables? "config" seems a little overloaded here

packages/world/src/AccessControl.sol
```diff
@@ -44,7 +44,7 @@ library AccessControl {
   ) internal view returns (bytes32 resourceSelector) {
     resourceSelector = ResourceSelector.from(namespace, file);
 
-    if (NamespaceOwner.get(namespace) != msg.sender) {
+    if (NamespaceOwner.get(namespace) != caller) {
```
holic, line 47: this seems like a scary thing to have missed! do we have tests for this?

packages/world/src/World.sol
```diff
@@ -200,7 +188,21 @@ contract World is Store, IWorldCore, IErrors {
    */
   function registerSchema(uint256 tableId, Schema valueSchema, Schema keySchema) public virtual {
     bytes32 tableSelector = ResourceSelector.from(tableId);
-    _this.registerTable(tableSelector.getNamespace(), tableSelector.getFile(), valueSchema, keySchema);
+    (address systemAddress, ) = Systems.get(ResourceSelector.from(ROOT_NAMESPACE, REGISTRATION_SYSTEM_NAME));
+
+    // We can't call IWorld(this).registerSchema directly because it would lose the msg.sender
```
holic, line 193: ooc what do you mean "lose" the `msg.sender`?

packages/world/src/WorldContext.sol
```diff
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: MIT
 pragma solidity >=0.8.0;
 
+import { StoreSwitch } from "@latticexyz/store/src/StoreSwitch.sol";
+
 // Similar to https://eips.ethereum.org/EIPS/eip-2771, but any contract can be the trusted forwarder.
 // This should only be used for contracts without own storage, like Systems.
 contract WorldContext {
```
holic, line 8: should this be abstract?

packages/world/src/modules/index/IndexModule.sol
```diff
@@ -0,0 +1,63 @@
...
+ * from value to list of keys with this value. This mapping is stored in a table registered
+ * by the module at the `targetTableId` provided in the install methods arguments.
+ *
+ * Note: for now this module only supports tables with single keys, no composite keys.
+ * Support for composite keys can be added by using a parallel array to store the key in the target table.
+ *
+ * Note: this module currently expects to be `delegatecalled` via World.installRootModule.
+ * Support for installing it via `World.installModule` depends on `World.callFrom` being implemented.
+ */
+contract IndexModule is IModule, WorldContext {
```
holic, line 30: nit: `IndexModule` doesn't feel self explanatory to me, maybe `ReverseMappingModule`? or `IndexedDataModule`?

packages/world/src/modules/index/tables/ReverseMapping.sol
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  /* Delete all data for given keys */
+  function deleteRecord(uint256 _tableId, bytes32 valueHash) internal {
+    bytes32[] memory _primaryKeys = new bytes32[](1);
+    _primaryKeys[0] = bytes32((valueHash));
+
+    StoreSwitch.deleteRecord(_tableId, _primaryKeys);
+  }
+}
```
holic, line 111: broader note/thought, esp because we ran into world contract size limits, is letting you conditionally specify in the mud config -> autogen which methods you need so we can have only what we need for the core libs

even better would be an optimization step before deploy that 1) finds all the calls to these libs in the contracts being deployed and 2) removes any unused calls/dead code before compiling

packages/world/test/IndexModule.t.sol
```diff
@@ -0,0 +1,164 @@
...
+    // Assert that the list is correct
+    assertEq(keysWithValue.length, 0);
+
+    // Get the list of entities with value2 from the target table
+    keysWithValue = ReverseMapping.get(targetTableId, world, keccak256(abi.encode(value2)));
+
+    // Assert that the list is correct
+    assertEq(keysWithValue.length, 1);
+    assertEq(keysWithValue[0], key1);
+  }
```
holic, line 163: can we add some gas tests too?

holic (approved):
Raddd! Mostly nits around naming, but otherwise looks great!

packages/store/src/StoreCore.sol
alvrs: it saves a little bit of gas! I used to prefer `i = 0` too, but got used to not initializing zero values as an easy optimization. At this point we use the non-initialize pattern in most MUD contracts, so I changed it here for consistency.

packages/store/src/StoreCore.sol
alvrs: Agree it's kind of unnecessary to call the hook method if it's not implemented but I'm afraid checking if the hook implements a certain interface would be more gas expensive. We can make a note of this for the gas optimization pass though! Will add it to #444.

packages/world/src/AccessControl.sol
alvrs: Added some!

packages/world/src/World.sol
alvrs: `msg.sender` will be the address of the World contract, because calls to `IWorld(this)` are handled like external calls. We could use `delegatecall` to `this`, but that would defeat the purpose of having the `IWorld` interface, in which case using `Call.withSender` directly is cheaper. Will update the comment to make it more clear.

packages/world/src/WorldContext.sol
alvrs: Can you elaborate?

packages/world/src/modules/index/tables/ReverseMapping.sol
alvrs: Agreed! Sometimes we might still want the switch so the table can be used both from root systems/modules (= delegatecalled) and regular systems/modules (= called), but it's definitely useful to have a config option for those cases where it's clear how they'll be used.

packages/world/test/IndexModule.t.sol
alvrs: Yeah let's add some but don't look too close, the gas cost is gonna be bad. Will be useful for optimizing the implementation though.

packages/world/src/WorldContext.sol
holic: `abstract contract WorldContext` so it can't be instantiated on its own

alvrs: Oh i see, sure can't hurt

holic (dismissed):

holic (approved):

PR #488, opened 2023-03-14
feat(schema-type): add SchemaType -> primitive map, rearrange files

holic:
Pulled out of https://github.com/latticexyz/mud/pull/415 to reduce diff


alvrs (approved):
lgtm!

packages/schema-type/src/typescript/index.ts
```diff
@@ -0,0 +1,10 @@
+export type { StaticSchemaType } from "./StaticSchemaType.js";
+export type { DynamicSchemaType } from "./DynamicSchemaType.js";
+export type { ArraySchemaType } from "./ArraySchemaType.js";
+export type { SchemaTypeToPrimitive } from "./SchemaTypeToPrimitive";
```
alvrs, line 4: looks like this makes the CI fail:
```suggestion
export type { SchemaTypeToPrimitive } from "./SchemaTypeToPrimitive.js";
```

holic: yep, just fixed!

alvrs (approved):

PR #490, opened 2023-03-15
feat(world): ReverseMapping: infer target table id from source table id, add getKeysWithValue util

alvrs:
This PR removes the need to install a different ReverseMappingHook for each table the hook is registered by inferring the target table id from the source table id. This also removes the need for users to manually define the target table id in the mud config, and enables a `getKeysWithValue` function that doesn't require users to keep track of mapping between source table and target table (or require us to do a storage access to do the mapping).

### DIFF
--- a/packages/world/src/modules/reversemapping/constants.sol
+++ b/packages/world/src/modules/reversemapping/constants.sol
@@ -1,0 +1,6 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
+
+// Limiting the module namespace to 12 bytes so the last 4 bytes
+// can be used for a hash of the source table to avoid collisions
+bytes12 constant MODULE_NAMESPACE = "valuekeylist";
--- a/packages/world/src/modules/reversemapping/constants.sol
+++ b/packages/world/src/modules/reversemapping/constants.sol
@@ -4,4 +4,4 @@ pragma solidity >=0.8.0;
 // Limiting the module namespace to 12 bytes so the last 4 bytes
 // can be used for an identifier of the source table namespace to avoid
 // collisions between tables with the same name in different namespaces
-bytes8 constant MODULE_NAMESPACE = "vkeylist";
+bytes8 constant MODULE_NAMESPACE = "revmap";
--- a/packages/world/src/modules/reversemapping/constants.sol
+++ b/packages/world/src/modules/reversemapping/constants.sol
@@ -2,5 +2,6 @@
 pragma solidity >=0.8.0;
 
 // Limiting the module namespace to 12 bytes so the last 4 bytes
-// can be used for a hash of the source table to avoid collisions
+// can be used for an identifier of the source table namespace to avoid
+// collisions between tables with the same name in different namespaces
 bytes12 constant MODULE_NAMESPACE = "valuekeylist";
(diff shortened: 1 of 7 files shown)


packages/world/src/modules/reversemapping/getTargetTableSelector.sol
```diff
@@ -0,0 +1,19 @@
...
+ *  - The next 4 bytes are the first 4 bytes of the source table namespace
+ *    -- This is to avoid collisions between tables with the same name in different namespaces
+ *       (Note that collisions are still possible if the first 4 bytes of the namespace are the same, in which case installing the module fails)
+ *  - The last 16 bytes are the source table name
+ */
+function getTargetTableSelector(uint256 sourceTableId) pure returns (bytes32) {
+  bytes16 tableName = ResourceSelector.getFile(bytes32(sourceTableId));
+  bytes4 sourceTableNamespace = bytes4(bytes32(sourceTableId));
+  return bytes32(MODULE_NAMESPACE) | (bytes32(sourceTableNamespace) >> 96) | (bytes32(tableName) >> 128);
+}
```
holic, line 19: woah, a function that lives outside a contract/library? how does this get compiled? does this act like `internal`?

holic (dismissed):

packages/world/src/modules/reversemapping/getTargetTableSelector.sol
alvrs: yeah it's exactly like a library with an internal function (just no need to prefix with the lib name)

packages/world/src/modules/reversemapping/getTargetTableSelector.sol
```diff
@@ -0,0 +1,19 @@
...
+import { ResourceSelector } from "../../ResourceSelector.sol";
+import { MODULE_NAMESPACE } from "./constants.sol";
+
+/**
+ * Get a deterministic selector for the reverse mapping table for the given source table.
+ * The selector is constructed as follows:
+ *  - The first 12 bytes are the module namespace
+ *  - The next 4 bytes are the first 4 bytes of the source table namespace
+ *    -- This is to avoid collisions between tables with the same name in different namespaces
+ *       (Note that collisions are still possible if the first 4 bytes of the namespace are the same, in which case installing the module fails)
```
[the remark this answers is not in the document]
alvrs: yeah we could do that!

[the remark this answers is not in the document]
alvrs: changed it to 8 bytes / 8 bytes

packages/world/src/modules/reversemapping/constants.sol
```diff
@@ -0,0 +1,7 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
+
+// Limiting the module namespace to 12 bytes so the last 4 bytes
+// can be used for an identifier of the source table namespace to avoid
+// collisions between tables with the same name in different namespaces
+bytes8 constant MODULE_NAMESPACE = "vkeylist";
```
holic, line 7: why `vkeylist` instead of e.g. `revmap` to match the module name?

holic: makes sense! down for that rename, I agree that "reverse mapping" isn't super intuitive

holic (approved):

PR #492, opened 2023-03-15
feat: mode

authcall:
MVP of MODE service. For more technical overview refer to relevant MUD documentation + code comments


packages/services/pkg/mode/schema/constructors.go
```diff
@@ -0,0 +1,254 @@
...
+			"syncing": "boolean",
+		},
+		SolidityTypes: map[string]string{
+			"syncing": "bool",
+		},
+		Namespace: Namespace(chainId, ""),
+	}
+}
+
+func Internal__NamespacesTableSchema() *mode.TableSchema {
```
ludns, line 49: are we still usgin the namespace table?

packages/services/pkg/mode/storecore/encoding.go
```diff
@@ -0,0 +1,631 @@
...
+	case UINT16:
+		return fmt.Sprint(new(big.Int).SetBytes(encoding[bytesOffset : bytesOffset+2]).Uint64())
+	case UINT32:
+		return fmt.Sprint(new(big.Int).SetBytes(encoding[bytesOffset : bytesOffset+4]).Uint64())
+	case UINT64:
+		return fmt.Sprint(new(big.Int).SetBytes(encoding[bytesOffset : bytesOffset+8]).Uint64())
+	case UINT128:
+		return fmt.Sprint(new(big.Int).SetBytes(encoding[bytesOffset : bytesOffset+16]).Uint64())
+	case UINT256:
+		return fmt.Sprint(new(big.Int).SetBytes(encoding[bytesOffset : bytesOffset+32]).Uint64())
```
ludns, line 516: would the conversion to uint64 here fail if the number encoded is bigger than 2^64? (which can be the case with uint128 and uin256)

packages/services/pkg/mode/storecore/encoding.go
```diff
@@ -0,0 +1,631 @@
...
+//
+// Parameters:
+// - schemaType (SchemaType): The schema type of the field to decode.
+// - encoding ([]byte): The byte slice containing the encoded field.
+// - bytesOffset (uint64): The offset in bytes within the encoding byte slice where the field to be decoded starts.
+//
+// Returns:
+// - (string): The decoded field as a string.
+func DecodeStaticField(schemaType SchemaType, encoding []byte, bytesOffset uint64) string {
+	switch schemaType {
```
ludns, line 504: this is missing a lot of the types from the SchemaType enum (all signed integers, a bunch of uints, all arrays)

dk1a: also might be nice to extract the SchemaType stuff into `schema-type` package (not sure how much trouble that'd be with go)

alvrs:
Merging this to unblock #493 and template releases, fixes will be merged as a follow up

packages/services/pkg/mode/storecore/encoding.go
```diff
@@ -0,0 +1,631 @@
...
+//
+// Returns:
+// (string) - A string representing the PostgreSQL type for the specified SchemaType instance.
+func SchemaTypeToPostgresType(schemaType SchemaType) string {
+	if (schemaType >= UINT8 && schemaType <= UINT32) || (schemaType >= INT8 && schemaType <= INT32) {
+		// Integer.
+		return "integer"
+	} else if (schemaType >= UINT64 && schemaType <= UINT256) || (schemaType >= INT64 && schemaType <= INT256) {
+		// Big integer.
+		return "text"
```
holic, line 599: ooc why not bigint?

PR #494, opened 2023-03-16
feat(cli): add module config to CLI

alvrs:
- This PR adds support for the `modules` CLI config. It allows users to specify modules to be installed in the World during the deployment process. Modules can be either "default modules" (part of the World package, for now only `ReverseMappingModule`), or custom modules defined by users.
- In addition this PR adds the concept of "dynamically resolved config arguments", which are very useful for modules. Those can be things like "resolveTableId(<tableName>)" or "resolveSystemAddress(<systemName>)" and let users pass arguments to modules that will only be known during the deploy process.

- This MUD branch is integrated with the `alvrs/modules-cli` on `v2sandbox`. See https://github.com/latticexyz/v2sandbox/pull/6 for a usage example (including dynamically resolved argument). Note that v2sandbox also requires the follow up PR to this PR that renames `ReverseMappingModule` to `KeysWithValueModule` (#497), which I pulled out of this PR to reduce the scope of this PR.


packages/cli/src/utils/deploy-v2.ts
```diff
@@ -365,3 +419,12 @@ function toBytes16(input: string) {
...
 }
+
+function toResourceSelector(namespace: string, file: string): Uint8Array {
+  const namespaceBytes = toBytes16(namespace);
+  const fileBytes = toBytes16(file);
+  const result = new Uint8Array(32);
+  result.set(namespaceBytes);
+  result.set(fileBytes, 16);
+  return result;
+}
```
alvrs, line 430: We should use TableId from `@latticexyz/utils` here, but I wasn't able to import it, my guess is because utils doesn't use `.js` imports, so I'll leave that refactor for later

dk1a: yeah before cli ESM migration utils just couldn't be used by cli. Now with some refactoring they probably can

packages/cli/src/config/world/parseWorldConfig.ts
```diff
@@ -0,0 +1,52 @@
...
+      accessList: SystemAccessList,
+    }),
+  ])
+);
+
+const ValueWithType = z.object({
+  value: z.union([z.string(), z.number(), z.instanceof(Uint8Array)]),
+  type: z.string(),
+});
+const DynamicResulution = z.object({ type: z.nativeEnum(DynamicResolutionType), input: z.string() });
```
dk1a, line 29: typo here DynamicRes`u`lution

packages/cli/src/utils/deploy-v2.ts
holic: can we add a TODO + issue?

packages/store/src/StoreCore.sol
```diff
@@ -1,7 +1,6 @@
 // SPDX-License-Identifier: MIT
 pragma solidity >=0.8.0;
 
-import { console } from "forge-std/console.sol";
```
holic, line 4: whoops

packages/cli/src/utils/deploy-v2.ts
```diff
@@ -171,8 +199,34 @@ export async function deploy(mudConfig: MUDConfig, deployConfig: DeployConfig):
...
   }
 
-  // Await all promises
-  await Promise.all(promises);
+  // Wait for access to be granted before installing modules
+  await Promise.all(promises); // ----------------------------------------------------------------------------------------------
+
+  // Install modules
+  promises.push(
+    ...mudConfig.modules.map(async (module) => {
```
holic, line 207: haha, this was an interview question for me at Stripe! this can actually fail in JS because the list of arguments can get too long

should use `splice` instead

packages/cli/src/utils/deploy-v2.ts
alvrs: #499

dk1a (approved):
lgtm

dk1a (approved):

PR #495, opened 2023-03-16
refactor(network): rework entity/component ID normalizing

Kooshaba:


packages/network/src/utils.ts
```diff
@@ -8,15 +9,27 @@ export function messagePayload(msg: Message) {
...
 }
 
-// Remove zero padding from all entity ids
-export function formatEntityID(entityID: string | EntityID | BigNumber): EntityID {
-  if (BigNumber.isBigNumber(entityID) || entityID.substring(0, 2) === "0x") {
+// For v1 entities (BigNumber or hex strings) and single-key v2 entities (hex strings), strip zero padding
+// For composite-key v2 entities (hex strings concatenated with `:`), leave the ID as is
+// See `keyTupleToEntityID`
+export function normalizeEntityID(entityID: string | EntityID | BigNumber): EntityID {
+  if (BigNumber.isBigNumber(entityID)) {
```
alvrs, line 16: does this work if the entity id is concatenated with `:` or does ethers still think it's a bignumber but then error on `toHexString()`? (Asking bc I thought we had an issue with that yesterday)

holic: the issue came not from ethers, but our bad logic that assumed "string starts with 0x can be converted to BigNumber"

alvrs (approved):

packages/network/src/utils.ts
holic: here's what `BigNumber.isBigNumber` looks like: image

PR #496, opened 2023-03-16
feat(cli): add worldgen

dk1a:
A nice thing about solidity-parser is I can skip `build`, so neither forge nor hardhat are needed for worldgen


packages/cli/package.json
```diff
@@ -44,6 +44,7 @@
     "@latticexyz/services": "^1.41.0",
     "@latticexyz/solecs": "^1.41.0",
     "@latticexyz/std-contracts": "^1.41.0",
+    "@solidity-parser/parser": "^0.16.0",
```
holic, line 47: 🤯 Can't believe we're at this level of meta programming

alvrs (approved):
amazing!

packages/cli/src/render-solidity/worldgen.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    // prepare imports for IWorld
+    systemInterfaceImports.push({
+      symbol: systemInterfaceName,
+      path: `./${systemInterfaceName}.sol`,
+    });
+  }
+
+  // render IWorld
+  const worldInterfaceName = "IWorld";
```
alvrs, line 47: Should we make this configurable? Or make it `I${worldContractName}` and fall back to `IWorld` if `worldContractName` is undefined?

dk1a: I think that would be kind of a lie then, since the custom world might have its own functions, which this interface won't have. Imo this should be done together with support for extra interfaces as its own PR

alvrs: sounds good!

PR #498, opened 2023-03-16
feat(cli): add encode function to all tables

dk1a:


holic:
This was requested by @alvrs on Discord, just copying here for context:
> could we add a `encode` function to tablegen? We need that for `getKeysWithValue(tableId, value)` where `value` needs to be the bytes encoded version of the data, but since we do custom encoding `abi.encodePacked` would only work for static values
> so basically every table library needs an `encode` function that takes the typed arguments of a full record and returns the packed bytes

holic (approved):

alvrs (approved):
lgtm!

PR #500, opened 2023-03-16
feat(cli): improve storeArgument, refactor cli

dk1a:
A bunch of code improvements for solidity renderers and store config
And some fixes:

- storeArgument is now always 1st and present for all relevant functions
- singletons work
- added pure to encode

this is extracted from #459


packages/cli/src/config/parseStoreConfig.ts
```diff
@@ -39,16 +84,26 @@ const TableDataFull = z
...
+/************************************************************************
+ *
+ *    TABLES
+ *
+ ************************************************************************/
+
+export type TablesConfig = Record<string, TableConfig | z.input<typeof FieldData>>;
+
+export const TablesConfig = z.record(TableName, TableConfig).transform((tables) => {
+  // default route depends on tableName
```
alvrs, line 106: gonna rename `fileSelector` to `name` or sth soon, but for now let's use `fileSelector`
```suggestion
  // default fileSelector depends on tableName
```

packages/store/src/StoreCore.sol
```diff
@@ -95,8 +95,8 @@ library StoreCore {
    */
   function registerSchema(uint256 tableId, Schema valueSchema, Schema keySchema) internal {
     // Verify the schema is valid
-    valueSchema.validate();
-    keySchema.validate();
+    valueSchema.validate(false);
+    keySchema.validate(true);
```
alvrs, line 99: I'm not sure if the following compiles, but it would be more readable in this case. Should test if it compiles before committing though
```suggestion
    valueSchema.validate({ allowEmpty: false });
    keySchema.validate({ allowEmpty: true });
```

packages/world/mud.config.mts
```diff
@@ -65,15 +65,15 @@ const config = {
...
-    // TODO: This table is only used for testing, move it to `test/tables` via the directory config once supported
-    //   primaryKeys: {},
-    //   schema: {
-    //     value: SchemaType.BOOL,
-    //   },
-    //   storeArgument: true, // TODO Add support for store argument in setter function to table autogen
-    //   tableIdArgument: true,
-    // },
+    Bool: {
+      // TODO: This table is only used for testing, move it to `test/tables` via the directory config once supported
```
alvrs, line 69: this should be supported now right?
```suggestion
      directory: "test/tables",
```

packages/world/mud.config.mts
```diff
@@ -65,15 +65,15 @@ const config = {
...
+      primaryKeys: {},
+      schema: {
+        value: SchemaType.BOOL,
+      },
+      storeArgument: true,
+      tableIdArgument: true,
+    },
     AddressArray: {
       // TODO: This table is only used for testing, move it to `test/tables` via the directory config once supported
```
alvrs, line 78: same here
```suggestion
      directory: "../test/tables",
```

packages/world/mud.config.mts
dk1a: there're some problems with just doing this, so I left it for another PR, this one is big enough as it is

dk1a: I think forge doesn't see imports in `test`, unless we add `test` to `libs`, which is a questionable decision and I haven't looked into what to really do about it yet

alvrs: I see, fine to leave it for later!

alvrs (approved):
lgtm!

packages/world/mud.config.mts
dk1a: (like me, you might wonder how it works in `solecs` - because there `test` is in `src/test`)

packages/store/src/StoreCore.sol
dk1a: haha you merged it right  before I pushed the changes. It does compile

dk1a: (wanted to do as a manual commit since it's used in 2 other places)

packages/cli/src/config/parseStoreConfig.ts
```diff
@@ -87,45 +154,33 @@ export interface StoreUserConfig {
...
+  tables: TablesConfig,
+  userTypes: UserTypesConfig,
+});
 
-export type StoreConfig = z.output<typeof StoreConfig>;
+// finally validate global conditions
+export const StoreConfig = StoreConfigUnrefined.superRefine(validateStoreConfig);
 
-export function parseStoreConfig(config: StoreUserConfig) {
+export function parseStoreConfig(config: unknown) {
```
holic, line 174: Did this have to get reverted?

dk1a: this was an accidental revert actually - was adapting these changes was from code before your PR and missed this. But good to know all works well

PR #501, opened 2023-03-16
feat(cli): add registerFunctionSelectors to deploy cli

alvrs:
- Adds a `registerFunctionSelectors` flag to the world's system config (defaults to true), which registers (root) function selectors in the World for all system the system's function (including fallback function). In combination with #496 this allows for systems to be called via the combined World interface


packages/cli/src/utils/deploy-v2.ts
```diff
@@ -152,19 +152,54 @@ export async function deploy(mudConfig: MUDConfig, deployConfig: DeployConfig):
...
+              ? functionName + functionArgs
+              : `${namespace}_${fileSelector}_${functionName}${functionArgs}`;
+
+            console.log(chalk.blue(`Registering function "${functionSignature}"`));
+            if (isRoot) {
+              const worldFunctionSelector = toFunctionSelector(
+                functionSignature === ""
+                  ? { functionName: systemName, functionArgs } // Register the system's fallback function as `<systemName>(<args>)`
+                  : { functionName, functionArgs }
+              );
```
dk1a, line 182: note that `worldgen` currently ignores `fallback`, this should be added to TODOs for feature parity

dk1a (approved):
lgtm, nice catch with `types.sol`

packages/cli/src/utils/deploy-v2.ts
```diff
@@ -152,19 +152,54 @@ export async function deploy(mudConfig: MUDConfig, deployConfig: DeployConfig):
...
+          await contractPromises[systemName],
+          openAccess,
+        ]);
+        console.log(chalk.green(`Registered system ${systemName} at ${namespace}/${fileSelector}`));
+
+        // Register function selectors for the system
+        if (registerFunctionSelectors) {
+          const functionSignatures: FunctionSignature[] = await loadFunctionSignatures(systemName);
+          const isRoot = namespace === "";
+          for (const { functionName, functionArgs } of functionSignatures) {
```
holic, line 171: worth using a `Promise.all(functionSignatures.map(...))` here to parallelize?

holic (approved):

dk1a (approved):
lgtm, took me a sec to realize why a map that returns nothing works correctly

packages/cli/src/config/world/resolveWorldConfig.ts
```diff
@@ -50,12 +50,14 @@ export function resolveWorldConfig(config: ParsedWorldConfig, existingContracts?
...
  * @returns ResolvedSystemConfig object
  * Default value for fileSelector is `systemName`
+ * Default value for registerFunctionSelectors is true
  * Default value for openAccess is true
  * Default value for accessListAddresses is []
  * Default value for accessListSystems is []
  */
 export function resolveSystemConfig(systemName: string, config?: SystemUserConfig, existingContracts?: string[]) {
   const fileSelector = config?.fileSelector ?? systemName;
+  const registerFunctionSelectors = config?.registerFunctionSelectors ?? true;
```
holic, line 60: wondering about how best to include these in the system call stream:
- figure out the function selectors here and include in the return object?
- wait for the resulting world to compile then use that ABI to determine if its a registered system call?

alvrs: `worldgen` creates a world abi with the correct function selectors (without compiling because it parses solidity directly) - could we use that for they system call stream?

dk1a (approved):

holic (approved):

PR #502, opened 2023-03-16
fix(cli): use fileSelector in worldgen

dk1a:


packages/cli/src/render-solidity/worldgen.ts
```diff
@@ -26,9 +26,10 @@ export async function worldgen(
     }));
     const systemInterfaceName = `I${system.basename}`;
     // create an interface using the external functions and imports
+    const { fileSelector } = config.systems[system.basename];
     const output = renderSystemInterface({
       name: systemInterfaceName,
-      functionPrefix: config.namespace === "" ? "" : `${config.namespace}_${system.basename}_`,
+      functionPrefix: config.namespace === "" ? "" : `${config.namespace}_${fileSelector}_`,
```
holic, line 32: should this/does this default to `system.basename`?

dk1a: it does default to basename

alvrs (approved):

PR #493, opened 2023-03-15
feat(network): integrate initial sync from MODE

alvrs:
Integrates initial sync from MODE into the client libraries

Corresponding v2sandbox PR: https://github.com/latticexyz/v2sandbox/pull/22

PR #504, opened 2023-03-17
refactor: fix v2 decoding, refactor other things for clarity

holic:
Pulled out of https://github.com/latticexyz/mud/pull/493 to keep diff small


packages/network/src/v2/schemas/decodeStaticField.spec.ts
```diff
@@ -0,0 +1,63 @@
+import { SchemaType } from "@latticexyz/schema-type";
+import { hexToArray } from "@latticexyz/utils";
+import { decodeStaticField } from "./decodeStaticField";
+
+describe("decodeStaticField", () => {
```
holic, line 5: This isn't a full suite but I figured a few tests were better than nothing and helped me to get the decoding fixed

packages/network/src/v2/schemas/decodeValue.ts
```diff
@@ -0,0 +1,110 @@
...
+  switch (schemaType) {
+    case SchemaType.BOOL:
+    case SchemaType.UINT8:
+    case SchemaType.UINT16:
+    case SchemaType.UINT24:
+    case SchemaType.UINT32:
+    case SchemaType.UINT40:
+    case SchemaType.UINT48:
+    case SchemaType.UINT56:
+    case SchemaType.UINT64:
```
alvrs, line 16: For context - we need this instead of something like `getStaticByteLength(schemaType)` for proper types correct?

packages/utils/src/v2/TableId.ts
```diff
@@ -26,6 +27,10 @@ export class TableId {
     return arrayToHex(tableId);
   }
 
+  static fromHexString(tableId: string) {
+    return TableId.fromBytes32(hexToArray(tableId));
+  }
+
   static fromBytes32(tableId: Uint8Array) {
     // assumes tableId is a 32-byte hex string, otherwise it left-pads with zeros (for numbers)
     // this is scary, since zero padding is different depending on the type (bytes types vs number types)
```
alvrs, line 36: A comment about line 39 (can't comment on there bc it wasn't changed in this PR): the `tableId.reverse()` reverses the array in place, this tripped me up today when using the `fromBytes32` method as a util to print a table id I was debugging. Could we first copy it to a fresh array and do the `reverse` there to avoid that?

alvrs (approved):
lgtm! Just a quick question for context and one comment unrelated to this PR

packages/network/src/v2/schemas/decodeValue.ts
holic: yes, though this and the `decodeStaticField` method could probably be simplified with some `const` arrays instead of long case statements, but planning to come back to that as a "tidy" thing

holic: but also the types still don't exactly work due to some issues with mapping to primitives, this method still returns something like `string | number | bigint | boolean` for any `schemaType`

packages/utils/src/v2/TableId.ts
holic: oh yes, definitely! I also learned today that TypedArrays are mutable and Uint8Array, etc. can be "views" into larger arrays under the hood 🙈 which caused ~half of this refactor

alvrs (approved):

PR #505, opened 2023-03-17
wip system call streams

holic:


packages/std-client/src/setup/createSystemCallStreams.ts
```diff
@@ -0,0 +1,52 @@
+import { SystemCall } from "@latticexyz/network";
+import { Components, World, Type, EntityID, getComponentValue, Component } from "@latticexyz/recs";
+import { Contract, BigNumber } from "ethers";
+import toLower from "lodash/toLower";
+import compact from "lodash/compact";
+import { Subject } from "rxjs";
+import { DecodedSystemCall } from "./types";
+import { createDecodeNetworkComponentUpdate } from "./utils";
+
+export function createSystemCallStreams<C extends Components, SystemTypes extends { [key: string]: Contract }>(
```
holic, line 10: this is the same logic as before, just pulled out into its own file so I could see it better

packages/cli/src/commands/deploy-v2.ts
```diff
@@ -1,6 +1,6 @@
 import chalk from "chalk";
 import glob from "glob";
-import path, { basename } from "path";
+import path, { basename } from "node:path";
```
holic, line 3: this is me attempting to fix build issues

I believe vite, etc. bundle in their own polyfill so seems fine to use the node lib directly rather than whatever polyfill we were manually adding as a dep cc @dk1a

holic:
@alvrs looking at this again, it might be better for you to start a fresh branch since I was fiddling a lot with build issues here that may or may not be an issue anymore

packages/std-client/src/setup/v2/createV2SystemCallStreams.ts
```diff
@@ -0,0 +1,67 @@
...
+import { Subject } from "rxjs";
+
+type SystemsToCallStreams<Config extends MUDUserConfig> = Config["overrideSystems"] extends undefined
+  ? undefined
+  : {
+      // TODO: pass through total components, not just store components from config
+      [K in keyof Config["overrideSystems"]]: Subject<SystemCall<StoreUserConfigToComponents<Config>>>;
+    };
+
+export function createV2SystemCallStreams<Config extends MUDUserConfig>(
```
holic, line 16: this is the new logic

PR #506, opened 2023-03-17
feat(cli): add prototypes to tablegen

dk1a:
Simplified replacement for #459
(this is just prototypes now, without all the unrelated refactoring/fixes, which were done in #500)


alvrs:
Closing this for now since it's out of date. Once modules/plugins are in a stable state we should add a plugin for this (@yonadaaa was working on something similar in #749)

PR #507, opened 2023-03-17
feat(cli): use abi types in store config

dk1a:
Config now uses strings for both abi and user types
I also added a function to define configs, because otherwise user types can't really be typehinted
When using the function any missing user types will show a typescript error
(to simplify the logic for this I also removed `userTypes`, so `enums` are in config root now)


packages/cli/src/config/index.ts
```diff
@@ -1,9 +1,18 @@
...
 import { StoreUserConfig, StoreConfig } from "./parseStoreConfig.js";
 import { WorldUserConfig, ResolvedWorldConfig } from "./world/index.js";
 
-export type MUDUserConfig = StoreUserConfig & WorldUserConfig;
+export type MUDUserConfig<EnumNames extends StringForUnion = StringForUnion> = StoreUserConfig<EnumNames> &
+  WorldUserConfig;
 export type MUDConfig = StoreConfig & ResolvedWorldConfig;
 
+/** Type helper for defining MUDUserConfig */
+export function defineMUDUserConfig<EnumNames extends StringForUnion = StringForUnion>(
```
holic, line 10: nit: would be great if we could just call this `mudConfig({...})`

packages/cli/src/render-solidity/renderTypesFromConfig.ts
```diff
@@ -2,9 +2,9 @@ import { StoreConfig } from "../config/parseStoreConfig.js";
 import { renderTypes } from "./renderTypes.js";
 
 export function renderTypesFromConfig(config: StoreConfig) {
-  const enums = Object.keys(config.userTypes.enums).map((name) => ({
+  const enums = Object.keys(config.enums).map((name) => ({
     name,
-    memberNames: config.userTypes.enums[name],
+    memberNames: config.enums[name],
```
holic, line 7: 😍

packages/cli/src/utils/typeUtils.ts
```diff
@@ -0,0 +1,5 @@
+export type RequireKeys<T extends Record<string, unknown>, P extends string> = T & Required<Pick<T, P>>;
+
+// This allows unions between string literals and `string` without sacrificing autocompletion.
+// Workaround for https://github.com/Microsoft/TypeScript/issues/29729
+export type StringForUnion = string & Record<never, never>;
```
holic, line 5: nice!

packages/store/mud.config.mts
```diff
@@ -1,37 +1,34 @@
...
+        a32: "uint32[]",
+        s: "string",
       },
     },
     Vector2: {
       schema: {
-        x: SchemaType.UINT32,
-        y: SchemaType.UINT32,
+        x: "uint32",
+        y: "uint32",
```
holic, line 30: ohhh this is sooo much better!

holic (commented):
just one small nit about simplifying the name the config function as we gear up towards release!

holic (approved):

packages/cli/src/config/parseStoreConfig.ts
```diff
@@ -151,22 +169,29 @@ export interface StoreUserConfig {
    * The key is the table name (capitalized).
    *
    * The value:
-   *  - `SchemaType | userType` for a single-value table (aka ECS component).
+   *  - abi or user type for a single-value table (aka ECS component).
```
alvrs, line 172: not part of this PR, but I think we should remove `aka ECS component`. Having a single value is not the distinguishing factor for an ECS component, since components can have multiple (related) values (eg. `PositionComponent = {x: int32, y: int32 }`). The distinguishing factor for ECS components vs regular tables would rather be having a single key vs a composite key.

packages/recs/src/v2/abiTypesToRecsTypes.ts
```diff
@@ -0,0 +1,203 @@
...
+  uint16: RecsType.Number,
+  uint24: RecsType.Number,
+  uint32: RecsType.Number,
+  uint40: RecsType.Number,
+  uint48: RecsType.Number,
+  uint56: RecsType.Number,
+  uint64: RecsType.Number,
+  uint72: RecsType.Number,
+  uint80: RecsType.Number,
+  uint88: RecsType.Number,
```
alvrs, line 15: Is anything above `uint53` really represented as `number` in `recs`? Since those don't fit into javascript `number` and `recs` in its current state doesn't support `bigint`, I would expect them to be represented as a hex string (= `RecsType.String`)

packages/cli/src/config/index.ts
alvrs: For consistency we should probably capitalize `MUD` here.

packages/recs/src/v2/abiTypesToRecsTypes.ts
dk1a: Defer to @holic (this isn't really a part of this PR, as I just refactored the previous mapping of schema types)

packages/cli/src/config/index.ts
dk1a: should `storeConfig` be `StoreConfig` then to consistently capitalize them too?

packages/recs/src/v2/abiTypesToRecsTypes.ts
holic: does RECS consistently give us back a native number? iirc some things use `BigNumber` under the hood

packages/cli/src/config/index.ts
dk1a: doing a little refactoring PR rn, capitalized func names don't feel good. And with MUDConfig I also get conflicts with types and zod stuff

dk1a: `defineMUDConfig` would be a better alternative, but I'd still prefer `mudConfig` as it's shorter

PR #508, opened 2023-03-18
feat(cli): improve store config typehints, prepare for static array support

dk1a:
This mostly just improves typehints, in particular the specific field/key with an invalid type is now highlighted, instead of a vague ts error about enums not having correct keys.
The improvements also allow to easily plug in support for static arrays, so that they just work without defining them as user types (this PR does not actually add support for static arrays)

### DIFF
--- a/packages/cli/src/config/parseStoreConfig.ts
+++ b/packages/cli/src/config/parseStoreConfig.ts
@@ -1,8 +1,8 @@
 import { AbiType, AbiTypes, StaticAbiType, StaticAbiTypes } from "@latticexyz/schema-type";
 import { RefinementCtx, z, ZodIssueCode } from "zod";
-import { RequireKeys, StringForUnion } from "../utils/typeUtils.js";
+import { ExtractUserTypes, RequireKeys, StringForUnion } from "../utils/typeUtils.js";
 import { ObjectName, Selector, UserEnum, ValueName } from "./commonSchemas.js";
-import { getDuplicates } from "./validation.js";
+import { getDuplicates, parseStaticArray } from "./validation.js";
 
 const TableName = ObjectName;
 const KeyName = ValueName;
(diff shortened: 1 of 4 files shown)


packages/cli/src/config/parseStoreConfig.ts
```diff
@@ -131,18 +134,29 @@ export const zTablesConfig = z.record(TableName, zTableConfig).transform((tables
  *
  ************************************************************************/
 
-export type EnumsConfig<EnumNames extends StringForUnion = StringForUnion> = string extends EnumNames
+export type EnumsConfig<EnumNames extends StringForUnion> = never extends EnumNames
```
holic, line 137: how does this condition work?

packages/cli/src/config/parseStoreConfig.ts
```diff
@@ -172,13 +189,17 @@ export type StoreUserConfig<EnumNames extends StringForUnion = StringForUnion> =
...
+  tables: TablesConfig<StaticUserTypes, StaticUserTypes>;
   /** Path to the file where common user types will be generated and imported from. Default is "Types" */
   userTypesPath?: string;
 };
 
 /** Type helper for defining StoreUserConfig */
-export function storeConfig<EnumNames extends StringForUnion = StringForUnion>(config: StoreUserConfig<EnumNames>) {
+export function storeConfig<
+  // (`never` is overridden by inference, so only the defined enums can be used by default)
+  EnumNames extends StringForUnion = never,
```
holic, line 200: oh I see here!

packages/cli/src/config/validation.ts
```diff
@@ -151,3 +151,13 @@ export function validateSelector(name: string, ctx: RefinementCtx) {
     });
   }
 }
+
+/** Returns null if the type does not look like a static array, otherwise element and length data */
+export function parseStaticArray(abiType: string) {
+  const matches = abiType.match(/^(\w+)\[(\d+)\]$/);
+  if (matches === null) return null;
```
holic, line 158: slight preference for `== null` here to catch `undefined` (I know the docs say this should return `null` but just wanna be safe)

packages/cli/src/config/parseStoreConfig.ts
```diff
@@ -172,13 +189,17 @@ export type StoreUserConfig<EnumNames extends StringForUnion = StringForUnion> =
    *  - abi or user type for a single-value table (aka ECS component).
    *  - FullTableConfig object for multi-value tables (or for customizable options).
    */
-  tables: TablesConfig<EnumNames>;
+  tables: TablesConfig<StaticUserTypes, StaticUserTypes>;
```
holic, line 192: is this change intentional?

holic (commented):
just one tiny thing and a question but otherwise looks great!

packages/cli/src/config/validation.ts
dk1a: changed it to just `!matches`, might as well catch any falsey value then

packages/cli/src/config/parseStoreConfig.ts
dk1a: yes, for now it doesn't do much of anything, but the idea is to separate all types for fields, and static types for primary keys; since I'll be adding static user types probably. I imagine I'll be changing these generics again after figuring out how exactly I want to use them

holic (approved):

PR #509, opened 2023-03-20
feat(cli): allow static arrays as abi types in store config and tablegen

dk1a:
Now that schema types are strings of abi types, I can make e.g. `uint32[2]` just work, without user types
Also added some more tests

An annoying regression here is a bunch of `function mutability can be restricted to pure` warnings. `Memory.copy` uses identity precompile, which forces `view`. And detecting which vars use such memory copying is a lot of needless complexity I'd rather avoid.
This can be solved by changing `Memory.copy` (would likely cost more gas), or generalizing `TightCoder` to be usable directly for static arrays (would sidestep the need for copying altogether).
Regardless, for `Memory.copy` it would be best to use the native MCOPY if it ever happens: https://eips.ethereum.org/EIPS/eip-5656
related: https://github.com/ethereum/solidity/issues/12127

Also fixed a type inference bug which would happen in some complex configs with multiple user types


packages/cli/src/utils/typeUtils.ts
```diff
@@ -9,3 +9,9 @@ export type StringForUnion = string & Record<never, never>;
 export type StaticArray = `${StaticAbiType}[${number}]`;
 // static arrays and inferred enum names get mixed together - this helper separates them
 export type ExtractUserTypes<UnknownTypes extends StringForUnion> = Exclude<UnknownTypes, AbiType | StaticArray>;
+
+// When type inference sees multiple uses of 1 generic, it can only guess
+// which of those are supposed to define the generic (and it will be wrong in complex situations).
+// This helper explicitly makes a type that's dependent on some generic,
+// and will not be inferred as the generic's definition.
+export type AsDependent<T> = T extends infer P ? P : never;
```
alvrs, line 17: What's an example for this?

alvrs (approved):
Very cool! Nothing blocking, just two questions.

packages/cli/src/utils/typeUtils.ts
dk1a: check out [parseStoreConfig.test-d.ts](https://github.com/latticexyz/mud/pull/509/files#diff-c9b49da8abbb495550842f78d3966ae8959e6316d04eeaaf62cbc3ae026bab54)

alvrs (approved):

PR #511, opened 2023-03-26
[WIP] feat(monorepo) :construction_worker: migrate from yarn to pnpm

roninjin10:
TODO:
- [x] replace `workspaceRun.mjs` usage with `--no-bail`
- [ ] replace `workspace run` with recursive/parallel options
- [x] figure out package prepare ordering
- [x] change deps to use `workspace:*`


.github/actions/setup/action.yml
```diff
@@ -0,0 +1,22 @@
+name: Setup
```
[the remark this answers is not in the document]
holic: ohh I see, you just abstracted this out, this isn't an automatic thing that CI runs as part of the workflow

.npmrc
```diff
@@ -0,0 +1 @@
+shamefully-hoist=true
```
holic, line 1: what does this do?

README.md
```diff
@@ -91,16 +91,22 @@ The following steps are only necessary if you want to contribute to MUD. To use
...
 ```
 git clone https://github.com/latticexyz/mud
 ```
 
-4. Install MUD dependencies and setup local environment
+5. Install MUD dependencies and setup local environment
 
 ```
-cd mud && yarn
+cd mud && pnpm i
```
holic, line 109: slight preference for the more verbose `pnpm install` so its easier to understand/intuit

package.json
```diff
@@ -58,25 +61,25 @@
...
+    "link:packages": "lerna run link",
+    "docs": "pnpm docs:prepare && retype build",
+    "docs:prepare": "lerna run docs && pnpm retype:updateversion",
+    "docs:dev": "retype watch .",
+    "retype:updateversion": "sed \"s/label: .*/label: $(pnpm list --filter @latticexyz/solecs | grep -e @latticexyz | sed \"s/.*@//\")/\" retype.yml > retype.yml.tmp && mv retype.yml.tmp retype.yml",
+    "test": "pnpm recursive run test",
+    "entry:dist": "lerna run prepack",
+    "entry:src": "lerna run postpack",
+    "gas-report": "workspace @latticexyz/store run gas-report && pnpm recursive run --filter @latticexyz/world run gas-report",
+    "tablegen": "pnpm recursive run --filter @latticexyz/store run tablegen && pnpm recursive run --filter @latticexyz/world run tablegen"
```
holic, line 82: since we're on pnpm now, I think this can just be `pnpm --parallel gas-report` and `pnpm --parallel tablegen` (and pnpm is smart enough to skip packages where the script is not defined)

.npmrc
roninjin10: When yarn/npm installs they by default install to `node_modules`.   If one package has a different version of the same package, let's say `react@17` vs `react@18`.   Then the package will be at `packages/my_workspace_package/node_modules` instead of at the root level node_modules.    By default npm and yarn hoist.

Hoisting is bad and causes issues.   But npm and yarn installs would be way slower if they didn't hoist.

Pnpm doesn't have this same problem.  PNPM only installs to `node_modules/.pnmm-store` (might be different folder name forgot actual name).   It then symlinks in rest of repo.   SO it can do the better behavior of never hoisting.

Me adding shamefully-hoist to start is me keeping the behavior as close as possible to what yarn was doing before.    Makes it less likely I need to make changes or something will break.   We should try to remove it though... it is shameful

package.json
```diff
@@ -2,6 +2,9 @@
   "name": "mud",
   "version": "0.0.0",
   "description": "Mud is the fabric of autonomous worlds",
+  "engines": {
+    "pnpm": ">=7"
```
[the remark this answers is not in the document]
roninjin10: nested nvms in the repo are node 17.   Node 18 is current.   Node 18 is likely the best choice assuming it works as node 16 eol is coming soon

package.json
roninjin10: Ah I see yea yarn workspaces are annoying about that in yarn classic

.github/actions/setup/action.yml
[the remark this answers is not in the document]
roninjin10: Oh just noticed forge is already here but you know what I mean

.github/actions/setup/action.yml
```diff
@@ -0,0 +1,22 @@
...
+        cache: pnpm
+
+    - name: Install Foundry
+      uses: foundry-rs/foundry-toolchain@v1
+      with:
+        version: nightly
+
+    - name: Install node modules
+      run: pnpm install --frozen-lockfile
+      shell: bash
```
roninjin10, line 22: We could further cache this by reading from cache if package.jsons and pnpm-lock haven't changed.  I can add that in it's own pr

.npmrc
roninjin10: I can't remove it though until I verify everything is working with this here.   It removes a variable of what could go wrong making anything easier to debug so it has to be last

package.json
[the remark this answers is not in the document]
holic: down to change to node 18, we should just get everyone on the same page with their local env

holic:
I seem to recall issues with e.g. npm for certain dependencies using git URLs (where yarn/pnpm handle them fine), wondering if we can fix that here too for better npm support.

.github/actions/setup/action.yml
holic: is `shell: bash` needed here? what happens without it?

holic:
been fighting branches with this coming from your fork, so I made this a proper branch in this repo and opened a new PR here: https://github.com/latticexyz/mud/pull/591

PR #514, opened 2023-03-27
feat: 👷 Use vite instead of rollup

roninjin10:
Rather than opening an issue I figured I'd make a pr and see what everyone thought about this.

# Problems

- Packages are a mix of esm and cjs both in package module type and in build output type
- Jest struggles with modern tooling like es modules.   Esm is only supported experimentally
- Jest wasn't working in this react package.   We had a hack where we compiled the tests to js first
- Jest is generally slow
- Jest requires it's own seperate build process.  You configure your normal rollup build, but then you have to configure a completely seperate version of doing similar thing in jest.  This is hard to mantain
- USers tools cannot properly code split react output since this package happened to only output cjs
- Lack of a consistent pattern for monorepo/package management is the cause of some other issues I would suspect

# Solution

## Build outputs

- I changed this react package to output both cjs and esm as two different build outputs
- I included cjs build output as `main` in package.json and esm output as `module`.   This will make it work seemlessly no matter what build tool user is using
- I switched to vitest which is pretty close to the same api as jest but faster and supports modern tooling by default
- I also switched from rollup to vite.   Vite uses rollup under the hood and is a superset of rollup
- This allows the build and tests to share the same config


packages/react/package.json
```diff
@@ -27,18 +30,22 @@
...
+    "@latticexyz/schema-type": "^1.41.0",
+    "@latticexyz/utils": "^1.41.0",
     "@rollup/plugin-json": "^4.1.0",
-    "@rollup/plugin-node-resolve": "^13.1.3",
+    "@rollup/plugin-node-resolve": "^15.0.1",
     "@rollup/plugin-typescript": "^11.0.0",
     "@testing-library/react-hooks": "^8.0.1",
     "@types/react": "^18.0.12",
+    "@types/rollup-plugin-peer-deps-external": "^2.2.1",
+    "cli": "link:@lattice/xyz/cli",
```
roninjin10, line 46: This might be an accident including this
```suggestion
```

packages/react/src/useComponentValue.test.ts
```diff
@@ -1,10 +1,9 @@
 import { renderHook, act } from "@testing-library/react-hooks";
+import type { World, Component } from "@latticexyz/recs";
```
roninjin10, line 2: Autofixer did all these changes

packages/react/vite.config.ts
```diff
@@ -0,0 +1,27 @@
...
+      formats: ["cjs", "es"],
+    },
+    outDir: "dist",
+    minify: false,
+    sourcemap: true,
+  },
+  test: {
+    environment: "jsdom",
+    globals: true,
+  },
```
roninjin10, line 25: this config is the test config too which is very nice

packages/react/package.json
```diff
@@ -4,7 +4,10 @@
   "description": "React tools for MUD client.",
   "license": "MIT",
   "source": "src/index.ts",
-  "main": "src/index.ts",
+  "type": "module",
+  "types": "dist/index.d.ts",
+  "main": "dist/index.js",
+  "module": "dist/index.mjs",
```
roninjin10, line 10: The way this repo in general handles setting all this up is a bit of a mess.  This is what build tools use to find the types, esm, or cjs.   I would clean this up everywhere.

packages/react/.eslintrc
```diff
@@ -1,4 +1,7 @@
 {
   "extends": ["../../.eslintrc", "plugin:react/recommended", "plugin:react-hooks/recommended"],
-  "plugins": ["react", "react-hooks"]
+  "plugins": ["react", "react-hooks"],
+  "rules": {
+    "@typescript-eslint/consistent-type-imports": "error"
```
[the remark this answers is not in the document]
holic: can we separate this into its own PR and turn this on everywhere?

packages/react/vite.config.ts
```diff
@@ -0,0 +1,27 @@
...
+  plugins: [nodeResolve() as any, typescript(), commonjs(), peerDepsExternal(), pluginJson()],
```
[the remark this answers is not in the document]
holic: any way we can test for which ones we actually need and remove the rest? would be nice to slim this down as I don't know when we'll come back to this

holic:
ooc why did you resolve your own comments? makes it a little hard to have a conversation in them now that they're all collapsed

roninjin10:
@holic reopening any that conversations are happening in.   Since I tend to leave tons of comments on all my prs that are mostly there to just provide extra context my general habit is to comment and then resolve it so the comment is there but the comment is also not getting in the way creating extra noise in the review

packages/react/.eslintrc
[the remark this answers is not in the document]
roninjin10: Yea it should be seperate pr.   This pr in general mostly exists just to try to solicit early feedback before I go about refactoring anything.   I was going to open an issue but this pr was quick to whip up and I figured it would be better than an issue

packages/react/vite.config.ts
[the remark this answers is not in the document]
roninjin10: Yes would be easy just need to remove them 1 at a time and see if they break or don't break.   I'm guessing we don't need nodeResolve.   I'm guessing we don't need commonjs().   I'm guessing we don't need pluginJson

holic:
> @holic reopening any that conversations are happening in. Since I tend to leave tons of comments on all my prs that are mostly there to just provide extra context my general habit is to comment and then resolve it so the comment is there but the comment is also not getting in the way creating extra noise in the review

the context is helpful for us reviewing too, so I have a slight preference to leave em open!

packages/react/.eslintrc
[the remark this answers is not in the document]
roninjin10: removed from this pr I can open a seperate pr later this week to add it.   I'll see if there are any other lint rules useful to add too while I'm at it.   I will only add a lint rule if it has an autofixer and is useful

packages/react/vite.config.ts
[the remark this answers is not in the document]
roninjin10: I was right removed everything except typescript and peerDepsExternal

packages/react/vite.config.ts
roninjin10: Note how I didn't need to change any of the tests.  This is because jest and vitest are almost the same api

packages/utils/package.json
```diff
@@ -22,7 +22,6 @@
     "release": "npm publish --access=public"
   },
   "devDependencies": {
-    "@rollup/plugin-node-resolve": "^13.1.3",
```
roninjin10, line 25: ooops wrong package.json

packages/react/jest.config.js
```diff
@@ -1,9 +0,0 @@
-/** @type {import('ts-jest/dist/types').InitialOptionsTsJest} */
-module.exports = {
-  preset: "ts-jest",
-  moduleNameMapper: {
-    // fix TS build issues
-    "^(..?/.*).js$": "$1",
```
roninjin10, line 6: Here can see jest wasn't working as intended

packages/react/package.json
alvrs: Currently we're pointing `main` to the typescript source in order to have instant updates during development, and then only set it to `dist/index.js` in the `prerelease` step. Down to change this if find an equally performant alternative to pointing to the ts source for development though

holic:
moved this to a draft until we can rebase/merge main

holic:
we're moving to tsup, see #660

PR #518, opened 2023-03-27
fix(cli): add back in resolveTableId export for use with mudConfig

holic:
Removed in #517 but didn't realize we needed this in https://github.com/latticexyz/v2sandbox/pull/7


holic:
Confirmed locally that this works without issue (the exported module doesn't use any node deps), so gonna self-approve and merge so I can unblock some follow up work

PR #522, opened 2023-03-28
feat(network): add option to sync in main thread instead of worker

alvrs:
Adds an option to `createSyncWorker` (and `setupMUDNetwork`) to instantiate the "SyncWorker" in the main thread instead of a worker thread.

### DIFF
--- a/packages/network/src/createSyncWorker.ts
+++ b/packages/network/src/createSyncWorker.ts
@@ -2,7 +2,7 @@ import { Components } from "@latticexyz/recs";
 import { fromWorker } from "@latticexyz/utils";
 import { map, Observable, Subject, timer } from "rxjs";
 import { NetworkEvent } from "./types";
-import { Input, Ack, ack } from "./workers/SyncWorker";
+import { Input, Ack, ack, SyncWorker } from "./workers/SyncWorker";
 
 /**
  * Create a new SyncWorker ({@link Sync.worker.ts}) to performn contract/client state sync.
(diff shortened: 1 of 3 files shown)


dk1a (approved):
Do we even need workers still? Given the troubles of using them in e.g. vite I'd just remove them unless UI noticeably freezes without using workers.

Otherwise a more explicit default for `thread` would be nice, rn it's not obvious that it's `main`

packages/network/src/createSyncWorker.ts
```diff
@@ -14,22 +14,39 @@ import { Input, Ack, ack } from "./workers/SyncWorker";
...
+      ackSub?.unsubscribe();
+    };
+  } else {
+    // Otherwise sync the state in the main thread
+    // Pass in a "config stream", receive a stream of ECS events
+    const subscription = new SyncWorker<C>().work(input$).subscribe(ecsEvents$);
+    dispose = () => {
+      subscription?.unsubscribe();
+      ackSub?.unsubscribe();
+    };
```
holic, line 48: Not blocking here but a potential solution if we find we need to lean on `requestIdleCallback` to avoid long blocks on the main thread: https://ncjamieson.com/how-to-use-requestidlecallback/

holic (approved):

holic:
> Do we even need workers still? Given the troubles of using them in e.g. vite I'd just remove them unless UI noticeably freezes without using workers

I'm curious how vite builds this now that this is conditional.

dk1a:
> I think the vite issue only occurs when trying to use a worker (it just doesn't start and fails silently), not when building with vite.

yes, but there's another one with imports in firefox, not sure if that still remains

holic:
Another thing to consider is by separating the import path, we could also remove the need to add `threads` etc as peer deps.

PR #526, opened 2023-03-28
feat(std-client): move v2 setup to its own function/file

holic:


packages/std-client/src/setup/setupMUDV2Network.ts
```diff
@@ -0,0 +1,196 @@
...
+    // If no initial gas price is provided, check the gas price once and add a 30% buffer
+    options?.initialGasPrice || Math.ceil((await signerOrProvider.get().getGasPrice()).toNumber() * 1.3)
+  );
+
+  const { txQueue, dispose: disposeTxQueue } = createTxQueue(contracts, network, gasPriceInput$, {
+    devMode: options.networkConfig.devMode,
+  });
+  world.registerDisposer(disposeTxQueue);
+
+  const { systems, registerSystem, getSystemContract } = createSystemExecutor<SystemTypes>(
```
holic, line 132: @alvrs do we need this in v2 if we have the central world + interfaces? (didn't wanna do the work of resolving TS issues with this if we could just remove it)

packages/std-client/src/setup/setupMUDV2Network.ts
```diff
@@ -0,0 +1,196 @@
...
+  );
+
+  const decodeNetworkComponentUpdate = createDecodeNetworkComponentUpdate(world, components, mappings);
+  const { systemCallStreams, decodeAndEmitSystemCall } = createSystemCallStreams(
+    world,
+    keys(SystemAbis),
+    SystemsRegistry,
+    getSystemContract,
+    decodeNetworkComponentUpdate
+  );
```
holic, line 150: same question here: wondering if we can just replace this with the separate v2 call streams approach to avoid having to make it compatible with v1 types, etc.

packages/std-client/src/setup/setupMUDV2Network.ts
alvrs: we don't need the system executor in v2!

packages/std-client/src/setup/setupMUDV2Network.ts
alvrs: yeah I think that's reasonable

packages/std-client/src/setup/setupMUDV2Network.ts
```diff
@@ -0,0 +1,186 @@
...
+
+  const network = await createNetwork(networkConfig);
+  world.registerDisposer(network.dispose);
+
+  const signerOrProvider = computed(() => network.signer.get() || network.providers.get().json);
+
+  const { contracts, config: contractsConfig } = await createContracts<{ World: WorldContract }>({
+    config: { World: { abi: WorldAbi, address: networkConfig.worldAddress } },
+    signerOrProvider,
+  });
```
alvrs, line 119: In a v2 environment the provided worldAddress won't correspond to the `solecs` `WorldAbi`. I'm happy with still merging this though and iterating on it, and eventually replacing the network package with a fresh sync package for v2.

alvrs (dismissed):
There are a couple things in `setupMUDV2Network` that are only needed for `v1`, but for I'm good with merging this for now (since v2 support in `network` is mostly for the transition period).

packages/std-client/src/setup/setupMUDV2Network.ts
holic: Would it be a bad idea to use the v2world ABI/contract here? Not sure what the txQueue and other downstream stuff depends on.

image

alvrs: Unfortunately the interface from `world` doesn't include the user defined function selectors (which are generated by `mud worldgen && mud worldtypes` in the project repo)

alvrs: we can require users to pass their world interface as an argument though, similar to before they had to pass the system types object

holic: Since both this setup function and worldgen depend on the mud config, I wonder if we have enough info to avoid having to pass through the artifact of worldgen into this?

Ideally, our mud config approach looks something like:
- user supplies simplified config (with optional shorthands, etc.)
- method + TS types that "expand" the config into something more verbose and useful for downstream things like codegen
  - the result of this would be used to populate things like component definitions and types, etc.
- the expanded config is used by things like worldgen, etc.

The only thing that breaks this is that system methods (and thus the world interface) are based on the filesystem rather than config.

dk1a: #536 reimagines how setup works, so if we go with tsgen it can do all the work and setup just gets readymade tables+systems

PR #527, opened 2023-03-28
feat(cli): add `set-version` to upgrade all MUD dependencies in a project

alvrs:
- adds a simple `set-version <version>` cli command to upgrade all mud packages in a package to the given version
- if `canary` is given as version, the cli resolves to the current `canary` version number and puts the exact number in the package.json to avoid unexpected breaking changes on later installs. (Also there seems to be an issue with yarn's cache and `canary` versions where you have to clear the yarn cache to get the latest canary if you already have a canary installed. This command fixes this by avoiding the `canary` keyword in package.json)

Examples:

```
yarn mud set-version -v canary --backup
yarn mud set-version --restore
```

### DIFF
--- a/packages/cli/src/commands/set-version.ts
+++ b/packages/cli/src/commands/set-version.ts
@@ -1,34 +1,48 @@
-import { readFileSync, rmSync, writeFileSync } from "fs";
+import { existsSync, readFileSync, rmSync, writeFileSync } from "fs";
 import path from "path";
 import type { CommandModule } from "yargs";
 import { logError, MUDError } from "../utils/errors.js";
 
 type Options = {
-  branch?: string;
+  github?: string;
   backup?: boolean;
+  force?: boolean;
   restore?: boolean;
+  npm?: string;
 };
 
-const BACKUP_FILE = ".mudinstall";
+const BACKUP_FILE = ".mudbackup";
 
 const commandModule: CommandModule<Options, Options> = {
-  command: "install",
+  command: "set-version",
 
   describe: "Install a custom MUD version (local or GitHub) and backup the previous version",
 
   builder(yargs) {
     return yargs.options({
-      branch: { type: "string" },
+      github: { type: "string", description: "The MUD GitHub branch to install from" },
+      npm: { type: "string", description: "The MUD NPM version to install" },
       backup: { type: "boolean", description: "Back up the current MUD versions to `.mudinstall`" },
+      force: {
+        type: "boolean",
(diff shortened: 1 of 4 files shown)


packages/cli/src/commands/set-version.ts
```diff
@@ -0,0 +1,169 @@
...
+
+  async handler(options) {
+    try {
+      if (!options.mudVersion && !options.restore) {
+        throw new MUDError(`Version parameter is required unless --restore is provided.`);
+      }
+
+      // Resolve the `canary` version number if needed
+      options.mudVersion =
+        options.mudVersion === "canary" ? await getCanaryVersion("@latticexyz/world") : options.mudVersion;
```
holic, line 41: is this just an arbitrary package for finding the current version?

packages/cli/src/commands/set-version.ts
```diff
@@ -0,0 +1,169 @@
...
+const BACKUP_FILE = ".mudbackup";
+
+const commandModule: CommandModule<Options, Options> = {
+  command: "set-version",
+
+  describe: "Install a custom MUD version and optionally backup the previously installed version",
+
+  builder(yargs) {
+    return yargs.options({
+      backup: { type: "boolean", description: "Back up the current MUD versions to `.mudinstall`" },
```
holic, line 23: should this be `.mudbackup` (value of `BACKUP_FILE` variable above)?

packages/cli/src/commands/set-version.ts
```diff
@@ -0,0 +1,169 @@
...
+
+  const packageJson = readPackageJson(filePath);
+
+  // Load .mudinstall if `restore` is true
+  const backupJson = restore ? readPackageJson(backupFilePath) : undefined;
+
+  // Find all @latticexyz dependencies
+  const mudDependencies: Record<string, string> = {};
+  for (const key in packageJson.dependencies) {
+    if (key.startsWith("@latticexyz")) {
```
holic, line 81: kinda feel like this is worth lifting into a variable/constant, to make it easier to change if we decide later to move to e.g. `@muddev`

holic (commented):
couple of small things but otherwise lgtm

packages/cli/src/commands/set-version.ts
alvrs: good idea, will do that!

packages/cli/src/commands/set-version.ts
alvrs: indeed, will use the variable

holic (approved):

PR #528, opened 2023-03-28
chore: check in build artifacts

alvrs:
TODO:
- [ ] set up an action to check if the build artifacts are up to date, like #468


holic:
hmm, what's the reason for this?

alvrs:
this would be required to install packages directly from github, see #527

holic:
> this would be required to install packages directly from github, see #527. Otherwise we have to somehow build the target package on the consumer's machine, which then requires all dependencies to be installed in the consumer project, and potentially has cross-dependencies from other unreleased mud packages.

@roninjin10 does the nx approach help with this at all?

alvrs:
> @roninjin10 does the nx approach in https://github.com/latticexyz/mud/issues/529 help with this at all?

I think the use case is slightly different. Installing from github would be to allow installing "nightly" package versions that are not released to npm yet, without a local MUD setup. (The reason why this is high on my list right now is that I expect we have a couple hot fixes we have to do during the AW hackathon this weekend and allow participants to update to the latest nightly version, without a full npm alpha release cycle). The `nx` approach on the other hand would be an alternative to local linking (as far as my limited nx knowledge goes), for which we currently use a not-super-stable yarn setup.

Note that this draft PR is mostly an experiment to see if installing directly from github is even feasible.

roninjin10:
nx would help with making sure everything is built that needs to be built.   It's a much more robust way of smartly caching and building the build graph [than github actions paths](https://github.com/latticexyz/mud/pull/468/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721L10).  Otherwise unrelated.

I don't think this is a good approach.  Installing from github is definitely doable, but it's an escape hatch.   What I recomend is adding a github action that publishes every develop branch and/o has a `workflow_trigger` to manually trigger it on any branch at any time.   This workflow would then run `npm publish --tag canary` to publish a under @latticexyz/mudcli@canary

roninjin10:
`npm install @latticexyz/package@latest` will get latest official release
`npm install @latticexyz/package@canary` gets you latest code whatever is on main branch

roninjin10:
forge calls it nightly so I'd maybe better to call it @nightly than @canary?

alvrs:
Came back here to add that installing from GitHub might not be necessary if we automatically release a canary version from main, but @roninjin10 beat me to it.

@roninjin10 would you have time to set up such an action before the weekend by any chance? If not I'll give it a shot tomorrow. For context: we have a hackathon this weekend where many people will work with mud v2 and automated canary releases would be very helpful.

roninjin10:
Yea I can look at it later tonight should be quick

roninjin10:
Might need changes left tons of comments but this should be pretty much all you need for canary releases https://github.com/latticexyz/mud/pull/531/files

alvrs:
Amazing, thank you! Closing this PR in favor of https://github.com/latticexyz/mud/pull/531

PR #535, opened 2023-03-29
feat(mud): add global mud package and release cli under it

alvrs:
~~This adds configuration to the monorepo's root package to be released under `mud`. For now only the `cli` package is included (binary and regular imports), but it's very easy to add all other entry points in the future.~~

Since publishing the root of a yarn workspace is prohibited by yarn, I went with a workaround: creating a new sub workspace `mud`, installing `cli` as a dependency with `nohoist` (so the files are in `mud/node_modules`) and re-exporting those files from the `mud` package. The `mud` package is prepared last, after all other workspace packages, so it's dependencies should always be up to date during the release process.

Note: The initial approach of releasing the root as package would have been way cleaner, but is impossible with our current yarn workspace setup. We should revisit this after we migrated the monorepo to `pnpm` (see #511)


package.json
```diff
@@ -6,7 +6,10 @@
     "type": "git",
     "url": "https://github.com/latticexyz/mud.git"
   },
-  "private": true,
+  "exports": {
+    "./cli": "./packages/cli/dist/index.js"
+  },
+  "bin": "./packages/cli/dist/mud.js",
```
alvrs, line 12: For now we only release the `cli` package under the global `mud` package, but by adding more entry points here (and adding exceptions for the respective packages to `.npmignore`) we can easily start publishing all packages under `mud/<package>`!

packages/cli/package.json
```diff
@@ -6,9 +6,7 @@
   "types": "dist/index.d.ts",
   "type": "module",
   "license": "MIT",
-  "bin": {
-    "mud": "./dist/mud.js"
-  },
+  "bin": "./dist/mud.js",
```
alvrs, line 9: This change is to avoid a binary conflict of `mud` coming from the cli package directly and the global `mud` package.

alvrs:
Well, according to yarn:

> Note that the private: true is required! Workspaces are not meant to be published, so we’ve added this safety measure to make sure that nothing can accidentally expose them.

This is unfortunate because it would suit our use case pretty well. Will see how we can work around this.

holic:
> This is unfortunate because it would suit our use case pretty well. Will see how we can work around this.

I wonder if pnpm would make this easier?

Alternatively, could add a new package that has the cli as a dep and re-exports the bin in its package?

alvrs:
> I wonder if pnpm would make this easier?

Most likely yes!

> Alternatively, could add a new package that has the cli as a dep and re-exports the bin in its package?

This is the approach I went with now - it seems to work but we should definitely revisit once we migrate to pnpm.

holic (approved):

alvrs:
1. currently the cli only works from within a package that has `ts-node` installed as a (dev) dependency (see https://github.com/latticexyz/mud/issues/538)
2. if you're within a package you can just install the mud cli as a dev dependency and have its version synced with the other mud packages in that project, so that's the recommended way

Since the only motivation for releasing the mud cli under the mud package is to be able to run `npx mud` outside of a mud package, which is blocked by #538, there is no need to rush this PR before the pnpm migration (see https://github.com/latticexyz/mud/pull/511), which will most likely allow us to go with the cleaner approach of publishing the workspace root as `mud` package.

Gonna close this PR until then.

PR #536, opened 2023-03-29
feat(cli/recs/std-client): add ts definitions generator

dk1a:
[v2sandbox PR](https://github.com/latticexyz/v2sandbox/pull/15)

this is a rather hacky solution. Eventually I want tsgen separate from cli, so it can depend on recs (or even be the root source of ts table types)

Having `tablegen` plop `contractComponents.ts` into another package feels weird.
- Config can be in root and defining client/contract folders might be better
- Or ts can have a 2nd config, which just does `import config from "../contracts/mud.config.mts"; export default config;` And then you run a separate `tsgen` on the client


holic:
I agree that it does feel weird, also feels weird to have another config param for this, esp since its mostly a v1 compat thing. Ideally we'd keep them separate.

How painful would it be to add a separate CLI command for running TS codegen? Something like
```
mud tsgen path/to/mud.config.mts --out tsgen
```
(where `--out` is the output directory for generated TS files, relative to where the command was run)

dk1a:
> How painful would it be to add a separate CLI command for running TS codegen? Something like
> 
> ```
> mud tsgen path/to/mud.config.mts --out tsgen
> ```
> 
> (where `--out` is the output directory for generated TS files, relative to where the command was run)

Actually this'd be the best solution for now, and not that hard. I just didn't think of explicitly specifying config path in the command itself for some reason

packages/recs/package.json
```diff
@@ -13,6 +13,7 @@
   "scripts": {
     "prepare": "yarn build",
     "lint": "eslint . --ext .ts",
+    "codegen": "renderRecsV1Tables",
```
holic, line 16: is this an executable on its own?

packages/cli/src/commands/tsgen.ts
```diff
@@ -0,0 +1,33 @@
...
+import { loadStoreConfig } from "../config/loadStoreConfig.js";
+import { tsgen } from "../render-ts/tsgen.js";
+
+type Options = {
+  configPath: string;
+  out: string;
+};
+
+const commandModule: CommandModule<Options, Options> = {
+  command: "tsgen",
```
alvrs, line 11: I wonder if this should be `recsgen` or something since it might not be the only ts we generate

dk1a: that's actually why I called it tsgen - since I imagine it will generate enums and maybe other things aside from recs definitions

packages/std-client/src/mud-definitions/store/contractComponents.ts
```diff
@@ -0,0 +1,89 @@
...
+        },
+        {
+          metadata: {
+            contractId: tableId.toHexString(),
+            tableId: tableId.toString(),
+          },
+        }
+      );
+    })(),
+  };
```
alvrs, line 88: we should probably move out the `Mixed` and `Vector2` table definitions to a different config file in store's tests to not pollute the user's types with these tables

packages/std-client/src/setup/setupMUDV2Network.ts
```diff
@@ -9,36 +9,35 @@ import {
...
-  mudConfig,
+export async function setupMUDV2Network<C extends ContractComponents>({
   networkConfig,
+  world,
+  contractComponents,
   initialGasPrice,
   fetchSystemCalls,
   syncThread,
-}: SetupMUDV2NetworkOptions<M>) {
-  const world = createWorld();
```
holic, line 40: ooc why is it important to pass this in as an argument instead of creating here?

holic (approved):

packages/std-client/src/setup/setupMUDV2Network.ts
dk1a: I need world to create contractComponents to pass them to setup, so world has to come b4 setup

PR #540, opened 2023-03-29
build(world): report call with weird gas

dk1a:
dug into the cause of it too, seems to be when `isStore` gets called on `WorldTestSystem`, which doesn't have that method. A try-catch catches the error, but the invalid call creates the huge gas cost. Might just be a foundry glitch, not sure what's up with that

### DIFF
--- a/packages/world/gas-report.txt
+++ b/packages/world/gas-report.txt
@@ -9,6 +9,7 @@
 (test/KeysWithValueModule.t.sol) | install keys with value module [world.installRootModule(keysWithValueModule, abi.encode(sourceTableId))]: 600319
 (test/KeysWithValueModule.t.sol) | set a field on a table with KeysWithValueModule installed [world.setField(namespace, sourceFile, keyTuple1, 0, abi.encodePacked(value1))]: 177731
 (test/KeysWithValueModule.t.sol) | change a field on a table with KeysWithValueModule installed [world.setField(namespace, sourceFile, keyTuple1, 0, abi.encodePacked(value2))]: 142076
+(test/World.t.sol) | Call an autonomous system [world.call("namespace", "testSystem", abi.encodeWithSelector(WorldTestSystem.emitCallType.selector))]: 8797746687694640000
 (test/World.t.sol) | Delete record [world.deleteRecord(namespace, file, singletonKey)]: 16115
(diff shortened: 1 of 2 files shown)


packages/world/gas-report.txt
```diff
@@ -9,6 +9,7 @@
 (test/KeysWithValueModule.t.sol) | install keys with value module [world.installRootModule(keysWithValueModule, abi.encode(sourceTableId))]: 600319
 (test/KeysWithValueModule.t.sol) | set a field on a table with KeysWithValueModule installed [world.setField(namespace, sourceFile, keyTuple1, 0, abi.encodePacked(value1))]: 177731
 (test/KeysWithValueModule.t.sol) | change a field on a table with KeysWithValueModule installed [world.setField(namespace, sourceFile, keyTuple1, 0, abi.encodePacked(value2))]: 142076
+(test/World.t.sol) | Call an autonomous system [world.call("namespace", "testSystem", abi.encodeWithSelector(WorldTestSystem.emitCallType.selector))]: 8797746687694640000
```
holic, line 12: It's weird that `gasleft()` has the same effect here (large gas amount).

Is there a minimum repro we can report to the foundry team?

dk1a:
The current state of a minimal repro. Now I'm confused whether this is even a bug, and whether it's just foundry or EVM in general. Will shelve it for now.
As a solution we should be careful to pass explicit gas when staticcalling `IStore(address(this)).isStore()` because `fallback`

```solidity
pragma solidity >=0.8.0;

contract HelperContract {
  bool b;

  fallback() external {
    // anything to violate a staticcall is enough here
    b = false;
  }
}

contract GasTest {
  event GasUsed(uint256 gas);

  function testGas() public returns (uint256 gas) {
    gas = gasleft();

    HelperContract helper = new HelperContract();
    assembly {
      let res := staticcall(
        // any number you put here will be entirely consumed
        12345678,
        helper,
        0,
        0,
        0,
        0
      )
    }

    gas = gas - gasleft();
    emit GasUsed(gas);
  }
}
```

alvrs:
Let's keep this PR open for now until we've investigated more where this comes from and how to avoid it

alvrs:
closing this for now, will be addressed as part of https://github.com/latticexyz/mud/issues/940

PR #541, opened 2023-03-30
fix(network): handle singleton/empty keys

holic:


packages/network/src/utils.ts
```diff
@@ -13,6 +13,9 @@ export function messagePayload(msg: Message) {
 // For composite-key v2 entities (hex strings concatenated with `:`), leave the ID as is
 // See `keyTupleToEntityID`
 export function normalizeEntityID(entityID: string | EntityID | BigNumber): EntityID {
+  if (entityID === "") {
+    return BigNumber.from(0).toHexString() as EntityID;
+  }
```
alvrs, line 18: are there any cases where this would come through as a string if we have the fix with `keyTuple.length === 0 -> SingletonID` below?

holic: I'm not sure, but without this, an empty string has an obscure error because it passes our `isHex` check and then fails to convert to a BigNumber.

alvrs: `entityID === ""` seems unexpected though, maybe we should throw an explicit error here instead? `0` is a valid entity/tableId, but if it's `""` there is something weird going on

holic: changed to `throw` for now

alvrs (approved):

PR #542, opened 2023-03-30
fix(std-client): properly print errors and prevent failure when an action errors

Kooshaba:


packages/std-client/src/systems/ActionSystem/createActionSystem.ts
```diff
@@ -205,7 +206,7 @@ export function createActionSystem<M = undefined>(world: World, txReduced$: Obse
    */
   function remove(actionId: EntityID) {
     const action = actionData.get(actionId);
-    if (!action) throw new Error("Trying to remove an action that does not exist.");
+    if (!action) return;
```
alvrs, line 209: do you have context on in which situations this error was thrown?

Kooshaba:
@alvrs 

1. handleError was getting called in the catch statement, which ended up calling `remove`. `remove` is also called at the end of the execute function, so this double removal was always throwing a useless error. i think it's fine that two things are trying to remove an action, one just becomes a no-op.
2. i'm passing in the exception to handle error so it is always printed. the two situations are a) in the catch block and b) waiting from the confirmation promise in case we aren't waiting for txConfirmed

tested this out with the new `worldSend` pattern. the execute function should be async and return `await worldSend(...)`

alvrs (approved):

PR #543, opened 2023-03-30
feat(network): add fastTxExecute util

alvrs:
Moving some of the magic of `createTxQueue` into a simpler helper function. Totally optional for consumers, but here are some quick benchmarks for sending tx on the testnet:

With manual gas limit:
CleanShot 2023-03-30 at 16 29 48@2x

Without manual gas limit:
CleanShot 2023-03-30 at 16 26 55@2x

Usually consumers in MUD don't have to await the result but only sending the transaction, so the first of the two numbers is more relevant

Corresponding v2sandbox branch: https://github.com/latticexyz/v2sandbox/pull/20

### DIFF
--- a/packages/network/package.json
+++ b/packages/network/package.json
@@ -26,8 +26,7 @@
   },
   "dependencies": {
     "debug": "^4.3.4",
-    "lodash": "^4.17.21",
-    "vitest": "^0.29.8"
+    "lodash": "^4.17.21"
   },
   "peerDependencies": {
     "@ethersproject/providers": "^5.6.1",
--- a/packages/network/package.json
+++ b/packages/network/package.json
@@ -26,7 +26,8 @@
   },
   "dependencies": {
     "debug": "^4.3.4",
-    "lodash": "^4.17.21"
+    "lodash": "^4.17.21",
+    "vitest": "^0.29.8"
   },
   "peerDependencies": {
     "@ethersproject/providers": "^5.6.1",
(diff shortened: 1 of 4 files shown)


Kooshaba:
i love this!

packages/network/src/createFastTxExecutor.ts
```diff
@@ -0,0 +1,139 @@
+import { ArgumentsType } from "vitest";
+import { BigNumber, Contract, Overrides, Signer } from "ethers";
+import { JsonRpcProvider } from "@ethersproject/providers";
+
+/**
+ * Create a stateful util to execute transactions as fast as possible.
+ * Internal state includes the current nonce and the current gas price.
+ */
+export async function createFastTxExecutor(
```
holic, line 9: I assume we'd ideally only instantiate one of these per signer? Otherwise lots of nonce issues? Is that worth noting here?

holic (dismissed):

holic (approved):

packages/network/package.json
```diff
@@ -26,7 +26,8 @@
   },
   "dependencies": {
     "debug": "^4.3.4",
-    "lodash": "^4.17.21"
+    "lodash": "^4.17.21",
+    "vitest": "^0.29.8"
```
holic, line 30: this should prob be a dev dep if we need it?

alvrs: it's used for `ArgumentsType`, but should probably just pull that out from vitest

alvrs:
Gonna self-approve this since it was approved before and the last change was very minor + test are passing + works locally

PR #544, opened 2023-03-30
fix(recs): overridden component update stream should return the overridden component

holic:


holic:
cc @Kooshaba @yonadaaa in case this impacts Sky Strife (I'm hopeful it improves things though!)

alvrs (approved):

PR #546, opened 2023-03-30
feat(std-client): add getBurnerWallet util

holic:


packages/std-client/src/getBurnerWallet.ts
```diff
@@ -0,0 +1,38 @@
...
+      // loop issues, and just warn the user. A refresh will go through the logic above to
+      // create a new burner wallet.
+      console.warn("Burner wallet removed from cache! You may need to reload to create a new wallet.");
+      return;
+    }
+
+    subject.next(event.newValue);
+  });
+
+  return subject;
```
holic, line 37: I've seen in some of our code that we return this using `asObservable`. Under what conditions does this make sense over returning the subject itself? I assume it would make it somewhat "read only" (but then we'd lose out on the `.get()` for the current value)?

alvrs: That's exactly right, `asObservable` makes it readonly but loses some of the `Subject` functionality. In this case `.get()` is pretty important so makes sense to not turn it into an observable.

alvrs (approved):

PR #547, opened 2023-03-30
fix(std-client): export getBurnerWallet

holic:


holic:
(low risk so just merging to fix my issues)

PR #552, opened 2023-03-30
feat(world): add UniqueEntityModule

dk1a:


packages/world/mud.config.mts
```diff
@@ -105,6 +105,12 @@ export default mudConfig({
       tableIdArgument: true,
       storeArgument: true,
     },
+    UniqueEntity: {
+      directory: "/modules/uniqueentity/tables",
+      primaryKeys: {},
+      schema: "uint256",
```
holic, line 111: do we expect entity IDs to be uint256 vs bytes32? I had started using bytes32 in my v2 emojimon project (since it more closely matches a key)

dk1a: ultimately I wanna try some built-in user-defined `EntityId` as the type.
I also want to allow default metaconfig, so users can choose what their default keys are.
For now whatever is fine for me, should I switch to bytes32?

packages/world/src/modules/uniqueentity/UniqueEntityModule.sol
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * Note: this module currently expects to be `delegatecalled` via World.installRootModule.
+ * Support for installing it via `World.installModule` depends on `World.callFrom` being implemented.
+ */
+contract UniqueEntityModule is IModule, WorldContext {
+  function getName() public pure returns (bytes16) {
+    return bytes16("uniqueEntity");
+  }
+
+  // The namespace argument is not used because the module is always installed in the root namespace
```
alvrs, line 22: this comment is outdated (same is true for the comment in RegistrationModule where this comment presumably comes from): the input here is not necessarily a namespace but rather arbitrary bytes passed in as arguments

packages/world/src/modules/uniqueentity/getUniqueEntity.sol
```diff
@@ -0,0 +1,25 @@
...
+  uniqueEntity = UniqueEntity.get() + 1;
+  UniqueEntity.set(uniqueEntity);
+}
+
+/**
+ * Increment and get an entity nonce.
+ */
+function getUniqueEntity(IStore store) returns (uint256 uniqueEntity) {
+  uniqueEntity = UniqueEntity.get(store) + 1;
+  UniqueEntity.set(store, uniqueEntity);
```
alvrs, line 24: One limitation with this is that the util can only be used by Systems with access to the `UniqueEntity` table. What do you think about adding a public system to increment the UniqueEntity table instead of writing to it directly?

packages/world/src/modules/uniqueentity/UniqueEntityModule.sol
```diff
@@ -0,0 +1,27 @@
...
+import { WorldContext } from "../../WorldContext.sol";
+
+import { UniqueEntity } from "./tables/UniqueEntity.sol";
+
+/**
+ * This module creates a table that stores a nonce, and `getUniqueEntity`
+ * returns an incremented nonce each time.
+ *
+ * Note: this module currently expects to be `delegatecalled` via World.installRootModule.
+ * Support for installing it via `World.installModule` depends on `World.callFrom` being implemented.
```
alvrs, line 15: The `UniqueEntity` table doesn't have to be owned by the caller of this module. Instead we can create a new namespace for the module in which the module installs the table (and thereby owns the table). Then we can create a public system to increment the value, and add a util to call the system (instead of putting the write logic in the util directly).

packages/world/mud.config.mts
alvrs: I agree with changing to bytes32 for now to more closely match a key

packages/world/src/modules/uniqueentity/UniqueEntityModule.sol
dk1a: yeah the system stuff you mention is why this is a draft - I've been thinking about adding a system as well

packages/world/mud.config.mts
```diff
@@ -99,8 +99,17 @@ export default mudConfig({
       },
       tableIdArgument: true,
     },
+    UniqueEntity: {
+      directory: "/modules/uniqueentity/tables",
+      primaryKeys: {},
+      schema: "bytes32",
```
holic, line 105: now that this has a system that things would funnel through, I feel like it makes sense to move this back to `uint256` to avoid converting back/forth, then just have the system return `bytes32(incremented ID)`

packages/world/src/modules/uniqueentity/UniqueEntitySystem.sol
```diff
@@ -0,0 +1,20 @@
...
+import { NAMESPACE, TABLE_NAME } from "./constants.sol";
+import { ResourceSelector } from "../../ResourceSelector.sol";
+
+contract UniqueEntitySystem is System {
+  /**
+   * Increment and get an entity nonce.
+   */
+  function getUniqueEntity() public virtual returns (bytes32 uniqueEntity) {
+    uint256 tableId = uint256(ResourceSelector.from(NAMESPACE, TABLE_NAME));
+    uniqueEntity = bytes32(uint256(UniqueEntity.get(tableId)) + 1);
```
holic, line 17: does this guarantee uniqueness in any way?

wonder if we should keccak256 this with some other entropy (table ID maybe) to reduce chance of collisions?

dk1a: no, but it also shouldn't be an issue as entities aren't user-supplied, the only collision I can think of is SingletonId (which shouldn't be needed with singleton tables)

packages/cli/src/utils/deploy-v2.ts
```diff
@@ -102,6 +108,7 @@ export async function deploy(mudConfig: MUDConfig, deployConfig: DeployConfig):
   console.log(chalk.blue("Installing core World modules"));
   await fastTxExecute(WorldContract, "installRootModule", [await modulePromises.CoreModule, "0x"]);
   await fastTxExecute(WorldContract, "installRootModule", [await modulePromises.RegistrationModule, "0x"]);
+  await fastTxExecute(WorldContract, "installModule", [await modulePromises.UniqueEntityModule, "0x"]);
```
alvrs, line 111: I think while this is a very common usecase, we should still allow users to decide to include or not include it, as it's not strictly necessary for the World contract to function (unlike `CoreModule` and `RegistrationModule`). If we remove this line, consumers have to add the module to their `mud.config.mts`, which is slightly inconvenient, but can be made more convenient with proper config plugin support in the near future.

alvrs (dismissed):

alvrs (approved):

PR #554, opened 2023-03-31
feat(cli): add `mud test-v2` command

alvrs:
* This PR adds the `mud test-v2` cli command and corresponding `MudV2Test.t.sol` base class
* Without arguments, the command uses the project's mud config to create a deployment on a fresh anvil node, and then executes fork test on that fresh deployment
* With a `--worldAddress` argument, the deployment is skipped and the fork test are executed on the provided world address (and rpc url from the `foundry.toml`)

Corresponding v2sandbox PR: https://github.com/latticexyz/v2sandbox/pull/24

### DIFF
--- a/packages/cli/src/commands/deploy-v2.ts
+++ b/packages/cli/src/commands/deploy-v2.ts
@@ -1,7 +1,7 @@
 import chalk from "chalk";
 import glob from "glob";
 import path, { basename } from "path";
-import type { CommandModule } from "yargs";
+import type { CommandModule, Options } from "yargs";
 import { loadWorldConfig } from "../config/world/index.js";
 import { deploy } from "../utils/deploy-v2.js";
 import { logError, MUDError } from "../utils/errors.js";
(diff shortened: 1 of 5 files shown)


packages/cli/src/commands/deploy-v2.ts
alvrs, line 1: Refactored this to make it possible to import form `test-v2`

packages/std-contracts/src/test/MudV2Test.t.sol
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
+
+import "forge-std/Test.sol";
+
+contract MudV2Test is Test {
+  address worldAddress;
+
+  function setUp() public virtual {
+    worldAddress = vm.parseAddress(vm.readFile(".mudtest"));
```
alvrs, line 10: this temporary file is generated by the `mud test-v2` command.

ludns (approved):

PR #555, opened 2023-03-31
fix(network): fall back to RPC sync if MODE is not available

alvrs:


alvrs:
self-approving - it's a simple change, tested locally and works, and it's time critical

PR #557, opened 2023-03-31
feat: mode mvp followups

authcall:
- Handling of arrays / dynamic data / setField code paths
- Complete decoding handling of static types

### DIFF
--- a/packages/services/pkg/mode/ops/update/update.go
+++ b/packages/services/pkg/mode/ops/update/update.go
@@ -53,7 +53,12 @@ func (builder *UpdateBuilder) BuildUpdateRowFromKV(row map[string]string, fieldN
 		if _, ok := row[field]; !ok {
 			continue
 		}
-		rowStr = rowStr + field + ` = ` + `'` + row[field] + `'`
+		// Handle array fields.
+		if strings.Contains(builder.TableSchema.PostgresTypes[field], "[]") {
+			rowStr = rowStr + field + ` = ` + `ARRAY['` + row[field] + `']`
+		} else {
+			rowStr = rowStr + field + ` = ` + `'` + row[field] + `'`
+		}
 		if idx != len(fieldNames)-1 {
 			rowStr = rowStr + `, `
 		}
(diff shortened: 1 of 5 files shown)


packages/services/pkg/mode/data.go
```diff
@@ -66,6 +84,16 @@ func SerializeRow(row []interface{}, colNames []string, colEncodingTypes []*abi.
 				if err != nil {
 					return nil, err
 				}
+			} else if colEncodingTypes[i].String() == "bytes" || colEncodingTypes[i].String() == "string" {
+				// Handle bytes / string specially to avoid the offset missing issue.
+				encodedField, err = EncodeParameters(
+					[]string{colEncodingType.String()},
+					[]interface{}{row[i]},
+				)
```
holic, line 92: are there tests for these sorts of things to demonstrate what the expected output is for given input?

packages/services/pkg/mode/ops/update/update.go
```diff
@@ -53,7 +53,16 @@ func (builder *UpdateBuilder) BuildUpdateRowFromKV(row map[string]string, fieldN
...
+		// Handle array fields.
+		if strings.Contains(builder.TableSchema.PostgresTypes[field], "[]") {
+			rowStr = rowStr + field + ` = ` + `ARRAY['` + row[field] + `']`
+		} else
+		// Handle bytea raw byte fields.
+		if builder.TableSchema.PostgresTypes[field] == "bytea" {
+			rowStr = rowStr + field + ` = ` + "'\\" + row[field][1:] + `'`
+		} else {
+			rowStr = rowStr + field + ` = ` + `'` + row[field] + `'`
+		}
```
holic, line 65: is this prone to sql injection?

holic (approved):

packages/services/pkg/mode/data.go
authcall: not yet, good point though. i'll add a bunch when we implement a proper encoding solution to make sure sync works (will be separate pr)

PR #562, opened 2023-04-01
fix(network): fix mode decoding

alvrs:
- `decodeAbiParameters` returns an array, but we're interested in the individual value


ludns (approved):
lgtm 🚢

PR #564, opened 2023-04-01
chore(cli): update autogenerated test tables

alvrs:


holic:
these were updated in #574

PR #565, opened 2023-04-01
fix(cli): use esbuild to load mud config

alvrs:
* Remove the `ts-node` requirement from the mud cli by loading config files via esbuild

### DIFF
--- a/packages/cli/src/config/loadConfig.ts
+++ b/packages/cli/src/config/loadConfig.ts
@@ -1,20 +1,21 @@
 import { findUp } from "find-up";
 import path from "path";
-import { NotESMConfigError, NotInsideProjectError } from "../utils/errors.js";
+import { NotInsideProjectError } from "../utils/errors.js";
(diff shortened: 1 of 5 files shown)


packages/cli/src/config/loadConfig.ts
```diff
@@ -1,20 +1,21 @@
...
   try {
+    await esbuild.build({ entryPoints: [configPath], format: "esm", outfile: TEMP_CONFIG });
+    configPath = await resolveConfigPath(TEMP_CONFIG);
     return (await import(configPath)).default;
-  } catch (error) {
-    if (error instanceof SyntaxError && error.message === "Cannot use import statement outside a module") {
-      throw new NotESMConfigError();
-    } else {
-      throw error;
-    }
```
alvrs, line 17: this is not necessary anymore because esbuild transforms all types of js files into esm

ludns (approved):
windows

PR #567, opened 2023-04-01
feat(cli): add mud2 cli entrypoint with only v2 commands

alvrs:
- windows users are reporting errors when using the CLI that are caused by deprecated v1 commands
- This PR adds a `mud2` cli entry point that omits all v1 commands from the binary
- This should fix the windows issues, and is a transition step towards sunsetting the v1 commands (the next step could be to rename `mud -> mud1` and `mud2 -> mud`)

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -7,7 +7,8 @@
   "type": "module",
   "license": "MIT",
   "bin": {
-    "mud": "./dist/mud.js"
+    "mud": "./dist/mud.js",
+    "mud2": "./dist/mud2.js"
   },
   "repository": {
     "type": "git",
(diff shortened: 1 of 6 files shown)


packages/cli/package.json
```diff
@@ -7,7 +7,8 @@
   "type": "module",
   "license": "MIT",
   "bin": {
-    "mud": "./dist/mud.js"
+    "mud": "./dist/mud.js",
+    "mud2": "./dist/mud2.js"
```
holic, line 11: `mud2` without v2 commands is confusing, what about `mud-v1` or `mud-deprecated`?

alvrs: `mud2` is _exclusively_ v2 commands

holic: ohhh, the PR title was misleading!

alvrs: oops true! good catch

PR #569, opened 2023-04-01
feat(world): allow KeysWithValueModule on tables with composite keys

alvrs:
- Before `KeysWithValueHook` required tables to have a single key
- With this PR also tables with composite keys are supported, but only the first key will be indexed

### DIFF
--- a/packages/world/src/modules/keyswithvalue/KeysWithValueHook.sol
+++ b/packages/world/src/modules/keyswithvalue/KeysWithValueHook.sol
@@ -16,15 +16,14 @@ import { getTargetTableSelector } from "./getTargetTableSelector.sol";
  * We can optimize this by adding support for `setIndexOfField` in Store
  * and then replicate logic from solecs's Set.sol.
  * (See https://github.com/latticexyz/mud/issues/444)
+ *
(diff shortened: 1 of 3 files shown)


packages/world/src/modules/keyswithvalue/KeysWithValueHook.sol
```diff
@@ -16,15 +16,14 @@ import { getTargetTableSelector } from "./getTargetTableSelector.sol";
  * We can optimize this by adding support for `setIndexOfField` in Store
  * and then replicate logic from solecs's Set.sol.
  * (See https://github.com/latticexyz/mud/issues/444)
+ *
+ * Note: if a table with composite keys is used, only the first key is indexed
```
holic, line 20: Even though it doesn’t throw now, does this help with the issue @yonadaaa had? Is there a quick way to support composite keys? (ie hashing)

alvrs: > Even though it doesn’t throw now, does this help with the issue @yonadaaa had?

Can you elaborate which issue you're referring to?

holic (approved):

PR #570, opened 2023-04-01
feat(cli): interactive test-runner

ludns:
This adds an interactive test-runner (used with `mud test-v2 --watch`; it's also turned off when the `CI` env var is set, regardless of `--watch`)
The runner allows users to re-run their test suite without redeploying the world and all that stuff. It supports file watching or explicitly re-running tests.
- The deployer runs when the config or the systems change (there is a TODO to run all our code-generation tools when the config changes -- we don't have a `codegen` command yet).
- The tests run (without the deployer running) when the tests change.
The runner also warns the user when their Forge verbosity is too low (if their tests fail), given they won't see logs and errors from their asserts statements.

### DIFF
--- a/packages/cli/src/commands/test-v2.ts
+++ b/packages/cli/src/commands/test-v2.ts
@@ -21,7 +21,7 @@ import { execLog } from "../utils/execLog.js";
 
 type Options = DeployOptions & {
   port?: number;
-  interactive?: boolean;
+  watch?: boolean;
   forgeOptions?: string;
 };
 
@@ -78,7 +78,7 @@ const commandModule: CommandModule<Options, Options> = {
         description:
           "Address of an existing world contract. If provided, deployment is skipped and the RPC provided in the foundry.toml is used for fork testing.",
       },
-      interactive: { type: "boolean", alias: "i" },
+      watch: { type: "boolean", alias: "w" },
       configPath: { type: "string", desc: "Path to the config file" },
       forgeOptions: { type: "string", description: "Options to pass to forge test" },
     });
@@ -118,7 +118,7 @@ const commandModule: CommandModule<Options, Options> = {
       console.log("TODO: rerun all codegen if stale detected");
     }
 
-    if (process.env.CI || !args.interactive) {
+    if (process.env.CI || !args.watch) {
       await runCodegen();
       const worldAddress = await deployWorld();
       writeFileSync(WORLD_ADDRESS_FILE, worldAddress);
--- a/packages/cli/src/commands/test-v2.ts
+++ b/packages/cli/src/commands/test-v2.ts
@@ -21,39 +21,44 @@ import { execLog } from "../utils/execLog.js";
 
 type Options = DeployOptions & {
   port?: number;
-  worldAddress?: string;
   interactive?: boolean;
-  configPath?: string;
   forgeOptions?: string;
 };
 
 const WORLD_ADDRESS_FILE = ".mudtest";
 
-const TESTS_CHANGED = "TESTS_CHANGED";
-const SOURCES_CHANGED = "SOURCES_CHANGED";
-const CONFIG_CHANGED = "CONFIG_CHANGED";
+enum Events {
+  TESTS_CHANGED = "TESTS_CHANGED",
+  SOURCES_CHANGED = "SOURCES_CHANGED",
+  CONFIG_CHANGED = "CONFIG_CHANGED",
+}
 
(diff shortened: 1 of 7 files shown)


packages/cli/src/utils/execLog.ts
```diff
@@ -1,8 +1,9 @@
 import chalk from "chalk";
 import { execa, Options } from "execa";
+import { CommandFailedError } from "./errors.js";
 
 /**
- * Executes the given command, returns the stdout, and logs the command to the console.
+ * Executes the given command, returns the stdout, and logs the command to the logger.
```
ludns, line 6:
```suggestion
 * Executes the given command, returns the stdout, and logs the command to the console.
```

dk1a:
there should probably remain an option to use non-interactive tests for CI

ludns:
good point, i will add a CI check and make it non interactive there.

packages/cli/src/commands/test-v2.ts
```diff
@@ -36,35 +92,199 @@ const commandModule: CommandModule<Options, Options> = {
...
+              break;
+            } else {
+              console.log(chalk.green("Tests changed. rerunning forge tests."));
+              for (const filesChanged of signals.testsChanged) {
+                console.log(chalk.gray("changed: ", filesChanged));
+              }
+              resetSignals();
+              continue;
+            }
+          } else {
```
dk1a, line 233: seems like you could reduce nesting depth by removing the first condition and doing `else if (signals.testsChanged.length > 0)` and then `else`

dk1a (approved):
lgtm. Some parts of the big func can perhaps be split off, but not important and possibly less convenient

packages/cli/src/commands/test-v2.ts
```diff
@@ -1,15 +1,69 @@
...
+const listenForKeyPressesOrEvents = (keys: string[], exit: string, events: EventEmitter) => {
+  return new Promise<{ keyPressed?: string; filesChanged: string[] }>((resolve, reject) => {
+    const removeListenersAndResolve = (returnValue: { keyPressed?: string; filesChanged: string[] }) => {
+      events.removeAllListeners();
+      resolve(returnValue);
+    };
+
+    events.on(Events.TESTS_CHANGED, (filesChanged) => removeListenersAndResolve({ filesChanged }));
+    events.on(Events.SOURCES_CHANGED, (filesChanged) => removeListenersAndResolve({ filesChanged }));
+    events.on(Events.CONFIG_CHANGED, (filesChanged) => removeListenersAndResolve({ filesChanged }));
```
holic, line 45: I know we're removing these immediately, but I tend to use `.once` for cases like these just to show intent and to reduce the chance of memory leaks if this behavior changes in the future

holic (commented):
I wonder if we should separate the meat of the test runner (run codegen, deploy world, write address file, etc.) into its own thing that we can call either 1) once or 2) repeatedly during watch mode

holic (approved):
whoops, meant for that to be an approval (none of my comments are blocking)

packages/cli/src/utils/foundry.ts
```diff
@@ -1,6 +1,12 @@
 import { execa, Options } from "execa";
 import { execLog } from "./execLog.js";
 
+export const IGNORED_ERROR_CODES = [2018 /** state mutability can be restricted to pure */];
+export const FOUNDRY_OPTIONS_IGNORE_CODES = IGNORED_ERROR_CODES.flatMap((errorCode) => [
+  "--ignored-error-codes",
+  errorCode.toString(),
+]);
```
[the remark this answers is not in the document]
dk1a: not really, it requires either a significant complexity increase in tablegen to keep tract of what's pure/view (I'd rather avoid that), or wait till solidity rolls out a `MCOPY` (which will be great and reduce gas costs to boot but not soon)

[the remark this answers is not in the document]
holic: it's strange that `forge fmt` or similar linters couldn't autofix these, then we could just run this through the linter to fix up things after codegen

alvrs:
gonna turn this into a draft for now to clean up some more based on the comments above

alvrs:
Gonna close this with a heavy heart because it diverted quite a bit from main and we align the hot reloading logic for this one with `dev-contracts` (probably rip it out into a shared file watcher script). Added an issue to track here https://github.com/latticexyz/mud/pull/570, contributors welcome!

Issue #559, opened 2023-04-01
Add error signatures to System, IWorldCore


holic:
We're missing these on the worldgen ABI, so it's ~impossible to parse custom errors that come from our Store/World internals.

PR #573, opened 2023-04-02
feat: use IErrors in IStore and IWorldCore

dk1a:
fixes https://github.com/latticexyz/mud/issues/559

Since systems are called via IWorld, I assume `System` doesn't need to inherit `IErrors`?

### DIFF
--- a/packages/store/src/IErrors.sol
+++ b/packages/store/src/IErrors.sol
@@ -1,0 +1,13 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
+
+interface IErrors {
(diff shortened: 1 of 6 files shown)


holic:
> fixes #559
> 
> Since systems are called via IWorld, I assume `System` doesn't need to inherit `IErrors`?

in autonomous mode, isn't a system a store? or is a store deployed separately?

dk1a:
> or did we end up centralizing all storage on the world and controlling it with access/namespace stuff?

^ this
Store could be used standalone, but then systems just aren't a thing because they need World

holic (approved):

Issue #560, opened 2023-04-01
Store.get() on a non-existent record where schema has an array fails


holic:
My table config:
```
    Encounter: {
      dataStruct: false,
      primaryKeys: {
        player: "bytes32",
      },
      schema: {
        actionCount: "uint256",
        monsters: "bytes32[]",
      },
    },
```

The table function that fails when called
```
  function get(bytes32 player) internal view returns (uint256 actionCount, bytes32[] memory monsters) {
    bytes32[] memory _primaryKeys = new bytes32[](1);
    _primaryKeys[0] = bytes32((player));

    bytes memory _blob = StoreSwitch.getRecord(_tableId, _primaryKeys, getSchema());
    return decode(_blob);
  }
```

I suspect it has something to do inside `decode`:
```
  /** Decode the tightly packed blob using this table's schema */
  function decode(bytes memory _blob) internal view returns (uint256 actionCount, bytes32[] memory monsters) {
    // 32 is the total byte length of static data
    PackedCounter _encodedLengths = PackedCounter.wrap(Bytes.slice32(_blob, 32));

    actionCount = (uint256(Bytes.slice32(_blob, 0)));

    uint256 _start;
    uint256 _end = 64;

    _start = _end;
    _end += _encodedLengths.atIndex(0);
    monsters = (SliceLib.getSubslice(_blob, _start, _end).decodeArray_bytes32());
  }
```

This doesn't seem to be an issue when loading each field individually.

PR #574, opened 2023-04-02
fix(cli): account for getRecord's trimming

dk1a:
fixes https://github.com/latticexyz/mud/issues/560
replaces #564 

the gist of the issue was: in `getRecord` when dynamic fields are empty, Store trims their 32-byte length word, so the blob is only static data.
`tablegen` didn't account for the trimming, and `Slice`'s sanity check detected the invalid subslice [64:64] on a blob of length 32
(an empty subslice is fine, what's invalid is that the indexes exceed the length)

### DIFF
--- a/packages/cli/contracts/src/tables/Dynamics.sol
+++ b/packages/cli/contracts/src/tables/Dynamics.sol
@@ -73,10 +73,21 @@ library Dynamics {
     StoreSwitch.registerSchema(_tableId, getSchema(), getKeySchema());
   }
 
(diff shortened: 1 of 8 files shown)


holic (approved):

holic:
does this replace the updated autogen'ed files in https://github.com/latticexyz/mud/pull/564?

dk1a:
> does this replace the updated autogen'ed files in #564?

yes, forgot to mention, this does make #564 unnecessary

PR #575, opened 2023-04-02
fix(world): give World contract access to root namespace

alvrs:
- The World didn't have access to the root namespace. This doesn't matter for situations where the World writes to its own storage, but it matters for root systems who want to call other root systems via the World. Since root systems are delegatecalled from the World, `address(this)` in the System is the world address. If these systems try to call a method on the World, it's considered an external call, so access control checks apply, and fail if the World doesn't have access to the root namespace.


ludns (approved):
lgtm

PR #576, opened 2023-04-03
feat: use viem when creating burner wallet

holic:
see #582


packages/std-client/src/getBurnerWallet.ts
```diff
@@ -1,16 +1,16 @@
-import { Wallet } from "ethers";
+import { generatePrivateKey, privateKeyToAccount } from "viem/accounts";
+import { isHex, Hex } from "viem";
 import { BehaviorSubject } from "rxjs";
 
-export function getBurnerWallet(cacheKey = "mud:burnerWallet"): BehaviorSubject<string> {
+export function getBurnerWallet(cacheKey = "mud:burnerWallet"): BehaviorSubject<Hex> {
   const cachedPrivateKey = localStorage.getItem(cacheKey);
-  const subject = cachedPrivateKey
+  const subject = isHex(cachedPrivateKey)
```
holic, line 7: are we okay with this being a destructive action? it used to just add a burner wallet if no value was set for the given key, but now it checks if it's hex and will recreate one if not

which means if someone has a different value stored at this key, we will end up overriding it with a new burner wallet private key

alvrs: I think throwing if it's an invalid value would be better bc it would tell users who are trying to override the burner wallet (but put an invalid value) why it fails

alvrs (dismissed):
lgtm 👍 (just one comment about throwing)

Issue #335, opened 2023-01-10
Ensure mud packages can be installed with npm, pnpm, etc.


holic:
While I was working on [create-mud](https://github.com/latticexyz/mud/pull/336), I noticed an `npm install` failed due to missing peer deps.

holic:
I think the errors I was seeing before is just for missing peer deps, and can be worked around with `--force` or `--legacy-peer-deps`. To help a bit with the peer dep install, I've aligned versions in #577 to hopefully reduce the chances of this error.

Gonna close this for now.

PR #577, opened 2023-04-03
feat: align git dep versions

holic:
These were a bit inconsistent, which might be the source of some issues related to https://github.com/latticexyz/mud/issues/335, but otherwise should help with weird edge cases.

Ideally we'd have a linter for this sort of thing but I couldn't find any off-the-shelf things for this.


packages/cli/package.json
```diff
@@ -63,7 +63,6 @@
     "path": "^0.12.7",
     "prettier": "^2.8.4",
     "prettier-plugin-solidity": "^1.1.2",
-    "solmate": "https://github.com/Rari-Capital/solmate.git#9cf1428245074e39090dceacb0c28b1f684f584c",
```
holic, line 66: This isn't used in CLI directly, but the CLI does use solecs, which has this as a peer dep (among others that aren't included here), so I'm unsure if this is actually needed.

cc @alvrs @dk1a

dk1a: I'm pretty sure solmate is used exclusively for DSTestPlus in solecs tests, so it's fine to remove in all other packages

alvrs: let's remove it!

alvrs (dismissed):

alvrs (approved):

PR #578, opened 2023-04-03
chore: add engines to root package.json

holic:
Specify node, pnpm, yarn versions in prep for pnpm migration

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,4 +1,5 @@
 {
+  "private": true,
   "name": "mud",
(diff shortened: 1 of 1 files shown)


alvrs:
I think we have to update the node version in our actions to 18 too to make the CI pass

alvrs:
there are also a couple `.nvmrc` files in the repo, we should probably update ( or remove?) those too

alvrs (approved):

PR #581, opened 2023-04-03
feat(network,recs,std-client): support StoreSetField before StoreSetRecord

### LINKED ISSUE #479: decide on StoreSetRecord vs StoreSetField for single value tables
Currently single value tables autogenerate their `.set(value)` call to use `store.setField`. This broke an assumption I had in https://github.com/latticexyz/mud/pull/415/files#diff-a6b694ab744d9808772b7fb1cfc0a286ee20f371069cf905c4e1af8b70bf5d68

I can work around this in the client to check if a `StoreSetField` event actually has all the data defined by the schema (and thus is equivalent to a `StoreSetRecord`), but I am wondering if this change would be more "correct" by doing it at the autogen level.

### LINKED ISSUE #523: Support setField before setRecord in MUD and MODE
We can't assume a setRecord before setField (as per convo in #479); and hence should add support for creating a default row before setting a field in MODE and MUD `network`. 
We should probably put the default values of each Solidity type in `schema-type` for now (and for MODE in the mode package before we move that stuff to a `go` folder in the `schema-type` package).

holic:
closes #479
closes #523

related to #582


alvrs (approved):
lgtm!

Issue #489, opened 2023-03-15
Update world/store to use `mud tablegen` rather than local mud


holic:
May need to update CI to link mud CLI locally.

dk1a:
fixed in https://github.com/latticexyz/mud/pull/679 by using tablegen and worldgen via scripts, which avoids both local linking and dependency on cli

PR #583, opened 2023-04-04
feat(world): add KeysInTable module

yonadaaa:
Adds a new module, `KeysWithTable`, that indexes all of the keys in a given table on-chain. This is similar to [KeysWithValue](https://github.com/latticexyz/mud/pull/487), except that it is per-table instead of per-table-and-value. Similarly, the implementation is naive, but this reimplements important missing functionality from MUD V1.

This PR also updates the gas-report scripts to use local MUD, as the MUD CLI is used inconsistently in this repo, ie. `mud` vs `../cli/dist/mud.js`. However, this conflicts with an [existing issue](https://github.com/latticexyz/mud/issues/489), so would be good to hear your thoughts.


yonadaaa:
Sure @dk1a I would be happy to generalise this and implement KeySet instead. 

To be clear, is `KeySet` a "set intersection" operator that returns an array of keys that exist in ALL of the given tables, or a "union" that returns keys that exist is ANY of the given tables?

dk1a:
Though I could also see a usecase for "intersection" in very ECS-centric apps where sets of components are used to define entities

alvrs:
Nit: should we call this `KeysInTable` or `TableKeys` instead of `KeysWithTable`? (`KeysWithTable` makes sense when thinking of it in terms of ECS, where a "table is attached to a key", but in the general sense of tables I find it confusing)

packages/world/src/modules/keyswithtable/KeysWithTableModule.sol
```diff
@@ -0,0 +1,69 @@
...
+ * provided in the install methods arguments. The hook keeps track of the keys that are in a given table.
+ * This mapping is stored in a table registered by the module at the `targetTableId` provided in the
+ * install methods arguments.
+ *
+ * Note: if a table with composite keys is used, only the first key is indexed
+ *
+ * Note: this module currently expects to be `delegatecalled` via World.installRootModule.
+ * Support for installing it via `World.installModule` depends on `World.callFrom` being implemented.
+ */
+contract KeysWithTableModule is IModule, WorldContext {
```
holic, line 29: I realize `KeysWith` is probably trying to align naming with `KeysWithValue` but I feel like it's a little wonky for a table (since a table is "higher level" than keys/values).

Some other name ideas:
- `KeysInTable`
- `KeysForTable`
- `TableKeys`

dk1a:
> As to a table union vs intersection operator, intersection would be most useful for me with ECS games ie. "Get all entities that are in the Player table AND the Position Table". This replicate the [on-chain queries](https://mud.dev/packages/solecs/#queries) from MUD1, but perhaps we can make both.

I hadn't considered replacing queries with static indexes, yeah that could be useful and being able to choose union/intersection would be good imo

yonadaaa:
I think this is good to go! But there are some nice to haves:

- Adding a `.length` view for array tables (now the module maintains that itself, duplicating the state)
- Adding a `getAtIndex` view for array tables (now the hook reads the whole table from memory then indexes it)

packages/store/gas-report.txt
```diff
@@ -0,0 +1,566 @@
+[
```
dk1a, line 1: I assume this is a leftover from before gas-report got changed, you should delete this file, it's `.json` now
(same for the one in `world`)

packages/world/mud.config.ts
```diff
@@ -99,6 +99,26 @@ export default mudConfig({
...
+      primaryKeys: {},
+      schema: "uint256",
+      tableIdArgument: true,
+    },
+    UsedKeysIndex: {
+      directory: "/modules/keysintable/tables",
+      primaryKeys: {
+        keysHash: "bytes32",
+      },
+      schema: { has: "bool", index: "uint256" },
```
dk1a, line 119: `uint32` should be enough, MUD2 arrays are limited anyways and this will fit both cols into 1 slot

packages/world/src/modules/keysintable/KeysInTableHook.sol
```diff
@@ -0,0 +1,80 @@
...
+
+  function onDeleteRecord(bytes32 tableId, bytes32[] memory key) public {
+    bytes32 keysInTableTableId = getTargetTableSelector(MODULE_NAMESPACE, tableId);
+    bytes32 keysInTableLengthTableId = getTargetTableSelector(KEYS_LENGTH_NAMESPACE, tableId);
+    bytes32 usedIndexTableId = getTargetTableSelector(USED_KEYS_NAMESPACE, tableId);
+
+    bytes32 keysHash = keccak256(abi.encode(key));
+    UsedKeysIndexData memory data = UsedKeysIndex.get(usedIndexTableId, keysHash);
+
+    // If the key not has yet been set in the table...
```
dk1a, line 65:
```suggestion
    // If the key has not been set in the table...
```

dk1a (approved):
lgtm, just some minor nits
(assuming composite keys and length optimizations will come later)

yonadaaa:
Thanks for the comments @dk1a, this is ready for another review!

packages/world/gas-report.json
```diff
@@ -1,4 +1,10 @@
 [
+  {
+    "source": "test/KeysInTableModule.t.sol",
+    "name": "compute the target table selector",
+    "functionCall": "bytes32 targetTableSelector = getTargetTableSelector(MODULE_NAMESPACE, sourceTableId)",
+    "gasUsed": 2240
+  },
```
alvrs, line 7: should we add another gas report for setting a value on a table with this module installed, and for getting the keys from a table? (Should be as easy as adding the `!gasreport` comment in the existing tests in `KeysInTableModule.t.sol`)

packages/world/mud.config.ts
```diff
@@ -99,6 +99,26 @@ export default mudConfig({
...
+      primaryKeys: {},
+      schema: "bytes32[]", // For now only supports 1 key per value,
+      tableIdArgument: true,
+    },
+    KeysInTableLength: {
+      directory: "/modules/keysintable/tables",
+      primaryKeys: {},
+      schema: "uint32",
+      tableIdArgument: true,
+    },
```
alvrs, line 113: Why don't we combine these two to a single table with schema `{ keys: "bytes32[]", length: "uint32" }`?

packages/world/src/modules/keysintable/KeysInTableHook.sol
```diff
@@ -0,0 +1,80 @@
...
+ *
+ * Note: if a table with composite keys is used, only the first key is indexed
+ */
+contract KeysInTableHook is IStoreHook {
+  using ArrayLib for bytes32[];
+  using ResourceSelector for bytes32;
+
+  function handleSet(bytes32 tableId, bytes32[] memory key) internal {
+    bytes32 keysInTableTableId = getTargetTableSelector(MODULE_NAMESPACE, tableId);
+    bytes32 keysInTableLengthTableId = getTargetTableSelector(KEYS_LENGTH_NAMESPACE, tableId);
```
alvrs, line 29: instead of using a different singleton table per source table to index, why don't we use a single `KeysInTableTable` with a `{ sourceTable: "bytes32" }` primary key?

packages/world/src/modules/keysintable/KeysInTableHook.sol
```diff
@@ -0,0 +1,80 @@
...
+    bytes32 usedIndexTableId = getTargetTableSelector(USED_KEYS_NAMESPACE, tableId);
```
alvrs, line 30: same here, why not use a single UsedIndex table with `{ sourceTable: "bytes32", keyHash: "bytes32" }` primary key?

packages/world/src/modules/keysintable/KeysInTableHook.sol
```diff
@@ -0,0 +1,80 @@
...
+
+    bytes32 keysHash = keccak256(abi.encode(key));
+
+    // If the key not has yet been set in the table...
```
alvrs, line 34:
```suggestion
    // If the key has not yet been set in the table...
```

packages/world/src/modules/keysintable/KeysInTableHook.sol
```diff
@@ -0,0 +1,80 @@
...
+    if (!UsedKeysIndex.getHas(usedIndexTableId, keysHash)) {
+      uint32 len = KeysInTableLength.get(keysInTableLengthTableId);
```
alvrs, line 36: Nit: let's avoid abbreviations if possible
```suggestion
      uint32 length = KeysInTableLength.get(keysInTableLengthTableId);
```

packages/world/src/modules/keysintable/KeysInTableHook.sol
```diff
@@ -0,0 +1,80 @@
...
+    handleSet(table, key);
+  }
+
+  function onDeleteRecord(bytes32 tableId, bytes32[] memory key) public {
+    bytes32 keysInTableTableId = getTargetTableSelector(MODULE_NAMESPACE, tableId);
+    bytes32 keysInTableLengthTableId = getTargetTableSelector(KEYS_LENGTH_NAMESPACE, tableId);
+    bytes32 usedIndexTableId = getTargetTableSelector(USED_KEYS_NAMESPACE, tableId);
+
+    bytes32 keysHash = keccak256(abi.encode(key));
+    UsedKeysIndexData memory data = UsedKeysIndex.get(usedIndexTableId, keysHash);
```
alvrs, line 63: For simple tables like this I have a slight style preference to set `structArgument` to false in the config and then work with tuples directly - seems easier to read:

```suggestion
    (bool has, uint32 index) = UsedKeysIndex.get(usedIndexTableId, keysHash);
```

No strong opinion though, feel free to ignore

packages/world/src/modules/keysintable/KeysInTableHook.sol
```diff
@@ -0,0 +1,80 @@
...
+
+    // If the key has not been set in the table...
+    if (data.has) {
+      uint32 len = KeysInTableLength.get(keysInTableLengthTableId);
+      bytes32 lastKey = KeysInTable.get(keysInTableTableId)[len - 1];
+
+      // Remove the key from the list of keys in this table
+      KeysInTable.update(keysInTableTableId, data.index, lastKey);
+      KeysInTable.pop(keysInTableTableId);
```
alvrs, line 72: You should be able to simplify this to:

```suggestion
      // Remove the key from the list of keys in this table
      bytes32 lastKey = KeysInTable.pop(keysInTableTableId);
      KeysInTable.update(keysInTableTableId, data.index, lastKey);
```

packages/world/src/modules/keysintable/KeysInTableHook.sol
```diff
@@ -0,0 +1,80 @@
...
+import { KeysInTableLength } from "./tables/KeysInTableLength.sol";
+import { UsedKeysIndex, UsedKeysIndexData } from "./tables/UsedKeysIndex.sol";
+import { ArrayLib } from "../utils/ArrayLib.sol";
+import { getTargetTableSelector } from "../utils/getTargetTableSelector.sol";
+
+/**
+ * This is a very naive and inefficient implementation for now.
+ * We can optimize this by adding support for `setIndexOfField` in Store
+ * and then replicate logic from solecs's Set.sol.
+ * (See https://github.com/latticexyz/mud/issues/444)
```
alvrs, line 19: This comment seems to be out of date

packages/world/src/modules/keysintable/KeysInTableHook.sol
```diff
@@ -0,0 +1,80 @@
...
+    bytes32 keysInTableTableId = getTargetTableSelector(MODULE_NAMESPACE, tableId);
+    bytes32 keysInTableLengthTableId = getTargetTableSelector(KEYS_LENGTH_NAMESPACE, tableId);
+    bytes32 usedIndexTableId = getTargetTableSelector(USED_KEYS_NAMESPACE, tableId);
+
+    bytes32 keysHash = keccak256(abi.encode(key));
+    UsedKeysIndexData memory data = UsedKeysIndex.get(usedIndexTableId, keysHash);
+
+    // If the key has not been set in the table...
+    if (data.has) {
+      uint32 len = KeysInTableLength.get(keysInTableLengthTableId);
```
alvrs, line 67:
```suggestion
      uint32 length = KeysInTableLength.get(keysInTableLengthTableId);
```

packages/world/src/modules/keysintable/KeysInTableModule.sol
```diff
@@ -0,0 +1,91 @@
...
+    world.grantAccess(targetTableSelector.getNamespace(), targetTableSelector.getName(), address(hook));
+  }
+
+  function install(bytes memory args) public override {
+    // Extract source table id from args
+    bytes32 sourceTableId = abi.decode(args, (bytes32));
+
+    bytes32 targetTableSelector1 = getTargetTableSelector(MODULE_NAMESPACE, sourceTableId);
+    bytes32 targetTableSelector2 = getTargetTableSelector(KEYS_LENGTH_NAMESPACE, sourceTableId);
+    bytes32 targetTableSelector3 = getTargetTableSelector(USED_KEYS_NAMESPACE, sourceTableId);
```
alvrs, line 52: if we change the tables from singleton to use the target table ids (see comment above) we don't need to register a new table for every source table here

packages/world/src/modules/keysintable/KeysInTableModule.sol
```diff
@@ -0,0 +1,91 @@
...
+    (string memory tableName1, string[] memory fieldNames1) = KeysInTable.getMetadata();
+    world.setMetadata(targetTableSelector1.getNamespace(), targetTableSelector1.getName(), tableName1, fieldNames1);
+    (string memory tableName2, string[] memory fieldNames2) = KeysInTableLength.getMetadata();
+    world.setMetadata(targetTableSelector2.getNamespace(), targetTableSelector2.getName(), tableName2, fieldNames2);
+    (string memory tableName3, string[] memory fieldNames3) = UsedKeysIndex.getMetadata();
+    world.setMetadata(targetTableSelector3.getNamespace(), targetTableSelector3.getName(), tableName3, fieldNames3);
+
+    grantAccess(world, targetTableSelector1);
+    grantAccess(world, targetTableSelector2);
+    grantAccess(world, targetTableSelector3);
```
alvrs, line 86: why is this needed?

packages/world/src/modules/keysintable/getKeysInTable.sol
alvrs, line 1: should we add a util for `function hasKey(bytes32[] key) view returns (bool)` as a wrapper around `UsedKeysIndex`?

packages/world/test/KeysInTableModule.t.sol
```diff
@@ -0,0 +1,200 @@
...
+
+  function setUp() public {
+    sourceTableSchema = SchemaLib.encode(SchemaType.UINT256);
+    sourceTableKeySchema = SchemaLib.encode(SchemaType.BYTES32);
+    world = IBaseWorld(address(new World()));
+    world.installRootModule(new CoreModule(), new bytes(0));
+    keyTuple1 = new bytes32[](1);
+    keyTuple1[0] = key1;
+    keyTuple2 = new bytes32[](1);
+    keyTuple2[0] = key2;
```
alvrs, line 46: should we add a test for a key tuple with 2+ entries? (with the expectation of `hasTable` working for the full key, but `getKeysInTable` returning an array of the first key only)

packages/world/test/KeysInTableModule.t.sol
```diff
@@ -0,0 +1,200 @@
...
+    // Assert that the list is correct
+    assertEq(keysInTable.length, 2, "5");
+    assertEq(keysInTable[1], key2, "6");
+
+    // Get the list of keys in the target table
+    keysInTable = KeysInTable.get(world, targetTableId);
+
+    // Assert that the list is correct
+    assertEq(keysInTable.length, 2, "7");
+    assertEq(keysInTable[0], key1, "8");
```
alvrs, line 116: why getting the `keysInTable` twice here?

packages/world/test/KeysInTableModule.t.sol
```diff
@@ -0,0 +1,200 @@
...
+
+    // Assert that the list is correct
+    assertEq(keysInTable.length, 1);
+
+    // Get the list of keys in the target table
+    keysInTable = KeysInTable.get(world, targetTableId);
+
+    // Assert that the list is correct
+    assertEq(keysInTable.length, 1);
+    assertEq(keysInTable[0], key1);
```
alvrs, line 159: same here, getting `keysInTable` once should be sufficient

packages/world/test/KeysInTableModule.t.sol
```diff
@@ -0,0 +1,200 @@
...
+    assertEq(bytes8(targetTableSelector), MODULE_NAMESPACE);
+
+    // followed by the first 4 bytes of the source table namespace
+    assertEq(bytes8(targetTableSelector << 64), bytes8(namespace));
+
+    // The last 16 bytes are the source file
+    assertEq(targetTableSelector.getName(), sourceFile);
+  }
+
+  function testGetkeysInTable(uint256 value1) public {
```
alvrs, line 176: typo
```suggestion
  function testGetKeysInTable(uint256 value1) public {
```

packages/world/test/KeysInTableModule.t.sol
```diff
@@ -0,0 +1,200 @@
...
+
+    // !gasreport Get list of keys in a given table
+    bytes32[] memory keysInTable = getKeysInTable(world, sourceTableId);
+
+    // Assert that the list is correct
+    assertEq(keysInTable.length, 1);
+    assertEq(keysInTable[0], key1);
+
+    // Set another key with the same value
+    world.setRecord(namespace, sourceFile, keyTuple2, abi.encodePacked(value1));
```
alvrs, line 190: Nit - I feel like setting a different value here would represent the purpose of this module slightly better (if we set the same value the difference to `KeysWithValue` is a bit less clear)

packages/world/mud.config.ts
dk1a: I was thinking that since `KeysInTableLength` is a workaround until there's a Store-native length getter, this way is easier to refactor

packages/world/src/modules/keysintable/KeysInTableHook.sol
yonadaaa: I wanted to do that, but `pop()` doesn't currently return the key.

packages/world/mud.config.ts
yonadaaa: Exactly. I have unified the tables for now anyway

packages/world/src/modules/keysintable/KeysInTableHook.sol
yonadaaa: I copied this pattern from `KeysWithValue`. Is there any reason for that module to do it differently?

packages/world/src/modules/keysintable/KeysInTableModule.sol
yonadaaa: This grants the hook access to the target tables - without it I get an `AccessDenied` error.

packages/world/src/modules/keysintable/getKeysInTable.sol
yonadaaa: Great idea! This is useful by itself

packages/world/test/KeysInTableModule.t.sol
yonadaaa: The flags an important point; the API of `getKeysInTable` is wrong. It only returns an array of `bytes32` (single keys instead of a composite array).

packages/world/mud.config.ts
alvrs: > I was thinking that since KeysInTableLength is a workaround until there's a Store-native length getter, this way is easier to refactor

Oh I see! That makes sense, but also fine with still unifying for now

packages/world/src/modules/keysintable/KeysInTableHook.sol
alvrs: > I copied this pattern from KeysWithValue. Is there any reason for that module to do it differently?

context from discord: it probably should be a single table for `KeysWithValue` too.

> In that case, they have a singleton table of the values in each source table, but it could be { sourceTable: "bytes32", keyHash: "bytes32" } 

You mean `{ sourceTable: "bytes32", valueHash: "bytes32" }` as the primary key for the table in `KeysWithValue`? If so, I agree!

packages/world/src/modules/keysintable/KeysInTableModule.sol
alvrs: hm interesting, the world should have access to everything by default (see #609)

packages/world/src/modules/keysintable/KeysInTableHook.sol
```diff
@@ -0,0 +1,58 @@
...
+
+  function onAfterSetField(bytes32 table, bytes32[] memory key, uint8, bytes memory) public {
+    handleSet(table, key);
+  }
+
+  function onDeleteRecord(bytes32 tableId, bytes32[] memory key) public {
+    bytes32 keysHash = keccak256(abi.encode(key));
+    (bool has, uint32 index) = UsedKeysIndex.get(tableId, keysHash);
+
+    // If the key has not been set in the table...
```
alvrs, line 43:
```suggestion
    // If the key was part of the table...
```

packages/world/src/modules/keysintable/KeysInTableHook.sol
```diff
@@ -0,0 +1,58 @@
...
+    if (has) {
+      uint32 length = KeysInTable.getLength(tableId);
+      bytes32 lastKey = KeysInTable.getKeys(tableId)[length - 1];
+
+      // Remove the key from the list of keys in this table
+      KeysInTable.updateKeys(tableId, index, lastKey);
+      KeysInTable.popKeys(tableId);
```
alvrs, line 50: let's add a comment here with a link to an issue that tracks that `.pop` should return the popped field so we're reminded to change this once that ships

alvrs: Related to a recent discussion on discord: should we call `deleteRecord` here if `length` was 1 (so is 0 now)? This would slightly reduce garbage data (empty arrays for tables with no keys) on indexers/clients. It's less important here than for `KeysWithValue` though, because there is a much smaller number of tables compared to every possible value in a table.

packages/world/src/modules/keysintable/KeysInTableModule.sol
```diff
@@ -0,0 +1,77 @@
...
+
+import { KeysInTableHook } from "./KeysInTableHook.sol";
+import { KeysInTable, KeysInTableTableId } from "./tables/KeysInTable.sol";
+import { UsedKeysIndex, UsedKeysIndexTableId } from "./tables/UsedKeysIndex.sol";
+
+/**
+ * This module deploys a hook that is called when a value is set in the `sourceTableId`
+ * provided in the install methods arguments. The hook keeps track of the keys that are in a given table.
+ * This mapping is stored in a table registered by the module at the `targetTableId` provided in the
+ * install methods arguments.
```
alvrs, line 22: This part of the comment seems to be outdated

```suggestion
 * This mapping is stored in a global table that is registered when the module is first installed.
```

packages/world/src/modules/keysintable/KeysInTableModule.sol
```diff
@@ -0,0 +1,77 @@
...
+ */
+contract KeysInTableModule is IModule, WorldContext {
+  using ResourceSelector for bytes32;
+
+  // The KeysInTableHook is deployed once and infers the target table id
+  // from the source table id (passed as argument to the hook methods)
+  KeysInTableHook immutable hook = new KeysInTableHook();
+
+  function getName() public pure returns (bytes16) {
+    return bytes16("index");
```
alvrs, line 37: same for `KeysWithValue`, but this name should be something more unique than `index`. It's used to track the installed modules in `World.installModule` / `World.installRootModule`, so right now we'd have a clash between `KeysWithValue` and `KeysInTable` if they're called with the same arguments

alvrs, line 37: same for `KeysWithValue`, but this name should be something more unique than `index`. It's used to track the installed modules in `World.installModule` / `World.installRootModule`, so right now we'd have a clash between `KeysWithValue` and `KeysInTable` if they're called with the same arguments

packages/world/src/modules/keysintable/KeysInTableModule.sol
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  function install(bytes memory args) public override {
+    // Extract source table id from args
+    bytes32 sourceTableId = abi.decode(args, (bytes32));
+
+    IBaseWorld world = IBaseWorld(_world());
+
+    bytes32 resourceSelector = ResourceSelector.from(KeysInTableTableId.getNamespace(), KeysInTableTableId.getName());
```
alvrs, line 46: The `KeysInTableTableId` itself is equivalent with the resource selector already, so we can just use that here

packages/world/src/modules/keysintable/KeysInTableModule.sol
```diff
@@ -0,0 +1,77 @@
...
+      (string memory tableName2, string[] memory fieldNames2) = UsedKeysIndex.getMetadata();
+      world.setMetadata(UsedKeysIndexTableId.getNamespace(), UsedKeysIndexTableId.getName(), tableName2, fieldNames2);
+
+      // Grant the hook access to the tables
+      world.grantAccess(KeysInTableTableId.getNamespace(), KeysInTableTableId.getName(), address(hook));
+      world.grantAccess(UsedKeysIndexTableId.getNamespace(), UsedKeysIndexTableId.getName(), address(hook));
+    }
+
+    // Register a hook that is called when a value is set in the source table
+    StoreSwitch.registerStoreHook(sourceTableId, hook);
```
alvrs, line 75: tiny nit, but since we call the other methods above on `world`, I'd call `registerStoreHook` on `world` as well for consistency

packages/world/src/modules/keysintable/getKeysInTable.sol
```diff
@@ -0,0 +1,61 @@
...
+}
+
+/**
+ * Get whether the key is in the given table for the given store.
+ */
+function hasKey(IStore store, bytes32 tableId, bytes32[] memory key) view returns (bool) {
+  bytes32 keysHash = keccak256(abi.encode(key));
+
+  return UsedKeysIndex.getHas(store, tableId, keysHash);
+}
```
alvrs, line 61: nit - we should either move these utils to a `hasKey.sol` file, or rename this file to `utils.sol`

packages/world/src/modules/keysintable/getKeysInTable.sol
```diff
@@ -0,0 +1,61 @@
...
+function getKeysInTable(bytes32 tableId) view returns (bytes32[][] memory keysInTable) {
+  // Get the keys with the given table
+  bytes32[] memory keysInTableRaw = KeysInTable.getKeys(tableId);
+
+  keysInTable = new bytes32[][](keysInTableRaw.length);
+
+  for (uint256 i; i < keysInTableRaw.length; i++) {
+    keysInTable[i] = new bytes32[](1);
+    keysInTable[i][0] = keysInTableRaw[i];
+  }
```
alvrs, line 24: let's add a comment here about this being temporary until we support composite keys in `KeysInTable` (otherwise without context this operation is confusing)

packages/world/test/KeysInTableModule.t.sol
```diff
@@ -0,0 +1,246 @@
...
+
+    // The first 8 bytes are the module namespace
+    assertEq(bytes8(targetTableSelector), MODULE_NAMESPACE);
+
+    // followed by the first 4 bytes of the source table namespace
+    assertEq(bytes8(targetTableSelector << 64), bytes8(namespace));
+
+    // The last 16 bytes are the source file
+    assertEq(targetTableSelector.getName(), sourceFile);
+  }
```
alvrs, line 196: we can probably remove this test from here now

packages/world/mud.config.ts
```diff
@@ -99,6 +99,20 @@ export default mudConfig({
       },
       tableIdArgument: true,
     },
+    KeysInTable: {
+      directory: "/modules/keysintable/tables",
+      primaryKeys: { sourceTable: "bytes32" },
+      schema: { length: "uint32", keys: "bytes32[]" }, // For now only supports 1 key per value,
```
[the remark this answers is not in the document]
dk1a: made one #735

packages/world/gas-report.json
alvrs: Okay figured out why the other gas reports don't show up here: it's because forge doesn't print logs for fuzz tests (gas-report relies on console logs to parse the gas usage) and this is the only non-fuzz test. I think fuzz tests are nice, but it would also be useful to track the gas cost of this module for optimizations in the future, so maybe we should add a couple non-fuzz test instances for gas-report purposes?

packages/world/src/modules/keysintable/KeysInTableHook.sol
```diff
@@ -42,17 +42,23 @@ contract KeysInTableHook is IStoreHook {
...
 
-      KeysInTable.setLength(tableId, length - 1);
+        // Remove the key from the list of keys in this table
+        KeysInTable.updateKeys(tableId, index, lastKey);
+        KeysInTable.popKeys(tableId);
 
-      // Delete the index as the key is not in the table
-      UsedKeysIndex.deleteRecord(tableId, keysHash);
+        KeysInTable.setLength(tableId, length - 1);
+      }
```
holic, line 61: :chefkiss:

yonadaaa: no memories leaks for us

packages/world/gas-report.json
yonadaaa: Thanks for figuring this out, it stumped me. I will add a couple normal tests.

packages/std-client/src/mud-definitions/world/contractComponents.ts
```diff
@@ -113,5 +113,37 @@ export function defineContractComponents(world: World) {
...
+          index: RecsType.Number,
+        },
+        {
+          metadata: {
+            contractId: tableId.toHexString(),
+            tableId: tableId.toString(),
+          },
+        }
+      );
+    })(),
```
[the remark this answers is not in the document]
yonadaaa: As discussed in Discord, these are produced because codegen doesn't differentiate between MUD internals and modules. The other modules are ignored because codegen skips tables with `tableIdArgument`.

For now, I think we should leave these and (possibly) raise an issue for codegen to ignore them.

packages/world/gas-report.json
```diff
@@ -1,4 +1,34 @@
...
+    "source": "test/KeysInTableModule.t.sol",
+    "name": "install keys in table module",
+    "functionCall": "world.installRootModule(keysInTableModule, abi.encode(sourceTableId))",
+    "gasUsed": 1032191
+  },
+  {
+    "source": "test/KeysInTableModule.t.sol",
+    "name": "set a record on a table with keysInTableModule installed",
+    "functionCall": "world.setRecord(namespace, sourceFile, keyTuple1, abi.encodePacked(val1))",
+    "gasUsed": 239738
```
alvrs, line 12: surprisingly expensive 😱

alvrs (approved):
Let's go! 🚢 ✅

PR #584, opened 2023-04-04
feat(plugins): create plugins package

dk1a:
Stub so that boilerplate diffs and discussion don't have to be in the actual plugins PR

@roninjin10 I tried to account for #514 already

The idea with the dep chain: `schema-type` -> `store` -> `world` -> `config` -> `world-plugins` -> `config defaults`.
`cli` would be after `config`, and probably after `config defaults`.

`cli` and `config` separation is unimplemented.
`config defaults` is unimplemented, I think the ECS wrapper @ludens had in mind would fit here, as well as some more general default for any kind of tables


packages/world-plugins/.nvmrc
```diff
@@ -0,0 +1 @@
+17
```
alvrs, line 1: we're moving to 18 in https://github.com/latticexyz/mud/pull/578, should account for that here already

alvrs (dismissed):
good call on moving this to a separate PR!

ludns:
> I think the ECS wrapper @ludens had in mind would fit here

so would the ECS plugin basically hook into the default config and change stuff there when imported?
I specifically don't think we should have ECS stuff by default, but if an import is found in the config file (eg: `mud/ecs`) then it should be able to set a default config, do config rewriting (ie: go from `components` -> `tables` with some rewriting), and change the type of the `createStoreConfig` function (or maybe expose another entry function? maybe that would be easier).

dk1a:
> (or maybe expose another entry function? maybe that would be easier).

yeah I think a separate entry would be easier to both implement and see what's happening.
`ecsConfig` could then restructure the config and pass it to `mudConfig` to reuse its most validation

packages/world-plugins/package.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "@latticexyz/world-plugins",
```
holic, line 2: can we mark this private for now? or otherwise "hide" this from the lerna release flow so canaries can work until we have an actual package here?

holic:
what kind of world plugins are we thinking about? have we considered how the plugin API will look?

dk1a:
> what kind of world plugins are we thinking about?

current modules, tsgen, prototypes

> have we considered how the plugin API will look?

not in any details, I'll probably improvise a PR and we can go from there

alvrs:
i'd propose waiting with merging this one until we have a clearer picture of the plugin system

dk1a:
Updated this to match other `package.json`s and renamed to `plugins` because https://github.com/latticexyz/mud/pull/679 is not specific to `world`, which is also just a plugin

alvrs:
I think the order of operations should be
1. #1025
2. hooks
3. create separate plugin package (this)
4. move existing plugins out of world/store into this 

So i'm gonna temporarily close this to declutter the PR list, but generally think this is good and we can just reopen when it's time to merge it

PR #585, opened 2023-04-04
feat(cli): use a central codegen dir for tablegen and worldgen

Kooshaba:
1. there is a configurable central codegen dir where all tables and world interfaces are placed. (Default `codegen`)
2. all tables are exported in an index file `Tables.sol`
3. side effect: old tables are cleaned up before generating new ones by cleaning out all the table directories before generation


holic:
From @alvrs on the problems we're solving for:
> I think the problem here is:
> 1. we want to be able to generate tables into subdirectories to allow for any project structure -> could be solved by configuring an output dir, and creating a tables dir inside the output dir
> 2. we want to clean up the previous auto generated files -> we could parse the config file for the current output directories, but we could miss the ones for previously generated tables that have been removed from the config

Kooshaba:
i don't think generated code colocated with user code is a good idea, so im going with a separate codegen dir.

one question is do we do the one codegen command to rule them all? the order of operations is currently:
1. `tablegen`
2. `forge build`
3. `worldgen`
4. `worldtypes` and `tsgen`

i'm worried if we clear the whole codegen directory on start and then you have have forge build errors your project is borked until you solve your build errors. for now i was going to have `tablegen` clean up its known directories (like `worldgen` currently does).

the one downside of this is that changing the output directory of a table has the chance to leave around an old table directory. this seems like a small loss, just noting.

holic:
> i'm worried if we clear the whole codegen directory on start and then you have have forge build errors your project is borked until you solve your build errors
we can work around this by doing all the codegen in a temporary directory, then hot swapping

dk1a:
tablegen and worldgen should always come before forge build, tsgen can happen whenever (can even be async), worldtypes must come after

tablegen should happen first, because it can break both worldgen and forge build by changing tables that systems use.
worldgen should be before forge build because it doesn't need abi, and generates solidity files which themselves should be included in the current build, not the next one

packages/cli/src/commands/tablegen.ts
```diff
@@ -19,11 +19,11 @@ const commandModule: CommandModule<Options, Options> = {
   },
 
   async handler({ configPath }) {
-    const srcDirectory = await getSrcDirectory();
+    const codegenDir = await getCodegenDirectory();
 
     const config = await loadStoreConfig(configPath);
 
-    await tablegen(config, srcDirectory);
+    await tablegen(config, codegenDir);
```
dk1a, line 26: `src` doesn't need a config option because you can just change it in `foundry.toml`,
you can't change `codegen` though, would be nice to have it be configurable

packages/cli/src/utils/codegen.ts
```diff
@@ -0,0 +1,10 @@
+import path from "path";
+import { getForgeConfig } from "./foundry.js";
+
+/**
+ * The directory that all MUD auto-generated files are placed in.
+ */
+export async function getCodegenDirectory(profile?: string) {
+  const srcDirectory = (await getForgeConfig(profile)).src;
+  return path.join(srcDirectory, "codegen");
+}
```
dk1a, line 10: this might be better of as part of store config, rather than a hardcoded standalone utility

dk1a (commented):
tests use `tablegen` from utils, not commands. You should change `scripts/codgen` to reflect the new default so it's easier to see if problems arise.

A few nits aside I really like this and think we should merge it soon, as it's useful already. Plugins will require modifications, but that can be a separate PR

holic:
made a follow up issue to track consolidating codegen under a `mud codegen` command

packages/cli/contracts/src/Tables.sol
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
+
+/* Autogenerated file. Do not edit manually. */
```
holic, line 4: does this need to regenerated now that the codegen paths have changed?

also how would you feel about this being a `codegen.sol` entry point? then all codegen things could live here
(this is assuming that we'll keep the "all identifiers must be unique" constraint)

Kooshaba:
the `world` package had a folder structure that made sense for handwritten and generated code to sit side by side. i was able to keep this pattern by passing `codegenDir: ""` in mud config. woo flexible configs!

Kooshaba:
@holic cool if we hold off on further changes for now? feels like the plugin PR will heavily influence this

btw i tested against the game templates and everything is working

packages/store/mud.config.mts
```diff
@@ -1,7 +1,7 @@
 import { storeConfig } from "@latticexyz/cli";
 
 export default storeConfig({
-  storeImportPath: "../",
+  storeImportPath: "../../",
```
holic, line 4: is this because these things now live in `codegen`?

Kooshaba: Yep

holic (approved):

packages/cli/contracts/src/Tables.sol
dk1a: I very slightly prefer tables and types separate, but everything in codegen seems fine too.
No strong opinion on `tables` vs  `Tables` either, I'd probably go with `tables` in an empty repo, but in MUD for consistency `Tables` would fit better, unless we want to change some casings in `store`/`world` too

packages/store/mud.config.mts
dk1a: yeah that path is used for default local resolution. It pretty much exists purely to support `tablegen` within `store` package itself

packages/cli/contracts/src/Tables.sol
holic: > unless we want to change some casings in store/world too
do we have imports where the file is uppercase but doesn't export an uppercase thing with that same name?

my biggest beef with the uppercase `Tables` is I would expect it to export a `Tables` contract or something, but it doesn't, so lowercase is more clear

(this can all happen in a follow up too, doesn't need to block this PR)

packages/store/mud.config.mts
holic: added a new issue around this here https://github.com/latticexyz/mud/issues/596

dk1a (approved):

packages/cli/contracts/src/Tables.sol
dk1a: actually after looking into it not really, only `Types.sol` (for store it's not autogenerated), so we should then call it `types.sol` too

holic: now that we're calling these `enums` in the config, I wonder if the file should be `enums.sol` 

can still also re-export these from a `codegen.sol`

PR #588, opened 2023-04-05
build: refactor github actions

holic:


package.json
```diff
@@ -69,7 +69,7 @@
     "prettier": "prettier --write 'packages/**/*.ts'",
     "lint": "eslint . --ext .ts",
     "release": "yarn && lerna publish --no-private --force-publish && yarn retype:updateversion",
-    "release:ci": "yarn && lerna publish --no-private --force-publish --yes --no-verify-access && yarn retype:updateversion",
```
holic, line 72: - don't need to re-run `yarn` in ci
- `--no-verify-access` is now the default
- don't need to update retype docs version (because we don't publish docs in this ci step)

package.json
```diff
@@ -69,7 +69,7 @@
...
+    "release:ci": "lerna publish --no-private --force-publish --yes",
```
holic, line 72: - don't need to re-run `yarn` in ci
- `--no-verify-access` is now the default
- don't need to update retype docs version (because we don't publish docs in this ci step)

package.json
```diff
@@ -83,7 +83,9 @@
     "entry:dist": "yarn lerna run prepack",
     "entry:src": "yarn lerna run postpack",
     "gas-report": "yarn workspace @latticexyz/store run gas-report && yarn workspace @latticexyz/world run gas-report",
-    "tablegen": "yarn workspace @latticexyz/store run tablegen && yarn workspace @latticexyz/world run tablegen"
+    "codegen": "yarn codegen:tablegen && yarn codegen:worldgen",
+    "codegen:tablegen": "yarn workspace @latticexyz/store run tablegen && yarn workspace @latticexyz/world run tablegen",
+    "codegen:worldgen": "yarn workspace @latticexyz/world run worldgen"
```
holic, line 88: This simplifies codegen to one command for CI + us, and prepares for moving these all under one CLI command

package.json
holic: hmm, for the CI-based release, how should we handle docs updates?

I think what's missing from this is committing/pushing docs update on a manual release?

alvrs: I think it's fine for now. We'll have to find a way to have properly versioned docs soon though

alvrs (approved):

PR #589, opened 2023-04-05
feat(config): create config package

dk1a:
similar to https://github.com/latticexyz/mud/pull/584 but config has no solidity/forge


packages/config/.gitignore
```diff
@@ -0,0 +1,12 @@
+cache
+abi
+out
+dist
+types/ethers-contracts
```
alvrs, line 5: abi, out, ethers-contracts is just needed in solidity packages

packages/config/vite.config.ts
```diff
@@ -0,0 +1,23 @@
...
+  build: {
+    lib: {
+      entry: "src/index.ts",
+      formats: ["cjs", "es"],
+    },
+    outDir: "dist",
+    minify: false,
+    sourcemap: true,
+  },
+  test: {
```
roninjin10, line 18:
```suggestion
  /**
   * @see https://vitest.dev/config/
   */
  test: {
```

packages/config/.nvmrc
```diff
@@ -0,0 +1 @@
+18
```
alvrs, line 1: Don't think we need a local .nvmrc, @holic moved this to a root one recently

packages/config/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "main": "dist/index.js",
+  "module": "dist/index.mjs",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/latticexyz/mud.git",
+    "directory": "packages/config"
+  },
+  "scripts": {
+    "prepare": "yarn build",
+    "test": "echo 'TODO tests'",
```
roninjin10, line 17:
```suggestion
    "test": "vitest",
```
Could just use vitest (loooks like you were headed that way)

alvrs (approved):
lgtm, just tiny nits

packages/config/yarn.lock
```diff
@@ -0,0 +1,645 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
roninjin10, line 1: Wierd yarn generated this

roninjin10: You will want to remove this and run yarn again from root btw

dk1a: oh right I forgot to add the package to "workspaces"

packages/config/package.json
dk1a: can't do that rn - I have no actual tests and vitest errors because of it, wouldn't want to break our CI

dk1a: but yeah the plan is to do `vitest` when I make tests

holic (approved):

PR #591, opened 2023-04-05
build: yarn -> pnpm

holic:
Continued from #511 

TODO: 
- [x] replace `workspace run` commands with `--recursive` ones
- [x] fix cyclic deps
    ```
    WARN  There are cyclic workspace dependencies: /Users/kevin/Projects/latticexyz/mud/packages/cli, /Users/kevin/Projects/latticexyz/mud/packages/store; /Users/kevin/Projects/latticexyz/mud/packages/cli, /Users/kevin/Projects/latticexyz/mud/packages/store, /Users/kevin/Projects/latticexyz/mud/packages/cli, /Users/kevin/Projects/latticexyz/mud/packages/world; /Users/kevin/Projects/latticexyz/mud/packages/cli, /Users/kevin/Projects/latticexyz/mud/packages/store, /Users/kevin/Projects/latticexyz/mud/packages/cli, /Users/kevin/Projects/latticexyz/mud/packages/world, /Users/kevin/Projects/latticexyz/mud/packages/store
    ```
- [x] figure out if we need `types: ['mocha']` in `services`
- [x] try including cli in monorepo root so we can use `mud` command without local pointer in store/world, and without circular deps


.github/workflows/build.yml
```diff
@@ -16,28 +16,31 @@ jobs:
...
           version: nightly
 
+      - uses: pnpm/action-setup@v2
+        with:
+          version: 8.x
+      - name: Setup node
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18.x
+          cache: pnpm
```
holic, line 31: the order of these steps matters, it needs to be:
- checkout (so pnpm lock file is present)
- pnpm setup (so next step can use pnpm to determine cache location)
- node setup

will clean this up once https://github.com/latticexyz/mud/pull/588 lands

packages/cli/src/commands/deprecated/trace.ts
```diff
@@ -37,7 +36,7 @@ const commandModule: CommandModule<Options, Options> = {
     const labels = [];
 
     const rpcUrl = rpc || "http://localhost:8545";
-    const provider = new JsonRpcProvider(rpcUrl);
+    const provider = new providers.StaticJsonRpcProvider(rpcUrl);
```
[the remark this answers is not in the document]
roninjin10: nvm this is deprecated anyways

packages/solecs/package.json
```diff
@@ -10,22 +10,21 @@
...
-    "link": "yarn link",
+    "lint": "pnpm prettier && pnpm solhint",
     "docs": "rimraf API && hardhat docgen && echo 'label: API\norder: 50' > API/index.yml",
     "release": "npm publish --access=public"
   },
   "devDependencies": {
     "@ethersproject/abi": "^5.7.0",
     "@ethersproject/providers": "^5.7.2",
-    "@rari-capital/solmate": "https://github.com/Rari-Capital/solmate.git#9cf1428245074e39090dceacb0c28b1f684f584c",
+    "@rari-capital/solmate": "https://github.com/transmissions11/solmate.git#9cf1428245074e39090dceacb0c28b1f684f584c",
```
roninjin10, line 27: Ah yea I had to do this same thing in optimism repo

roninjin10 (approved):

package.json
```diff
@@ -64,32 +42,32 @@
...
   },
   "scripts": {
-    "prepare": "husky install && (forge --version || yarn foundryup) && yarn workspaces run prepare",
+    "prepare": "husky install && (forge --version || foundryup)",
     "commit": "cz",
     "prettier:check": "prettier --check 'packages/**/*.ts'",
     "prettier": "prettier --write 'packages/**/*.ts'",
     "lint": "eslint . --ext .ts",
-    "release": "yarn && lerna publish --no-private --force-publish && yarn retype:updateversion",
+    "release": "pnpm && lerna publish --no-private --force-publish && pnpm retype:updateversion",
```
alvrs, line 50: Nit; I saw we use the explicit `pnpm install` now, so we should also use it here for consistency
```suggestion
    "release": "pnpm install && lerna publish --no-private --force-publish && pnpm retype:updateversion",
```

package.json
```diff
@@ -64,32 +42,32 @@
     }
   },
   "scripts": {
-    "prepare": "husky install && (forge --version || yarn foundryup) && yarn workspaces run prepare",
+    "prepare": "husky install && (forge --version || foundryup)",
```
alvrs, line 45: If `forge --version` fails, we probably have to assume `foundryup` is not installed yet and should run the package script to install it?
```suggestion
    "prepare": "husky install && (forge --version || pnpm foundryup)",
```

package.json
```diff
@@ -64,32 +42,32 @@
...
+    "entry:src": "lerna run postpack",
+    "gas-report": "pnpm recursive run gas-report",
+    "codegen": "pnpm codegen:tablegen && pnpm codegen:worldgen",
+    "codegen:tablegen": "pnpm recursive run tablegen",
+    "codegen:worldgen": "pnpm recursive run worldgen"
   },
   "lint-staged": {
     "*.ts": "eslint --cache --fix",
-    "*.{ts,css,md,sol}": "prettier --write"
+    "*.{ts,css,md}": "prettier --write"
```
alvrs, line 71: why this change?

packages/cli/src/utils/deploy-v2.ts
```diff
@@ -105,7 +105,11 @@ export async function deploy(mudConfig: MUDConfig, deployConfig: DeployConfig):
   const contractPromises: Record<string, Promise<string>> = { ...worldPromise, ...systemPromises, ...modulePromises };
 
   // Create World contract instance from deployed address
-  const WorldContract = new ethers.Contract(await contractPromises.World, IBaseWorldData.abi, signer) as IBaseWorld;
+  const WorldContract = new ethers.Contract(
+    await contractPromises.World,
+    IBaseWorldData.abi,
+    signer
+  ) as any as IBaseWorld;
```
alvrs, line 112: why do we need to cast to `any` first here?

packages/network/package.json
```diff
@@ -4,24 +4,23 @@
   "version": "1.41.0",
   "source": "src/index.ts",
   "main": "src/index.ts",
+  "type": "module",
   "engines": {
-    "node": ">=16"
+    "node": ">=18"
```
alvrs, line 9: ooc what is this used for? do we need it in the package's `package.json` in addition to the root `package.json` and if so do we also need the `pnpm` record here?

packages/noise/package.json
```diff
@@ -55,5 +54,6 @@
   "peerDependencies": {
     "@flatten-js/interval-tree": "^1.0.19"
   },
-  "gitHead": "914a1e0ae4a573d685841ca2ea921435057deb8f"
+  "gitHead": "914a1e0ae4a573d685841ca2ea921435057deb8f",
+  "module": "dist/index.js"
```
alvrs, line 58: is this change intended? what is it used for?

alvrs (commented):
very excited for this! Just a couple smaller nits/questions

.vscode/settings.json
```diff
@@ -1,15 +0,0 @@
...
-    "solmate/=node_modules/@rari-capital/solmate/src",
-    "persona/=node_modules/@latticexyz/persona/src/",
-    "gsn/=node_modules/@opengsn/contracts/src/",
-    "ds-test/=node_modules/ds-test/src/",
-    "solecs/=node_modules/@latticexyz/solecs/src/",
-    "royalty-registry/=node_modules/@manifoldxzy/royalty-registry/contracts/",
-    "std-contracts/=node_modules/@latticexyz/std-contracts/src/",
-    "@latticexyz/=node_modules/@latticexyz/"
-  ]
-}
```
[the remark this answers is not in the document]
alvrs: Just tested, this is what I see in vscode without the root vscode settings: 
CleanShot 2023-04-12 at 13 31 30@2x

Context for why the root settings are necessary with the juanblanco extension: https://github.com/juanfranblanco/vscode-solidity#remappings

Speaking of which - are you using this extension as well or a different one? Experiencing quite some lags in error highlighting etc with it.

packages/noise/package.json
alvrs: I see, would it make sense to leave that this change for a follow up where we align it everywhere? Might make sense to bundle it with the `nx` followup since it's related to how we handle local changes during development (eg pointing to the ts source vs having a build watcher)

lerna.json
```diff
@@ -3,7 +3,7 @@
     "packages/*"
   ],
   "version": "1.41.0",
-  "npmClient": "yarn",
+  "npmClient": "pnpm",
   "useWorkspaces": true,
```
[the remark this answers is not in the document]
holic: yes according to this: https://lerna.js.org/docs/recipes/using-pnpm-with-lerna#getting-started

unfortunately still using lerna in a few spots to run commands on all packages, like for linking (pnpm had `pnpm recursive link` and then it got removed in later versions for some reason)

package.json
holic: this doesn't actually work iirc, I made a solidity change somewhere and it wouldn't let me commit without removing this because prettier would fail

packages/cli/src/utils/deploy-v2.ts
holic: might have been related to me chasing down bugs with circular dependencies + ethers version misalignment, lemme revert and see if this still works

packages/network/package.json
holic: I tried to remove this, but then parcel failed to build due to some issue with `threads`. Adding this back in magically solved it and unclear why.
https://github.com/latticexyz/mud/pull/591/commits/98305000dbac100d5e68a8d4afb785034165f238
This is also why I added https://github.com/latticexyz/mud/issues/612

packages/noise/package.json
holic: > is this change intended? what is it used for?
I think @roninjin10 might have added this? I can't recall needing this

> if we use those fields we should probably standardize their order/position in package.json
100%! I've now twice tried to add [package.json eslint plugin](https://github.com/zetlen/eslint-plugin-package-json) to get consistent ordering in our package.json keys but it fails with a parse error as if it can't parse JSON.

.vscode/settings.json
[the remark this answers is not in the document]
holic: from discord
> there is a related ongoing discussion on the extensions github about this: https://github.com/juanfranblanco/vscode-solidity/issues/401
> Edit - multiple discussions: https://github.com/juanfranblanco/vscode-solidity/issues/400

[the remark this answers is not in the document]
holic: worked around this for now in https://github.com/latticexyz/mud/commit/f238f67834c2062fdaf672003e6466b6e33488a0

package.json
alvrs: just tested by changing the formatting of a solidity file and trying to commit the change, it does work for me (commit was prevented because it was empty because lint-staged fixed the formatting I changed):

CleanShot 2023-04-12 at 18 54 03@2x

holic: added back in

packages/cli/src/utils/deploy-v2.ts
holic: was able to revert this

packages/noise/package.json
holic: reverted the `module` change and added a follow up issue for the package.json order: https://github.com/latticexyz/mud/issues/617

packages/store/package.json
```diff
@@ -3,38 +3,37 @@
   "license": "MIT",
   "version": "1.41.0",
   "description": "Store",
-  "types": "./types/ethers-contracts/",
```
holic, line 6: I am unclear on the side effects of this, can no longer remember why I removed it 🙈

holic: reverted for now to see if everything still builds, etc.

alvrs (approved):
let's go!

PR #597, opened 2023-04-06
build(noise): always restore package.json after build

holic:
keep coming across this when adjusting build stuff and noise package fails but leaves a bunch of `package.json.bk` files around

### DIFF
--- a/packages/noise/package.json
+++ b/packages/noise/package.json
@@ -20,7 +20,7 @@
     "prepack:cjs": "mv package.json package.json.bak && jq \".main = \\\"dist/cjs/index.cjs\\\"\" package.json.bak > package.json.tmp && jq \".module = \\\"dist/index.js\\\"\" package.json.tmp > package.json.tmp2 && jq \"del(\".type\")\" package.json.tmp2 > package.json",
     "prepack": "mv package.json package.json.bak && jq \".main = \\\"dist/index.js\\\"\" package.json.bak > package.json",
     "postpack": "(mv package.json.bak package.json || true) && (rm package.json.tmp || true) && (rm package.json.tmp2 || true)",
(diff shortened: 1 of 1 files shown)


packages/noise/package.json
```diff
@@ -20,7 +20,7 @@
     "prepack:cjs": "mv package.json package.json.bak && jq \".main = \\\"dist/cjs/index.cjs\\\"\" package.json.bak > package.json.tmp && jq \".module = \\\"dist/index.js\\\"\" package.json.tmp > package.json.tmp2 && jq \"del(\".type\")\" package.json.tmp2 > package.json",
     "prepack": "mv package.json package.json.bak && jq \".main = \\\"dist/index.js\\\"\" package.json.bak > package.json",
     "postpack": "(mv package.json.bak package.json || true) && (rm package.json.tmp || true) && (rm package.json.tmp2 || true)",
-    "build": "rm -rf && yarn asbuild:release && yarn prepack:cjs && yarn parcel build && yarn postpack",
```
holic, line 23: I don't think this `rm -rf` is doing anything without a file arg?

holic: looks like it used to remove `dist` https://github.com/latticexyz/mud/commit/fb61b0326a84fe1e11dafeae240a0c1fe6bd0f2f

packages/noise/package.json
```diff
@@ -20,7 +20,7 @@
...
+    "build": "rimraf dist && yarn asbuild:release && (yarn prepack:cjs; yarn parcel build; yarn postpack)",
```
dk1a, line 23: `a; b` breaks on windows, `(a || true) && (b || true)` doesn't

dk1a (dismissed):

PR #599, opened 2023-04-06
feat(create-mud): use pnpm in templates, move to root so they can be installed/run

holic:
Some notes:
- Initially wanted to put templates in the same workspace as the rest of the packages (i.e. add `templates/*` to root `pnpm-workspace.yaml`), but I felt like the hoisting of certain dependencies made it a little hard to ensure these templates would run the same way once "cloned" with `create-mud`. I also wasn't sure about a "workspace in a workspace" (each template is also a pnpm workspace).
- I started by checking in `pnpm-lock.yaml` for each template then `.gitignore`'d it for a smaller diff. I think we might be better off leaving this out, so that dependencies can bump patch versions for security needs. CI should catch any changes for us. But curious if you feel strongly in another direction.
- Since each template is a monorepo, and pnpm is (IMO) the best package manager for a monorepo, I think it makes sense to be opinionated and just use pnpm for the templates, rather than trying to create some normalizing layer between package managers and how they define/manage monorepos/workspaces.


packages/create-mud/package.json
```diff
@@ -6,24 +6,22 @@
...
+    "build": "tsup src/cli.ts --minify && ./scripts/copy-templates.sh",
+    "clean": "rimraf dist",
     "dev": "tsup src/cli.ts --watch",
     "prepublishOnly": "npm run clean && npm run build",
     "test": "echo \"TODO\"",
     "release": "npm publish --access=public"
   },
   "dependencies": {
-    "create-create-app": "git+https://github.com/holic/create-create-app#e5c8bb449033418ad8518fc73f72c1a07eab2577"
+    "create-create-app": "git+https://github.com/holic/create-create-app#74376c59b48a04aabbe94d9cacfe9cb1cecccd63"
```
holic, line 21: this version bump improves package manager detection

packages/network/src/workers/Recover.worker.ts
```diff
@@ -1,5 +1,5 @@
 import { Message } from "@latticexyz/services/protobuf/ts/ecs-relay/ecs-relay";
-import { expose } from "threads";
+import { expose } from "threads/worker";
```
holic, line 2: gpt-4 figured this out after a couple rounds of pasting in build errors and trying out its suggestions 🤯

.gitignore
```diff
@@ -11,3 +11,7 @@ lerna-debug.log
 .retype
 .env
 yarn-error.log
+
+# We don't projects created from templates to ignore their lockfiles, but we don't
```
alvrs, line 15:
```suggestion
# We don't want projects created from templates to ignore their lockfiles, but we don't
```

packages/create-mud/package.json
```diff
@@ -6,24 +6,24 @@
...
-    "dist",
-    "templates"
+    "dist"
   ],
   "scripts": {
     "prepare": "pnpm build",
-    "build": "tsup src/cli.ts --minify",
-    "clean": "shx rm -rf lib",
+    "build": "tsup src/cli.ts --minify && ./scripts/copy-templates.sh",
+    "clean": "rimraf dist",
```
alvrs, line 14: do we need to add rimraf as a dev dependency in this package?

packages/network/src/workers/Recover.worker.ts
alvrs: 🤯

packages/create-mud/src/cli.ts
```diff
@@ -16,7 +16,8 @@ create("create-mud", {
   promptForEmail: false,
   promptForLicense: false,
   promptForTemplate: true,
-  caveat: ({ answers }) => `Done! Play in the MUD with \`cd ${answers.name}\` and \`yarn dev\``,
+  caveat: ({ answers, packageManager }) =>
+    `Done! Play in the MUD with \`cd ${answers.name}\` and \`${packageManager} run dev\``,
```
alvrs, line 20: We still require folks to use pnpm in the templates right? Should we add a note about that here?

alvrs (approved):
This is great!

roninjin10:
> Initially wanted to put templates in the same workspace as the rest of the packages (i.e. add templates/* to root pnpm-workspace.yaml), but I felt like the hoisting of certain dependencies made it a little hard to ensure these templates would run the same way once "cloned" with create-mud. I also wasn't sure about a "workspace in a workspace" (each template is also a pnpm workspace).

One of the big benifits of pnpm is this isn't much of a concern.  The store can be in node modules, somewhere globally on your system ect. and it will always work the same.  So I wouldn't worry too much

packages/create-mud/package.json
```diff
@@ -6,24 +6,24 @@
...
     "dev": "tsup src/cli.ts --watch",
     "prepublishOnly": "npm run clean && npm run build",
-    "test": "echo \"TODO\"",
+    "test": "pnpm test:minimal && pnpm test:react",
+    "test:minimal": "(dist/cli.js test-project --template minimal || true) && rimraf test-project",
+    "test:react": "(dist/cli.js test-project --template react || true) && rimraf test-project",
```
alvrs, line 19: would this let the CI fail if creating the template fails or would it still pass bc of the `|| true`?

holic: good catch!

packages/create-mud/src/cli.ts
holic: probably, I think I can also force the package manager here too

holic:
> One of the big benifits of pnpm is this isn't much of a concern. The store can be in node modules, somewhere globally on your system ect. and it will always work the same. So I wouldn't worry too much

are you suggesting we _should_ get this to live within the same workspace? how does pnpm handle workspaces-within-workspaces? how do we protect against implicit dependencies (a template workspace not accidentally using a root workspace dep)

alvrs (approved):

PR #600, opened 2023-04-08
feat(config): separate config from cli

dk1a:


packages/cli/src/commands/deploy-v2.ts
```diff
@@ -2,12 +2,12 @@ import chalk from "chalk";
 import glob from "glob";
 import path, { basename } from "path";
 import type { CommandModule, Options } from "yargs";
-import { loadWorldConfig } from "../config/world/index.js";
+import { loadStoreConfig, loadWorldConfig } from "@latticexyz/config";
+import { MUDError } from "@latticexyz/config";
```
dk1a, line 6: initially wanted to put `MUDError` into `utils` package, but importing it doesn't work yet so left this for another PR

packages/cli/src/index.ts
```diff
@@ -1,15 +1 @@
...
-  StoreUserConfig,
-  StoreConfig,
-  WorldUserConfig,
-  ResolvedWorldConfig,
-  MUDUserConfig,
-  MUDConfig,
-} from "./config/index.js";
-
-export { storeConfig, mudConfig } from "./config/index.js";
+export {};
```
holic, line 1: 🎉

holic (approved):

PR #603, opened 2023-04-10
feat(store,world,cli): extract DynamicPartial from StoreCore

dk1a:
This separates push, update and future methods for dealing with parts of dynamic fields to avoid max size for World.
After this PR World should go down to 20
https://github.com/latticexyz/mud/pull/525#pullrequestreview-1365868473


packages/store/src/IStoreDynamicPartial.sol
```diff
@@ -0,0 +1,18 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
+
+import { IErrors } from "./IErrors.sol";
+
+interface IStoreDynamicPartial is IErrors {
```
holic, line 6: is this ~always going to be used for arrays? just wondering if we should name it more by what it's designed to be used for rather than how it's implemented

packages/world/gas-report.txt
```diff
@@ -1,26 +1,26 @@
...
-(test/UniqueEntityModule.t.sol) | install unique entity module [world.installModule(uniqueEntityModule, new bytes(0))]: 773726
-(test/UniqueEntityModule.t.sol) | get a unique entity nonce (non-root module) [uint256 uniqueEntity = uint256(getUniqueEntity(world))]: 65723
-(test/UniqueEntityModule.t.sol) | installRoot unique entity module [world.installRootModule(uniqueEntityModule, new bytes(0))]: 780192
-(test/UniqueEntityModule.t.sol) | get a unique entity nonce (root module) [uint256 uniqueEntity = uint256(getUniqueEntity(world))]: 65723
-(test/World.t.sol) | Delete record [world.deleteRecord(namespace, file, singletonKey)]: 16137
-(test/World.t.sol) | Push data to the table [world.pushToField(namespace, file, keyTuple, 0, encodedData)]: 96529
+(test/KeysWithValueModule.t.sol) | install keys with value module [world.installRootModule(keysWithValueModule, abi.encode(sourceTableId))]: 600366
+(test/KeysWithValueModule.t.sol) | set a record on a table with KeysWithValueModule installed [world.setRecord(namespace, sourceFile, keyTuple1, abi.encodePacked(value))]: 191672
```
holic, line 5: afaict, this added 22k gas? 😵‍💫

holic:
I defer to @alvrs on this one. I think we need to decide if 1) that gas trade-off is worth it and 2) if we want to move more of store calls into a "system"?

packages/world/gas-report.txt
```diff
@@ -1,26 +1,26 @@
...
+(test/KeysWithValueModule.t.sol) | set a field on a table with KeysWithValueModule installed [world.setField(namespace, sourceFile, keyTuple1, 0, abi.encodePacked(value1))]: 199906
+(test/KeysWithValueModule.t.sol) | change a field on a table with KeysWithValueModule installed [world.setField(namespace, sourceFile, keyTuple1, 0, abi.encodePacked(value2))]: 155751
+(test/UniqueEntityModule.t.sol) | install unique entity module [world.installModule(uniqueEntityModule, new bytes(0))]: 773660
+(test/UniqueEntityModule.t.sol) | get a unique entity nonce (non-root module) [uint256 uniqueEntity = uint256(getUniqueEntity(world))]: 65679
+(test/UniqueEntityModule.t.sol) | installRoot unique entity module [world.installRootModule(uniqueEntityModule, new bytes(0))]: 780148
+(test/UniqueEntityModule.t.sol) | get a unique entity nonce (root module) [uint256 uniqueEntity = uint256(getUniqueEntity(world))]: 65679
+(test/World.t.sol) | Delete record [world.deleteRecord(namespace, file, singletonKey)]: 16115
+(test/World.t.sol) | Push data to the table [world.pushToField(namespace, file, keyTuple, 0, encodedData)]: 119235
```
alvrs, line 17: Around 20k gas increase here 😢 I'm not sure how much data was pushed here though (with the additional delegatecall indirection we also pay for calldata for calling the DynamicPartial method). Also it seems like we're missing a gas report for `updateInField`.

packages/world/src/interfaces/IWorldCore.sol
```diff
@@ -8,8 +8,8 @@ import { ISystemHook } from "./ISystemHook.sol";
 import { IModule } from "./IModule.sol";
 
 /**
- * The IWorldCore interfaces includes all World methods with a static function selector.
- * Consumers should use the `IBaseWorld` interface instead, which includes
+ * The IWorldCore interface only includes core World methods with a static function selector.
+ * Consumers should use the `IBaseWorld` interface instead, which includes other static and
```
alvrs, line 12: `IWorldCore` has all the "static" function selectors (those that are part of the World bytecode), while `IBaseWorld` also includes function selectors that are registered dynamically (those that are part of the bytecode of another contract but are called via the World's fallback function)
```suggestion
 * Consumers should use the `IBaseWorld` interface instead, which includes
```

packages/world/src/modules/dynamicpartial/DynamicPartialModule.sol
```diff
@@ -0,0 +1,94 @@
...
+        value: 0,
+        funcSelectorAndArgs: abi.encodeWithSelector(
+          RegistrationSystem.registerRootFunctionSelector.selector,
+          ROOT_NAMESPACE,
+          DYNAMIC_PARTIAL_SYSTEM_NAME,
+          rootFunctionSelectors[i], // Use the same function selector for the World as in DynamicPartialSystem
+          rootFunctionSelectors[i]
+        )
+      });
+    }
```
alvrs, line 92: Have you tried if it works if we just call `world.registerRootFunctionSelector`? With https://github.com/latticexyz/mud/pull/575 the World should have access to the root namespace, so if the module is delegatecalled it should work

packages/world/src/modules/registration/tables/SystemRegistry.sol
```diff
@@ -8,6 +8,7 @@ import { SchemaType } from "@latticexyz/schema-type/src/solidity/SchemaType.sol"
 
 // Import store internals
 import { IStore } from "@latticexyz/store/src/IStore.sol";
+import { IStoreDynamicPartial } from "@latticexyz/store/src/IStoreDynamicPartial.sol";
```
alvrs, line 11: Is it an issue if we import the dynamics library in table libraries that don't need it?

packages/cli/src/render-solidity/common.ts
```diff
@@ -80,23 +80,28 @@ export function renderImports(imports: ImportDatum[]) {
...
+function _renderWithStoreFactory(storeInterfaceName: string) {
+  return (
+    storeArgument: boolean,
+    callback: (
+      _typedStore: string | undefined,
+      _store: string,
+      _commentSuffix: string,
+      _untypedStore: string | undefined
+    ) => string
+  ) => {
```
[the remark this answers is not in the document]
dk1a: where the method is used is a lot of stuff happening, so it's just clearer this way imo. Config param can easily get lost there, especially if it comes after the huge callback

packages/world/src/interfaces/IWorldCore.sol
dk1a: by "other static" I meant DynamicPartial funcs, but yea they are actually dynamic too

packages/world/src/modules/dynamicpartial/DynamicPartialModule.sol
dk1a: I have tried, but my issue started with registerSystem, then I changed both, haven't tried `registerRootFunctionSelector` on its own, will see

packages/world/src/modules/registration/tables/SystemRegistry.sol
dk1a: afaik no, solidity is good with this sort of tree shaking (specifically I was testing adding unused methods to libraries, and they didn't get included in bytecode; I assume imports are even easier to shake)

alvrs:
Did some measurements around delegatecalling store methods here: https://github.com/latticexyz/mud/issues/608

tl;dr: there should be around 4k overhead for storing data via `delegatecall` instead of directly internally. I wonder where the 20k increase from `pushToField` comes from here? Maybe because we do reads from storage in there too?

There might also be an overhead of translating the dynamic function selector to `resourceSelector` and then using `call`. Since calling systems via the function selector is probably much more common than calling via `call`, we might be able to optimize this path by storing a flag for whether or not the system is public in the function selector table, and if it is, calling the system directly from the fallback function instead of loading the resource selector first.

Also it gets cheaper the more often a System accesses a method in the DynamicPartial system, because of the fixed `address_access_cost` -> this also means more code we can bundle in the same core module system the better because only one external address has to be accessed.

packages/world/gas-report.txt
alvrs: Agree this PR is necessary to reduce the World size. How many more DynamicPartial method do you think we'll need? If there are other less used methods like `registerStoreHook` that we could externalize that might be a good alternative too.

I'm down to merge as is though to unblock you and then try to reduce gas overhead in a followup by reducing the general overhead of calling dynamic function selectors and by moving frequently called methods to the World internal code and externalizing less often used methods instead.

packages/world/src/modules/dynamicpartial/DynamicPartialModule.sol
dk1a: non-delegated `registerRootFunctionSelector` also gives AccessDenied

alvrs: This should work now with #609 merged (both here and above for `registerSystem`)

alvrs:
context from github: we'll wait with merging this until we refactored World to move all non-hot-path methods into a module, so that World will have space for the hot-path ones (like DynamicPartial). Making this PR a draft for now.

holic:
done in https://github.com/latticexyz/mud/pull/622

PR #606, opened 2023-04-10
fix(cli): wait for tx confirmation on deploy txs

holic:


packages/cli/src/utils/deploy-v2.ts
```diff
@@ -408,16 +408,16 @@ export async function deploy(mudConfig: MUDConfig, deployConfig: DeployConfig):
...
     try {
       const gasLimit = await contract.estimateGas[func].apply(null, args);
       console.log(chalk.gray(`executing transaction: ${functionName} with nonce ${nonce}`));
-      const txPromise = contract[func].apply(null, [
-        ...args,
-        { gasLimit, nonce: nonce++, maxPriorityFeePerGas, maxFeePerGas },
-      ]);
+      const txPromise = contract[func]
+        .apply(null, [...args, { gasLimit, nonce: nonce++, maxPriorityFeePerGas, maxFeePerGas }])
+        .then((tx: any) => tx.wait(confirmations));
```
holic, line 420: I originally was doing this per call at `await fastTxExecute`, but I wanted this ~everywhere. Since this function is local to the deploy call, I decided to apply this wait here, but made the number of confirmations configurable in case there are usages where we don't need to wait for confirmation.

alvrs (approved):
Lgtm!

PR #607, opened 2023-04-11
feat(cli): use json for gas report output

holic:


holic:
also made gas reports faster by creating all the files then running them once rather than individually

before:
image

after:
image

packages/cli/src/commands/gas-report.ts
```diff
@@ -52,43 +54,39 @@ const commandModule: CommandModule<Options, Options> = {
...
-          const name = compareGasReportMatch[2];
-          const functionCall = compareGasReportMatch[3];
-          const gasUsed = compareGasReportMatch[4];
-
-          compareGasReport.push({ source, name, functionCall, gasUsed: parseInt(gasUsed) });
-        }
+        const compareGasReport: GasReport = JSON.parse(readFileSync(compare, "utf8"));
+        // Merge the previous gas report with the new one
+        gasReport = gasReport.map((entry) => {
+          console.log("looking for prev value");
```
alvrs, line 78: Do we still need this log?

```suggestion
```

alvrs (approved):
Lgtm! Unrelated to this PR but we should also make sure to remove the temp files if a `SIGINT` is caught (by a user cancelling the gas report via `CTRL+C`)

PR #609, opened 2023-04-11
fix(world): allow world to access own functions via external calls

alvrs:
- Calls that require `AccessControl.requireOwner` (eg. `registerRootFunctionSelector` or `registerSystem`) were failing if called externally by the World itself, because the World is not the owner of any `ROOT_FILE`.
- Similarly calls that require `AccessControl.requireAccess` were failing if called externally by the World for the same reason.
- However, the World can always call these methods internally, so it should also be allowed for the World to call them externally (eg. when the World is delegatecalling a `RootModule` to install it, which then tries to install Systems in the World by calling `registerSystem` _as the World_)
- This makes #575 obsolete (since World has access to every namespace now)

### DIFF
--- a/packages/world/src/AccessControl.sol
+++ b/packages/world/src/AccessControl.sol
@@ -15,7 +15,8 @@ library AccessControl {
    */
   function hasAccess(bytes16 namespace, bytes16 file, address caller) internal view returns (bool) {
     return
-      ResourceAccess.get(ResourceSelector.from(namespace, 0), caller) || // First check access based on the namespace
+      address(this) == caller || // First check if the World is calling itself
+      ResourceAccess.get(ResourceSelector.from(namespace, 0), caller) || // Then check access based on the namespace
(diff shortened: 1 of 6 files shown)


packages/world/src/AccessControl.sol
```diff
@@ -44,7 +44,7 @@ library AccessControl {
   ) internal view returns (bytes32 resourceSelector) {
     resourceSelector = ResourceSelector.from(namespace, file);
 
-    if (NamespaceOwner.get(namespace) != caller) {
+    if (NamespaceOwner.get(namespace) != caller && caller != address(this)) {
```
holic, line 47: tiny thing but this might be more readable if `caller` is in the same position for each conditional check, also might be a little more gas efficient if the `address(this)` check is first
```suggestion
    if (caller != address(this) && caller != NamespaceOwner.get(namespace)) {
```

dk1a (dismissed):
would it be better to name it `requireOwnerOrWorld` to better reflect the conditions?

dk1a (approved):

PR #611, opened 2023-04-11
build: split out ci build steps again

holic:
github actions don't show steps within a composite action, just the single action as a whole step, so it's hard to tell where in the build step a failure happened without digging into the logs

see https://github.com/actions/runner/issues/1877

before
image

after
image

I left "Setup" alone, because the extra granularity there isn't as important

And it's nice to now see the release steps separated too (from #588):
image


holic:
gonna merge since CI passed and this is blocking #591

PR #613, opened 2023-04-12
refactor: uint256 tableId -> bytes32 tableId

alvrs:
* We originally used `bytes32` for tableIds, but switched to `uint256` in #416 because `bytes32` had a type overlap with `string`, which caused issues in the first version of the `World` contract (where we used strings for routes instead of the current namespace/name approach, see #467 for context on this change).
* Now that we don't use strings for routes in the World anymore, there is no reason for `tableId` to be `uint256` anymore, while `resourceSelector` in `World` is `bytes32`. A `resourceSelector` is a superset of a `tableId` (also including systems), and converting from `bytes32` to `uint256` whenever we use a `resourceSelector` as a `tableId` is unnecessary overhead.

### DIFF
--- a/packages/network/src/v2/schemas/tableSchemas.ts
+++ b/packages/network/src/v2/schemas/tableSchemas.ts
@@ -2,6 +2,7 @@ import { Contract } from "ethers";
 import { TableId } from "@latticexyz/utils";
 import { TableSchema } from "../common";
 import { decodeSchema } from "./decodeSchema";
+import { IStore } from "@latticexyz/store/types/ethers-contracts/IStore";
 
 // worldAddress:tableId => schema
 // TODO: add chain ID to namespace?
@@ -9,7 +10,7 @@ const schemaCache: Partial<Record<`${string}:${string}`, Promise<TableSchema>>> 
 
 // the Contract arguments below assume that they're bound to a provider
 
-export function getSchema(world: Contract, table: TableId): Promise<TableSchema> | undefined {
+export function getSchema(world: IStore, table: TableId): Promise<TableSchema> | undefined {
   const cacheKey = `${world.address}:${table.toHexString()}` as const;
   return schemaCache[cacheKey];
 }
@@ -45,7 +46,7 @@ export function registerSchema(world: Contract, table: TableId, rawSchema?: stri
   // TODO: populate from ECS cache before fetching from RPC
 
   console.log("fetching schema for table", { table: table.toString(), world: world.address });
-  const schema = world.getSchema(table).then((rawSchema: string) => {
+  const schema = (world as IStore).getSchema(table.toHexString()).then((rawSchema: string) => {
     const decodedSchema = decodeSchema(rawSchema);
     if (decodedSchema.isEmpty) {
       console.warn("Schema not found for table", { table: table.toString(), world: world.address });
--- a/packages/store/mud.config.mts
+++ b/packages/store/mud.config.mts
@@ -8,7 +8,7 @@ export default storeConfig({
     Callbacks: "bytes24[]",
     StoreMetadata: {
       primaryKeys: {
-        tableId: "uint256",
+        tableId: "bytes32",
       },
       schema: {
         tableName: "string",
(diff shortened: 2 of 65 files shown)


packages/store/package.json
```diff
@@ -15,10 +15,11 @@
   },
   "scripts": {
     "prepare": "yarn build",
-    "codegen": "ts-node ./scripts/codegen.ts",
+    "codegen": "yarn tsgen && yarn tablegen",
+    "tsgen": "ts-node ./scripts/codegen.ts",
     "tablegen": "../cli/dist/mud.js tablegen",
     "test": "yarn codegen && forge test",
-    "build": "yarn codegen && rimraf out && forge build && yarn dist && yarn types && tsup",
+    "build": "rimraf out && forge build && yarn dist && yarn types && tsup",
```
alvrs, line 22: Using `tablegen` here caused an error in the CI, because `../cli/dist/mud.js` does not exist yet while store is being built (CLI can't be built first because it depends on `World`'s ABI, which depends on `Store`). There are (at least) two approaches to solve this:
1. Move the `tablegen` scripts into `Store` (and `worldgen` scripts into `World`), so these packages can use them locally (and consumers of these packages don't have to install the MUD CLI if they only want to use eg `Store` without the rest of the framework). CLI can still expose a `tablegen` command by importing the scripts from `Store` (and a `worldgen` command by importing from `World`)
2. Check in the World/Store ABIs, so CLI can be built independently from building World/Store (this alone doesn't fix the circular dependency though)

packages/utils/src/v2/TableId.ts
```diff
@@ -35,6 +35,7 @@ export class TableId {
     // assumes tableId is a 32-byte hex string, otherwise it left-pads with zeros (for numbers)
     // this is scary, since zero padding is different depending on the type (bytes types vs number types)
     // TODO: fix this after we move tableIds to bytes32 instead of uint256
+    // TODO: check if this is still relevant
     const tableIdBytes = new Uint8Array(32);
     tableIdBytes.set(tableId.slice().reverse());
     tableIdBytes.reverse();
```
alvrs, line 41: @holic can you help me figure out how to change this logic to make it work with `tableIds` that are `bytes32`? Do we assume `tableId` to be `uint256` anywhere else in the client code?

packages/store/mud.config.mts
```diff
@@ -8,7 +8,7 @@ export default storeConfig({
     Callbacks: "bytes24[]",
     StoreMetadata: {
       primaryKeys: {
-        tableId: "uint256",
```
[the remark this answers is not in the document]
holic: in client, `StoreMetadata` is inferred since its a proper table

packages/store/package.json
holic: since this has come up a bunch, mind opening an issue for this?

packages/utils/src/v2/TableId.ts
holic: I think we can just change this to
```
const tableIdBytes = new Uint8Array(32);
tableIdBytes.set(tableId);
```
which ensures right-zero padding. But if ethers is already passing bytes32 values as zero padded hex, then we won't even need this.

(before it would reverse, then right-pad, then reverse again, so that the result is left-padded)

We could also use the viem `pad` helpers for this to make it more clear what's going on.

packages/store/package.json
alvrs: https://github.com/latticexyz/mud/issues/616

packages/network/src/v2/schemas/tableSchemas.ts
```diff
@@ -45,7 +46,7 @@ export function registerSchema(world: Contract, table: TableId, rawSchema?: stri
   // TODO: populate from ECS cache before fetching from RPC
 
   console.log("fetching schema for table", { table: table.toString(), world: world.address });
-  const schema = world.getSchema(table).then((rawSchema: string) => {
+  const schema = (world as IStore).getSchema(table.toHexString()).then((rawSchema: string) => {
```
holic, line 49: good catch!

holic: should we make the `world` argument an `IStore` here too?

alvrs: I tried that first but it requires to change it all the way upstream too, so I waited for now bc it's mostly unrelated to this PR. Just thought having type hints where we call the World methods is useful

holic (approved):

PR #614, opened 2023-04-12
refactor(std-client): deprecate devMode, add disableCache flag instead

alvrs:
* `devMode` was used to both set gas price to 0 (to use dev chains with 0 gas) and to disable the cache
* Therefore using the `devMode` flag to disable the cache in a production deployment didn't work, because in this situation the side effect of setting gas price to 0 was unintended.
* This PR deprecates the `devMode` flag and adds a new `disableCache` flag to be used to disable the cache. (We don't remove the `devMode` flag yet to avoid a breaking change in MUD v1, since the config is shared by both v1 and v2 network setup functions)
* To send transactions with gas price 0 during development, the `fastTxExecute` function should be used instead of `txQueue` (it's much leaner and has logic to not set a gas price / priority fee if the chain's base fee is 0)

### DIFF
--- a/packages/std-client/src/setup/setupMUDNetwork.ts
+++ b/packages/std-client/src/setup/setupMUDNetwork.ts
@@ -155,7 +155,8 @@ export async function setupMUDNetwork<C extends ContractComponents, SystemTypes 
         provider: providerConfig,
         worldContract: contractsConfig.World,
         initialBlockNumber: networkConfig.initialBlockNumber ?? 0,
-        disableCache: networkConfig.devMode, // Disable cache on local networks (hardhat / anvil)
+        // TODO: `networkConfig.devMode` is deprecated, only use `networkConfig.disableCache`
+        disableCache: networkConfig.devMode || networkConfig.disableCache,
         fetchSystemCalls: options?.fetchSystemCalls,
       },
     });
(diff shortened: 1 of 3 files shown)


packages/std-client/src/setup/setupMUDNetwork.ts
```diff
@@ -155,7 +155,8 @@ export async function setupMUDNetwork<C extends ContractComponents, SystemTypes
         provider: providerConfig,
         worldContract: contractsConfig.World,
         initialBlockNumber: networkConfig.initialBlockNumber ?? 0,
-        disableCache: networkConfig.devMode, // Disable cache on local networks (hardhat / anvil)
+        // TODO: `networkConfig.devMode` is deprecated, only use `networkConfig.disableCache`
+        disableCache: networkConfig.devMode || networkConfig.disableCache,
```
holic, line 159: I'd be fine just removing dev mode!

holic (approved):

packages/std-client/src/setup/setupMUDNetwork.ts
alvrs: unfortunately removing it would be a breaking change if we ever want to do another patch release for v1. I removed it in `setupMUD2Network` though.

PR #618, opened 2023-04-12
refactor: consolidate prettier config

holic:
- moves all prettier usage to the root package
- adjusts prettier command to match lint-staged extensions (and runs prettier across repo)
- adds lint step to CI


.prettierrc.js
```diff
@@ -0,0 +1,9 @@
+/** @type {import('prettier').Config} */
+module.exports = {
+  plugins: ["prettier-plugin-solidity"],
+  printWidth: 120,
+  semi: true,
+  tabWidth: 2,
+  useTabs: false,
+  bracketSpacing: true,
+};
```
holic, line 9: moved this to a `.js` file so other packages can include/extend in the future with

```
/** @type {import('prettier').Config} */
module.exports = {
  ...require('../../.prettierrc.js'),
  // extended config values or overrides here
};
```

alvrs (approved):
very nice!

holic:
unclear why prettier at the root is disagreeing with cli-generated output (which runs through the same prettier version using the same config + plugin)

@dk1a do you recall any issues with this?

dk1a:
yes I remember it happening, but I hadn't really looked into the cause

holic:
Oh these are a separate autogen pipeline (ejs -> sol, not using prettier), so that's why they differ.

Issue #612, opened 2023-04-11
Consolidate JS build tooling


holic:
We currently have a wide mix of these, and we should probably consolidate:
- ~~parcel, rollup, vite~~
- ~~tsup, tsc, esbuild~~
- ~~ts-node, tsx~~ #619 

If we're able to get rid of parcel in `network` package, we should remove `engines` from `package.json` (it's only there to fix parcel build issues).

alvrs:
Context from discord: 

>Vite 
Pro - 1 tool that can build everything
Pro - Doubles as a test config don't need to configure babel or anything like with jest
Pro - Should have less hacky than otherwise ways of configuring dev builds to work with hot reloading
Con - Bit of overkill
>
>Tsup
Pro - Really fast even for prod build since it uses esbuild instead of rollup 
Pro - Really simple to configure
Pro - Definitely much simpler tool than vite
Con - You are forced to use different tool once you need to bundle an actual website or non typescript or struggle configuring it with tsup
>
>Tsc
Pro - It's typescript it's kinda the most obvious tool to use
Pro - Simplist to configure (sometimes)
Con - less forgiving than using esbuild
Con - slow
Con - boring
>
>Parcel
con - meh
>
>Rollup
Pro - most well documented build tool in js ecosystem
Pro - Plugin for everything
Pro - one build tool can do everything
Con - vite is basically a better version of rollup

Based on this I think we should try to consolidate to use `vite` everywhere

holic:
vite -> tsup happened in https://github.com/latticexyz/mud/pull/660

just need to do parcel/rollup ones now!

PR #619, opened 2023-04-12
build: remove ts-node usages

holic:
We no longer need ts-node at runtime like we did before (this was either a cli thing or a config thing), we can remove ts-node from most packages.

That leaves only a couple that actually used it:
- store for codegen, but this was easy to swap with tsx (used elsewhere)
- noise for `test:hardhat` but the command wasn't working or in use anyway, so I just reduced it to what it _should_ be assuming hardhat adds TS support
  - we could also just migrate the test cases over to foundry

Helps a bit with #612


holic:
gonna merge this since this is mostly clean up and want this for templates PR that I am starting on next

PR #620, opened 2023-04-13
[VERY WIP] :construction_worker: feat(ci): start adding nx to mud repo

roninjin10:
- added initial nx.json file
- switched some scripts over
- it currently works correctly but it's cache for many packages is broken most of the time so it's cache is mostly useless atm until we finetune it

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,6 +19,9 @@ jobs:
       - name: Setup
         uses: ./.github/actions/setup
 
+      - name: Build
+        run: pnpm nx run-many --target=build --parallel
+
       - name: Run codegen
         run: pnpm codegen
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,7 +18,7 @@ jobs:
         uses: ./.github/actions/setup
 
       - name: Run tests
-        run: pnpm test
+        run: pnpm nx run-many --target=test --parallel
 
       - name: Lint
-        run: pnpm lint
+        run: pnpm nx run-many --target=lint --parallel
--- a/nx.json
+++ b/nx.json
@@ -1,0 +1,54 @@
+{
+  "tasksRunnerOptions": {
+    "default": {
+      "runner": "@nrwl/nx-cloud",
+      "options": {
+        "cacheableOperations": [
+          "lint",
+          "build",
+          "test",
+          "test:contracts",
+          "prettier",
+          "solhint",
+          "test:forge",
+          "test:hardhat",
+          "test:solidity",
+          "test:typescript",
+          "build:solidity",
+          "build:typescript"
+        ]
+      }
+    }
+  },
+  "targetDefaults": {
+    "dev": {
+      "dependsOn": ["^build"]
+    },
+    "build": {
+      "dependsOn": ["^build"],
+      "outputs": ["{projectRoot}/dist"]
+    },
+    "release": {
+      "dependsOn": ["^build", "build", "^release"]
+    },
+    "test": {
+      "dependsOn": ["^build", "build"]
+    },
+    "test:contracts": {
+      "dependsOn": ["^build", "build"]
+    },
+    "test:forge": {
+      "dependsOn": ["^build", "build"]
+    },
+    "test:hardhat": {
+      "dependsOn": ["^build", "build"]
+    },
+    "test:solidity": {
+      "dependsOn": ["^build", "build"]
+    },
+    "test:typescript": {
+      "dependsOn": ["^build", "build"]
+    }
+  },
+  "defaultBase": "main"
+}
--- a/package.json
+++ b/package.json
@@ -34,7 +34,9 @@
     "rimraf": "^3.0.2",
     "run-pty": "^3.0.0",
     "shelljs": "^0.8.5",
-    "typescript": "^4.9.5"
+    "typescript": "^4.9.5",
+    "nx": "15.9.2",
+    "@nrwl/nx-cloud": "latest"
   },
   "config": {
     "commitizen": {
@@ -42,6 +44,7 @@
     }
   },
   "scripts": {
+    "build": "pnpm nx run-many --target=build",
     "prepare": "husky install && (forge --version || pnpm foundryup)",
     "commit": "cz",
     "prettier:check": "prettier --check 'packages/**/*.{ts,css,md,sol}'",
@@ -58,13 +61,13 @@
     "docs:prepare": "lerna run docs && pnpm retype:updateversion",
     "docs:dev": "retype watch .",
     "retype:updateversion": "sed \"s/label: .*/label: $(cat packages/cli/package.json | jq -r '.version')/\" retype.yml > retype.yml.tmp && mv retype.yml.tmp retype.yml",
-    "test": "pnpm recursive run test",
+    "test": "pnpm nx run-many --target=test",
     "entry:dist": "lerna run prepack",
     "entry:src": "lerna run postpack",
-    "gas-report": "pnpm recursive run gas-report",
+    "gas-report": "pnpm nx run-many --target=gas-report",
     "codegen": "pnpm codegen:tablegen && pnpm codegen:worldgen",
-    "codegen:tablegen": "pnpm recursive run tablegen",
-    "codegen:worldgen": "pnpm recursive run worldgen"
+    "codegen:tablegen": "pnpm nx run-many --target=tablegen",
+    "codegen:worldgen": "pnpm nx run-many --target=worldgen"
   },
   "lint-staged": {
     "*.ts": "eslint --cache --fix",
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -14,6 +14,9 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
 
+      - name: Build
+        run: pnpm build
+
       - name: Setup Go
         uses: actions/setup-go@v3
         with:
--- a/README.md
+++ b/README.md
@@ -103,10 +103,16 @@ npm install pnpm --global
 git clone https://github.com/latticexyz/mud
 ```
 
-5. Install MUD dependencies and setup local environment
+5. Install MUD dependencies
 
 ```
 cd mud && pnpm install
+```
+
+6. Setup the local environment
+
+```
+pnpm build
 ```
 
 ### Pull requests
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -10,6 +10,9 @@ importers:
       '@commitlint/config-conventional':
         specifier: ^16.2.4
         version: 16.2.4
+      '@nrwl/nx-cloud':
+        specifier: latest
+        version: 15.3.2
       '@parcel/optimizer-data-url':
         specifier: 2.8.3
         version: 2.8.3(@parcel/core@2.8.3)
@@ -52,6 +55,9 @@ importers:
       lint-staged:
         specifier: '>=10'
         version: 13.0.3
+      nx:
+        specifier: 15.9.2
+        version: 15.9.2
       prettier:
         specifier: ^2.8.4
         version: 2.8.4
@@ -3820,6 +3826,22 @@ packages:
       tslib: 2.5.0
     dev: true
 
+  /@nrwl/nx-cloud@15.3.2:
+    resolution: {integrity: sha512-fga4duqcrkyT3djQpyRp8god3trb18hb+Tv0CYRrCcdNDHhzb6LOrxnenrFsjPjt38hyCdyu/ZYmHI/lQXyhAw==}
+    hasBin: true
+    dependencies:
+      axios: 0.21.4
+      chalk: 4.1.0
+      dotenv: 10.0.0
+      fs-extra: 10.1.0
+      node-machine-id: 1.1.12
+      strip-json-comments: 3.1.1
+      tar: 6.1.11
+      yargs-parser: 21.1.1
+    transitivePeerDependencies:
+      - debug
+    dev: true
+
   /@nrwl/nx-darwin-arm64@15.9.2:
     resolution: {integrity: sha512-Yv+OVsQt3C/hmWOC+YhJZQlsyph5w1BHfbp4jyCvV1ZXBbb8NdvwxgDHPWXxKPTc1EXuB7aEX3qzxM3/OWEUJg==}
     engines: {node: '>= 10'}
@@ -6192,6 +6214,14 @@ packages:
   /at-least-node@1.0.0:
     resolution: {integrity: sha512-+q/t7Ekv1EDY2l6Gda6LLiX14rU9TV20Wa3ofeQmwPFZbOMo9DXrLbOjFaaclkXKWidIaopwAObQDqwWtGUjqg==}
     engines: {node: '>= 4.0.0'}
+    dev: true
+
+  /axios@0.21.4:
+    resolution: {integrity: sha512-ut5vewkiu8jjGBdqpM44XxjuCjq9LAKeHVmoVfHVzy8eHgxxq8SbAVQNovDA8mVi05kP0Ea/n/UzcSHcTJQfNg==}
+    dependencies:
+      follow-redirects: 1.15.2(debug@4.3.4)
+    transitivePeerDependencies:
+      - debug
     dev: true
 
   /axios@1.3.4:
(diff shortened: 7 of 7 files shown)


.github/workflows/build.yml
```diff
@@ -19,6 +19,9 @@ jobs:
       - name: Setup
         uses: ./.github/actions/setup
 
+      - name: Build
+        run: pnpm nx run-many --target=build --parallel
```
roninjin10, line 23: pnpm nx --help
 WARN  Unsupported engine: wanted: {"node":"18.x"} (current: {"node":"v16.17.1","pnpm":"8.1.1"})
Smart, Fast and Extensible Build System

Commands:
  nx generate <generator> [_..]                    Generate or update source code (e.g., nx generate @nrwl/js:lib mylib). [aliases: g]
  nx run [project][:target][:configuration] [_..]  Run a target for a project
                                                   (e.g., nx run myapp:serve:production).

                                                   You can also use the infix notation to run a target:
                                                   (e.g., nx serve myapp --configuration=production)

                                                   You can skip the use of Nx cache by using the --skip-nx-cache option.
  nx run-many                                      Run target for multiple listed projects
  nx affected                                      Run target for affected projects
  nx affected:apps                                 Print applications affected by changes
                      [deprecated: Use `nx print-affected --type=app --select=projects` instead. This command will be removed in v15.]
  nx affected:libs                                 Print libraries affected by changes
                      [deprecated: Use `nx print-affected --type=lib --select=projects` instead. This command will be removed in v15.]
  nx affected:graph                                Graph dependencies affected by changes                [aliases: affected:dep-graph]
  nx print-affected                                Prints information about the projects and targets affected by changes
  nx daemon                                        Prints information about the Nx Daemon process or starts a daemon process
  nx graph                                         Graph dependencies within workspace                            [aliases: dep-graph]
  nx format:check                                  Check for un-formatted files
  nx format:write                                  Overwrite un-formatted files                                      [aliases: format]
  nx workspace-lint [files..]                      Lint nx specific workspace files (nx.json, workspace.json)
  nx workspace-generator [name]                    Runs a workspace generator from the tools/generators directory
                                                                                                        [aliases: workspace-schematic]
  nx migrate [packageAndVersion]                   Creates a migrations file or runs migrations from the migrations file.
                                                   - Migrate packages and create migrations.json (e.g., nx migrate
                                                   @nrwl/workspace@latest)
                                                   - Run migrations (e.g., nx migrate --run-migrations=migrations.json). Use flag
                                                   --if-exists to run migrations only if the migrations file exists.
  nx report                                        Reports useful version numbers to copy into the Nx issue template
  nx init                                          Adds nx.json file and installs nx if not installed already
  nx list [plugin]                                 Lists installed plugins, capabilities of installed plugins and other available
                                                   plugins.
  nx reset                                         Clears all the cached Nx artifacts and metadata about the workspace and shuts down
                                                   the Nx Daemon.                                               [aliases: clear-cache]
  nx connect                                       Connect workspace to Nx Cloud                        [aliases: connect-to-nx-cloud]
  nx repair                                        Repair any configuration that is no longer supported by Nx.
  nx view-logs                                     Enables you to view and interact with the logs via the advanced analytic UI from Nx
                                                   Cloud to help you debug your issue. To do this, Nx needs to connect your workspace
                                                   to Nx Cloud and upload the most recent run details. Only the metrics are uploaded,
                                                   not the artefacts.
  nx exec                                          Executes any command as if it was a target on the project
  nx watch                                         Watch for changes within projects, and execute commands
  nx show <object>                                 Show information about the workspace (e.g., list of projects)

Options:
  --help     Show help                                                                                                       [boolean]
  --version  Show version number

roninjin10, line 23:
```suggestion
        run: pnpm nx run-many --target=build --parallel 3
```
Meant to put a 3 here

.github/workflows/test.yml
```diff
@@ -18,7 +18,7 @@ jobs:
         uses: ./.github/actions/setup
 
       - name: Run tests
-        run: pnpm test
+        run: pnpm nx run-many --target=test --parallel
```
roninjin10, line 21: should have a 3

.github/workflows/test.yml
```diff
@@ -18,7 +18,7 @@ jobs:
...
 
       - name: Lint
-        run: pnpm lint
+        run: pnpm nx run-many --target=lint --parallel
```
roninjin10, line 24: should have a 3

nx.json
```diff
@@ -0,0 +1,54 @@
+{
+  "tasksRunnerOptions": {
+    "default": {
+      "runner": "@nrwl/nx-cloud",
```
roninjin10, line 4: This makes it so when main branch has new commits it caches the build to the cloud.  Any ci process or developer can then read from that cannonical remote cache

nx.json
```diff
@@ -0,0 +1,54 @@
...
+          "test",
+          "test:contracts",
+          "prettier",
+          "solhint",
+          "test:forge",
+          "test:hardhat",
+          "test:solidity",
+          "test:typescript",
+          "build:solidity",
+          "build:typescript"
```
roninjin10, line 18: This is the list of operations nx should check cache for.    

For case of build it will automatically use cached outputs
In case of tests or lint it will just cache whether the tests passed or not

nx.json
```diff
@@ -0,0 +1,54 @@
+{
```
roninjin10, line 1: This root level nx.json configures stuff project wide. 

We can also add nx files to individual packages to do more granular configuration

nx.json
```diff
@@ -0,0 +1,54 @@
...
+          "test:hardhat",
+          "test:solidity",
+          "test:typescript",
+          "build:solidity",
+          "build:typescript"
+        ]
+      }
+    }
+  },
+  "targetDefaults": {
```
roninjin10, line 23: These are default configurations that apply to all packages.   Individual packages can override them.

nx.json
```diff
@@ -0,0 +1,54 @@
...
+    "dev": {
+      "dependsOn": ["^build"]
```
roninjin10, line 25: The `^build` means it depends on dependencies builds finishing before it can run
`build` without the ^ would mean it depends on THIS packages builds not dependencies

nx.json
```diff
@@ -0,0 +1,54 @@
...
+    },
+    "build": {
+      "dependsOn": ["^build"],
+      "outputs": ["{projectRoot}/dist"]
+    },
```
roninjin10, line 30: Build depends on dependencies builds and it outputs to `dist`.  This is currently not true for all packages and we need to override this config in those packages or make them output only to dist whichever is easiest on a per package basis.

nx.json
```diff
@@ -0,0 +1,54 @@
...
+      "dependsOn": ["^build", "build"]
+    },
+    "test:solidity": {
+      "dependsOn": ["^build", "build"]
+    },
+    "test:typescript": {
+      "dependsOn": ["^build", "build"]
+    }
+  },
+  "defaultBase": "main"
```
roninjin10, line 53: `nx affected test` or `nx affected foo` will run only the tests or builds that have changed since last main branch once ci starts caching main branch

package.json
```diff
@@ -58,13 +61,13 @@
...
+    "test": "pnpm nx run-many --target=test",
     "entry:dist": "lerna run prepack",
     "entry:src": "lerna run postpack",
-    "gas-report": "pnpm recursive run gas-report",
+    "gas-report": "pnpm nx run-many --target=gas-report",
     "codegen": "pnpm codegen:tablegen && pnpm codegen:worldgen",
-    "codegen:tablegen": "pnpm recursive run tablegen",
-    "codegen:worldgen": "pnpm recursive run worldgen"
+    "codegen:tablegen": "pnpm nx run-many --target=tablegen",
+    "codegen:worldgen": "pnpm nx run-many --target=worldgen"
```
roninjin10, line 70: nx has a lot of features but I tend to only use 3 day to day

`nx build` or `nx foo` is the same as `pnpm foo` but nx is smart enough to do things the script depends on first (like build sub dependencies) if it needs to.  Otherwise skip.  And it's shorter to type than pnpm foo.

`nx run-many --targets=foo,bar,baz` runs foo script bar script and baz script in parallel in every package

`nx affected foo` runs foo in only packages that have changed since main branch

packages/cli/package.json
```diff
@@ -16,7 +16,6 @@
     "directory": "packages/cli"
   },
   "scripts": {
-    "prepare": "pnpm build",
```
roninjin10, line 19: remove all prepare scripts for `yarn build`

.github/workflows/build.yml
holic: beware that foundry seems to have trouble with parallel tasks

nx.json
holic: thoughts on doing this without cloud for now?

.github/workflows/build.yml
holic: we prob need this step in release too

nx.json
holic: if we enumerated more outputs here and some package didn't put any files in those dirs, would that be okay?

e.g.
```
"outputs": ["{projectRoot}/dist", "{projectRoot}/out", "{projectRoot}/abi", "{projectRoot}/types"]
```

packages/cli/package.json
holic: we could prob just do this ahead of nx changes, just to make install experience better (noting that folks will have to manually run `pnpm build` after if they need built files)

nx.json
roninjin10: This actually isn't using cloud now.   IT doesn't start using cloud until you add an access token.   But it's set up to turn on cloud right away.

nx.json
roninjin10: Yup.   What we want is
- Very confident we overshot the outputs
- very confident we overshot the inputs too

Since safety is more important than overoptimizing cache hits

roninjin10: The other thing we need to find is anything that is actually an output but is outputed to folders captured by inputs.   E.g. if a step generates a config file in the root that could break the cache

.github/workflows/build.yml
roninjin10: That's why I added --paralell 3 here even though 3 is the default.  Safer to be explicit in CI.

@holic can foundry run before everything else?   e.g. can the build pipeline be like

pnpm install => pnpm nx run-many --target=prebuild --parallel 1 => pnpm nx run-many --target=build --parallel 3

holic:
I am curious how nx would tie into my approach with templates in https://github.com/latticexyz/mud/pull/599

since packages in templates are linked with `link:../../../../packages/*` but each template is its own workspace (and not part of the root workspace), I wonder if we can make nx aware of this part of the dependency graph so that it would know that building a template's client dir would need to build the root package's network package, for example

.github/workflows/build.yml
holic: seems like 3 is still too much for forge/solc
image

it might be possible to run all forge tasks first but we'd need to comb through our packages and make some consistent script changes between em, and some do have an autogen step before a forge build step, so we'd need to introduce three total steps: pre-forge build, forge build, post-forge build

how slow does this get if we just do `--parallel 1` everywhere for now?

roninjin10: I set parallel false in nx.json so it never runs parallel to start.  That's fixable later though

packages/cli/package.json
roninjin10: I think in general the strategy is to keep this pr a wip and pull out tiny safe prs from it into seperate prs

roninjin10:
> I am curious how nx would tie into my approach with templates in #599
> 
> since packages in templates are linked with `link:../../../../packages/*` but each template is its own workspace (and not part of the root workspace), I wonder if we can make nx aware of this part of the dependency graph so that it would know that building a template's client dir would need to build the root package's network package, for example

Yup nx is powerful enough to set up any build graph.  It has features like [named inputs](https://nx.dev/more-concepts/customizing-inputs#customizing-inputs-and-named-inputs)

.github/workflows/build.yml
dk1a: there are also some things (like tsgen) that can be run in parallel to all 3 forge steps, if that helps

nx.json
dk1a: `gas-report.json` is a root output in `store` and `world`

package.json
holic: do these need `pnpm` prefix?

packages/cli/package.json
holic: extracted it here: https://github.com/latticexyz/mud/pull/623

roninjin10:
Can't remove nx-cloud because that's the task runner.   By default it doesn't actually push to cloud but it is the thing that runs the local cache.   It's the default task-runner for nx

holic:
For nx to be effective and manageable, I think we're gonna need to create some consistency in our script/task naming between packages.

For example:
- `build` creates all artifacts for publishing and includes things like mud codegen, typechain, forge build, TS -> JS, etc.
- some number of common sub-build steps like `build:forge` that do specific/common things like `forge build`
  - each project might need to specify its task ordering/dependencies like "mud codegen before forge build" (so codegen'ed solidity can be compiled) and "typechain after forge build" (so ABIs are available)

Guessing the list will look something like:
- `build` includes some combo of everything below, depending on the package
- `build:mudcodegen` runs `mud codegen`
- `build:typechain` runs `typechain`
- `build:forge` runs `forge build`
- `build:js` runs whatever TS -> JS pipeline (sometimes vite, sometimes tsup, sometimes parcel)

roninjin10:
This nx.json that I am adding to optimism repo post bedrock.  Sharing it to explain namedInputs which can help clean up these config files, make them more maintainable, and give you a much smarter cache.

```js
{
  "defaultBase": "develop",
  "affected": {
    "defaultBase": "develop"
  },
  // implicit dependencies will break the cache of everything if they change
  "implicitDependencies": {
    "nx.json": "*"
  },
  "tasksRunnerOptions": {
    "default": {
      "runner": "@nrwl/nx-cloud",
      "options": {
        "cacheableOperations": [
          "lint",
          "build",
          "test",
          "build:contracts",
          "autogen:artifacts"
        ]
      }
    }
  },
  // named inputs are a way of adding a group of files to a name
  "namedInputs": {
    // configs in root of repo.  THese names are names I gave it they don't mean anything magical just variable names
    "configsWorkspace": [
      "{workspaceRoot}/package.json",
      "{workspaceRoot}/tsconfig.json"
    ],
    // configs in individual packages
    "configsProject": [
      "{projectRoot}/foundry.toml",
      "{projectRoot}/tsconfig.json",
      "{projectRoot}/hardhat.config.ts",
      "{projectRoot}/package.json"
    ],
    // these are the default files we want to bust the cache on
    "default": ["{projectRoot}/**/*", "configsWorkspace"],
    // optimism builds stuff into src directory in contracts-bedrock (shame!!!).  This file group I will use later to fix it
    "srcGenerated": [
      "{projectRoot}/src/contract-artifacts.ts",
      "{projectRoot}/src/contract-deployed-artifacts.ts"
    ],
   // this is what most src looks like.  Remove the .spec files ect.   Without this it will depend on literally everything
    "productionSrc": [
      "{projectRoot}/src/**/*",
      "!{projectRoot}/src/contract-artifacts.ts",
      "!{projectRoot}/src/contract-deployed-artifacts.ts",
      "!{projectRoot}/src/**/*.spec.ts"
    ],
    "productionContracts": ["{projectRoot}/contracts/**/*"],
    // combining filegroups into names aggregated file groups
    "production": ["productionSrc", "productionContracts"],
    "testing": ["configsWorkspace", "default"]
  },
  "targetDefaults": {
    "lint": {
      // now I can use my filegroupsin my inputs and outputs
      "inputs": ["{workspaceRoot}/.markdownlint.json", "default"]
    },
    "test": {
      "inputs": ["default", "testing", "^production"],
      "dependsOn": ["^build"]
    },
    "build:contracts": {
      "inputs": [
        "configsProject",
        "productionContracts",
        "^productionContracts"
      ],
      "dependsOn": ["^build"],
      "outputs": ["{projectRoot}/artifacts", "{projectRoot}/forge-artifacts"]
    },
    "autogen:artifacts": {
      "inputs": [
        "configsWorkspace",
        "configsProject",
        "productionContracts",
        "^productionContracts"
      ],
      "dependsOn": ["^build", "build:contracts"],
      "outputs": ["srcGenerated"]
    },
    "build": {
      "inputs": [
        "configsWorkspace",
        "configsProject",
        "production",
        "^production"
      ],
      "dependsOn": ["^build", "autogen:artifacts", "build:contracts"],
      "outputs": ["{projectRoot}/dist"]
    }
  }
}
```

holic:
I've started working on the build scripts in a new branch. Here's what I've got so far:

- `build:mudcodegen` eventually runs `mud codegen` but for now is `tablegen && worldgen && tsgen`
- `build:abi` runs `forge build` for abi files to be used by typechain
- `build:typechain` runs `typechain` to generate TS types
- `build:js` runs `tsup`

(Each `build:*` script also has a corresponding `clean:*` script, where applicable)

I'll update this list as I go.

PR #623, opened 2023-04-13
build: remove prepare in favor of build script

holic:
Speed up installs by removing `prepare`. For now we'll have to run `pnpm build` before using (this happened implicitly after install with `prepare`).

Pulled out of https://github.com/latticexyz/mud/pull/620 to reduce scope of that PR


alvrs (approved):

roninjin10:
nice!

PR #625, opened 2023-04-13
Allows components, systems, and initializers to be placed in subdirectories and specified in deploy.json

emersonhsieh:
Allows components, systems, and initializers to be placed in subdirectories. This helps with organizing components if there is a lot of them.

For example, "components/base/PositionComponent.sol" in the components directory will be imported as `import { PositionComponent, ID as PositionComponentID } from "components/base/PositionComponent.sol";` in `LibDeploy.sol`.

Sample deploy.json:

```
  "components": [
    "debug/CounterComponent",
    "base/GameConfigComponent",
    "base/PositionComponent",
    "base/TileComponent",
    "base/OwnedByComponent",
    "resource/IronComponent",
    "resource/CopperComponent",
    "resource/IridiumComponent",
    "crafted/BulletComponent",
    "crafted/FastMinerComponent"
  ]
```

### DIFF
--- a/packages/cli/src/contracts/LibDeploy.ejs
+++ b/packages/cli/src/contracts/LibDeploy.ejs
@@ -60,9 +60,9 @@ library LibDeploy {
     if (!_reuseComponents) {
       IComponent comp;
 <% components.forEach(component => { -%>
-
-      console.log("Deploying <%= component %>");
-      comp = new <%= component %>(address(result.world));
+      <% const componentName=component.split("/").pop(); %>
+      console.log("Deploying <%= componentName %>");
+      comp = new <%= componentName %>(address(result.world));
       console.log(address(comp));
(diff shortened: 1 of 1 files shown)


dk1a:
MUD2 doesn't have this problem, I'd suggest you use it; for MUD1 you can also look at https://github.com/latticexyz/mud/pull/309 (which was abandoned in favor of v2)

alvrs:
Thanks for this PR @emersonhsieh! Unfortunately this change alone is not sufficient for full support of subdirectories, since `system-types` and other CLI command also assume a flat setup. Changing all these would touch a lot of v1 code, and since v2 (currently in alpha) is very close a stable release, we're not planning to add new features to v1. As @dk1a mentioned, v2 doesn't have the problem this PR addresses for v1.

PR #626, opened 2023-04-14
build: allow forge to cache build

roninjin10:
- Don't rimraf out so forge can notice files haven't changed
- Add clean scripts


packages/solecs/package.json
```diff
@@ -10,8 +10,9 @@
     "directory": "packages/solecs"
   },
   "scripts": {
+    "clean": "rimraf out rimraf dist rimraf cache",
```
holic, line 13: missing && in these commands?

packages/world/package.json
```diff
@@ -14,10 +14,11 @@
     "directory": "packages/world"
   },
   "scripts": {
+    "clean": "rimraf abi && rimraf cache && rimraf dist",
```
holic, line 17: should this include out?

packages/store/package.json
```diff
@@ -14,10 +14,11 @@
     "directory": "packages/store"
   },
   "scripts": {
+    "clean": "rimraf abi && rimraf cache && rimraf dist",
```
holic, line 17: should this also include out?

holic:
iirc the clearing of out was to make sure old files got purged from the build output (and thus also typechain output)

cc @alvrs in case he can’t recall why

packages/store/package.json
roninjin10:
```suggestion
    "clean": "rimraf out && rimraf abi && rimraf cache && rimraf dist",
```

packages/world/package.json
roninjin10:
```suggestion
    "clean": "rimraf out && rimraf abi && rimraf cache && rimraf dist",
```

packages/solecs/package.json
roninjin10:
```suggestion
    "clean": "rimraf out && rimraf dist && rimraf cache",
```
wierd

package.json
```diff
@@ -48,10 +48,10 @@
...
     "prettier": "prettier --write 'packages/**/*.{ts,css,md,sol}'",
     "lint": "pnpm prettier:check && eslint . --ext .ts",
-    "release": "pnpm install && lerna publish --no-private --force-publish && pnpm retype:updateversion",
-    "release:ci": "lerna publish --no-private --force-publish --yes",
-    "release:canary": "lerna publish premajor --canary --no-private --force-publish",
-    "release:manual": "pnpm --recursive --no-bail run release",
+    "release": "pnpm recursive run clean && pnpm install && lerna publish --no-private --force-publish && pnpm retype:updateversion",
+    "release:ci": "pnpm recursive run clean && lerna publish --no-private --force-publish --yes",
+    "release:canary": "pnpm recursive run clean lerna publish premajor --canary --no-private --force-publish",
+    "release:manual": "pnpm recursive run clean pnpm --recursive --no-bail run release",
```
roninjin10, line 54:
```suggestion
    "release": "pnpm recursive run clean && pnpm install && lerna publish --no-private --force-publish && pnpm retype:updateversion",
    "release:ci": "pnpm recursive run clean && lerna publish --no-private --force-publish --yes",
    "release:canary": "pnpm recursive run clean && lerna publish premajor --canary --no-private --force-publish",
    "release:manual": "pnpm recursive run clean && pnpm --recursive --no-bail run release",
```

package.json
```diff
@@ -48,10 +48,10 @@
     "prettier:check": "prettier --check 'packages/**/*.{ts,css,md,sol}'",
     "prettier": "prettier --write 'packages/**/*.{ts,css,md,sol}'",
     "lint": "pnpm prettier:check && eslint . --ext .ts",
-    "release": "pnpm install && lerna publish --no-private --force-publish && pnpm retype:updateversion",
-    "release:ci": "lerna publish --no-private --force-publish --yes",
-    "release:canary": "lerna publish premajor --canary --no-private --force-publish",
-    "release:manual": "pnpm --recursive --no-bail run release",
+    "release": "pnpm recursive run clean && pnpm install && lerna publish --no-private --force-publish && pnpm retype:updateversion",
```
alvrs, line 51: not directly related to this PR, but before #623 we had a build step before `lerna publish` (implicit via the prepare script) to double check the build succeeds before attempting to publish. The reason for this was that otherwise lerna already created tags etc even if later building one of the packages failed, resulting in a messy manual clean up process.

I'm not sure if this is still the case with the current lerna version, but would prefer to re-add that build step to be sure (unless you know a better way to prevent this)

Also small nit: should we create a `clean`/`clean:forge` script and invoke it here to make these scripts slightly less verbose?

packages/solecs/package.json
```diff
@@ -10,8 +10,9 @@
     "directory": "packages/solecs"
   },
   "scripts": {
+    "clean": "rimraf out && rimraf dist && rimraf cache",
```
alvrs, line 13: I don't think `solecs` has a `dist` folder, but there are a couple other folders we could clean here:

```suggestion
    "clean": "rimraf out && rimraf cache && rimraf types && rimraf abi",
```

(`types` and `abi` is already cleaned by calling `dist` and `types` though, so not sure if relevant here)

I guess the main purpose of this is to remove the forge output/cache, so maybe we could just call it `clean:forge` and use the built-in `forge clean`?

```suggestion
    "clean:forge": "forge clean",
```

packages/std-contracts/package.json
```diff
@@ -10,8 +10,9 @@
     "directory": "packages/std-contracts"
   },
   "scripts": {
+    "clean": "rimraf out && rimraf abi && rimraf types && rimraf cache",
```
alvrs, line 13: same as in `solecs`, `abi` and `types` are cleared when calling `dist` and `types`, so we might not need them here and instead can call this `clean:forge` too?

```suggestion
    "clean:forge": "forge clean",
```

packages/store/package.json
```diff
@@ -14,10 +14,11 @@
     "directory": "packages/store"
   },
   "scripts": {
+    "clean": "rimraf out && rimraf abi && rimraf cache && rimraf dist",
```
alvrs, line 17: same as above
```suggestion
    "clean:forge": "forge clean",
```

packages/world/package.json
```diff
@@ -14,10 +14,11 @@
     "directory": "packages/world"
   },
   "scripts": {
+    "clean": "rimraf out && rimraf abi && rimraf cache && rimraf dist",
```
alvrs, line 17: same as above

```suggestion
    "clean:forge": "forge clean",
```

holic:
fwiw I am working on similar things over at https://github.com/latticexyz/mud/pull/633

happy to bring changes from here over to that PR (if I haven't already)

holic:
I think I've addressed all the use cases in https://github.com/latticexyz/mud/pull/633 so gonna close this in favor of the new PR!

PR #627, opened 2023-04-14
style: Add package.json linter

roninjin10:
Adds a really basic linter for package.json


.github/workflows/test.yml
```diff
@@ -22,3 +22,6 @@ jobs:
 
       - name: Lint
         run: pnpm lint
+
+      - name: Lint package.json
+        run: yarn lint:package-json && git diff --exit-code
```
roninjin10, line 27: should rename this sort-package-json or just use the npx command maybe

package.json
```diff
@@ -1,15 +1,47 @@
...
+    "docs": "pnpm docs:prepare && retype build",
+    "docs:dev": "retype watch .",
+    "docs:prepare": "lerna run docs && pnpm retype:updateversion",
+    "entry:dist": "lerna run prepack",
+    "entry:src": "lerna run postpack",
+    "foundryup": "curl -L https://foundry.paradigm.xyz | bash && bash ~/.foundry/bin/foundryup",
+    "gas-report": "pnpm recursive run gas-report",
+    "link:packages": "lerna exec 'pnpm link --global'",
+    "lint": "pnpm prettier:check && eslint . --ext .ts",
+    "lint:package-json": "npx sort-package-json package.json && npx sort-package-json packages/*/package.json",
```
roninjin10, line 25: I don't like this name looking at this now

.github/workflows/test.yml
```diff
@@ -22,3 +22,6 @@ jobs:
 
       - name: Lint
         run: pnpm lint
+
+      - name: Lint package.json
+        run: yarn sort-package-json
```
roninjin10, line 27:
```suggestion
        run: pnpm sort-package-json
```

packages/services/OPCraftDataSchema.json
alvrs, line 1: Weird that it updated this file and `typedoc.json` too instead of just `package.json`?

alvrs (dismissed):
Very nice! Couple questions, but nothing blocking

.github/workflows/test.yml
```diff
@@ -22,3 +22,6 @@ jobs:
 
       - name: Lint
         run: pnpm lint
+
+      - name: Lint package.json
+        run: pnpm sort-package-json && git diff --exit-code
```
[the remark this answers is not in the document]
roninjin10: oh nice didn't notice that action I'll use that instead to be consistent

packages/services/OPCraftDataSchema.json
roninjin10: I think this was just user error by me typing the command wrong but I'll revert these files and try again to test that

.github/workflows/test.yml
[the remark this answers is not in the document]
holic: ohh could easily swap to this command too, can’t recall why I landed on the git diff approach I am using

holic (approved):

PR #629, opened 2023-04-14
feat(common-codegen): create package

dk1a:
this package is meant for some utilities from `render-solidity`/`render-ts`
not sure what's happening with pnpm-lock.yaml, I just ran `pnpm install` with pnpm@8.2.0


packages/codegen-utils/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "@latticexyz/codegen-utils",
```
alvrs, line 2: nit - should we call this package `common-codegen`? (Also wanna rename the `utils` package to `common`, so it makes more sense for it to also include shared types etc in addition to "utils" in the sense of utility functions)

packages/codegen-utils/tsconfig.vitest-temp.json
alvrs, line 1: ooc why do we need a separate tsconfig for vitest?

packages/codegen-utils/vite.config.ts
```diff
@@ -0,0 +1,27 @@
...
+    sourcemap: true,
+  },
+  /**
+   * @see https://vitest.dev/config/
+   */
+  test: {
+    environment: "node",
+    globals: true,
+  },
+  plugins: [(typescript as any)(), peerDepsExternal()],
```
alvrs, line 26: why do we need to cast to `any` here?

pnpm-lock.yaml
alvrs, line 1: weird, @holic @roninjin10 any idea what might cause this?

alvrs (commented):
nothing really blocking, but I'd like to wait for @holic or @roninjin10 to have a look at what might cause the `pnpm-lock.yaml` diff

packages/codegen-utils/tsconfig.vitest-temp.json
dk1a: we don't, removed and gitignored it. That seems to be a weird consequence of using `--passWithNoTests`

packages/codegen-utils/vite.config.ts
dk1a: I just copied from @roninjin10's PR, the plugin has bad types for some reason

roninjin10 (commented):
lockfile just looks liek prettier formatted it

alvrs (dismissed):
Happy to merge as is, but would you mind removing the `pnpm-lock` changes? Adding prettier formatting to .yaml files in https://github.com/latticexyz/mud/pull/629, which should hopefully normalise these changes before committing them

holic:
what pnpm version are you on?

holic (approved):

PR #630, opened 2023-04-14
style: run prettier on all files

alvrs:
Context in https://github.com/latticexyz/mud/pull/630#issuecomment-1509704464


holic (approved):
Did no yaml files get changed after this was enabled?

holic:
Oh if this is for pnpm yaml, we should prob ignore that in linting and let pnpm be the owner of formatting that

dk1a (dismissed):
It does fix my issue, but feels weird to use prettier for pnpm-lock

alvrs:
yeah the reason for this PR was the inconsistent formatting of the `pnpm-lock.yaml` file in https://github.com/latticexyz/mud/pull/629. Generally agree that pnpm should just be responsible for formatting that file, but for some reason formatting was different on linux, and prettier yaml formatting is the same as pnpm formatting on mac. So this PR adds prettier yaml formatting to the pre-commit hook as a last line of defence.

package.json
```diff
@@ -45,8 +45,8 @@
     "prepare": "husky install && (forge --version || pnpm foundryup)",
     "build": "pnpm recursive run build",
     "commit": "cz",
-    "prettier:check": "prettier --check 'packages/**/*.{ts,css,md,sol}'",
-    "prettier": "prettier --write 'packages/**/*.{ts,css,md,sol}'",
+    "prettier:check": "prettier --check '**/*.{ts,css,md,sol,yaml}'",
```
holic, line 48: should remove yaml from here too

Issue #615, opened 2023-04-12
Clean up build scripts


holic:
We should have a consistent set of `package.json` scripts for reliable building across the monorepo. Right now, these are spread out among a few different things and doesn't consistently capture everything.

- `codegen`: should this regenerate all code? or just run mud-specific codegen? we have things like tablegen, worldgen, tsgen (hopefully being consolidated in #590) but we also have things like typechain and ts-proto
- `build`: should build everything needed to use the package locally (including codegen) and mirrors what the package from npm might look like
  - this should probably include compiling ts down to js?

If we also consolidated "steps" like "compile TS to JS", it'd help us verify that they're all using the same tooling (e.g. vite, see #612)
- `build:ts`: should compile ts to js
- `build:abi`: should build with forge and output what we need for typechain

I'm sure I am missing a bunch, but just wanted to start the convo and getting notes down.

alvrs:
Some more thoughts:

- Some codegen commands are required for the build step (`tablegen`, `worldgen`), some have to run after the build step (`typechain`, `ts-proto`) and for some it doesn't matter `tsgen`.
- I agree that `build` should run all the necessary steps to run the package, and therefore should probably include the `codegen` steps that are required before build
- I think we should:
  * Move `tablegen` and `worldgen` to Store and World (see #616) so World and Store don't depend on CLI anymore
  * Add a `codegen` command to CLI that runs `tablegen`, `worldgen` and `tsgen` (all scripts must/can run before the build step)
  * Make sure `codegen` never depends on build output so it can always run before the build step

holic:
> and for some it doesn't matter `tsgen`
worth noting that `tsgen` is used to generate types for components, which should be type checked along other code, so should prob be included to be part of tests (but you're right that order doesn't matter inside the build step)

alvrs:
Linking here for context: https://github.com/latticexyz/mud/pull/620#issuecomment-1507767752

holic:
working on this in https://github.com/latticexyz/mud/pull/633

Issue #616, opened 2023-04-12
Move `tablegen` to `Store` and `worldgen` to `World` to resolve circular dependencies


alvrs:
Context in https://github.com/latticexyz/mud/pull/613#discussion_r1164263579_ :

Using `tablegen` here caused an error in the CI, because `../cli/dist/mud.js` does not exist yet while store is being built (CLI can't be built first because it depends on `World`'s ABI, which depends on `Store`). There are (at least) two approaches to solve this:
1. Move the `tablegen` scripts into `Store` (and `worldgen` scripts into `World`), so these packages can use them locally (and consumers of these packages don't have to install the MUD CLI if they only want to use eg `Store` without the rest of the framework). CLI can still expose a `tablegen` command by importing the scripts from `Store` (and a `worldgen` command by importing from `World`)
   - this may require moving common utils (render solidity) into a common package
2. Check in the World/Store ABIs, so CLI can be built independently from building World/Store (this alone doesn't fix the circular dependency though)

holic:
I burned a bunch of time today on the circular dependency (cli <> world/store) issue again while working on https://github.com/latticexyz/mud/pull/620. So I'm eager to get this fixed so we don't have to fight this anymore.

holic:
done in https://github.com/latticexyz/mud/commit/60e62ebf70b0fe2fce39cd15b3d5b46120c34dd4

PR #633, opened 2023-04-17
build: consistent build/clean scripts

holic:
Fixes https://github.com/latticexyz/mud/issues/615

This will help us create a dependency graph of scripts, better caching, etc. (See #620)

The format I took with these build steps is to keep them focused on their output (e.g. `build:js` is for TS -> JS). These are also in execution order.

- `build:mud` runs mud-specific codegen
  - eventually runs `mud codegen` but for now is `tablegen && worldgen && tsgen` (see #616)
- `build:abi` runs `forge build --extra-output-files abi --out abi --skip test script` for abi files to be used by typechain
  - this is separate from a regular `forge build` in that it 1) also generates abi files, for better imports 2) outputs to a separate directory (for easier typechain targeting) and 3) skips generating files for tests/scripts to avoid cluttering typechain
- `build:typechain` runs `typechain` to generate typed ethers contracts/interfaces
- `build:wasm` runs `asc` (just for noise package)
- `build:js` runs `tsup` or `rollup` (see #612)

Each `build:*` script also has a corresponding `clean:*` script, where applicable. Each package also has a corresponding ` build` and `clean` that is a combo of any of the above (depending on what the package is doing). I expect `build` will only be run locally, and nx will use the steps themselves (to resolve dependent steps and run in parallel)  (see #620)

Some other things to note:
- I moved one-off scripts (store's tightcoder, cli's test tables) out of the build pipeline, since they should only be run on an as-needed basis. These should be validated as up-to-date by a future lint step (like we do for gas reports).
- `clean:abi` uses `rimraf abi` instead of `forge clean` because the output directory is separate from the one configured by `foundry.toml` (used by tests, scripts)

---

TODO:
- packages
  - [x] root package
  - [x] cli
  - [x] common-codegen
  - [x] config
  - [x] create-mud
  - [x] ecs-browser
  - [x] network
  - [x] noise
  - [x] phaserx
  - [x] react
  - [x] recs
  - [x] schema-type
  - [x] solecs
  - [x] std-client
  - [x] std-contracts
  - [x] store
  - [x] utils
  - [x] world
  - [x] docs
- [x] fix circular dependencies (#616) to fix build
- [x] after circ deps fix, update `../cli/dist/mud.js` to `mud` with proper dep
  - [x] if we still can't, may need to add back in `chmod +x` step
- [ ] update templates to match?
- [x] move store's custom codegen into a one-off script and out of the build/test pipeline?
- [x] do we need a `build:forge` for general building of solidity (i.e. tests, scripts) or let it get handled by e.g. `forge test`? do we need a `clean:forge` for `out` dir?
- [ ] rethink prepack/postpack so that pack failures don't leave bad files around, or remove the need for it entirely
- [ ] output typechain to `./typechain` instead of default `./types/ethers-contracts`?
- [x] do solecs, std-contracts still need to do the out -> exports.txt -> abi conversion (prev `dist` task)?
- [x] can we combine `build:codegen` and `build:mudcodegen` steps? I mostly separated because 1) some mud codegen depended on local cli to be build (thus diff dep tree) and 2) to separate from other types of codegen (e.g. typechain), maybe we move it all into a `build:mud` step?
- [x] std-client can't clean up its mud definitions because it has a `.test-d.ts`  in the same dir, consider relocating?
- [x] remove `build:abi` step from schema-type? not currently in use
- [ ] consider foundry profiles for `abi` output vs regular `out`
  - these are separated because tests/scripts need the full compiled output and abi output only builds what is needed to export abis for downstream use
- [x] make sure docs are not part of the build pipeline


packages/std-client/package.json
```diff
@@ -8,16 +8,21 @@
     "directory": "packages/std-client"
   },
   "license": "MIT",
+  "type": "module",
```
holic, line 11: I changed this so I could move `codegen.mts` to `codegen.ts` and align this with the rest of our packages. I ran `templates/minimal` with this and it seemed to work just fine.

packages/cli/package.json
```diff
@@ -18,15 +18,15 @@
...
   "scripts": {
-    "build": "pnpm run build:js && chmod +x ./dist/mud.js",
-    "build:js": "vite build",
-    "codegen": "tsx ./scripts/codegen.ts",
+    "build": "pnpm run build:js",
+    "build:js": "vite build && chmod +x ./dist/mud.js ./dist/mud2.js",
+    "clean": "pnpm run clean:js",
+    "clean:js": "rimraf dist",
     "dev": "vite build --watch",
+    "generate-test-tables": "tsx ./scripts/generate-test-tables.ts",
```
holic, line 26: This looked more like a one-off script so I moved it out of the build path (doesn't need to be regenerated each time and is expected to be run locally and changes committed).

packages/store/package.json
```diff
@@ -16,23 +16,21 @@
...
-    "build:tightcoder": "tsx ./ts/scripts/codegenTightcoder.ts && prettier --write '**/*.sol'",
+    "build:mud": "tsx ./ts/scripts/tablegen.ts",
     "build:typechain": "typechain --target ethers-v5 'abi/**/*.sol/!(*.abi).json'",
     "clean": "pnpm run clean:abi && pnpm run clean:typechain && pnpm run clean:js",
     "clean:abi": "rimraf abi",
     "clean:js": "rimraf dist",
     "clean:typechain": "rimraf types",
     "docs": "rimraf API && hardhat docgen",
     "gas-report": "../cli/dist/mud.js gas-report --path test/* --path test/**/* --save gas-report.json",
+    "generate-tightcoder": "tsx ./scripts/generate-tightcoder.ts && prettier --write '**/tightcoder/*.sol'",
```
holic, line 30: This looked more like a one-off script so I moved it out of the build path (doesn't need to be regenerated each time and is expected to be run locally and changes committed).

packages/store/package.json
```diff
@@ -16,23 +16,21 @@
...
     "lint": "solhint --config ./.solhint.json 'src/**/*.sol'",
     "release": "npm publish --access=public",
-    "test": "pnpm run test:solidity && pnpm run test:typescript",
-    "test:solidity": "forge test",
-    "test:typescript": "tsc --noEmit"
+    "test": "tsc --noEmit && forge test"
```
holic, line 33: I'm planning to come back to this in https://github.com/latticexyz/mud/issues/637

pnpm-workspace.yaml
```diff
@@ -1,3 +1,2 @@
 packages:
   - packages/*
-  - docs
```
holic, line 3: I moved this out for now to avoid running `build` on docs when it doesn't depend on anything in the workspace.

Instead, we may want to treat this as a separate app and adjust our `docs` scripts throughout to generate files into this `docs/`.

packages/store/package.json
```diff
@@ -16,23 +16,21 @@
...
     "build:js": "vite build",
-    "build:tightcoder": "tsx ./ts/scripts/codegenTightcoder.ts && prettier --write '**/*.sol'",
+    "build:mud": "tsx ./ts/scripts/tablegen.ts",
     "build:typechain": "typechain --target ethers-v5 'abi/**/*.sol/!(*.abi).json'",
     "clean": "pnpm run clean:abi && pnpm run clean:typechain && pnpm run clean:js",
     "clean:abi": "rimraf abi",
     "clean:js": "rimraf dist",
     "clean:typechain": "rimraf types",
     "docs": "rimraf API && hardhat docgen",
     "gas-report": "../cli/dist/mud.js gas-report --path test/* --path test/**/* --save gas-report.json",
```
holic, line 29: 😭  Still need CLI here, but at least this isn't in the build path. I tried to add CLI as a dep, but that creates circular deps again. I also tried to add CLI as a dep in the root package, but there are no bins to link when it attempts to install from the workspace.

We could use whatever npm version is published, but that seems not ideal if we ever want to test changes to the gas report and want to use the local/workspace version.

We could also move the gas report out to common?

holic:
Ready for review! I left a few non-blocking TODOs undone, so we can talk about em and decide if we should create follow up issues/PRs.

In terms of the build scripts defined above (and in this PR), is it worth adding a little CI script to validate that no new types of `build:*` scripts are added (without also extending the CI check)? This would at least make it a little more clear about the intent behind these.

packages/cli/package.json
dk1a: I agree, though it's easy to forget to run this (and this one changes often, as it's meant to test any changes to tablegen), which is why it was part of `test`, even though it doesn't really fit there too well.
If not part of `test`, it would be nice to have it somewhere else in CI at least

packages/store/package.json
dk1a: the tablegen treatment for gas-report would make sense imo, except yeah it'd go to common rather than store/world. And cli just gets a wrapper over the imported script

packages/store/package.json
dk1a: this doesn't really change/matter, but if we have some localgen script for `generate-test-tables`, tightcoder could also use it

dk1a (dismissed):
lgtm, though I would prefer to not put off automating `generate-test-tables` in some way as it changes too often and cli's tests rely on it (making it part of the build/test pipeline again temporarily would be easiest I guess)

packages/cli/package.json
holic: added back into `test` script for now, will refactor in a follow up as part of #637

packages/store/package.json
holic: added an issue to track this here: https://github.com/latticexyz/mud/issues/667

dk1a (approved):

PR #635, opened 2023-04-17
refactor(docs): migrate to Nextra

alvrs:
WIP


alvrs:
Previously we were building (=autogenerating from code comments) docs in the `docs` github action, but with this approach we can't use a regular vercel integration. So I changed the approach to check in the autogenerated docs in the `docs` folder, and changed the docs action to check if they are up to date (the same approach as we use with other checked in generated files like gas-reports or tables). However `typedoc` includes a link to the github commit in the autogenerated docs, which makes it impossible to check if the files are up to date in the github action, since some files will always have a changed reference to the latest commit (eg the commit that tried to update the references to the latest commit before that).

Since in the medium term I'd like to move away from typedoc / generated api docs anyway, I think a good intermediate solution is to check in the docs and skip the CI action that verifies whether these docs are up do date.

docs/pages/index.mdx
```diff
@@ -0,0 +1,124 @@
+# MUD
```
alvrs, line 1: For now this file is just a copy of the root level README.md. Soon we'll have separate specific front page for mud.dev.

holic (approved):
🎉🎉

holic:
are the API docs still using hardhat stuff to autogenerate? or can we move to `forge doc`?

README.md
```diff
@@ -60,7 +60,7 @@ MUD is MIT-licensed, open source and free to use.
 pnpm create mud my-project
 ```
 
-![Scaffolding a new project with the MUD CLI.](./public/docs/mud-create.gif)
+![Scaffolding a new project with the MUD CLI.](./docs/public/docs/mud-create.gif)
```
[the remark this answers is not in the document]
alvrs: good catch!
```suggestion
![Scaffolding a new project with the MUD CLI.](./docs/public/mud-create.gif)
```

README.md
```diff
@@ -34,7 +34,7 @@ MUD is MIT-licensed, open source and free to use.
 
 ## Features
 
-![MUD features](./public/docs/features.png)
+![MUD features](./docs/public/docs/features.png)
```
alvrs, line 37:
```suggestion
![MUD features](./docs/public/features.png)
```

alvrs:
> are the API docs still using hardhat stuff to autogenerate? or can we move to forge doc? (doesn't have to happen here, just curious)

Right now yes, this PR doesn't change how the individual package's docs are generated. Down to switch to `forge docs` so we don't need the double setup, but would keep that for a follow up PR

docs/pages/index.mdx
alvrs: I didn't find an option to specify a default theme in the docs, i guess we could hack it by eg. setting the `theme` key in the local storage, but we could also just wait to publish the nextra docs under [mud.dev](https://mud.dev) until we have the new design (right now they're under [nextra.mud.dev](https://nextra.mud.dev))

alvrs:
Only change since last approval: [8fbcbdc](https://github.com/latticexyz/mud/pull/635/commits/8fbcbdc6db84b20d7f09799a43393243d9264007)

holic:
did the above in https://github.com/latticexyz/mud/pull/654

PR #639, opened 2023-04-17
build: update canary preid

holic:
Our canary release pipeline assumed the next release would be a major version, so when we did a manual minor release, the alpha counter reset:

image

In ~30 main commits, this will conflict with previously released alphas.

We'll have probably lots of minor/patch releases until 2.0 (and after), so let's switch canary/alpha releases to use prepatch. It won't exactly follow semver but at least we won't have to fight this issue every manual release.


alvrs (approved):
This should work! Tested locally (without `--yes` which is appended in the action):

CleanShot 2023-04-18 at 15 48 50@2x

PR #640, opened 2023-04-18
refactor: move tablegen to store and worldgen to world, update javascript build/export setup

dk1a:


packages/cli/package.json
```diff
@@ -9,34 +9,39 @@
   },
   "license": "MIT",
   "type": "module",
-  "main": "dist/index.js",
-  "types": "dist/index.d.ts",
+  "exports": {
+    ".": "./dist/index.js"
+  },
+  "main": "./src/index.ts",
```
alvrs, line 15: The interesting thing with this setup (main pointing to `src/index.ts`, exports pointing to `dist/index.js`) is that TypeScript updates types immediately during development (because it looks at `main` and ignores `exports`), but bundlers look at `exports` and resolve to the build file, so we might not have to switch around main entries (between js/ts) when releasing the package but still have the same local devex

alvrs:
The original reason for this PR was to refactor tablegen/worldgen to be in store/world respectively. Most of this was done in the first commit: https://github.com/latticexyz/mud/pull/640/commits/9179ed89d7c933fd752d5656296c2f3cf7a39f10

Unfortunately this refactor uncovered a bunch of inconsistencies and issues with out javascript build/export setup, so the rest of this PR was aligning `package.json`, bundlers, settings to fix these issues.

The streamlined setup uses `vite` as a bundler (#612), `exports` in `package.json` to declare the package's build exports (which enables code splitting), `types`/`typesVersions` to point to the source files (for Typescript support), and aligned build commands (`build` = `build:js + build:codegen` + build:abi + build:typechain` à la #615)

The setup of the following packages has been update to the new setup:
- schema-type
- config
- common
- cli
- world
- store

The setup of the following packages has not been updated to not increase the scope of this PR even more:
- recs
- react
- esc-browser (deprecated as soon as we have the new dev dashboard)
- network (deprecated as soon as the new sync package is ready)
- noise
- phaserx
- services
- std-client (deprecated soon)
- std-contracts (deprecated soon)
- solecs (deprecated soon)
- utils (deprecated soon in favor of common)

packages/common/package.json
```diff
@@ -0,0 +1,50 @@
...
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/latticexyz/mud.git",
+    "directory": "packages/common"
+  },
+  "license": "MIT",
+  "type": "module",
+  "exports": {
+    "./codegen": "./dist/codegen/index.js",
+    "./foundry": "./dist/foundry/index.js"
```
holic, line 14: 👏  this is nice, so we can import just what we need

I wonder if we should have some policy around deps for this package, to keep it slim/focused

packages/common/package.json
```diff
@@ -0,0 +1,50 @@
...
+        "./src/foundry/index.ts"
+      ]
+    }
+  },
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "build": "vite build",
+    "release": "npm publish --access=public",
```
holic, line 31: I noticed this sprinkled everywhere - is this needed?

packages/common/src/codegen/render-solidity/userType.ts
```diff
@@ -1,25 +1,25 @@
 import { AbiTypeToSchemaType, getStaticByteLength, SchemaType, SchemaTypeToAbiType } from "@latticexyz/schema-type";
 import { StoreConfig, parseStaticArray } from "@latticexyz/config";
-import { RelativeImportDatum, RenderTableType } from "./types.js";
+import { RelativeImportDatum, RenderType } from "./types.js";
```
holic, line 3: I noticed in some areas we're able to remove the `.js` and others not, is this intentional?

packages/config/package.json
```diff
@@ -9,9 +9,10 @@
   },
   "license": "MIT",
   "type": "module",
-  "main": "dist/index.js",
```
holic, line 12: does the removal of `main` in these packages mean we can import plainly as TS? or does importing end up using the JS build?

packages/config/package.json
```diff
@@ -21,16 +22,15 @@
     "@latticexyz/schema-type": "^1.42.0",
     "chalk": "^5.2.0",
     "esbuild": "^0.17.15",
-    "ethers": "^6.3.0",
+    "ethers": "^5.7.2",
```
holic, line 25: oof, that would have been bad

packages/store/package.json
```diff
@@ -8,43 +8,57 @@
...
   },
-  "types": "./types/ethers-contracts/",
+  "types": "ts/index.ts",
   "scripts": {
-    "build": "pnpm codegen && rimraf out && forge build && pnpm dist && pnpm types && tsup",
-    "codegen": "tsx ./scripts/codegen.ts && prettier --write '**/*.sol'",
-    "dist": "rimraf abi && mkdir abi && cp -rf out/*.sol/*.json abi/ && pnpm rimraf abi/*.metadata.json",
+    "build": "pnpm run build:js && pnpm run build:tightcoder && pnpm run build:codegen && pnpm run build:abi && pnpm run build:typechain",
+    "build:abi": "forge build --out abi --skip test script",
+    "build:codegen": "node ./dist/tablegen.js",
```
holic, line 21: should we use `tsx` and the source here?

packages/store/.npmignore
```diff
@@ -2,6 +2,7 @@
 
 !abi/**
 !src/**
+!ts/**
```
holic, line 5: why is this ignored?

holic (dismissed):
looks great! thanks everyone for slogging through this mess to get this over the finish line!

we should probably standardize how we organize packages with both solidity exports and ts/js exports

this PR now adds a `ts` dir to packages that previously only had solidity in `src`, which is different from e.g. schema-type that uses `src/typescript` and `src/solidity`

edit: added an issue to follow up here: https://github.com/latticexyz/mud/issues/652

holic:
see https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c for CJS vs ESM

packages/world/ts/scripts/tablegen.ts
```diff
@@ -0,0 +1,9 @@
+import path from "path";
+import { loadStoreConfig } from "@latticexyz/config";
+import { tablegen } from "@latticexyz/store";
+import { getSrcDirectory } from "@latticexyz/common/foundry";
+
+const config = await loadStoreConfig();
+const srcDir = await getSrcDirectory();
+
+tablegen(config, path.join(srcDir, config.codegenDirectory));
```
dk1a, line 9: also missing await?

packages/common/package.json
alvrs: you mean the `release` script? It's used by `release:manual` in the root package.json, which is useful if a lerna release fails midway (after increasing all versions etc) to just push the current state to npm

packages/common/src/codegen/render-solidity/userType.ts
alvrs: we should be able to remove it everywhere now!

packages/config/package.json
alvrs: in vscode it's gonna resolve types immediately (via the `types` key), but during build it's gonna resolve to `dist/index.js` (via the `exports` key). This means to run tests etc we need to build first (which should be less of a pain once nx is added)

packages/common/package.json
holic: in that case (assuming all `release` scripts are the same in each package), we can just recursively exec this same command but from the root package

dk1a (commented):
haha now I'm getting a failed build locally
Screenshot 2023-04-20 181336

packages/common/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "type": "git",
+    "url": "https://github.com/latticexyz/mud.git",
+    "directory": "packages/common"
+  },
+  "license": "MIT",
+  "type": "module",
+  "exports": {
+    "./codegen": "./dist/codegen/index.js",
+    "./foundry": "./dist/foundry/index.js"
+  },
```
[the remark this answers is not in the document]
holic: only benefit I can see is if we need to introduce heavy deps, splitting out one of the "subpackages" into its own proper package may make sense

dk1a:
also `formatAndWriteSolidity` is missing await in `store/ts/render-solidity/tablegen.ts` seems to be missing await, but that's not even related to this PR. I guess I forgot it but things still work cause those functions are quick

packages/store/package.json
alvrs: done in https://github.com/latticexyz/mud/pull/640/commits/77487c03bc674acebf1d7501810259c128e4a256

packages/common/src/codegen/render-solidity/userType.ts
alvrs: done in https://github.com/latticexyz/mud/pull/640/commits/77487c03bc674acebf1d7501810259c128e4a256

packages/std-client/scripts/codegen.mts
```diff
@@ -1,8 +1,8 @@
 import { parseStoreConfig } from "@latticexyz/config";
-import { tsgen } from "@latticexyz/cli/dist/render-ts";
+import { tsgen } from "@latticexyz/cli";
```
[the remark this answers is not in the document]
alvrs: sounds good!

packages/store/ts/scripts/tablegen.ts
```diff
@@ -0,0 +1,9 @@
+import path from "path";
+import { loadStoreConfig } from "@latticexyz/config";
+import { getSrcDirectory } from "@latticexyz/common/foundry";
+import { tablegen } from "../render-solidity";
+
+const config = await loadStoreConfig();
+const srcDir = await getSrcDirectory();
+
+tablegen(config, path.join(srcDir, config.codegenDirectory));
```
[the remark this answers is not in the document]
alvrs: good catch

packages/store/ts/modules.d.ts
```diff
@@ -0,0 +1 @@
+declare module "prettier-plugin-solidity";
```
dk1a, line 1: I think this shouldn't be needed in `store` and `world`, doing this in just `common` should be sufficient, as the other packages don't even directly install this plugin (at least locally for me it seems to work)

packages/store/package.json
```diff
@@ -8,43 +8,57 @@
...
+    "./*": "./dist/*",
+    "./abi/*": "./abi/*"
   },
-  "types": "./types/ethers-contracts/",
+  "types": "ts/index.ts",
   "scripts": {
-    "build": "pnpm codegen && rimraf out && forge build && pnpm dist && pnpm types && tsup",
-    "codegen": "tsx ./scripts/codegen.ts && prettier --write '**/*.sol'",
-    "dist": "rimraf abi && mkdir abi && cp -rf out/*.sol/*.json abi/ && pnpm rimraf abi/*.metadata.json",
+    "build": "pnpm run build:js && pnpm run build:tightcoder && pnpm run build:codegen && pnpm run build:abi && pnpm run build:typechain",
```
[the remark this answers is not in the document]
dk1a: right this wasn't part of my original commit, but as I was wrangling CI I tried building scripts, which had to happen before codegen, and so `build:js` became first
So unless we move `"build:codegen": "node ./dist/tablegen.js",` to tsx (I think that had some issues?) it still has to go before codegen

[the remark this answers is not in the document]
dk1a: I don't think vite uses codegen's or typechain's results in any way, so it should be fine to just run the whole `build:js` 1st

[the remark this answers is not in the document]
holic: a step like this going before is totally fine, we're just going to either need to
- be careful about defining dependencies between different steps (`build:js` happens first here but last in other packages)
- separate the "build JS for internal use" and "build JS for external use"

PR #641, opened 2023-04-18
[moved to #805] feat(store-cache): add typescript datastore with support for composite keys

alvrs:
This PR adds a new client data store with support for composite keys, typed based on a given `StoreConfig`. The intention is for this to be used in combination with with the network package to mirror the state of a `Store` contract to the client. (Previously we used `recs` for this purpose, but `recs` does not support composite keys).

Todos in this PR:
- [x] add default values for each abi type
- [x] add mapping to primitive type for every abi type

Followup:
- [ ] add `recs` wrapper around a store cache (to be able to interact with contract ecs data via `recs` as before)
- [ ] update `setupMUDV2Network` to pipe contract data into store core rather than `recs`


packages/store-client/tests/Database.spec.ts
```diff
@@ -0,0 +1,84 @@
+import { describe, expect, it } from "vitest";
```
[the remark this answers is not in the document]
dk1a: +1 to `.test.ts`, vitest docs also recommend it. I use it in some of my recent PRs with ts tests

packages/store-cache/src/defaults.ts
```diff
@@ -0,0 +1,10 @@
+import { AbiType } from "@latticexyz/schema-type";
+
+// TODO: move to schema-type?
+// TODO: add default values for all abi types
+export const AbiDefaults = {
+  bytes: "0x00",
+  string: "",
+  uint256: 0n,
+  int32: 0,
+} as any satisfies { [key in AbiType]: unknown };
```
alvrs, line 10: does something like this exist in other packages? viem?

packages/store-cache/src/types.ts
```diff
@@ -0,0 +1,74 @@
...
+
+// TODO: add mappings for remaining Solidity types
+// TODO: move to schema-type
+type AbiTypeToPrimitiveTypeLookup = {
+  bytes32: `0x${string}`;
+  uint256: bigint;
+  int32: number;
+} & {
+  [_ in FieldData<string>]: unknown;
+};
```
alvrs, line 12: does something like this exist in other packages? viem? (we might have to extend the definition though to support mapping enums in `FieldData` to `uint8`/`number`)

packages/store-cache/src/types.ts
```diff
@@ -0,0 +1,74 @@
...
+  table: T & string;
+  key?: { [key in "gt" | "gte" | "lt" | "lte" | "eq"]?: Partial<Key<C, T>> };
+};
+
+// TODO: figure out how to turn this into a proper typed union (where typescript can infer the type of set/remove from the type of table)
+export type Update<C extends StoreConfig = StoreConfig, Table extends keyof C["tables"] = keyof C["tables"]> = {
+  table: Table;
+  set: KeyValue<C, Table>[];
+  remove: Key<C, Table>[];
+};
```
alvrs, line 74: I want to be able to do `if(update.table === "Position") update.set[0].value.x` and have typescript infer that `x` is a number. Right now it still treats `set` and `remove` as arrays with "a value of any table in the config" in that situation.

alvrs, line 74: I want to be able to do if(update.table === "Position") update.set[0].value.x and have typescript infer that x is a number. Right now it still treats set and remove as arrays with "a value of any table in the config" in that situation.

packages/store-cache/src/defaults.ts
holic: not that I know of

packages/store-cache/src/types.ts
holic: I believe you can use abitype's `AbiTypeToPrimitiveType`: https://abitype.dev/api/utilities.html#abitypetoprimitivetype

holic: this might also be helpful/relevant: https://abitype.dev/api/types.html#solidity-types

packages/store-cache/src/types.ts
dk1a: you can do something like this:
```typescript
type ConvertToUnion<T> = T[keyof T];

export type Update<C extends StoreConfig = StoreConfig, Table extends keyof C["tables"] = keyof C["tables"]> = ConvertToUnion<{
  [key in Table]: {
    table: key;
    set: KeyValue<C, key>[];
    remove: Key<C, key>[];
  }
}>;
```

packages/store-cache/src/utils.test.ts
```diff
@@ -0,0 +1,130 @@
...
+import { mudConfig } from "@latticexyz/store/register";
+import { subscribe } from "./utils";
+import { KeyValue } from "./types";
+
+const config = mudConfig({
+  tables: {
+    MultiKey: { primaryKeys: { first: "bytes32", second: "uint32" }, schema: "int32" },
+    Position: { schema: { x: "int32", y: "int32" } },
+  },
+} as const);
```
dk1a, line 12: `as const` here too

packages/store/ts/library/config/storeConfig.ts
```diff
@@ -32,7 +32,7 @@ const zUserEnumName = zObjectName;
 // (user types are refined later, based on the appropriate config options)
 const zFieldData = z.string();
 
-type FieldData<UserTypes extends StringForUnion> = AbiType | StaticArray | UserTypes;
+export type FieldData<UserTypes extends StringForUnion> = AbiType | StaticArray | UserTypes;
```
dk1a, line 35: this might be better solved by adding a `FieldType` aggregate to `schema-type`, and having `storeConfig` extend it with `UserTypes`.
You would then want to export/import `schema-types`'s version because I don't think anything outside of config should be aware of UserTypes, or have to deal with an unnecessary generic here

holic (commented):
So clean and nice 🎉  I'm glad that this ended up as a fairly lightweight wrapper with strong types and we could lean on another package to do the heavy lifting. Looking forward to having this!

None of my questions/feedback are strictly blocking and mostly thinking about the long-term maintainability here (API design choices, data format choices, etc.):
- where tables should live on the client object relative to other possible APIs
- how tables are referenced/used as keys (unclear if it's already namespace+name and think it prob should be)
- some more room to narrow/strongly type things
- how we serialize/deserialize values (bigints and other things)

packages/store-cache/src/utils.test.ts
```diff
@@ -0,0 +1,130 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import { createDatabase, createDatabaseClient } from ".";
+import { mudConfig } from "@latticexyz/store/register";
```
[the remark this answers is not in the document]
dk1a: rule of thumb is to use it only if you need `mudConfig` (which is the case here). It doesn't even export much else so you can't use it interchangeably with `"@latticexyz/store"` anyway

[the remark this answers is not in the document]
holic: but why wouldn't I use `mudConfig` from `@latticexyz/config`?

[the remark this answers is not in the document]
dk1a: after https://github.com/latticexyz/mud/pull/752 is merged, `@latticexyz/config` won't export `mudConfig`, so you can't. Because `mudConfig` depends on global types which kinda break if you don't register them properly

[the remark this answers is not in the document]
dk1a: and `@latticexyz/config` exports only `mudCoreConfig`, which is different and shouldn't really be used outside of mud internals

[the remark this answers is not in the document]
holic: I have lost the plot

packages/store/ts/library/config/storeConfig.ts
alvrs: tried this but it feels better to import the actual `FieldType` type from store than reconstructing it here (we need to account for enum strings here anyway by mapping them to `uint8`/`number`)

packages/store-cache/src/createDatabaseClient.ts
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Create a typed database client from a tuple database and a store config
+ */
+export function createDatabaseClient<C extends StoreConfig>(database: TupleDatabase, config: C) {
+  const _tupleDatabaseClient = new TupleDatabaseClient(database);
+  const client: Record<string, unknown> = { _tupleDatabaseClient };
+
+  for (const table in config.tables) {
+    client[table] = {
```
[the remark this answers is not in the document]
alvrs: yeah this is a good point! right now the types are only based on the local `mud.config.ts`. I think eventually we should add a cli command to load table schemas from an existing world into a local file to infer ts types from it. To prepare for that I agree it makes sense to take the namespace into account here (and also to allow the network package to figure out where to set the values to). Will add this.

packages/store-cache/src/createDatabaseClient.ts
```diff
@@ -0,0 +1,37 @@
...
+  SubscriptionFilterOptions,
+  Value,
+} from "./types";
+import { set, get, remove, getDefaultValue, subscribe } from "./utils";
+import { StoreConfig } from "@latticexyz/store";
+
+/**
+ * Create a typed database client from a tuple database and a store config
+ */
+export function createDatabaseClient<C extends StoreConfig>(database: TupleDatabase, config: C) {
```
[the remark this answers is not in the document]
alvrs: Currently only exposes the user defined tables as types, but I'll add a method to set values in any table so we can also store records of tables not defined in the local config.

packages/store-cache/src/utils.ts
```diff
@@ -0,0 +1,194 @@
...
+ * @returns Transaction
+ */
+export function set<C extends StoreConfig = StoreConfig, T extends keyof C["tables"] = keyof C["tables"]>(
+  client: TupleDatabaseClient,
+  table: T,
+  key: Key<C, T>,
+  value: Partial<Value<C, T>>,
+  options?: SetOptions
+) {
+  const keyTuple = [String(table), ...recordToTuple(key)];
```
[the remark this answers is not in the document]
alvrs: > why do we need to convert table to string here? I would have expected keyof C['tables'] to be a string already?

Just to maks ts happy (it thought `table` can be `string | number | symbol`), but instead of casting here I just made the `table` type stronger now

packages/store-cache/src/utils.ts
```diff
@@ -0,0 +1,194 @@
...
+      record[key] = value;
+    }
+  }
+  return record;
+}
+
+/**
+ * Helper to serialize bigints
+ */
+function flatten(value: unknown) {
```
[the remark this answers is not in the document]
alvrs: This function is exclusively used in `recordToTuple`, which is exclusively used to turn an object with many keys into an array of objects with one key each, which is the format expected by `tuple-database`. It's only used for "input types", so we don't need to turn the key back into a bigint value (at least for now - if we do need to, we could think about appending `n` to the string and then parsing that in a reverse function, but for now that would be unnecessary complexity).

The type is `unknown` because the function accepts any value and only serializes it if it's a value that's not supported by `tuple-database` (only `bigint` for now). We could just remove this function and do the conversion inline, but that would make the `recordToTuple` function a bit more complex. No strong opinion here though.

For now I changed the name to `serializeKey` and added a better comment to make this a bit more clear!

packages/store-cache/src/types.ts
```diff
@@ -0,0 +1,74 @@
...
+      filter?: Omit<SubscriptionFilterOptions<C, table>, "table">
+    ) => Unsubscribe;
+  };
+} & {
+  _tupleDatabaseClient: TupleDatabaseClient;
+};
+
+export type SetOptions = {
+  defaultValue?: Record<string, unknown>;
+  appendToTransaction?: TupleRootTransactionApi;
```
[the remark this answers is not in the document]
alvrs: changed it to `transaction`!

packages/store-cache/src/createDatabaseClient.ts
```diff
@@ -10,28 +10,40 @@ import {
...
-  const client: Record<string, unknown> = { _tupleDatabaseClient };
+  const { namespace } = config;
+  const tables: Record<string, unknown> = {};
 
+  // Create utils with client argument prefilled
+  const utilsWithClient = {
+    set: curry(set<C>, _tupleDatabaseClient),
+    get: curry(get<C>, _tupleDatabaseClient),
+    remove: curry(remove<C>, _tupleDatabaseClient),
+    subscribe: curry(subscribe<C>, _tupleDatabaseClient),
```
holic, line 28: omg, I think this is the util I've been looking for in templates/examples

packages/store-cache/src/createDatabaseClient.ts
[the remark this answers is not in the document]
alvrs: we now have the user defined tables including typescript types under `tables`, and also expose `set, get, remove, subscribe` one level higher with `namespace` and `table` as arguments to interact with tables that are not part of the user defined mud config (eg root tables, third party tables)

packages/store-cache/src/createDatabaseClient.ts
```diff
@@ -10,28 +10,40 @@ import {
...
       set: (key: Key<C, typeof table>, value: Value<C, typeof table>, options: SetOptions) =>
-        set(_tupleDatabaseClient, table, key, value, {
+        utilsWithClient.set(namespace, table, key, value, {
           defaultValue: getDefaultValue(config.tables?.[table].schema),
           ...options,
         }),
-      get: (key: Key<C, typeof table>) => get(_tupleDatabaseClient, table, key),
-      remove: (key: Key<C, typeof table>, options?: RemoveOptions) => remove(_tupleDatabaseClient, table, key, options),
+      get: curry(utilsWithClient.get, namespace, table),
+      remove: curry(utilsWithClient.remove, namespace, table),
```
holic, line 40: :chefkiss:

packages/store-cache/src/createDatabaseClient.ts
```diff
@@ -10,28 +10,40 @@ import {
...
+    subscribe: curry(subscribe<C>, _tupleDatabaseClient),
+  };
+
+  // Create utils with client, namespace and table argument prefilled
   for (const table in config.tables) {
-    client[table] = {
+    tables[table] = {
       set: (key: Key<C, typeof table>, value: Value<C, typeof table>, options: SetOptions) =>
-        set(_tupleDatabaseClient, table, key, value, {
+        utilsWithClient.set(namespace, table, key, value, {
```
holic, line 35: afaik, tables can have a namespace override in config, so we may need to figure out how to handle that or forbid that e.g. one config per namespace

(preference for the former)

holic: I guess that brings up another question: should this use the resolved/expanded config, rather than the user defined one, because of things like this?

Imagining a fully resolved/expanded config would have a namespace defined per table (table namespace if defined, otherwise config-level namespace, otherwise empty), then we'd just use something like `table.namespace` here

packages/store-cache/src/utils.ts
```diff
@@ -116,31 +119,37 @@ export function subscribe<C extends StoreConfig = StoreConfig, T extends keyof C
...
     for (const update of write.set ?? []) {
       // The first tuple element is always the table name
-      const table = update.key[0];
-      if (typeof table !== "string") {
-        console.warn("store-cache: Expected first tuple element to be the table id, ignoring set operation:", set);
+      const [namespace, table] = update.key;
+      if (typeof namespace !== "string" || typeof table !== "string") {
+        console.warn(
+          "store-cache: Expected first tuple elements to be namespace and table, ignoring set operation:",
+          set
```
holic, line 126: where does `set` come from? should this be `update`?

packages/store-cache/src/utils.ts
```diff
@@ -116,31 +119,37 @@ export function subscribe<C extends StoreConfig = StoreConfig, T extends keyof C
...
     for (const removedKey of write.remove ?? []) {
       // The first tuple element is always the table name
-      const table = removedKey[0];
-      if (typeof table !== "string") {
-        console.warn("store-cache: Expected first tuple element to be the table id, ignoring remove operation:", set);
+      const [namespace, table] = removedKey;
+      if (typeof namespace !== "string" || typeof table !== "string") {
+        console.warn(
+          "store-cache: Expected first tuple elements to be namespace and table, ignoring remove operation:",
+          set
```
holic, line 142: like above , should this be `write` or `removedKey`?

packages/store-cache/src/utils.ts
```diff
@@ -205,3 +215,10 @@ function serializeKey(key: unknown) {
   if (typeof key === "bigint") return String(key);
   return key;
 }
+
+/**
+ * Helper to concat namespace and table with a separator
+ */
+function toSelector(namespace: string, table: string) {
+  return namespace + "/" + table;
```
holic, line 223: how does this manifest in our infra as a whole? I've been using `:` as a separator in the string representation of a `TableId` object (currently utils but planning to move to common) and in dev tools displays

alvrs: this is just a local helper to combine updates of a table in `subscribe`, but it never leaves the `subscribe` util since we do `Object.values` in line 152

packages/store-cache/src/utils.ts
alvrs: it comes from calling `set` on the tuple database (kept the names that `tuple-database` is using here, which also kinda corresponds to the naming used in `Store`)

packages/store-cache/src/createDatabaseClient.ts
holic: oh weird, it looks like the TS hints I was looking at were in a repo with broken TS, so my mud config was letting me set ~any key on the table (I had tried `namespace` to see if it was there)

image

packages/store-cache/package.json
```diff
@@ -0,0 +1,37 @@
...
+    "release": "npm publish --access=public",
+    "test": "vitest --run --passWithNoTests"
+  },
+  "dependencies": {
+    "@latticexyz/common": "workspace:^",
+    "@latticexyz/config": "workspace:*",
+    "@latticexyz/schema-type": "workspace:*",
+    "@latticexyz/store": "workspace:*",
+    "abitype": "^0.8.0",
+    "tsup": "^6.7.0",
```
holic, line 30: should be a dev dep

packages/store-cache/package.json
```diff
@@ -0,0 +1,37 @@
...
+  "type": "module",
+  "exports": {
+    ".": "./dist/index.js"
+  },
+  "types": "src/index.ts",
+  "scripts": {
+    "build": "tsup",
+    "clean": "rimraf dist",
+    "dev": "tsup --watch",
+    "release": "npm publish --access=public",
```
holic, line 21: don't need this anymore, but should add

```
  "publishConfig": {
    "access": "public"
  }
```

packages/store-cache/package.json
```diff
@@ -0,0 +1,37 @@
...
+    "test": "vitest --run --passWithNoTests"
+  },
+  "dependencies": {
+    "@latticexyz/common": "workspace:^",
```
holic, line 25: why `^` and not `*`?

alvrs: interesting, pnpm did this automatically when installing the common package

alvrs:
Seems like Github CI just doesn't like this PR, opened a new PR in #805 and CI is passing there. Closing this one in favor of #805.

PR #648, opened 2023-04-19
feat: add examples

holic:
(wasn't sure what conventional commit prefix to use here)

Adds a top-level `examples` directory with code from [v2sandbox](https://github.com/latticexyz/v2sandbox) so we can continue working out of the monorepo instead of a separate repo.

We can do follow up PRs to bring over v2sandbox branches (react, emojimon, phaser, etc.)

These are intentionally separated from templates as they aren't really meant to be a starting point for projects, but more of a place for us to 1) test new MUD code and 2) show examples of how you might build a thing with MUD.


alvrs (approved):
lgtm!

PR #649, opened 2023-04-19
feat(store,world): add popFromField

dk1a:
Just changing Store/World API, not tablegen, to avoid conflicts there. Can be another PR a bit later, or wait and update this one.
After the tablegen part this should unblock https://github.com/latticexyz/mud/pull/583 (unless @yonadaaa also needs a partial `getItem` method)

### DIFF
--- a/packages/store/src/StoreCore.sol
+++ b/packages/store/src/StoreCore.sol
@@ -291,6 +291,39 @@ library StoreCore {
     }
   }
 
+  function popFromField(bytes32 tableId, bytes32[] memory key, uint8 schemaIndex, uint256 byteLengthToPop) internal {
+    Schema schema = getSchema(tableId);
+
+    if (schemaIndex < schema.numStaticFields()) {
+      revert IErrors.StoreCore_NotDynamicField();
+    }
+
+    // TODO add pop-specific event and hook to avoid the storage read? (https://github.com/latticexyz/mud/issues/444)
+    bytes memory fullData;
+    {
+      bytes memory oldData = StoreCoreInternal._getDynamicField(tableId, key, schemaIndex, schema);
+      fullData = SliceLib.getSubslice(oldData, 0, oldData.length - byteLengthToPop).toBytes();
+    }
+
+    // Emit event to notify indexers
+    emit StoreSetField(tableId, key, schemaIndex, fullData);
+
+    // Call onBeforeSetField hooks (before modifying the state)
+    address[] memory hooks = Hooks.get(tableId);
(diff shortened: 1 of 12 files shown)


packages/store/src/StoreCore.sol
```diff
@@ -291,6 +291,39 @@ library StoreCore {
...
   }
 
+  function popFromField(bytes32 tableId, bytes32[] memory key, uint8 schemaIndex, uint256 byteLengthToPop) internal {
+    Schema schema = getSchema(tableId);
+
+    if (schemaIndex < schema.numStaticFields()) {
+      revert IErrors.StoreCore_NotDynamicField();
+    }
+
+    // TODO add pop-specific event and hook to avoid the storage read? (https://github.com/latticexyz/mud/issues/444)
```
holic, line 301: before we start promoting "MUD the protocol", we should probably figure out all the events we're gonna want, because it sounds dangerous to add something like a push event down the line where some old clients may not properly support it?

holic: though I guess we could still have a backwards compatibility layer or hook that does the full read + `emit StoreSetField` for supporting old clients

packages/store/test/StoreCoreDynamicUpdate.t.sol
```diff
@@ -0,0 +1,142 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
+
+import { Test, console } from "forge-std/Test.sol";
```
dk1a, line 4: do we want to update everything to use explicitly specified imports? (console is needed because gas-report doesn't seem to autoinject it)

packages/store/src/StoreCore.sol
alvrs: Agree we should add these events before the first stable release

packages/store/test/StoreCoreDynamicUpdate.t.sol
alvrs: Personally I think using file imports for `Test` and `Script` is fine, but no strong opinion.

For reference - [Foundry best practices](https://book.getfoundry.sh/tutorials/best-practices)
CleanShot 2023-04-21 at 19 07 34@2x

alvrs (approved):
lgtm! How does the world contract size look with this addition?

dk1a:
> lgtm! How does the world contract size look with this addition?

20.444 -> 21.535 similar increase to updateInField at ~1kB

PR #650, opened 2023-04-19
fix(create-mud): replace linked deps

### LINKED ISSUE #647: create-mud templates have broken mud deps
Our build pipeline that replaces the `link:../../..` defaults for MUD deps with actual version numbers doesn't seem to be working. Thus installing with `npm create mud` results in something like
```
 WARN  Local dependency not found at /Users/kevin/packages/network
 WARN  Local dependency not found at /Users/kevin/packages/react
 WARN  Local dependency not found at /Users/kevin/packages/recs
 WARN  Local dependency not found at /Users/kevin/packages/std-client
 WARN  Local dependency not found at /Users/kevin/packages/utils
```

I need to 1) figure out a fix and 2) add tests to make sure this doesn't break again.

holic:
Thanks GPT-4 for helping with the original code and fixing this bug.

The script now will 1) output more information so we can debug this easier in the future and 2) exit with a non-zero code if any linked dependencies are found after running the replacement logic.

Fixes #647


alvrs (approved):
not blocking but wonder if we should turn this into a node script for easier maintainability

PR #654, opened 2023-04-20
docs: move blog, clean up

holic:
I noticed a few lingering files left from Retype during the Nextra migration (#635).

I moved the blog article so that it renders as a standalone page and looks like this:

image

Because we're using the docs theme, there isn't really a built in "blog index" type of thing, so I left the blog index page for future us to solve. At least the article being here won't break backlinks.

In the meantime, I hid the blog from the sidebar.


holic:
also fixed the logo dark/light theme

image

image

alvrs (approved):
amazing!

PR #655, opened 2023-04-20
refactor: normalize vite configs, use tsx for store/world scripts

dk1a:
- vite config changes are about minify from https://github.com/latticexyz/mud/pull/640#discussion_r1172766764 and some minor normalization
- tsx requires to also bundle solidity parser, which ups `common`'s bundle size from ~50kB to ~1000kB. Probably not much of an issue since it's nodeside. Also moved `build:js` back to the end.
- Tried to remove `modules.d.ts` with prettier plugin stuff, but they're still required in the downstream packages unfortunately

### DIFF
--- a/packages/store/package.json
+++ b/packages/store/package.json
@@ -16,9 +16,9 @@
   },
   "types": "ts/index.ts",
   "scripts": {
-    "build": "pnpm run build:js && pnpm run build:tightcoder && pnpm run build:codegen && pnpm run build:abi && pnpm run build:typechain",
+    "build": "pnpm run build:tightcoder && pnpm run build:codegen && pnpm run build:abi && pnpm run build:typechain && pnpm run build:js",
     "build:abi": "forge build --out abi --skip test script",
-    "build:codegen": "node ./dist/tablegen.js",
+    "build:codegen": "tsx ./ts/scripts/tablegen.ts",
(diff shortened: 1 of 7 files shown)


packages/store/package.json
```diff
@@ -16,9 +16,9 @@
   },
   "types": "ts/index.ts",
   "scripts": {
-    "build": "pnpm run build:js && pnpm run build:tightcoder && pnpm run build:codegen && pnpm run build:abi && pnpm run build:typechain",
+    "build": "pnpm run build:tightcoder && pnpm run build:codegen && pnpm run build:abi && pnpm run build:typechain && pnpm run build:js",
     "build:abi": "forge build --out abi --skip test script",
-    "build:codegen": "node ./dist/tablegen.js",
+    "build:codegen": "tsx ./ts/scripts/tablegen.ts",
```
holic, line 21: yay! this resolves my comment about the script order

holic (approved):

holic:
> which ups common's bundle size from ~50kB to ~1000kB.
just to clarify, is this just `common/codegen`'s bundle?

dk1a:
> just to clarify, is this just common/codegen's bundle?

yes, forge for example is unchanged (much less than 1000kB)

Issue #634, opened 2023-04-17
Add support for struct arguments to `deploy-v2`


alvrs:
Currently we don't support structs as arguments for systems in https://github.com/latticexyz/mud/blob/dc76f2ea99e9d417ca8b216cea84e1ad8a1782eb/packages/cli/src/utils/deploy-v2.ts#L498-L502, which makes it impossible to call those systems via the World function selector

alvrs:
done in #https://github.com/latticexyz/mud/pull/658

PR #658, opened 2023-04-21
fix(cli): parse struct function arguments

smallbraingames:
Previously, function arguments that were structs (as opposed to base solidity types) were not parsed correctly by the Mud v2 deploy script. For example, a struct Coord (int32 x, int32 y) would parse as `tuple` instead of `(int32,int32)`.

This change parses structs (& nested structs) in function args


alvrs (approved):
Amazing, thank you!

Fixes https://github.com/latticexyz/mud/issues/634

PR #659, opened 2023-04-21
build(config): downgrade find-up

dk1a:
`tablegen` wasn't working for me because of some cjs/esm conflicts in the dependency chain. `find-up@5.0.0` and below should support both cjs and esm, which also will help building for both if we want that later

### DIFF
--- a/packages/config/package.json
+++ b/packages/config/package.json
@@ -23,7 +23,7 @@
     "chalk": "^5.2.0",
     "esbuild": "^0.17.15",
     "ethers": "^5.7.2",
-    "find-up": "^6.3.0",
+    "find-up": "^5.0.0",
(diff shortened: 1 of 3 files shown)


alvrs:
would be curious if it works for you with #660

dk1a:
> would be curious if it works for you with https://github.com/latticexyz/mud/pull/660

yeah just tested, it does. This PR still can be useful if we want to support cjs, up to you

holic:
seems reasonable to downgrade and then upgrade again if it works with tsup? looks like there's not many changes since 5.0.0 that we'd need take advantage of

dk1a:
closed as it is unnecessary, for now at least

PR #660, opened 2023-04-21
build: use tsup and do not bundle dependencies

alvrs:
- we are currently bundling dependencies (except for peerDependencies), which leads to various issues in the CLI, and makes the life of consumers harder (since, unlike dependencies, peerDependencies have to be installed manually)
- this PR changes the general approach to
  - avoid using `peerDependencies` unless actually necessary
  - avoid bundling dependencies into the libraries, since dependencies are automatically installed by consumer's package managers
- this PR also switches to `tsup` (from `vite`), since `tsup` is 1. faster and 2. the more natural choice for libraries (vite is more focussed on end-user apps/websites and needs more configuration than `tsup` if the goal is to bundle a library)


packages/store/package.json
```diff
@@ -34,40 +34,31 @@
...
-  "devDependencies": {
+  "dependencies": {
     "@ethersproject/abi": "^5.7.0",
     "@ethersproject/providers": "^5.7.2",
     "@latticexyz/common": "workspace:*",
     "@latticexyz/config": "workspace:*",
     "@latticexyz/schema-type": "workspace:*",
+    "ds-test": "https://github.com/dapphub/ds-test.git#c9ce3f25bde29fc5eb9901842bf02850dfd2d084",
+    "ethers": "^5.7.2",
+    "forge-std": "https://github.com/foundry-rs/forge-std.git#b4f121555729b3afb3c5ffccb62ff4b6e2818fd3"
```
dk1a, line 45: I think ds-test and forge can be devDeps as they're just for tests? (console is also imported in normal files, but that's just for debugging and should be eventually removed)

packages/world/package.json
```diff
@@ -34,46 +34,34 @@
...
-  "devDependencies": {
     "@ethersproject/abi": "^5.7.0",
     "@ethersproject/providers": "^5.7.2",
     "@latticexyz/common": "workspace:*",
     "@latticexyz/config": "workspace:*",
     "@latticexyz/schema-type": "workspace:*",
     "@latticexyz/store": "workspace:*",
+    "ds-test": "https://github.com/dapphub/ds-test.git#c9ce3f25bde29fc5eb9901842bf02850dfd2d084",
+    "ethers": "^5.7.2",
+    "forge-std": "https://github.com/foundry-rs/forge-std.git#b4f121555729b3afb3c5ffccb62ff4b6e2818fd3"
```
dk1a, line 45: (same here with devDeps)

dk1a (approved):
tsup looks much more convenient for our use-case. The config is simplified, no plugins and `noExternal` hacks

packages/world/vite.config.ts
```diff
@@ -1,21 +1,7 @@
...
-      formats: ["es"],
-    },
-    outDir: "dist",
-    minify: true,
-    sourcemap: true,
-    target: "esnext",
-  },
-  plugins: [peerDepsExternal()] as any,
-});
+export default defineConfig({});
```
roninjin10, line 7: You could just remove this file vitest doesn't need a config

roninjin10 (approved):

packages/config/src/store/loadStoreConfig.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
holic, line 1: not blocking but I would prefer to not use whole-file disables

we should either 1) decide to disable this globally or 2) ignore individual lines/usages

packages/cli/tsup.config.ts
```diff
@@ -0,0 +1,11 @@
+import { defineConfig } from "tsup";
+
+export default defineConfig({
+  entry: ["src/index.ts", "src/mud.ts", "src/mud2.ts"],
+  target: "esnext",
+  format: ["esm"],
+  dts: false,
```
holic, line 7: is dts not needed when doing tsup in this way? or is it because cli doesn't have anything to import?

holic (approved):
thank you for tackling this!

packages/world/vite.config.ts
holic: are we even using vitest here?

packages/cli/tsup.config.ts
holic: from @dk1a 
> I don't think we do, since package.json now has "types": "src/index.ts", so the sources themselves replaces dts

packages/world/vite.config.ts
dk1a: yep it's unused. Vitest was only for config, which is now in its own package

holic (approved):

PR #664, opened 2023-04-24
feat(cli): add pop to tablegen

dk1a:
follow-up for https://github.com/latticexyz/mud/pull/649
kind of depends on https://github.com/latticexyz/mud/pull/660 because cli is broken atm


yonadaaa:
Awesome. I realised #583 also needs the `length` of storage arrays (so I can get the index of the last pushed element). It would be good if we could automatically have the methods for non-primitive types.

dk1a:
good point about a length method, I want to add it and also a `getItem` of some sort. Store already has all this data, the downsides of adding 2 more methods are a more complicated interface and we're running low on World's available contract size (have space for 2-3 more methods)
I'm just adding them 1 by 1 to reduce scope of PRs

holic (approved):

PR #665, opened 2023-04-24
feat: ephemeral tables

### LINKED ISSUE #342: Proposal: Actions
_This is mostly in an idea phase rather than a well-formed proposal. Hoping to start a discussion around use cases and potential solutions._

There are many cases in which you might want to know when certain actions are being taken by players, that may or may not result in component updates, e.g. a player attacks another player and misses.

We have a system call stream, but it's just a different view on component updates (grouped by system calls), so it depends on a component updates being emitted for the system call to propagate to clients.

This also means that you have to infer what happened based on component updates, essentially duplicating logic on the client that already exists in the contracts (source of truth). "Attack system was called, and player health lowered, so they must have landed a hit and taken damage." We could expand the system call stream to include _all_ system calls, but it still requires inferring behavior in the client, and streaming this data from the RPC node isn't as efficient as filtering on events.

For cases where you might want to "listen" to things like a "missed attack", you can create/update an arbitrary component to get it into the system call stream, and then check that no damage was done (or however you model your combat system), but this feels more like a workaround and adds a lot more gas.

I'd like to propose we have some generic Solidity event type that you can emit from systems (and/or is emitted automatically on `execute` calls) that can represent named actions. Then expose an action stream so that UIs can trigger animations, effects, etc. We can create some patterns that help guide folks away from one-off modeling and preferring ECS where possible, but I think the cases above just can't be modeled well in our current setup with system call stream, etc.

dk1a:
Closes #342 
Docs in #758


Kooshaba:
this is the minimal version of ephemeral tables needed so we can start using it in examples.

question: is it a problem if we keep the new events as part of the base world for now? or should i already move it out into a module?

client integration does not cache or replay ephemeral events on load (do not want to replay all animations associated with them).

examples/v2-minimal/packages/contracts/mud.config.mts
```diff
@@ -15,6 +15,13 @@ export default mudConfig({
       },
       storeArgument: true,
     },
+    EventTable: {
+      name: "event",
+      schema: {
+        value: "bool",
```
holic, line 21: I'm a little confused by this example. A `bool` value type but the client renders as a counter. Is there another example or name we could use that might make this more clear?

examples/v2-minimal/packages/contracts/mud.config.mts
```diff
@@ -15,6 +15,13 @@ export default mudConfig({
...
+      },
+      ephemeral: true,
```
holic, line 23: Still not entirely sold on "ephemeral" as the name for this pattern but happy to go with this for now (haven't come up with anything better)

examples/v2-minimal/packages/contracts/src/codegen/tables/EventTable.sol
```diff
@@ -0,0 +1,114 @@
...
+  function setMetadata() internal {
+    (string memory _tableName, string[] memory _fieldNames) = getMetadata();
+    StoreSwitch.setMetadata(_tableId, _tableName, _fieldNames);
+  }
+
+  /** Set the table's metadata (using the specified store) */
+  function setMetadata(IStore _store) internal {
+    (string memory _tableName, string[] memory _fieldNames) = getMetadata();
+    _store.setMetadata(_tableId, _tableName, _fieldNames);
+  }
```
holic, line 66: wondering if `registerSchema` and `setMetadata` should also be ephemeral (only events) or not

feels weird to write these table definitions even though the underlying tables will never be written to

examples/v2-minimal/packages/contracts/types/ethers-contracts/IWorld.ts
```diff
@@ -56,6 +56,8 @@ export interface IWorldInterface extends utils.Interface {
     "registerTable(bytes16,bytes16,bytes32,bytes32)": FunctionFragment;
     "registerTableHook(bytes16,bytes16,address)": FunctionFragment;
     "revokeAccess(bytes16,bytes16,address)": FunctionFragment;
+    "setEphemeralRecord(bytes32,bytes32[],bytes)": FunctionFragment;
+    "setEphemeralRecord(bytes16,bytes16,bytes32[],bytes)": FunctionFragment;
```
holic, line 60: @alvrs @dk1a I think the answer is no, but does our module system support opt-in features like this but without the need to go through the hooks path? Just worried about extending the base world with ephemeral events if we're running low on contract size and have other array methods planned.

examples/v2-minimal/packages/contracts/types/ethers-contracts/IWorld.ts
```diff
@@ -472,11 +501,13 @@ export interface IWorldInterface extends utils.Interface {
 
   events: {
     "StoreDeleteRecord(bytes32,bytes32[])": EventFragment;
+    "StoreEphemeralSetRecord(bytes32,bytes32[],bytes)": EventFragment;
```
holic, line 504: we should probably align the names of these
- `StoreEphemeralSetRecord` event
- `setEphemeralRecord` function

packages/config/src/store/parseStoreConfig.ts
```diff
@@ -70,6 +70,8 @@ export interface TableConfig<
   storeArgument?: boolean;
   /** Include a data struct and methods for it. Default is false for 1-column tables; true for multi-column tables. */
   dataStruct?: boolean;
+  /** Generate only `setEphemeral` which emits an event without writing to storage. Default is false. */
+  ephemeral?: boolean;
```
holic, line 74: Hmm, rather than reducing the schema to just one `setEphemeral` field, I wonder if we can support any/all fields, with their underlying implementation calling out to `setEphemeralRecord`?

A big reason I was initially excited by the ephemeral table design is that it could look/act like a regular table (thus multiple fields), except only emit events for records rather than all the write methods.

examples/v2-minimal/packages/contracts/types/ethers-contracts/IWorld.ts
holic: in terms of API, I am wondering if it would be more clear to use the term "emit" instead of "set" and just call it an ephemeral record, e.g.
- `StoreEphemeralRecord` (since there is no notion of set/delete/field updates, etc)
- `emitEphemeralRecord` (rather than set)

packages/network/src/types.ts
```diff
@@ -113,12 +113,35 @@ export type SystemCall<C extends Components = Components> = {
...
+}[keyof C] & {
+  entity: EntityID;
+  lastEventInTx: boolean;
+  txHash: string;
+  txMetadata?: TxMetadata;
+  blockNumber: number;
+  // TODO: make this required, so we can later sort by logIndex when concatenating event types
+  //       would require updating the ECS snapshot, though
+  logIndex?: number;
+};
```
holic, line 133: do we need a separate event for this? why can't we lean on `NetworkComponentUpdate` as it is? maybe so we can discern whether or not this record should be cached downstream?

holic (commented):
Mostly left some high level thoughts/comments on API design.

If this is blocking hackathon things, I'm fine with landing this as is if we set the expectation that this API is likely to change before v2 stable release.

I'll defer to @alvrs for making a call on merging this as is vs. continuing to iterate.

packages/network/src/types.ts
Kooshaba: yes its purely for caching. i agree it's a bit dumb.

examples/v2-minimal/packages/contracts/types/ethers-contracts/IWorld.ts
Kooshaba: i agree with emit

examples/v2-minimal/packages/contracts/mud.config.mts
Kooshaba: i've been thinking about it and also haven't come up with anything

examples/v2-minimal/packages/contracts/mud.config.mts
Kooshaba: the increment system example is super contrived. the idea was that it represents that something happened (in this case with no information because literally nothing can happen)

packages/network/src/types.ts
holic: I added `partialValue` and `initialValue` to `NetworkComponentUpdate` for the purposes of v2. Another route is to extend `NetworkComponentUpdate` with something like `cache?: boolean` where it defaults to true and you can explicitly set to `false` for ephemeral events.

examples/v2-minimal/packages/contracts/mud.config.mts
ludns: ask chatgpt

packages/config/src/store/parseStoreConfig.ts
dk1a: The problem is probably that my comment was misleading, ephemeral option does allow multiple fields, it just removed ass `set` functions, and adds a new `setEphemeral`, which functions like `setRecord` but with a different event and no storage writes. `setEphemeral` supports the same variety/number of fields as `setRecord`

examples/v2-minimal/packages/contracts/types/ethers-contracts/IWorld.ts
dk1a: yeah if you mean the as-yet-unimplemented plugin hooks, modules don't need them for this. `core` module is a good example, as all it does is add more functions to `World` because of contract size issues

examples/v2-minimal/packages/contracts/src/codegen/tables/EventTable.sol
dk1a: that might be bad for sync, snap sync in particular. Also complicates tablegen a bit

alvrs: would tent to keeping this, since it might make it easier to sync the ephemeral events to off-chain indexers if this is ever wanted

examples/v2-minimal/packages/contracts/mud.config.mts
alvrs: I would have proposed "transient" table, but I feel like we'll need this name for tables that use Solidity's upcoming transient storage

packages/config/src/store/parseStoreConfig.ts
alvrs: If we go with `emit` (https://github.com/latticexyz/mud/pull/665#discussion_r1183035199) we could generate all `set*` with `emit*` and change the comment to something like "Generate a table that only emits `SetEphemeralRecord` events but doesn't write to storage."

packages/network/src/v2/ecsEventFromLog.ts
```diff
@@ -37,7 +37,17 @@ export const ecsEventFromLog = async (
     return {
       ...ecsEvent,
       value,
-    };
+    } as NetworkComponentUpdate;
```
alvrs, line 40: why do we need to cast here?

packages/network/src/types.ts
alvrs: would it work if we add `ephemeral: boolean` to `NetworkComponentUpdate` instead of a new event? Seems like that would simplify a lot of the logic/types below

(would tend to `ephemeral` instead of `cache` bc it would be used both for caching inside the sync logic, but also for downstream consumers like in `setupMUDV2Network`/ `applyNetworkUpdates`)

packages/network/src/workers/syncUtils.ts
```diff
@@ -350,7 +353,10 @@ export async function fetchStateInBlockRange(
     setPercentage
   );
 
-  storeEvents(cacheStore, events);
+  storeEvents(
+    cacheStore,
+    filter(events, (e) => isNetworkComponentUpdateEvent(e)) as NetworkComponentUpdate<Components>[]
```
alvrs, line 358: why `lodash` instead of the built in `Array.filter` (which also supports type guards)?

```suggestion
    events.filter(isNetworkComponentUpdateEvent)
```

packages/store/src/StoreCore.sol
```diff
@@ -372,6 +362,34 @@ library StoreCore {
...
+  /************************************************************************
+   *
+   *    EPHEMERAL SET DATA
+   *
+   ************************************************************************/
+
+  /**
+   * Emit the ephemeral event without modifying storage for the full data of the given tableId and key tuple
+   */
+  function setEphemeralRecord(bytes32 tableId, bytes32[] memory key, bytes memory data) internal {
```
alvrs, line 374: Do we need to add `setEphemeralField` too?

packages/world/src/World.sol
```diff
@@ -157,6 +157,29 @@ contract World is StoreRead, IStoreData, IWorldKernel {
...
+   *
+   *    WORLD EPHEMERAL METHODS
+   *
+   ************************************************************************/
+
+  /**
+   * Emit the ephemeral event without modifying storage at the given namespace and name.
+   * Requires the caller to have access to the namespace or name.
+   */
+  function setEphemeralRecord(
```
alvrs, line 170: I think we should move this to a module to avoid increasing the World contract's bytecode size - can even be the existing `CoreModule` (note that nothing changes for consumers since a root function selector is registered)

packages/world/src/World.sol
```diff
@@ -238,6 +261,21 @@ contract World is StoreRead, IStoreData, IWorldKernel {
...
+   *    STORE EPHEMERAL OVERRIDE METHODS
+   *
+   ************************************************************************/
+
+  /**
+   * Emit the ephemeral event without modifying storage at the given tableId.
+   * This overload exists to conform with the `IStore` interface.
+   * Access is checked based on the namespace or name (encoded in the tableId).
+   */
+  function setEphemeralRecord(bytes32 tableId, bytes32[] calldata key, bytes calldata data) public virtual {
```
alvrs, line 275: same here, should move this to `CoreModule`

examples/v2-minimal/packages/contracts/mud.config.mts
alvrs: agree an example where some explicit data is "set" to the ephemeral table and then displayed in the client would be a bit clearer. Maybe something like an input field that forwards a string to a system, which sets the ephemeral table?

examples/v2-minimal/packages/contracts/src/codegen/tables/EventTable.sol
alvrs: I guess `deleteRecord` is not really needed in this lib though (and a bit misleading)

alvrs:
> If this is blocking hackathon things, I'm fine with landing this as is if we set the expectation that this API is likely to change before v2 stable release.
>
>I'll defer to @alvrs for making a call on merging this as is vs. continuing to iterate.

Would prefer to resolve the open points in this PR before merging, but I think we're almost there! The open comments are mostly small things. Maybe except for the naming discussion, which we can iterate on later if we can't come up with a good alternative to "ephemeral" now.

examples/v2-minimal/packages/contracts/src/codegen/tables/EventTable.sol
dk1a: yeah, I missed deleteRecord because it has a separate rendering condition from the other record/field methods

packages/store/src/StoreCore.sol
dk1a: I'd favor only record, by analogy to events it'd be like emitting an event with partial arguments, which probably means you just want more events, or an argument in the event itself to specify which arguments to look at

packages/network/src/types.ts
Kooshaba: ok wil add ephemeral 👍🏼

packages/store/src/IStore.sol
```diff
@@ -54,6 +54,13 @@ interface IStoreWrite {
   function deleteRecord(bytes32 table, bytes32[] memory key) external;
 }
 
+interface IStoreEphemeral {
+  event StoreEphemeralSetRecord(bytes32 table, bytes32[] key, bytes data);
+
+  // Emit the ephemeral event without modifying storage
+  function setEphemeralRecord(bytes32 table, bytes32[] calldata key, bytes calldata data) external;
```
holic, line 61: Similar to above, if we think we'll have other operations (e.g. delete), then I think this is fine as is.

If we're only planning to have set (i.e. `emit StoreEphemeralRecord`), then I think it'd be clearer to have this as `emitEphemeralRecord` to indicate we're not storing anything.

packages/store/src/IStore.sol
```diff
@@ -54,6 +54,13 @@ interface IStoreWrite {
   function deleteRecord(bytes32 table, bytes32[] memory key) external;
 }
 
+interface IStoreEphemeral {
+  event StoreEphemeralSetRecord(bytes32 table, bytes32[] key, bytes data);
```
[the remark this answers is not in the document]
dk1a: I like `StoreEphemeralRecord`, and don't really expect a need for "delete"

examples/minimal/packages/contracts/src/systems/ChatSystem.sol
```diff
@@ -0,0 +1,13 @@
...
+import { System } from "@latticexyz/world/src/System.sol";
+import { MessageTable } from "../codegen/Tables.sol";
+
+bytes32 constant SingletonKey = bytes32(uint256(0x060D));
+
+contract ChatSystem is System {
+  function sendMessage(string memory message) public returns (uint32) {
+    MessageTable.setEphemeral(SingletonKey, message);
+  }
+}
```
alvrs, line 13: The `SingletonKey` pattern was confusing to folks at the last hackathon (esp. since tables also support `singleton: true` for "real singleton tables"). Should we just use the `msg.sender` as key? Then the frontend could even display the sender of the message.

packages/network/src/v2/common.ts
```diff
@@ -5,6 +5,7 @@ export const schemaTableId = new TableId("mudstore", "schema");
 export const metadataTableId = new TableId("mudstore", "StoreMetadata");
 
 export const storeEvents = ["StoreSetRecord", "StoreSetField", "StoreDeleteRecord"] as const;
+export const ephemeralEvents = ["StoreEphemeralSetRecord"] as const;
```
alvrs, line 8: Reminder to also change this if we change the event name to `StoreEphemeralRecord`

packages/network/src/v2/ecsEventFromLog.ts
```diff
@@ -40,6 +40,16 @@ export const ecsEventFromLog = async (
...
+  if (name === "StoreEphemeralSetRecord") {
+    const value = await decodeStoreSetRecord(contract, tableId, args.key, args.data);
+    console.log("StoreEphemeralSetRecord:", { table: tableId.toString(), component, entity, value });
+    return {
+      ...ecsEvent,
+      ephemeral: true,
+      value,
+    };
+  }
+
```
alvrs, line 52: just another reminder
```suggestion
  if (name === "StoreEphemeralRecord") {
    const value = await decodeStoreSetRecord(contract, tableId, args.key, args.data);
    console.log("StoreEphemeralRecord:", { table: tableId.toString(), component, entity, value });
    return {
      ...ecsEvent,
      ephemeral: true,
      value,
    };
  }

```

packages/store/src/StoreCore.sol
alvrs, line 374: let's call this `emitEphemeralRecord`?

alvrs (commented):
Only tiny nits left - mainly let's rename `setEphemeralRecord` to `emitEphemeralRecord` and `StoreEphemeralSetRecord` to `StoreEphemeralRecord`

examples/minimal/packages/contracts/src/systems/ChatSystem.sol
ludns: I think we should stop exposing the SingletonKey to user and surface APIs contract and client side to interact with Singleton (we already do have the API tablegen side, we need it RECS / mud-react side too)

examples/minimal/packages/contracts/src/systems/ChatSystem.sol
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
+import { System } from "@latticexyz/world/src/System.sol";
+import { MessageTable } from "../codegen/Tables.sol";
+
+bytes32 constant SingletonKey = bytes32(uint256(0x060D));
```
alvrs, line 6:
```suggestion
```
No need for this anymore

packages/store/ts/library/render-solidity/renderTable.ts
```diff
@@ -23,12 +24,15 @@ export function renderTable(options: RenderTableOptions) {
...
     dynamicFields,
+    withFieldMethods,
     withRecordMethods,
+    withEphemeralMethods,
     storeArgument,
     primaryKeys,
   } = options;
 
   const { _typedTableId, _typedKeyArgs, _primaryKeysDefinition } = renderCommonData(options);
+  const shouldRenderDelete = !withEphemeralMethods;
```
alvrs, line 35: For consistency we should add a `withDeleteMethods` field to the config (like `withFieldMethods`, `withRecordMethods`, `withEphemeralMethods`). Not strictly blocking for this PR, we can also do that in a follow up.

packages/world/src/modules/core/implementations/EphemeralRecordSystem.sol
```diff
@@ -0,0 +1,32 @@
...
+import { System } from "../../../System.sol";
+import { AccessControl } from "../../../AccessControl.sol";
+import { Call } from "../../../Call.sol";
+import { ResourceAccess } from "../../../tables/ResourceAccess.sol";
+import { InstalledModules } from "../../../tables/InstalledModules.sol";
+import { StoreCore } from "@latticexyz/store/src/StoreCore.sol";
+import { ResourceSelector } from "../../../ResourceSelector.sol";
+
+contract EphemeralRecordSystem is System {
+  using ResourceSelector for bytes32;
```
alvrs, line 14:
```suggestion
```
it seems like we're not doing anything in here that would require this

packages/world/src/modules/core/implementations/EphemeralRecordSystem.sol
```diff
@@ -0,0 +1,32 @@
...
+
```
alvrs, line 15:
```suggestion

contract EphemeralRecordSystem is System {
```
it seems like we're not doing anything in here that would require this

alvrs (approved):
just one nit left (https://github.com/latticexyz/mud/pull/665#discussion_r1187909532) but not blocking. Good to go imo!

alvrs (approved):
just one nit left (https://github.com/latticexyz/mud/pull/665#discussion_r1187909532) but not blocking. Good to go imo!

PR #661, opened 2023-04-23
feat: Turborepo build-scripts rebase

Handfish:
Apparently renaming a branch with the github web console closes the old pull request and thread. Just wanted to follow the convention of the repo by putting my username as the starting token of the branch.

Relevant pull requests
https://github.com/latticexyz/mud/pull/657
https://github.com/latticexyz/mud/pull/644

Suggestion to make a rebase to this branch
https://github.com/latticexyz/mud/pull/644#issuecomment-1517105608

Also a heads up - commit https://github.com/latticexyz/mud/commit/60e62ebf70b0fe2fce39cd15b3d5b46120c34dd4 broke some of the `hardhat` doc generation. Converting certain packages to [modules](https://github.com/latticexyz/mud/commit/60e62ebf70b0fe2fce39cd15b3d5b46120c34dd4#diff-9b298d2cadfb9600e8878ee44b4064e32664a2c00322309b5cf434e75ab87f31R11) causes some compilation errors. There is a [thread here](https://github.com/NomicFoundation/hardhat/issues/3385) with some claims support is coming soon.

PR #669, opened 2023-04-24
build: turborepo

holic:
continues https://github.com/latticexyz/mud/pull/661


.github/actions/setup/action.yml
```diff
@@ -20,17 +20,30 @@ runs:
...
       shell: bash
       run: pnpm install --frozen-lockfile
 
+    - name: Cache turbo build
+      uses: actions/cache@v3
+      with:
+        path: .turbo
+        key: ${{ runner.os }}-turbo-${{ github.sha }}
+        restore-keys: |
+          ${{ runner.os }}-turbo-
```
holic, line 42: pulled this from https://github.com/vercel/turbo/pull/2761

but `runner.os` being the only unique key for the cache seems a little aggressive? is turbo smart enough to figure out what cached things to use across all our builds/commits?

holic:
cc @Handfish

.github/actions/setup/action.yml
Handfish: I [forked austinwoon's demo](https://github.com/Handfish/turbo-repo-cache-with-github-actions), took a look at his commits and recreated his build examples so I could view them. I think turbo is smart enough to cache properly with that restore-key. It acts as if it would on your local machine.

Handfish: Branch name might be useful though because in the case two people are committing at the same time they will be fighting over the cache. According to the docs, multiple restore-keys can be provided and have a cascading fallback, so maybe it'd be good to take advantage of that.

package.json
```diff
@@ -8,9 +8,10 @@
     "url": "https://github.com/latticexyz/mud.git"
   },
   "scripts": {
-    "build": "pnpm recursive run build",
-    "clean": "pnpm recursive run clean",
+    "build": "turbo run build",
+    "clean": "turbo run clean",
     "commit": "cz",
+    "dev": "turbo run dev --concurrency 20",
```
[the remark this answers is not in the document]
alvrs: Related to the comment above (but not strictly related to this PR) - should we add `forge build --watch` to the `dev` commands of our contract packages to rebuild those too (so in the ideal case a rebuild of downstream dependencies, eg CLI or network, would be triggered)?

alvrs (approved):
had one question about `dev`, but this is a great improvement for already even just for build, so happy to merge as is and look at `dev` in a follow up

package.json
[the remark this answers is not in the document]
holic: > It seems like it runs dev independently for each package, so if I for example change things in world, only world rebuilds, but I would have expected all downstream dependencies (eg cli) to also rebuild? or is this not the intention of this command?

I'm not sure (first time using turbo/nx) but this is what I'd expect too, and something we can iterate on/improve as we go!

> should we add forge build --watch to the dev commands of our contract packages to rebuild those too (so in the ideal case a rebuild of downstream dependencies, eg CLI or network, would be triggered)

Probably! But there are other things ahead of `forge build` that don't have watch commands yet like tablegen, worldgen, etc. so I was kinda waiting on those first. But again, we can improve this as we go.

holic:
FYI, looks like Vercel attempts to detect turbo.json and uses `turbo run build` rather than `pnpm run build` as its default entry point, which doesn't work since I've moved docs out of the workspace. So I had to configure the mud docs vercel app to use `pnpm run build` as its build command.

image

I'll follow up soon to see if docs can safely move back into the workspace.

roninjin10:
Very nice!

.github/actions/setup/action.yml
```diff
@@ -20,17 +20,30 @@ runs:
       with:
         version: nightly
 
+    # Workaround for parallel building with forge:
+    # https://github.com/foundry-rs/foundry/issues/4736
+    - name: Install solc
+      shell: bash
+      run: cargo install svm-rs && svm install 0.8.13
```
holic, line 27: see https://github.com/foundry-rs/foundry/issues/4736

PR #671, opened 2023-04-24
ci: remove codegen step

holic:
should have been removed in #615


holic:
merging because this is blocking canary releases

Issue #673, opened 2023-04-24
build follow ups


holic:
We've done a lot of work with our build setup, but there's a handful of lingering things I came across/left undone:
- [ ] should `build:protobuf` in services package run `make protoc-go` or should the `protoc-ts` step have a dependency on `protoc-go` step?
- [ ] are there other outputs to define for services? currently have `bin`, `proto` and `protobuf`
- [ ] services package needs `clean` scripts
- [ ] should `tsup` command or `tsup.config.ts` explicitly define the output dir? (implicitly uses the tsup default of `dist`)
- [ ] move typechain outputs from default of `types/ethers-contracts` to `typechain`
- [ ] rethink prepack/postpack so that pack failures don't leave bad files around, or remove the need for it entirely
- [ ] consider foundry profiles for `abi` output vs regular `out`
  - these are separated because tests/scripts need the full compiled output and abi output only builds what is needed to export abis for downstream use

alvrs:
> rethink prepack/postpack so that pack failures don't leave bad files around, or remove the need for it entirely

this shouldn't be necessary anymore for the packages that use the new "exports" setup in package.json (see https://github.com/latticexyz/mud/pull/640#issuecomment-1516270855), but we need to update/deprecate the remaining packages

PR #674, opened 2023-04-25
docs: fix build

holic:
At some point `"private": true` became `"private": "true"` and whether or not a package is private means that pnpm seems to do different things when installing, running scripts, etc.

When we move back to `true` boolean (rather than `"true"` string), this issue becomes apparent in Vercel (see the failed preview deploy after the first commit in this PR).
 
To work around this, I've added a blank `pnpm-workspace.yaml` to the `docs` directory so that pnpm treats it as its own workspace (with no sub-packages) and installs work properly again.

I also had to add missing dev deps (since they were inferred from the root originally).

I opted to take this path instead of including this in the root workspace because I encountered issues when trying to figure out the build pipelines, especially for other monorepo-like directories (templates, examples), but I think we can safely return this to the workspace in the future once we finish sorting out the build clean up.

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -10,7 +10,7 @@
     "build": "next build",
     "build:prepare": "rm -rf pages/packages && mkdir pages/packages && for PKG in solecs network recs react services; do mv ../packages/$PKG/API ./pages/packages/$PKG; done"
   },
-  "private": "true",
+  "private": true,
   "license": "",
   "dependencies": {
     "next": "^13.3.0",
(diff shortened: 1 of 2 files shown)


holic:
> ooc would it help to add `docs` to the top level `pnpm-workspace.yaml` instead of making it its own workspace? (Noticed it's not listed there)
> 
> If not, happy to merge this

I left a note about this in the PR description:

> I opted to take this path instead of including this in the root workspace because I encountered issues when trying to figure out the build pipelines, especially for other monorepo-like directories (templates, examples), but I think we can safely return this to the workspace in the future once we finish sorting out the build clean up.

alvrs:
> because I encountered issues when trying to figure out the build pipelines, especially for other monorepo-like directories (templates, examples)

Do you have some more context on this?

Should we add this to https://github.com/latticexyz/mud/issues/673?

PR #679, opened 2023-04-26
feat: plugin system

dk1a:
- [x] plugins core (very minimal, part of the `config` package)
- [x] store as a plugin
- [x] world as a plugin
- [x] polish types (cleanup `any`, improve passthrough config types between plugins, particularly how hooks get mixed in, as they're only needed in the resolved config)
- [ ] wrap store/world plugins into `mudConfig` so they don't need to be manually specified?
- [ ] hooks (there're no core hooks, they'd be defined by each plugin for other plugins, not yet sure if they need a root `hooks` option, or if each plugin should make `storeHooks`, `worldHooks` etc)
- [ ] a place for plugins other than store/world to live (https://github.com/latticexyz/mud/pull/584)
- [ ] move a module into a plugin as an example (e.g. uniqueentity doesn't need to be a part of `world`)
- [ ] more endpoints for `config`? and move some stuff into `common`


packages/cli/src/commands/deploy-v2.ts
```diff
@@ -63,17 +63,15 @@ export async function deployHandler(args: Parameters<(typeof commandModule)["han
     // Get the basename of the file
     .map((path) => basename(path, ".sol"));
 
-  // Load and resolve the config
-  const worldConfig = await loadWorldConfig(configPath, existingContracts);
-  const storeConfig = await loadStoreConfig(configPath);
-  const mudConfig = { ...worldConfig, ...storeConfig };
+  // Load the config
+  const mudConfig = (await loadConfig(configPath)) as MUDCoreConfig;
```
alvrs, line 67: I noticed `MUDCoreConfig` here includes all types from store's and world's `typeExtensions` - how does that work? it it because they're part of the same monorepo?

packages/cli/scripts/generate-test-tables.ts
```diff
@@ -1,76 +1,74 @@
 import path from "path";
-import { storeConfig, parseStoreConfig } from "@latticexyz/config";
-import { tablegen } from "@latticexyz/store";
-import { logError } from "../src/utils/errors";
+import { mudConfig, storePlugin, tablegen } from "@latticexyz/store";
 import { getSrcDirectory } from "@latticexyz/common/foundry";
+import { logError } from "../src/utils/errors.js";
```
alvrs, line 4: does this need the `.js` extension?

packages/config/src/core.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export const zMUDCoreUserConfig = z.object({
+  plugins: zMUDPlugin,
+});
+
+export interface MUDCoreConfig {
+  plugins: MUDPlugin[];
+}
+
+/** Type helper for defining StoreUserConfig */
```
alvrs, line 20: the comment seems to be outdated

packages/config/src/core.ts
```diff
@@ -0,0 +1,27 @@
...
+// zod can't validate the config object as it's dynamically defined by the plugins themselves
+export const zMUDPlugin = z.function().args(z.any()).returns(z.any()).array();
+
+export interface MUDCoreUserConfig {
+  plugins: MUDPlugin[];
+}
+
+export const zMUDCoreUserConfig = z.object({
+  plugins: zMUDPlugin,
+});
```
alvrs, line 14: do we even need to use zod here?

packages/store/ts/config/parseStoreConfig.ts
```diff
@@ -170,45 +200,58 @@ export const zEnumsConfig = z.object({
...
  ************************************************************************/
+// Manual *UserConfig is used instead of `z.input<typeof zStoreConfig>` because zod can't preserve doc comments
 
-// zod doesn't preserve doc comments
-export type StoreUserConfig<
-  EnumNames extends StringForUnion = StringForUnion,
-  StaticUserTypes extends ExtractUserTypes<EnumNames> = ExtractUserTypes<EnumNames>
-> = EnumsConfig<EnumNames> & {
+/** Options without generics */
+export interface StoreSimpleOptions {
```
alvrs, line 206: why the separation into its own type?

packages/store/ts/config/parseStoreConfig.ts
```diff
@@ -170,45 +200,58 @@ export const zEnumsConfig = z.object({
...
+/** mudCoreConfig wrapper to use generics in some options for better type inference */
+export function mudConfig<
   // (`never` is overridden by inference, so only the defined enums can be used by default)
   EnumNames extends StringForUnion = never,
   StaticUserTypes extends ExtractUserTypes<EnumNames> = ExtractUserTypes<EnumNames>
->(config: StoreUserConfig<EnumNames, StaticUserTypes>) {
-  return config;
+>(config: MUDUserConfig<EnumNames, StaticUserTypes>): MUDConfig<EnumNames, StaticUserTypes> {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  return mudCoreConfig(config) as any;
```
alvrs, line 250: for context - why do we have to return this as any?

packages/store/ts/config/parseStoreConfig.ts
```diff
@@ -220,9 +263,13 @@ const StoreConfigUnrefined = z
 // finally validate global conditions
 export const zStoreConfig = StoreConfigUnrefined.superRefine(validateStoreConfig);
 
-export function parseStoreConfig(config: unknown) {
-  return zStoreConfig.parse(config);
-}
+// Catchall preserves other plguins' options
```
alvrs, line 266:
```suggestion
// Catchall preserves other plugins' options
```

packages/store/ts/config/typeExtensions.ts
```diff
@@ -0,0 +1,19 @@
...
+declare module "@latticexyz/config" {
+  // Extend the user config type, which represents the config as written by the users.
+  // Most things are optional here.
+  export interface MUDCoreUserConfig extends StoreSimpleOptions, EnumsConfig<string> {
+    tables: TablesConfig<string, string>;
+  }
+
+  // Also extend the config type, which represents the configuration after it has been resolved.
+  // It should not have any optional properties, with the default values applied instead.
+  // Other plugins may receive this resolved config as their input.
```
alvrs, line 16: Do they always?
```suggestion
  // Other plugins receive this resolved config as their input.
```

alvrs (commented):
About types in the config definition: right now the types are extended if the respective plugin is imported (via `typeExtension.ts`), but it only actually works if the plugin is also added to the `plugins` list. How can we make it more explicit to consumers that the plugin has to be added? (Either via typescript magic, or via an explicit error at runtime?)

packages/store/ts/config/typeExtensions.ts
dk1a: I think I meant that the plugins that come before don't receive that particular config as input, but otherwise they always do yeah

packages/store/ts/config/parseStoreConfig.ts
dk1a: also fixed the same typo in parseWorldConfig

packages/store/ts/config/parseStoreConfig.ts
dk1a: because zod doesn't have plugins (https://github.com/latticexyz/mud/pull/679#discussion_r1181621049), and for `parseWorldConfig` it doesn't have any of the store stuff either

packages/store/ts/config/parseStoreConfig.ts
dk1a: for `typeExtensions`, but I actually don't need it and can just use `z.input<typeof zStoreConfig>`. Fixed

dk1a:
> About types in the config definition: right now the types are extended if the respective plugin is imported (via typeExtension.ts), but it only actually works if the plugin is also added to the plugins list. How can we make it more explicit to consumers that the plugin has to be added? (Either via typescript magic, or via an explicit error at runtime?)

Yeah I've been thinking about this issue
Hardhat does this by using a magic global variable and not having a `plugins` array at all, but I think I want to try and avoid magic globals.
On the runtime side, the errors already usually happen, they're just confusing. I can do a double-zod-parse (so once within `mudConfig`, and again after `loadConfig`), and ensure logging of human-readable zod errors.
On the typescript side, ensuring a thing exists in an array is hard enough, but these are functions that have potentially the same signature and no clear identifier.
Another solution I've been thinking about - we just don't worry about this, and make a good wrapper which includes all the useful plugins.
Also for runtime - each plugin could inject an id, and the plugin-reliant functions can then check for the id's existence. That's dangerously close to magic global variables. Why not just use them then? that might be a good solution after all

holic:
> store as a plugin
> world as a plugin

Are we sure this is a useful direction? I feel like these things are so core to MUD that it's a little scary to me that they become plugins and thus may have to workaround whatever constraints the plugin system has

alvrs:
> Yeah I've been thinking about this issue
Hardhat does this by using a magic global variable and not having a plugins array at all, but I think I want to try and avoid magic globals.
On the runtime side, the errors already usually happen, they're just confusing. I can do a double-zod-parse (so once within mudConfig, and again after loadConfig), and ensure logging of human-readable zod errors.
On the typescript side, ensuring a thing exists in an array is hard enough, but these are functions that have potentially the same signature and no clear identifier.
Another solution I've been thinking about - we just don't worry about this, and make a good wrapper which includes all the useful plugins.
Also for runtime - each plugin could inject an id, and the plugin-reliant functions can then check for the id's existence. That's dangerously close to magic global variables. Why not just use them then? that might be a good solution after all

Yeah I feel like the current approach already has most of the magic, but also a couple tripwires that don't really help explicitness that much (= automatic types but having to manually define the plugins array). Since the alternative approach of making plugin config explicit inside a plugin-specific object has the limitation of plugins not being able to deeply integrate into existing configs, I tend to prefer the magic way, but then would go all the way and add the global variable too.

alvrs:
Here is another exploration to throw in the ring for how to combine plugin config types:

https://gist.github.com/alvrs/a3740bea53d2158811902b2a46c1549f

this would allow us to combine the types of plugins in the global config, but avoid magic type imports or global variables

dk1a:
> Here is another exploration to throw in the ring for how to combine plugin config types:
> 
> https://gist.github.com/alvrs/a3740bea53d2158811902b2a46c1549f
> 
> this would allow us to combine the types of plugins in the global config, but avoid magic type imports or global variables

I think this will significantly (perhaps too much so) complicate mudCoreConfig, and especially mudConfig (the one from store, which already has a bunch of generics)
I also worry about how well it will work with many plugins
It is probably worth exploring in a separate PR when plugins are a bit more stabilized and we have more of them to compare the 2 approaches, but I think it's best to go full hardhat for now as that works reliably

examples/minimal/packages/contracts/mud.config.mts
```diff
@@ -1,4 +1,4 @@
-import { mudConfig, resolveTableId } from "@latticexyz/config";
+import { mudConfig, resolveTableId } from "@latticexyz/world/register"
```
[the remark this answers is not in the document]
dk1a: technically any endpoint works, the point of `world/library` is to more explicitly separate stuff without side-effects from stuff with side-effects. And `world/register` is named thus mostly because of hardhat's conventions, config works as well I suppose, just isn't as clear to me that it has side-effects

packages/cli/src/commands/tablegen.ts
```diff
@@ -20,7 +20,7 @@ const commandModule: CommandModule<Options, Options> = {
   },
 
   async handler({ configPath }) {
-    const config = await loadStoreConfig(configPath);
+    const config = (await loadConfig(configPath)) as StoreConfig;
```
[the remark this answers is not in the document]
dk1a: no, we don't. Since the config is typescript, if you don't have the plugin, the config file will have the errors itself so validating it again doesn't seem necessary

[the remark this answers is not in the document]
dk1a: to clarify, we do get zod validation and the same zod errors, but in `mudConfig`/`mudCoreConfig` themselves, not in `tablegen`. When running `mud tablegen` the output of errors should remain unchanged. Things get kinda weird if you try to use a broken config in a frontend though, but you shouldn't do that

packages/store/ts/library/config/parseStoreConfig.ts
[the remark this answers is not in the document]
dk1a: good point, changed the names

packages/cli/src/commands/tablegen.ts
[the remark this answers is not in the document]
alvrs: got it. I think in order to get strong types on imports of the config we might have to move the zod validation somewhere else, but it's fine to think about that in a follow up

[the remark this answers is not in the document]
dk1a: you already get strong types from `mudConfig` actually, I added that as a little extra here. It works fine with zod, except for the quirks I mentioned if config is invalid

examples/minimal/packages/contracts/mud.config.mts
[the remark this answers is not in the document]
alvrs: got it - it wasn't immediately clear to me why it's called `register`, but if this is a convention from hardhat I think it's fine

packages/store/ts/scripts/tablegen.ts
```diff
@@ -1,9 +1,9 @@
 import path from "path";
-import { loadStoreConfig } from "@latticexyz/config";
+import { loadConfig } from "@latticexyz/config/library";
 import { getSrcDirectory } from "@latticexyz/common/foundry";
-import { tablegen } from "../render-solidity";
+import { tablegen, StoreConfig } from "../library";
 
-const config = await loadStoreConfig();
+const config = (await loadConfig()) as StoreConfig;
```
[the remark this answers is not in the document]
alvrs: answered in https://github.com/latticexyz/mud/pull/679/files#r1184945974

packages/world/ts/library/config/worldConfig.test-d.ts
```diff
@@ -1,8 +1,9 @@
 import { describe, expectTypeOf } from "vitest";
 import { z } from "zod";
-import { zWorldConfig, WorldUserConfig } from ".";
+import { zWorldConfig } from "./worldConfig";
+import { WorldUserConfig } from "./types";
 
-describe("loadWorldConfig", () => {
+describe("parseWorldConfig", () => {
```
alvrs, line 6: tiny nit

```suggestion
describe("worldConfig", () => {
```

alvrs (approved):
I'm good with merging this as is and iterate on adding strong config types for typescript imports in a followup!

Issue #681, opened 2023-04-27
Mud2 isn't compatible with webpack5


smallbraingames:
There are various issues when bundling mud packages with webpack5 (the default bundler for NextJS projects...although they are switching to Turbopack)

Here are some issues I ran into trying to figure this out:

1. **Named imports from JSON files**
Webpack5 requires imports from JSON files to be imported as default, and errors on named imports. In MUD, named imports from JSON are everywhere, especially when importing abis.
For example, this line [here](https://github.com/latticexyz/mud/blob/e636e9c9a2110664dfbc18397981126e816e597a/packages/network/src/workers/SyncWorker.ts#L52) should be changed from
`import { abi as IStoreAbi } from "@latticexyz/store/abi/IStore.sol/IStore.json";`
to 
`import IStore from "@latticexyz/store/abi/IStore.sol/IStore.json";`
and then 
`const abi = IStore.abi` to get the abi.
See this related issue: https://github.com/quasarframework/quasar/issues/7943

2. **Imports with ".js"**
Webpack5 couldn't handle these. Wrote this custom webpack config to deal with these (looks for lattice packages, removes .js from imports in the package during the resolve step
Get errors like this: `Module not found: Can't resolve './SchemaType.js'`
Here was the custom webpack config:
```
webpack: (webpackConfig, { webpack }) => {
    webpackConfig.plugins.push(
      new webpack.NormalModuleReplacementPlugin(new RegExp(/\.js$/), function (
        /** @type {{ request: string }} */
        resource
      ) {
        if (resource.request.startsWith(".")) {
          const path = resolve(resource.context, resource.request);
          if (path.includes("@latticexyz/schema-type") && !existsSync(path)) {
            resource.request = resource.request.replace(".js", "");
          }
        }
      })
    );
    return webpackConfig;
  }
  ```
  
  There might be other issues as well, these are the ones I ran into before switching to Vite

holic:
potential fix: https://github.com/latticexyz/mud/pull/682

still need to confirm/repro

holic:
this was optimistically closed via #682, feel free to reopen if this doesn't fix the issues you were seeing!

I'll plan to spin up an example project using Next.js soon that we can use to test against

rsodre:
> I'll plan to spin up an example project using Next.js soon that we can use to test against

yes, please do make all React templates with next.js, it's a huge quality-of-life improvement. And it supports server API routes, which are useful for all sort of things.

PR #682, opened 2023-04-27
build: remove .js from imports

holic:
see #681

### DIFF
--- a/packages/cli/scripts/generate-test-tables.ts
+++ b/packages/cli/scripts/generate-test-tables.ts
@@ -1,7 +1,7 @@
 import path from "path";
 import { storeConfig, parseStoreConfig } from "@latticexyz/config";
 import { tablegen } from "@latticexyz/store";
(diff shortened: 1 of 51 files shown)


dk1a:
fyi there's overlap with https://github.com/latticexyz/mud/pull/679 as I'm heavily refactoring config and also removing `.js` there and in a few other places

holic:
> fyi there's overlap with #679 as I'm heavily refactoring config and also removing `.js` there and in a few other places

it'd be great to pull this out sooner to help with #681 and to not block on the plugin system, which might take some time to refine/finalize

holic:
cc @smallbraingames

holic:
Just tested this with the example project and everything seems to be working and CI passed, so this is probably good to merge as-is as an improvement, but unclear if it fixes #681 yet.

dk1a (approved):
looks like an improvement either way, but definitely would like to see if this is sufficient to resolve @smallbraingames's issues

PR #685, opened 2023-04-28
build(noise): switch to tsup

holic:
see https://github.com/latticexyz/mud/issues/673 and https://github.com/latticexyz/mud/issues/612

leaving this as a draft as I'm not sure I did this right, might have to boot up opcraft again to verify

we could also just kill this package since it's only used by opcraft afaict?


alvrs:
> we could also just kill this package since it's only used by opcraft afaict?

It's used by a couple external projects too; but it doesn't really have to be in MUD, we can move it to its own repo.

holic:
@alvrs In that case, maybe we just land this and address any breakages later as they come up? I haven't been able to test this with OPCraft or anything else using the package yet and not sure it's worth the time right now.

holic:
from @alvrs in discord
> I’m out for today already but agree with merging as is and fixing later if issues come up
> can give this a thumbs up later when I’m home or you can just go ahead merging if you want

PR #688, opened 2023-04-28
ci: add token to fix protoc rate limiting

holic:
https://github.com/arduino/setup-protoc#usage

image


holic:
oof, CI doesn't like the `secrets.GITHUB_TOKEN` reference in a separate action:
- https://github.com/orgs/community/discussions/27054
- https://docs.github.com/en/actions/security-guides/automatic-token-authentication

alvrs (approved):

Issue #411, opened 2023-02-17
move away from `peerDependencies`


holic:
Some MUD packages depend on other MUD packages via `peerDependencies`, but the versions aren't incremented when we release (and looks like [lerna doesn't want to change this](https://github.com/lerna/lerna/issues/1575)), so we should figure out another approach to keep things in sync.

- [x] ecs-browser
- [x] network
- [x] noise
- [x] phaserx
- [x] react
- [x] recs
- [x] solecs
- [x] std-client
- [x] utils

holic:
With #660, we've started moving away from peer dependencies. Let's keep this going!

PR #689, opened 2023-04-28
wip dev ui

holic:


packages/world/src/interfaces/IBaseWorld.sol
```diff
@@ -4,8 +4,7 @@ pragma solidity >=0.8.0;
 /* Autogenerated file. Do not edit manually. */
 
 import { IStore } from "@latticexyz/store/src/IStore.sol";
-import { IWorldKernel } from "../interfaces/IWorldKernel.sol";
-
+import { IWorldKernel } from "./IWorldKernel.sol";
```
holic, line 7: @dk1a I noticed the import here didn't resolve this path completely/correctly. It's not an issue as this import still works, but is a little more verbose than intended. Is this intentional?

packages/world/src/interfaces/IStoreRegistrationSystem.sol
```diff
@@ -3,7 +3,7 @@ pragma solidity >=0.8.0;
 
 /* Autogenerated file. Do not edit manually. */
 
-import { Schema, IStoreHook } from "./../modules/core/implementations/StoreRegistrationSystem.sol";
+import { Schema, IStoreHook } from "../modules/core/implementations/StoreRegistrationSystem.sol";
```
holic, line 6: @dk1a Same question here as above with import path

holic:
I still need to rebase this once #759 lands and do a little clean up work, but I figured we could do at least the first-pass of async code review so I can button this up tomorrow morning.

I realize tests aren't passing, will fix that soon.

packages/dev-ui/src/tables/serializeWithoutIndexedValues.ts
```diff
@@ -0,0 +1,13 @@
+export function serializeWithoutIndexedValues(obj: any) {
+  return JSON.stringify(obj, (key, value) => {
+    // strip indexed values
+    if (/^\d+$/.test(key)) {
+      return;
+    }
```
alvrs, line 6: what are indexed values?

packages/dev-ui/src/tables/serializeWithoutIndexedValues.ts
```diff
@@ -0,0 +1,13 @@
...
+    // serialize bigints as strings
+    if (typeof value === "bigint") {
+      return value.toString();
+    }
+    return value;
+  });
```
alvrs, line 12: would it be worth to use the other `serialize` util here?

packages/network/package.json
```diff
@@ -9,9 +9,19 @@
...
+    "./dev": "./dist/dev/index.js"
+  },
+  "typesVersions": {
+    "*": {
+      "*": [
+        "src/index.ts"
+      ],
+      "dev": [
+        "./src/dev/index.ts"
+      ]
```
alvrs, line 22: in other packages we used `"index":` for the index types instead of `*` - do you know what's the difference / if there is a difference? Should we standardise on one of them everywhere?

Also the `src/index.ts` path is not prefixed with `./` while the others are - is that intended?

packages/network/package.json
```diff
@@ -79,8 +91,7 @@
     "nice-grpc-web": "^2.0.1",
     "observable-webworker": "^4.0.1",
     "rxjs": "^7.5.5",
-    "threads": "^1.7.0",
-    "viem": "^0.2.0"
+    "threads": "^1.7.0"
```
alvrs, line 94: not directly related to this PR but since we're changing the package.json anyway - should we move all those peer dependencies to regular dependencies?

packages/network/src/dev/emitter.ts
```diff
@@ -0,0 +1,27 @@
...
+// Expose a global emitter to ease in getting data from our internals to the dev UI
+
+export type EmitterEvents = {
+  storeEvent: {
+    chainId: number;
+    worldAddress: string;
+    transactionHash: string;
+    blockNumber: number;
+    logIndex: number;
+    event: "StoreSetRecord" | "StoreSetField" | "StoreDeleteRecord";
```
alvrs, line 13: should we add `StoreEphemeralRecord` here too? (Should we import this from `common/storeEvents` to not have to maintain it separately?)

alvrs (commented):
This is extremely cool!! I'm very hyped. 

For feedback, here are a couple things I noticed when exploring the dev UI as a user:

* Feels like block numbers would need a little less space here, and it would be nice to somehow be able to expand a row to show the full table name / key / value (although I just realized it works to copy the value and then paste somewhere to see the full value)
CleanShot 2023-05-09 at 11 25 40@2x

* Seems like the store data view doesn't update when the data updates but only when toggling back and forth to a different view?
CleanShot 2023-05-09 at 11 28 23@2x

* Very much a "nice-to-have", but it'd be nice to be able to change the width that the dev ui panel takes on the page (I caught myself multiple times trying to increase the width when I wanted to read the full value of a row)

* Do you know why there are a couple schema events in the beginning that don't have a value? 
CleanShot 2023-05-09 at 11 32 45@2x

* Tiny nit - on the overview page new transactions appear at the top of the list, on the tx details view they appear at the bottom. Should we always make them appear on the top for consistency?
CleanShot 2023-05-09 at 11 35 18@2x
CleanShot 2023-05-09 at 11 35 10@2x

* It seems like transactions that fail gas estimation are currently not displayed in the tx log, is that expected? (Edit: found the TODO in the code about this!)

packages/world/src/interfaces/IBaseWorld.sol
dk1a: kind of intentional, the path just isn't resolved (for simplicity, because I can avoid it and it's a package import with a path rather than a proper path)

packages/world/src/interfaces/IStoreRegistrationSystem.sol
dk1a: here the reason is different from `IWorldKernel` - I always want to prefix paths with `./` because local unix paths can start without it (and thus `path.join` etc can also reduce a path to e.g. `src/interfaces/...`), but local solidity paths must always start with `./`. I can check the string for `../` and conditionally not prefix it, but that's unnecessary complexity

packages/network/package.json
holic: I think I copied this pattern from elsewhere, so we should probably make all of these consistent so future copy+paste is easier to get right.

packages/network/package.json
holic: I'm gonna take a pass at this in a follow up: https://github.com/latticexyz/mud/issues/411

packages/network/src/dev/emitter.ts
holic: yes and yes, this PR started before that feature was finalized/merged :)

holic:
> * Feels like block numbers would need a little less space here, and it would be nice to somehow be able to expand a row to show the full table name / key / value (although I just realized it works to copy the value and then paste somewhere to see the full value)

I left block numbers a bit wider for cases like our testnet where the current block number is 13951282. I'm using a fixed table width in this case because letting the browser figure out the width based on the content ends up with wildly unbalanced column widths, and widths that move around as data flows in.

My intention is lay some groundwork here and revisit each section to flesh it out. With this one in particular, these lines are meant to be just a summary, not capture all information in one line, and add a way to expand each item in the log to show a more verbose view of what happened for a given event (with a better display of the value, like a diff from before/after and showing all fields).

> * Seems like the store data view doesn't update when the data updates but only when toggling back and forth to a different view?

I'll see if I can make some improvements here, but it's currently quite expensive to load from the cache store because of how the data is formatted (not tabular/relational). I'm assuming we'll be able to make a lot of improvements to this once we have a v2 client store.

> * Very much a "nice-to-have", but it'd be nice to be able to change the width that the dev ui panel takes on the page (I caught myself multiple times trying to increase the width when I wanted to read the full value of a row)
> * Do you know why there are a couple schema events in the beginning that don't have a value?
> 
> CleanShot 2023-05-09 at 11 32 45@2x
> 
> * Tiny nit - on the overview page new transactions appear at the top of the list, on the tx details view they appear at the bottom. Should we always make them appear on the top for consistency?
> 
> CleanShot 2023-05-09 at 11 35 18@2x
> 
> CleanShot 2023-05-09 at 11 35 10@2x
> 
> * It seems like transactions that fail gas estimation are currently not displayed in the tx log, is that expected? (Edit: found the TODO in the code about this!)

holic:
> Do you know why there are a couple schema events in the beginning that don't have a value?

Actually, I think this is because I strip the indexed values from the value object. When these events come in, there's missing metadata to map the indexed values to named values, so the object only contains the indexed values (number keys). These get stripped when displaying, because it's generally duplicative/hard to read.

packages/dev-ui/src/actions/getTransactionReceipt.ts
```diff
@@ -0,0 +1,17 @@
+import { Hex, TransactionReceipt, PublicClient, Chain } from "viem";
+
+// TODO: something about this fails when doing lots of simultaneous requests for transactions
+//       not sure if its viem or failed RPC requests or what, but the promises get stuck/never resolve
```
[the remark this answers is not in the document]
holic: reported here https://github.com/wagmi-dev/viem/issues/467

packages/dev-ui/src/tables/serializeWithoutIndexedValues.ts
holic: this is temporary until we're rendering table with columns instead of just dumping json

alvrs (approved):
None of the open points are blocking imo, happy to get this in and iterate on it over time! Should we move the open todos from the description / comments into a catchall issue?

PR #691, opened 2023-04-28
ci: reduce cache misses by forking main cache for each pull request

Handfish:
In the case where multiple developers are pushing branches simultaneously and running the `ci/cd`, the present action just checks package differences against the most recently built cache. With this change, pull_requests will have their own branch of cache.

By including [GITHUB_REF](https://docs.github.com/en/github-ae@latest/actions/learn-github-actions/variables#default-environment-variables) in our `key` and `restore-key`, we can essentially create a forked cache from the latest main cache. 

Information:
[The cache action searches the restore-keys in sequential order. When a key doesn't match directly, the action searches for keys prefixed with the restore key. If there are multiple partial matches for a restore key, the action returns the most recently created cache.](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#example-using-multiple-restore-keys)


.github/actions/setup/action.yml
```diff
@@ -20,25 +20,28 @@ runs:
...
+
     - name: Cache turbo build
       uses: actions/cache@v3
       with:
         path: .turbo
-        key: ${{ runner.os }}-turbo-${{ github.sha }}
+        key: ${{ runner.os }}-turbo-${{ github.ref_name }}-${{ github.sha }}
         restore-keys: |
+          ${{ runner.os }}-turbo-${{ github.ref_name }}-
+          ${{ runner.os }}-turbo-main
```
Handfish, line 44: In the case of branch of a branch, maybe removing `${{ runner.os }}-turbo-main` and just matching to the most recent `${{ runner.os }}-turbo-` might be better, as the CI/CD is becoming very fast and likely the devs approving these ci/cd processes will be focusing on individual problems at a time.

holic:
I also noticed that turbo supports env vars: https://turbo.build/repo/docs/core-concepts/caching#altering-caching-based-on-environment-variables

So another approach could be adding `GITHUB_REF` or `GITHUB_REF_NAME` to our turbo config?

alvrs (approved):
lgtm!

PR #692, opened 2023-04-29
docs: mud2 docs

ludns:


docs/pages/_meta.js
```diff
@@ -1,13 +1,19 @@
...
+  "what-is-mud": "What is MUD?",
+  "store": "Store",
+  "world": "World",
+  "mode": "MODE",
+  "cli": "CLI",
+  // blog: {
+  //   title: "Blog",
+  //   type: "page",
+  //   display: "hidden",
+  // },
```
holic, line 13: in my mind, blog posts are not evergreen and they're designed to be a snapshot in time, where the date stamp will be an indicator of whether or not something is no longer relevant (or we can add notices at the top of each post)

If we don't intend to have a blog on the docs site (I'm fine with this), we should at least take some care to redirect old URLs to the nearest relevant page in the new docs to not break backlinks, search engine results, etc.

holic (dismissed):

docs/pages/_meta.js
ludns: We can port the blog. The comments were from @alvrs exploration w/ Nextra.

PR #696, opened 2023-05-02
[WIP] Cascading Hot-reload / build script for Solidity

Handfish:
Hey again, Lattice,

I played around with making a hot-reload script for solidity files and downstream dependencies utilizing `turbowatch`. 

I produced a couple of different branches experimenting with the `Chokidar` file watching backend. I've chosen to focus my efforts on said backend mainly due to its ability to ignore paths and not index the entire project.

The branch with the present draft pull request is the first and better working branch with a more simple, stupid approach. It watches all packages and saves each package build time to a map. The map guarantees that no single package is built twice in a row in under a constant time span of seconds. The intention originally was to stop infinite build loops.

My [second, less successful, and more complicated approach is this branch](https://github.com/Handfish/mud/tree/handfish/turbowatch-events), which utilizes a [forked turbowatch](https://github.com/Handfish/turbowatch/commit/0255bc672fd69d677bda23a8483691200bf582a0) to take advantage of Chokidars' "unwatch" and "add" functions and adds the ability to pass an additional function to the spawn to run on process kill. Unfortunately, saving a file in rapid succession seems to break reactivity / listening of the file, [which I hoped to prevent with the introduction of the the mentioned changes](https://github.com/Handfish/mud/blob/handfish/turbowatch-events/turbowatch.ts#L62-L64). I included this dysfunctional branch as I think it is interesting and maybe it will spark some ideas.

While I don't like the idea of arbitrary waits in programs, and processes not being able to react to an event based system, I recognize that the simple approach might be working. I think the simple stupid approach might be worth a review or a glance since it seems to achieve hot-reload + cascading builds for solidity.

### How to run Turbowatch
`ROARR_LOG=true npx turbowatch` in project directory to run turbowatch in either branch.

Plays along with another terminal running `pnpm run dev`, the watch fighting will cause a few additional package builds but no infinite build loops as far as I know. 

### Other considerations

If this approach for solidity hot-reload is appreciated, maybe it might be worth looking into making turbowatch responsible for typescript files to reduce redundant builds with fighting watch tasks.

I will clean up git history if branch is decided valuable.


alvrs:
thank you for this @Handfish! Could you elaborate on the advantages of this approach versus something like using `forge build --watch` in a `dev` script in the solidity packages? The approach in this PR also seems to require manually keeping `turbowatch.ts` up to date with packages, could that be automated?

I'm going through all PRs right now with the goal of either merging or closing them to reduce the number of open/stale PRs, so I'm gonna close this one for now, but feel free to reopen if you intend to continue working on it.

PR #697, opened 2023-05-02
[WIP] react-example

ludns:


holic:
I pulled in the react minimal example here: https://github.com/latticexyz/mud/pull/726

Once that lands, I'll update the dev UI PR (https://github.com/latticexyz/mud/pull/689) to mount the dev UI to the react minimal client too

PR #725, opened 2023-05-02
docs(examples): add PostDeploy, test examples

dk1a:
Next would be good to add a workflow to actually run these tests in CI


holic (approved):
Added test command and CI step! It required me exposing the `.env` (anvil's default private key) but seems fine for now. Added an issue to follow up on these things here:
- https://github.com/latticexyz/mud/issues/728
- https://github.com/latticexyz/mud/issues/727

PR #730, opened 2023-05-03
feat(cli): add slow argument to wait until contract is deployed

vedhavyas:
I noticed that gas estimate is incorrect since the it was taken before previous contracts are deployed. This ensures contracts are deloyed one after the other.


dk1a:
do you need this because your contracts depend on each other in some way?

vedhavyas:
@dk1a we were trying to deploy world to our substrate based EVM chain and it was failing because the gas estimations for all the contracts deployed are fetched even before contracts are deployed. In our EVM, gas estimations are coming out initially and after some contracts are deployed.

Hence, the change to ensure the estimations are correct by ensuring the contracts are deployed synchronously.

jfrank-summit:
For more context see this [discord](https://discord.com/channels/865335009915961364/1029021674605129728/1103335818619265044) conversation.
> I think we figured out the issues we were having with gas estimations. The first issue we fixed on our side and this took care of the exceeds block gas limit error. However, we were still getting failures with deploy-v2 because the gas estimate was coming back 21,700 when it should have been 3.5 millionish. This was because the call to estimateGas was happening before the contract had actually been deployed. I believe you need to await deployed() after calling deploy() here: https://github.com/latticexyz/mud/blob/main/packages/cli/src/utils/deploy-v2.ts#L333-L339

vedhavyas:
@dk1a any thoughts on this so far? Would be great to get this change in some form or other :)

alvrs (commented):
We have similar logic to this for `fastTxExecute` (with the `confirmations` param, which is set to `1` by default) - I think we should combine the two in a single cli arg (`--disable-tx-wait`). By default we should await contract deployment and set `confirmations` to 1, if `--disable-tx-wait` is provided, then we should not await contract deployment or transactions.

This would be useful for local dev using `anvil`, where the tx are confirmed fast enough that we don't have to wait for confirmations.

For context on `confirmations`, see #606

alvrs (commented):
We have similar logic to this for `fastTxExecute` (with the `confirmations` param, which is set to `1` by default) - I think we should combine the two in a single cli arg (`--disable-tx-wait`). By default we should await contract deployment and set `confirmations` to 1, if `--disable-tx-wait` is provided, then we should not await contract deployment or transactions.

This would be useful for local dev using `anvil`, where the tx are confirmed fast enough that we don't have to wait for confirmations.

For context on `confirmations`, see #606

alvrs (commented):
We have similar logic to this for `fastTxExecute` (with the `confirmations` param, which is set to `1` by default) - I think we should combine the two in a single cli arg (`--disable-tx-wait`). By default we should await contract deployment and set `confirmations` to 1, if `--disable-tx-wait` is provided, then we should not await contract deployment or transactions.

This would be useful for local dev using `anvil`, where the tx are confirmed fast enough that we don't have to wait for confirmations.

For context on `confirmations`, see #606

vedhavyas:
@alvrs PTAL

packages/cli/src/utils/deploy-v2.ts
```diff
@@ -115,15 +123,28 @@ export async function deploy(
   // Create World contract instance from deployed address
   const WorldContract = new ethers.Contract(await contractPromises.World, IBaseWorldData.abi, signer) as IBaseWorld;
 
+  let confirmations = 1;
+  const retryCount = 0;
+  if (disableTxWait) {
+    confirmations = 0;
+  }
```
alvrs, line 130:
```suggestion
  const confirmations = disableTxWait ? 0 : 1;
```

also probably worth moving `retryCount` be the last argument in `fastTxExecute` so we don't have to provide a value when calling the method. (`retyCount = 0` should be implicit when calling the method for the first time and only be increased by recursive calls to the method itself)

alvrs (commented):
thank you for this! just a couple nits left

alvrs (commented):
thank you for this! just a couple nits left

alvrs:
I just tried this locally with `--disableTxWait` and there is one more change we need in line 462 in `utils/deploy-v2.ts`:

```ts
 .then((tx: any) => (confirmations === 0 ? tx : tx.wait(confirmations)));
```

(ethers errors when calling `tx.wait(0)`)

vedhavyas:
@alvrs do you have a solution for it? Else I'm leaning to set it back to `confirmations=1`

alvrs:
the solution is to change like `462` in `utils/deploy-v2.ts` from

```ts
 .then((tx: any) => tx.wait(confirmations));
```
to
```ts
 .then((tx: any) => (confirmations === 0 ? tx : tx.wait(confirmations)));
```

PR #732, opened 2023-05-03
feat(examples): todo example

Kooshaba:
CleanShot 2023-05-03 at 17 18 49


examples/todo/packages/contracts/src/codegen/world/IIncrementSystem.sol
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
+
+/* Autogenerated file. Do not edit manually. */
+
+interface IIncrementSystem {
+  function increment() external returns (uint32);
+}
```
holic, line 8: there's a bunch of leftover autogen in here, think we might need a `pnpm clean` script

examples/todo/packages/contracts/src/systems/TodoSystem.sol
```diff
@@ -0,0 +1,42 @@
...
+function addressToEntity(address a) pure returns (bytes32) {
+  return bytes32(uint256(uint160((a))));
+}
+
+contract TodoSystem is System {
+  uint256 totalTodos;
+
+  function add(string memory body) public {
+    bytes32 user = addressToEntity(_msgSender());
+    bytes32 todoKey = keccak256(abi.encodePacked(user, totalTodos++));
```
holic, line 17: I thought state inside a system was a MUD antipattern?

Could use we this module instead? https://github.com/latticexyz/mud/pull/552

examples/todo/packages/contracts/src/systems/TodoSystem.sol
```diff
@@ -0,0 +1,42 @@
...
+
+    TodoItem.set(todoKey, TodoItemData({ body: body, completed: false }));
+    Owned.set(todoKey, user);
+  }
+
+  function modify(bytes32 key, string memory body) public onlyOwner(key) {
+    require(Owned.get(key) == addressToEntity(_msgSender()), "Only owner can modify todo");
```
holic, line 24: I think this is handled by the `onlyOwner` modifier?

holic (commented):
Sweet! Left a couple comments, mostly clean up things.

examples/todo/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "mud-example-minimal",
```
holic, line 2: might wanna update this name

also might be a tiny bit more clear if this dir is `todo-app` and the package named after it like `mud-example-todo-app`

(just wanna make sure folks don't think this is a dir of our TODOs haha)

holic:
for now you'll need to add the new example dir name here: https://github.com/latticexyz/mud/blob/main/.github/workflows/examples.yml#L15
hoping to clean this up/make this dynamic in the future

examples/todo/package.json
Kooshaba: lmao true

examples/todo-app/packages/client-react/src/App.tsx
```diff
@@ -0,0 +1,77 @@
+import { useEntityQuery } from "@latticexyz/react";
+import { useMUD } from "./MUDContext";
+import { useState } from "react";
+import { Has, HasValue, getComponentValueStrict } from "@latticexyz/recs";
+
+function toKey(entityId: string) {
+  return "0x" + entityId.replace("0x", "").padStart(64, "0");
+}
```
alvrs, line 8: why is this needed? Shouldn't our entity ids be normalized on the client?

alvrs:
I really like this, but think we should first focus on getting v2 stable before adding new example projects, because each breaking v2 change requires us to update every example project, and also the template project setup is likely to change soon with the new sync stack. We should reopen/update this once v2 is stable.

PR #740, opened 2023-05-04
build(std-client): move to tsup

holic:
see https://github.com/latticexyz/mud/issues/673 and https://github.com/latticexyz/mud/issues/612

this will make it a little easier for me to split out a dev bundle for the dev UI


packages/std-client/package.json
```diff
@@ -9,20 +9,21 @@
...
     "clean": "pnpm run clean:mud && pnpm run clean:js",
     "clean:js": "rimraf dist",
     "clean:mud": "rimraf src/mud-definitions",
+    "dev": "tsup --watch",
     "docs": "rimraf API && typedoc src && find API -type f -name '*.md' -exec sed -E -i \"\" \"s/(#.*)(<.*>)/\\1/\" {} \\;",
-    "prepack": "mv package.json package.json.bak && jq \".main = \\\"dist/index.js\\\"\" package.json.bak > package.json ",
-    "postpack": "mv package.json.bak package.json || echo 'no package.json.bak'",
     "release": "npm publish --access=public",
-    "test": "pnpm run build:mud && vitest typecheck --run && echo 'todo: add tests'"
+    "test": "pnpm run build:mud && tsc --noEmit && vitest typecheck --run"
```
dk1a, line 26: having both tsc and vitest's typecheck seems redundant
I usually use tsc when there're no actual tests for vitest to run properly

packages/std-client/tsconfig.json
```diff
@@ -1,102 +1,18 @@
...
+    "outDir": "dist",
+    "isolatedModules": true,
+    "esModuleInterop": true,
+    "forceConsistentCasingInFileNames": true,
+    "strict": true,
+    "skipLibCheck": true,
+    "jsx": "react-jsx"
   },
-  "exclude": ["dist", "**/*.spec.ts", "scripts", "packages"]
+  "exclude": ["dist"]
```
dk1a, line 17: looked into the test failures a bit, it seems they were more so revealed by this changed, then caused by it

```
  "include": ["src"],
  "exclude": ["**/*.spec.ts"]
```
I'd consider an explicit include like in some other packages, and a hacky exclude until specs stop causing typecheck errors

dk1a (approved):
the build itself looks good, just 2 minor comments about tests

packages/std-client/tsconfig.json
holic: I was able to fix these (my preference rather than ignoring) but now seeing

```
mud/packages/std-client % pnpm test

> @latticexyz/std-client@1.42.0 test /Users/kevin/Projects/latticexyz/mud/packages/std-client
> pnpm run build:mud && tsc --noEmit && vitest typecheck --run


> @latticexyz/std-client@1.42.0 build:mud /Users/kevin/Projects/latticexyz/mud/packages/std-client
> tsx ./scripts/tsgen.ts

Generated ts definition files: src/mud-definitions/store/contractComponents.ts
Generated ts definition files: src/mud-definitions/world/contractComponents.ts
../common/src/codegen/utils/format.ts:3:36 - error TS7016: Could not find a declaration file for module 'prettier-plugin-solidity'. '/Users/kevin/Projects/latticexyz/mud/node_modules/.pnpm/prettier-plugin-solidity@1.1.2_prettier@2.8.4/node_modules/prettier-plugin-solidity/src/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/prettier-plugin-solidity` if it exists or add a new declaration (.d.ts) file containing `declare module 'prettier-plugin-solidity';`

3 import prettierPluginSolidity from "prettier-plugin-solidity";
                                     ~~~~~~~~~~~~~~~~~~~~~~~~~~


Found 1 error in ../common/src/codegen/utils/format.ts:3

 ELIFECYCLE  Test failed. See above for more details.
```

packages/std-client/package.json
holic: ahh does vitest also do a full typecheck?

packages/std-client/tsconfig.json
dk1a: that's even better, try copying `modules.d.ts` from another package

packages/std-client/package.json
dk1a: yeah it should be just strictly better than tsc for this purpose, it just gets a bit tricky if there're no test files at all because you need an extra flag and it can spawn a weird temporary file if it fails, so then tsc becomes an attractive option

dk1a (approved):
lgtm

PR #741, opened 2023-05-04
feat: phaser client in minimal example

Kooshaba:


examples/minimal/packages/client-phaser/src/MUDContext.tsx
```diff
@@ -0,0 +1,21 @@
+import React from "react";
+import { createContext, ReactNode, useContext } from "react";
+import { SetupResult } from "./mud/setup";
+
+const MUDContext = createContext<SetupResult | null>(null);
+
+type Props = SetupResult & {
+  children: ReactNode;
+};
```
holic, line 9: not super critical but this file also changed in templates fwiw

holic (dismissed):

examples/minimal/packages/client-phaser/src/MUDContext.tsx
Kooshaba: actually now that you mention it i don't use this

holic (approved):

Issue #680, opened 2023-04-26
use CREATE2 for world, emit event on deploy


holic:
Just an idea that might help us give more visibility into clients about what is going on under the hood while they wait for their `dev:contracts` service to boot up or `mud deploy` to execute:
- use CREATE2 to determine the world address before deploying, write the world address to disk as the first step in deploy
- emit an event (e.g. `emit Init()`) in the world constructor that represents a deploy, so we can query/poll the RPC for it rather than waiting for a block number (I wish ~all contracts did this)

ludns:
I agree with this
On 1) This is really useful, and we should move in that direction for all of MUD (systems and modules mostly). the tricky thing is we can't have CREATE2 because it still depends on the deployer address (and ideally the deployer address wouldn't matter).
Having a CREATE3 kind of situation where the address only depends on the the bytecode and a salt would be ideal. We can then get a fresh address for the same world bytecode by changing the salt, and we can set the salt to `0` for systems and modules. That avoids us from having to redeploy the same system or module, we can just check if it already exists!.
On 2) agreed too. Then we don't need `startBlockNumber` as a parameter ever again.

holic:
> Having a CREATE3 kind of situation where the address only depends on the the bytecode and a salt would be ideal. 
whoops, I think CREATE3 is what I meant here!

ludns:
Actually it's not event that!
CREATE3:
> Deterministic contract address based on msg.sender + salt

We want deterministic contract address based on bytecode + salt.
Something like this would work: https://github.com/Arachnid/deterministic-deployment-proxy

holic:
If we only have/need world addresses, then another thing we could do is write them all out to a single mapping/manifest file like `deploys.json` (rather than one per chain), which would make it easier to conditionally load them in Vite (we've had trouble with dynamic file imports in local and in Vercel)

holic:
got started on a single deploys file here: https://github.com/latticexyz/mud/pull/803

PR #742, opened 2023-05-04
feat(world): emit event on deploy

holic:
Adds an event to world deploy so we don't need to keep track of deploy block number, we can just grab it from the RPC by looking for this event.

see #680

### DIFF
--- a/packages/world/src/interfaces/IWorldKernel.sol
+++ b/packages/world/src/interfaces/IWorldKernel.sol
@@ -101,5 +101,5 @@ interface IWorldCall {
  * registered functions selectors from the `CoreModule`.
  */
 interface IWorldKernel is IWorldData, IWorldModuleInstallation, IWorldCall, IErrors {
-
+  event HelloWorld();
 }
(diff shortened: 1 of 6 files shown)


packages/world/src/interfaces/IWorldKernel.sol
```diff
@@ -101,5 +101,5 @@ interface IWorldCall {
  * registered functions selectors from the `CoreModule`.
  */
 interface IWorldKernel is IWorldData, IWorldModuleInstallation, IWorldCall, IErrors {
-
+  event HelloWorld();
```
holic, line 104: wasn't sure where to put this cc @alvrs @dk1a

holic: also just wanted to note that I considered event names like `Initialized` and `Constructed` and `Deployed`, but I thought `HelloWorld` was a nice callback to beginner programming tutorials and fits nicely into our "world" terminology :)

alvrs (approved):
lgtm!

packages/world/src/interfaces/IWorldKernel.sol
alvrs: Fine to put it on `IWorldKernel`. I like `HelloWorld`!

PR #744, opened 2023-05-05
feat: limit default selector names to 16 bytes

dk1a:
fixes https://github.com/latticexyz/mud/issues/706

### DIFF
--- a/packages/world/src/modules/core/tables/FunctionSelectors.sol
+++ b/packages/world/src/modules/core/tables/FunctionSelectors.sol
@@ -17,7 +17,7 @@ import { EncodeArray } from "@latticexyz/store/src/tightcoder/EncodeArray.sol";
 import { Schema, SchemaLib } from "@latticexyz/store/src/Schema.sol";
 import { PackedCounter, PackedCounterLib } from "@latticexyz/store/src/PackedCounter.sol";
 
-bytes32 constant _tableId = bytes32(abi.encodePacked(bytes16(""), bytes16("funcSelectors")));
+bytes32 constant _tableId = bytes32(abi.encodePacked(bytes16(""), bytes16("FunctionSelector")));
 bytes32 constant FunctionSelectorsTableId = _tableId;
 
 library FunctionSelectors {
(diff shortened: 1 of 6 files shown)


packages/world/src/modules/core/tables/FunctionSelectors.sol
```diff
@@ -17,7 +17,7 @@ import { EncodeArray } from "@latticexyz/store/src/tightcoder/EncodeArray.sol";
 import { Schema, SchemaLib } from "@latticexyz/store/src/Schema.sol";
 import { PackedCounter, PackedCounterLib } from "@latticexyz/store/src/PackedCounter.sol";
 
-bytes32 constant _tableId = bytes32(abi.encodePacked(bytes16(""), bytes16("funcSelectors")));
+bytes32 constant _tableId = bytes32(abi.encodePacked(bytes16(""), bytes16("FunctionSelector")));
```
holic, line 20: This caught me off guard because I would have expected this to be plural but apparently `FunctionSelectors` is one character too long.

I wonder of we should require an explicit name/selector if the table name is >16 characters, rather than automatically truncating?

holic (approved):

packages/world/src/modules/core/tables/FunctionSelectors.sol
dk1a: maybe, I was going off the issue, both ways seem fine to me, with truncation being more convenient, and requirement more explicit.
@alvrs what do you think?

alvrs: During the last hackathon in Berlin, a bunch of people got confused by error messages about the selector being too long in systems that were automatically included in the config (we include all contracts ending in `System` by default).  Currently we require people to create an entry in `overrideSystems` for these systems where the default selector is too long. In this case I think truncating the default selector is the better experience.

For tables I agree it might be slightly confusing to see a `FunctionSelector` table even though it is called `FunctionSelectors` in the config. I would also find it slightly confusing to have to specify a name in the table config that is different from the key of this table in the table config (even though we support this right now). The most restrictive but imo most clear developer experience would be to always use the table's key in the `tables` config as name, and restrict that to 16 characters (by throwing a validation error). But on the other hand the restriction to 16 characters is a technical restriction that maybe we should "hide" from users, since the table id is mostly used internally (in which case I think defaulting to truncating the key, but allowing explicit overrides as implemented in this PR would be the best option).

Issue #703, opened 2023-05-02
[Store] Add v2 query library


alvrs:
- Similar to v1's `LibQuery`, but no need for the complex `ProxyRead` / `ProxyExpand` query fragments
- Takes a list of query fragments (eg `Has`, `HasValue`, `Not`, `NotValue`), returns a list of keys that matches the query (for the mvp it's probably fine to just work with single-primary-key tables)
- The list of query fragments is concatenated with a logical AND: The first fragment creates a list of entities matching this fragment, the following fragments filter entities from this initial list

alvrs:
in progress in #747

PR #747, opened 2023-05-06
feat: query library

yonadaaa:
solves #703.

This PR adds a new utility function, `query()` that uses the `KeysInTable` and `KeysInValue` modules to provide functionality similar to MUD1's LibQuery.  Currently it uses array filtering to check if keys are in tables, but could be optimised for tables that have the `hasKey` index enabled.


packages/world/test/query.t.sol
```diff
@@ -0,0 +1,198 @@
...
+    assertEq(keyTuplesHasHasValue[0][0], keys[0][0]);
+    bytes32[][] memory keyTuplesHasNotvalue = query(world, fragmentsHasNotvalue);
+    assertEq(keyTuplesHasNotvalue.length, 0);
+    bytes32[][] memory keyTuplesHasvalueHas = query(world, fragmentsHasvalueHas);
+    assertEq(keyTuplesHasvalueHas.length, 1);
+    // !gasreport running a HasValue, HasValue query with four keys
+    bytes32[][] memory keyTuplesHasvalueHasvalue = query(world, fragmentsHasvalueHasValue);
+    assertEq(keyTuplesHasvalueHasvalue.length, 1);
+  }
+}
```
holic, line 198: would be nice to have some gas tests for much larger arrays (e.g. 100, 1k, 10k values) just to see how these all scale

(I assume they scale horribly because solidity and arrays are not friends)

holic (dismissed):

packages/world/test/query.t.sol
yonadaaa: great idea, thanks

yonadaaa: added some now. we can't test 10k values until #804

holic (approved):

Issue #704, opened 2023-05-02
[Plugins] Add Factories plugin


alvrs:
See https://github.com/latticexyz/mud/issues/534 for context on "Factories"

PR #749, opened 2023-05-07
feat: add Factories plugin

yonadaaa:
Solves #704. 

This PR adds Factories into MUD. Users define a `template.config` (which is strongly typed based via codegen on `mud.config`), like so:

```
const factoryConfig = templateConfig({
  templates: {
    Example: {
      Statics: {
        v1: 1n,
        v2: 1,
        v3: "wasd",
        v4: "0x71C7656EC7ab88b098defB751B7401B5f6d8976F",
        v5: true,
        v6: "E1",
        v7: "E1",
      },
      Counter: { value: 2n },
    },
  },
});
```

The templates are created in `PostDeploy`. Once the world is deployed, a method can be used to create instances from the template data.

## Outstanding issues
- This could perhaps be more cleanly implemented as some plugin. 
- We should replace the type codegen by deriving the types from the config in Typescript.
- Tables can have different composite keys, so it's unclear how to generate a new unique key for each, per instance.


alvrs:
Very much like this, but we should wait until the plugin foundation is done. Will close for now, but we should reopen and update this once plugins are ready.

PR #752, opened 2023-05-08
feat: strong types for config

dk1a:


packages/store/ts/library/config/index.ts
```diff
@@ -1,21 +1,29 @@
...
-  MUDCoreConfig,
   mudCoreConfig,
   // schemas
   zObjectName,
   zSelector,
   zUserEnum,
   zValueName,
 } from "@latticexyz/config";
+// typeExtensions are necessary for `ExpandMUDUserConfig`, but configExtensions can be avoided
+import "../../register/typeExtensions";
```
alvrs, line 26: could you provide some more context for why this is needed? feels a little wrong to have the separation into `library` and `register` but then also have global type side effects in `library`

packages/store/ts/library/config/index.ts
```diff
@@ -96,15 +107,31 @@ export type FullTableConfig<
...
+export interface ExpandTableConfig<T extends TableConfig<string, string>, TableName extends string>
+  extends OrDefaults<
+    T,
+    {
+      directory: "tables";
+      name: TableName;
+      tableIdArgument: false;
+      storeArgument: false;
+      // TODO should dataStruct conditional default be expanded?
+      dataStruct: boolean;
```
alvrs, line 119: probably shouldn't be expanded, since its value is conditional on the number of schema fields, which we can't replicate in typescript type logic afaik

packages/store/ts/library/config/index.ts
```diff
@@ -227,20 +263,28 @@ export type MUDUserConfig<
...
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
+export interface ExpandMUDUserConfig<T extends MUDCoreUserConfig>
+  extends OrDefaults<
+    T,
+    {
+      enums: Record<string, never>;
+      namespace: "";
+      storeImportPath: "@latticexyz/store/src/";
+      userTypesPath: "Types";
+      codegenDirectory: "codegen";
```
alvrs, line 275: I think it would be good to store these in some defaults const, so we can refer to them from the `zod` config expansion logic and have a single source of truth

packages/world/ts/library/config/types.ts
```diff
@@ -28,6 +29,24 @@ export type SystemUserConfig =
...
 
+export interface ExpandSystemConfig<T extends SystemUserConfig, SystemName extends string>
+  extends OrDefaults<
+    T,
+    {
+      name: SystemName;
+      registerFunctionSelectors: true;
+    }
+  > {
+  openAccess: T["openAccess"];
```
alvrs, line 40: default for `openAccess` should be true I think

packages/world/ts/register/typeExtensions.ts
```diff
@@ -1,4 +1,8 @@
-import { WorldConfig, WorldUserConfig } from "../library";
+import { OrDefaults } from "@latticexyz/common/type-utils";
+import { MUDCoreUserConfig } from "@latticexyz/config";
+import { ExpandSystemsConfig, WorldConfig, WorldConfigDefaults, WorldUserConfig } from "../library";
```
alvrs, line 3: Here we import from `library`, and `library/config/index.ts` imports from here. Does this circular import cause issues? Does it matter in which order the type extensions of store and world are imported? (I can't tell for sure which one would be imported first now, but I guess still `store` because it is listed before the type definition below?)

alvrs (commented):
just a couple nits

packages/store/ts/library/config/index.ts
dk1a: removed the need for it by moving `mudConfig` and `ExpandMUDUserConfig` to `/register` export. It was needed because `mudConfig`, while expanding strong types, makes assumptions about keys in `MUDUserConfig` which only work locally or if `typeExtensions` are imported.
For example running type tests in `cli` (which doesn't import `/register` anywhere) would break those assumptions. But things that import `/library` shouldn't really need `mudConfig` or the global expander, so moving those to `/register` seems like the best solution

packages/store/ts/library/config/index.ts
dk1a: agree, change the comment to reflect this

packages/store/ts/library/config/index.ts
dk1a: improved defaults for both `store` and `world` (took much of the code from your PR)

packages/world/ts/library/config/types.ts
dk1a: yep, but it wasn't reflected in zod or types, fixed

packages/world/ts/register/typeExtensions.ts
dk1a: unlike `store`, this wasn't really a problem in `world`. https://github.com/latticexyz/mud/pull/752#discussion_r1188523956 removes any circular imports there

Issue #751, opened 2023-05-08
Proposal: Tables should expose an `encodeKey()` view


yonadaaa:
As part of the Factories PR (#749), I need to generically pass around table keys as an array of `bytes32`. This is difficult because the logic for serialising keys is inlined in table contracts:

```
 /** Get the full data (using the specified store) */
  function get(...) internal view returns (StaticsData memory _table) {
    bytes32[] memory _primaryKeys = new bytes32[](7);
    _primaryKeys[0] = bytes32(uint256((k1)));
    _primaryKeys[1] = bytes32(uint256(uint32((k2))));
    _primaryKeys[2] = bytes32((k3));
    _primaryKeys[3] = bytes32(bytes20((k4)));
    _primaryKeys[4] = _boolToBytes32((k5));
    _primaryKeys[5] = bytes32(uint256(uint8(k6)));
    _primaryKeys[6] = bytes32(uint256(uint8(k7)));

    bytes memory _blob = _store.getRecord(_tableId, _primaryKeys, getSchema());
    return decode(_blob);
  }
```

Like the `encode()` method for data, this logic should be moved to a `encodeKey()` method like:
```

  function encodeKey(...) internal pure returns (bytes32[] memory) {
    bytes32[] memory _primaryKeys = new bytes32[](7);
    _primaryKeys[0] = bytes32(uint256((k1)));
    _primaryKeys[1] = bytes32(uint256(uint32((k2))));
    _primaryKeys[2] = bytes32((k3));
    _primaryKeys[3] = bytes32(bytes20((k4)));
    _primaryKeys[4] = _boolToBytes32((k5));
    _primaryKeys[5] = bytes32(uint256(uint8(k6)));
    _primaryKeys[6] = bytes32(uint256(uint8(k7)));

    return _primaryKeys;
  }
```

I experimentally added this functionality [here](https://github.com/latticexyz/mud/blob/9ea37b53405c0c8329a58e483ee87cf62f8bdb93/packages/store/ts/library/render-solidity/renderTable.ts#L132) but it should be moved to a separate PR.

PR #753, opened 2023-05-08
feat: add encodeKey function to tablegen

yonadaaa:
Addresses #751.

This simplifies the Table contracts, but replacing the inlined code with an internal function increases gas use :(

It also causes `stack too deep` error on worlds with too many keys + fields.

### DIFF
--- a/packages/store/ts/library/render-solidity/renderTable.ts
+++ b/packages/store/ts/library/render-solidity/renderTable.ts
@@ -5,6 +5,7 @@ import {
   renderedSolidityHeader,
   renderRelativeImports,
   renderTableId,
+  renderValueTypeToBytes32,
   renderWithStore,
   renderTypeHelpers,
   RenderDynamicField,
@@ -28,7 +29,7 @@ export function renderTable(options: RenderTableOptions) {
     primaryKeys,
   } = options;
 
-  const { _typedTableId, _typedKeyArgs, _primaryKeysDefinition } = renderCommonData(options);
+  const { _keyArgs, _typedTableId, _typedKeyArgs } = renderCommonData(options);
 
   return `${renderedSolidityHeader}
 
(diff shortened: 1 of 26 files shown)


dk1a:
> It also causes stack too deep error on worlds with too many keys + fields.

I'm pretty sure too many keys+fields always caused stack too deep

dk1a:
> This simplifies the Table contracts, but replacing the inlined code with an internal function increases gas use :(

You could add the function but not replace inlined code, the rendering function encapsulates the logic already, and the solidity code is autogenerated and code repetition doesn't matter

holic:
> the solidity code is autogenerated and code repetition doesn't matter

+1 to this, I think we have a lot of room to optimize gas by just repeating code in areas of codegen

yonadaaa:
> I'm pretty sure too many keys+fields always caused stack too deep

Yes, it will always happen eventually, but now it happens on the [`Statics`](https://github.com/latticexyz/mud/blob/main/packages/cli/scripts/generate-test-tables.ts#L14) table in the CLI tests,  which it didn't before. I'm not sure why because `_primaryKeys` was there before.

> the solidity code is autogenerated and code repetition doesn't matter

Fair, but the `encode` function for data is called and not inlined, right? IMO it's nice if the contracts stand alone as Solidity - but yes, in reality they are always codegen'ed.

dk1a:
> Yes, it will always happen eventually, but now it happens on the [Statics](https://github.com/latticexyz/mud/blob/main/packages/cli/scripts/generate-test-tables.ts#L14) table in the CLI tests, which it didn't before. I'm not sure why because _primaryKeys was there before.

yeah probably some nuances of additional stack space for internal method calls. Inlining would solve it, so even if you don't wanna optimize gas here you might still wanna go for it

holic (dismissed):

packages/store/ts/library/render-solidity/renderTable.ts
```diff
@@ -129,6 +130,15 @@ library ${libraryName} {
     ])});
   }
 
+  /** Encode keys as a bytes32 array using this table's schema */
+  function encodeKey(${renderArguments([_typedKeyArgs])}) internal pure returns (bytes32[] memory _primaryKeys) {
```
holic, line 134: thinking we will want to consolidate our language around "key" vs "primary key" vs "key tuple" and wondering if we wanna just go ahead and name this `encodeKeyTuple` here

yonadaaa: Agreed - I will use `encodeKeyTuple`. I much prefer "key tuple" as we represent primary keys in many ways. 

To be clear I don't think we should rename the `_primaryKeys` vars in this PR (not that you are proposing that).

holic: yep! agreed

holic (approved):

Issue #757, opened 2023-05-08
user defined custom errors, events don't get added to world


holic:
```solidity
contract IncrementSystem is System {
  error MyCustomError();

  function willRevert() public view {
    revert MyCustomError();
  }
}
```

when the above gets compiled by MUD, the `IWorld` interface is missing `MyCustomError` so there's no way to parse it in the frontend when calling the system method via the world

I assume the same would happen with custom events. And although custom events may be an anti-pattern in terms of MUD best practices, we should probably still support it?

Issue #712, opened 2023-04-10
[CLI] Decode custom errors in forge and client


alvrs:
Currently custom errors are not correctly decoded in forge tests or the client, but only surfaced as `bytes4` selector

holic:
Might be fixed by https://github.com/latticexyz/mud/pull/767 and https://github.com/latticexyz/mud/pull/768

holic:
I did see this pop up recently in some foundry tests, so I think this is fixed.



![Image](https://github.com/latticexyz/mud/assets/508855/21d6351b-a112-411b-97f9-33a013d637fc)

PR #759, opened 2023-05-08
feat(store,world): add errors to IStore, rename error interfaces

holic:
Addresses two issues:
- `IStore` (and thus the downstream generated `IWorld`) was missing store error definitions. Because we will use `IWorld` ABI as our typed interface into the world and used to decode errors, we'll need these here.
- Renames the two `IErrors` interfaces because forge flattens these ABIs by name, so only one was appearing in the ABI output for a given project using both world and store.

Might help with https://github.com/latticexyz/mud/issues/712 but unclear without a reproduction of the original issue.

There's another issue around user-defined errors/events (#757) that we should address separately but is also needed to ensure our parsing of errors etc. is robust when calling methods on the world.


alvrs:
> Might help with https://github.com/latticexyz/mud/issues/712 but unclear without a reproduction of the original issue.

We should be able to reproduce the forge error part by calling `willRevert` (from #689) in a forge test in the example project

alvrs (approved):
good catch!

PR #763, opened 2023-05-09
docs: use a logo with white background so it shows up in dark mode browsers

Kooshaba:


holic:
back when I was fixing dark mode things in the docs site, I tried to add dark mode support with `srcset` but wasn't seeing a dark background in github when I toggled darkmode on

this works for now but would be cool to have it detect+toggle between the two transparent PNGs

holic (approved):

PR #764, opened 2023-05-09
feat: create SnapSyncModule

yonadaaa:
SnapSyncModule registers a system that allows clients to sync by using view functions instead of events over RPC.

This requires every table in the world to use the `KeysInTable` module.

**ONE ISSUE:** The world interface generation does not handle adding optional modules. SnapSyncModule functions currently exist in the interface, but rely on users manually registering the module in their config to register them.


examples/minimal/packages/contracts/foundry.toml
```diff
@@ -2,7 +2,7 @@
 solc_version = "0.8.13"
 ffi = false
 fuzz_runs = 256
-optimizer = true
+optimizer = false
```
yonadaaa, line 5: Set this to `true` and `SyncSystem` fails to compile. Something weird about 2D arrays: https://github.com/ethereum/solidity/issues/11638

yonadaaa: Yes, definitely - we should NOT ship this. I need to look into it but wanted to get the code working

it might actually be fine now the function signature has changed

examples/minimal/packages/client-react/package.json
```diff
@@ -25,7 +25,7 @@
     "@latticexyz/world": "link:../../../../packages/world",
     "@wagmi/chains": "^0.2.22",
     "async-mutex": "^0.4.0",
-    "contracts": "workspace:*",
+    "contracts": "workspace:../contracts",
```
holic, line 28: any reason this changed?

Kooshaba: forget why i changed, just tried original and it worked, changing back

examples/minimal/packages/client-react/src/mud/setup.ts
```diff
@@ -48,11 +48,25 @@ export async function setup() {
...
 
+  const signerOrProvider = signer ?? result.network.providers.get().json;
   // Create a World contract instance
-  const worldContract = IWorld__factory.connect(
-    networkConfig.worldAddress,
-    signer ?? result.network.providers.get().json
-  );
+  const worldContract = IWorld__factory.connect(networkConfig.worldAddress, signerOrProvider);
+
+  const currentBlockNumber = await awaitStreamValue(result.network.blockNumber$);
```
alvrs, line 55: maybe we should just ping the RPC for the current block number here to not block for too long in case the chain has slow block times

examples/minimal/packages/client-react/src/mud/snapSync/getSnapSyncRecords.ts
```diff
@@ -0,0 +1,48 @@
...
+export async function getSnapSyncRecords(
+  worldAddress: string,
+  contractComponents: ReturnType<typeof defineContractComponents>,
+  currentBlockNumber: number,
+  signerOrProvider: Signer | providers.JsonRpcProvider
+) {
+  const snapSyncContract = new Contract(worldAddress, snapSyncSystemAbi, signerOrProvider);
+
+  const chunkSize = 100;
+  const tableIds = Object.values(contractComponents).map((c) => c.metadata.contractId);
```
alvrs, line 16: I'd propose passing in the table ids (and grabbing them directly from the mud config in `setup.ts`)

packages/world/src/modules/snapsync/SnapSyncSystem.sol
```diff
@@ -0,0 +1,50 @@
...
+
+import { getKeysInTable } from "../keysintable/getKeysInTable.sol";
+import { KeysInTable, KeysInTableTableId } from "../keysintable/tables/KeysInTable.sol";
+import { SyncRecord } from "./SyncRecord.sol";
+
+function singletonKey(bytes32 key) pure returns (bytes32[] memory) {
+  bytes32[] memory keyTuple = new bytes32[](1);
+  keyTuple[0] = key;
+  return keyTuple;
+}
```
alvrs, line 16: "singleton" is a really overloaded term, maybe we should just call it `keyToTuple` or something?

examples/minimal/packages/contracts/mud.config.ts
```diff
@@ -24,10 +24,20 @@ export default mudConfig({
     },
   },
   modules: [
+    {
+      name: "KeysInTableModule",
+      root: true,
+      args: [resolveTableId("CounterTable")],
+    },
```
alvrs, line 31: if we have capacity before wednesday we could add a super simple plugin that converts the config by adding this module for all tables in the config

examples/minimal/packages/client-react/src/mud/snapSync/getSnapSyncRecords.ts
alvrs: added this

examples/minimal/packages/client-react/src/mud/setup.ts
alvrs: added this

alvrs:
Merged with main and addressed my own feedback (see above). Open todos:
- [ ] update the `setup` functions of the templates and `client-vanilla`
- [ ] somehow snap sync is stuck now after merging with main, I can dig into this tomorrow

packages/world/src/modules/snapsync/SnapSyncSystem.sol
yonadaaa: This is similar logic to https://github.com/latticexyz/mud/blob/main/packages/world/src/modules/keysintable/query.sol#L23

alvrs (dismissed):
lgtm! We should revisit autogenerating the snap sync ABI after eth global (opened an issue to track: https://github.com/latticexyz/mud/issues/836)

PR #765, opened 2023-05-09
feat(world): delete record for empty keysWithValue array

yonadaaa:
@holic this will tidy the dev UI by removing redundant records.

We could also take this chance to refactor the internals of `KeysWithValue` to be like `KeysInTable` (ie. no `tableIdArgument`, removing unbounded array filter). But I don't think it is worth the time investment with MUD2 still in flux.

### DIFF
--- a/packages/world/gas-report.json
+++ b/packages/world/gas-report.json
@@ -57,7 +57,7 @@
     "source": "test/KeysWithValueModule.t.sol",
     "name": "set a record on a table with KeysWithValueModule installed",
     "functionCall": "world.setRecord(namespace, sourceName, keyTuple1, abi.encodePacked(value))",
-    "gasUsed": 169314
+    "gasUsed": 165686
   },
   {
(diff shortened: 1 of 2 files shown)


holic (approved):
thank you!

packages/world/gas-report.json
```diff
@@ -69,13 +69,13 @@
...
     "functionCall": "world.setRecord(namespace, sourceName, keyTuple1, abi.encodePacked(value2))",
-    "gasUsed": 135311
+    "gasUsed": 135337
   },
   {
     "source": "test/KeysWithValueModule.t.sol",
     "name": "delete a record on a table with KeysWithValueModule installed",
     "functionCall": "world.deleteRecord(namespace, sourceName, keyTuple1)",
-    "gasUsed": 57919
+    "gasUsed": 54294
```
yonadaaa, line 78: the storage refundoooor

PR #766, opened 2023-05-09
deploy example app

holic:


holic:
hmm example is a little difficult to deploy with linked deps, which means we need to install foundry and build everything

holic:
haha seems like @roninjin10 has been here recently: https://github.com/foundry-rs/foundry/issues/4337 https://github.com/evmts/evmts-monorepo/issues/23

holic:
changing to canary deps works but not ideal, because PR preview deploys won't show any package changes in a given PR (which is my ultimate goal with this)

image

the install command is
```
pnpm recursive exec npx --package=@latticexyz/cli@canary -y mud set-version --mudVersion canary && pnpm install --no-frozen-lockfile
```

PR #768, opened 2023-05-09
feat(common,world): include errors in worldgen's system interfaces

dk1a:
fixes https://github.com/latticexyz/mud/issues/757
only errors, events are not only an antipattern but wouldn't even work for any non-root system, since those aren't delegatecalled

### DIFF
--- a/packages/common/src/codegen/utils/contractToInterface.ts
+++ b/packages/common/src/codegen/utils/contractToInterface.ts
@@ -9,6 +9,11 @@ export interface ContractInterfaceFunction {
   returnParameters: string[];
 }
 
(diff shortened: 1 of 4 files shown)


holic:
Do we wanna add an example system to the examples dir for now to demonstrate this working?

dk1a:
> Do we wanna add an example system to the examples dir for now to demonstrate this working?

Yes, but I'm setting up a separate folder for integration tests, this can serve as a nice first example for them

> Or is there a way to write automated tests for this?

I'm thinking of trying to access an error's selector in a test. It will fail if the selector doesn't exist

holic (approved):

PR #769, opened 2023-05-09
refactor: standardise keysWithValue, keysInTable modules

yonadaaa:
## Proposal
This PR refactors `KeysInTable` and `KeysWithValue` to standardise the API.

Each module installs two tables:
- an array called `[name]Keys` that store the keys that match the criteria (ie. in table or has value) 
- a reverse mapping called `[name]Index` which stores the index of each key in the array

Accordingly, each module provides two getter functions:
- `getKeys()` returns all keys in the array
- `hasKey()` returns whether a specific key is in the array

These functions have the same name for but the "value" version has an additional argument, ie. 
`hasKey(bytes32 tableId, bytes32[] memory key)` vs 
`hasKey(bytes32 tableId, bytes memory value, bytes32[] memory key)`


packages/world/gas-report.json
```diff
@@ -2,98 +2,92 @@
...
+    "functionCall": "bytes32[][] memory keysWithValue = getKeysWithValue(world, tableId, abi.encode(value1))",
+    "gasUsed": 8679
   },
   {
     "source": "test/KeysWithValueModule.t.sol",
     "name": "install keys with value module",
-    "functionCall": "world.installRootModule(keysWithValueModule, abi.encode(sourceTableId))",
-    "gasUsed": 609250
+    "functionCall": "world.installRootModule(keysWithValueModule, abi.encode(tableId))",
+    "gasUsed": 1036002
```
yonadaaa, line 48: increase gas is expected because we now do a high (but constant) amount of storage writes. The cost amortises over many records - we should set up a gas report for this

packages/world/gas-report.json
```diff
@@ -2,98 +2,92 @@
...
   },
   {
     "source": "test/KeysWithValueModule.t.sol",
     "name": "change a record on a table with KeysWithValueModule installed",
     "functionCall": "world.setRecord(namespace, sourceName, keyTuple1, abi.encodePacked(value2))",
-    "gasUsed": 135337
+    "gasUsed": 340354
```
ludns, line 66: What's the trade-off space here? This is a significant bump in gas cost, and I wonder when this ends up being cheaper than the previous implementation (unbounded is bad in theory, but when does it get bad in practice?)

packages/world/mud.config.ts
```diff
@@ -91,19 +91,27 @@ export default mudConfig({
     KeysWithValue: {
       directory: "modules/keyswithvalue/tables",
       primaryKeys: {
+        sourceTable: "bytes32",
         valueHash: "bytes32",
       },
-      schema: {
-        keysWithValue: "bytes32[]", // For now only supports 1 key per value
+      schema: { length: "uint32", keys: "bytes32[]" }, // For now only indexes the first key of a primary key tuple
```
ludns, line 97: We are already storing the length of the array in Store, why do we need to write the length as a filed too? Is it because it's blocked on getting the low level "getLength" function from StoreCore?

dk1a: yep, I haven't had the time to implement it yet

packages/world/gas-report.json
yonadaaa: Good point. It's hard to evaluate these changes if we haven't measure the full space. 

I just opened #771 which adds gas reports for setting 10, 100, and 1000 keys. The costs for the original code are:

10 keys: **134842** gas
100 keys: **192968** gas
100 keys: **790975** gas

I merged those changes into this PR, which now has costs:

10 keys: **202529** gas
100 keys: **260655** gas
1000 keys: **858704** gas

Two observations: 
- the cost for the original code grows quite slowly.
- the cost for the new code is **NOT constant** and is always more expensive.

I think this is down to missing array methods. We cannot access index `Store` arrays, meaning we have to read the entire array unbounded from memory. We also have to maintain the array length, which uses an extra storage slot.

yonadaaa: If those two optimisations don't significantly amortise and reduce the constant cost, this says something about how we should program!

yonadaaa: Update: 

I added the new array methods, and tightly packed the storage slots in the `Index`. The new method now incurs an overhead of 41593 gas over the old method when setting 10, 100, 1000 keys.

packages/world/mud.config.ts
yonadaaa: I've updated this PR with those changes

alvrs:
let's revisit this once plugins are ready and v2 is stable

PR #770, opened 2023-05-09
feat(integration): initialize project directory for integration tests

dk1a:
most tests don't need a separate package and will go into `misc` (feel free to suggest a better name)
As the first example I test error generation from https://github.com/latticexyz/mud/pull/768


integration/misc/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "mud-integration-tests-misc",
```
holic, line 2: do we expect more than just one here? will it be useful to have isolated integration projects?

if so, it would be nice to come up with a better name than "misc"

if not, we could just make `integration` the project root (or better yet: `integration-tests`)

integration/misc/packages/contracts/src/systems/ErrorTestSystem.sol
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
+
+import { System } from "@latticexyz/world/src/System.sol";
+import { Position } from "../CustomTypes.sol";
+
+contract ErrorTestSystem is System {
```
holic, line 7: Since the directory here infers that this is a "test" system, how about calling this `SystemWithCustomErrors`?

integration/misc/package.json
dk1a: although I could just have a 2nd contracts package, it doesn't need to be multiple projects, will move it to root for now

integration/misc/packages/contracts/src/systems/ErrorTestSystem.sol
dk1a: changed, except so it ends with `System` to include it in worldgen by default

holic: oh right, I forgot about that requirement

integration/misc/package.json
holic: thoughts on renaming the integration dir to integration-sandbox to match?

integration/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "mud-integration-sandbox",
```
holic, line 2: thoughts on renaming the integration dir to integration-sandbox to match?

holic (approved):

integration/package.json
dk1a: tried it out, wdyt?

holic (approved):

PR #771, opened 2023-05-09
test(world): add module gas reports for setting many keys

yonadaaa:
Tests setting 10, 100, 1000, ~10000~ records with `KeysInTable` and `KeysWithValue`.


packages/world/test/KeysWithValueModule.t.sol
```diff
@@ -208,4 +208,79 @@ contract KeysWithValueModuleTest is Test {
...
+    for (uint256 i; i < amount - 1; i++) {
+      bytes32[] memory key = new bytes32[](1);
+      key[0] = bytes32(i);
+      world.setRecord(namespace, sourceName, key, abi.encodePacked(value));
+    }
+
+    // !gasreport Setting the last of 10 keys
+    world.setRecord(namespace, sourceName, lastKey, abi.encodePacked(value));
+
+    // !gasreport Get list of 10 keys with a given value
```
yonadaaa, line 230: It would be good if the gas report value was a parameter and not hardcoded in a comment - because I need to make 3 of these functions.

dk1a: yep, for now you could try putting
```solidity
uint256 amount = 100;
uint256 value = 1;

bytes32[] memory lastKey = new bytes32[](1);
lastKey[0] = bytes32(uint256(amount - 1));

for (uint256 i; i < amount - 1; i++) {
  bytes32[] memory key = new bytes32[](1);
  key[0] = bytes32(i);
  world.setRecord(namespace, sourceName, key, abi.encodePacked(value));
}
```
in a helper to reduce the issue of repetition

yonadaaa: Oh yup. Even that that would be better, thanks.

alvrs (dismissed):

alvrs:
unclear to me atm why the gas report step fails in the CI but works locally

alvrs:
alright seems like setting 10k keys was too much for the CI

alvrs:
Unfortunately this change increases the time for the `Build and validate artifacts` action from 5 minutes to 10 minutes (see https://github.com/latticexyz/mud/actions/runs/5294523734/jobs/9583964860 for comparison), which unfortunately is not a good tradeoff for adding these gas reports.

Because of that I'll close this PR for now, but we can look at it for reference when optimizing the `KeysInTable` / `KeysWithValue` modules.

We should also find a way to run these gas reports more efficiently, the `Build and validate artifacts` is already the slowest action without this change.

holic:
> We should also find a way to run these gas reports more efficiently, the Build and validate artifacts is already the slowest action without this change.

I'm hoping we can soon move over to Vulcan gas reports (or pull in the same strategy) once they land, which will mean the gas reports and tests are run at the same time: https://github.com/nomoixyz/vulcan/pull/104

PR #773, opened 2023-05-10
feat(cli): remove v1 mud bin in favor of v2

holic:
Fixes an issue where `npx` doesn't know which bin to use


alvrs (approved):
lgtm, should we also remove the deprecated commands folder altogether? I think there were some issues with stuff in there on windows even if it wasn't used in the cli (can't reproduce unfortunately)

alvrs:
also do you know why the template build is failing?

holic:
from @Kooshaba in Discord:

> i use some of them for the current Sky Strife deployment flow, but it is hard linked to a super old MUD version so should be totally fine
> good to remove 👍

holic:
> lgtm, should we also remove the deprecated commands folder altogether? I think there were some issues with stuff in there on windows even if it wasn't used in the cli (can't reproduce unfortunately)

I was able to remove a whole bunch of v1 stuff!

holic:
Oops, I think I made a circular dependency. Templates in this branch are still on v1 but this removes v1 commands, but I want these CLI changes in v2 templates branch.

PR #780, opened 2023-05-10
build: allow any mud version in peer deps

holic:
I am having trouble installing a template using a canary version because canary versions don't match the `latest` tag used in peer deps, thus peer deps are not resolving and pnpm complains.

Since our peer dep strategy is ~temporary (see https://github.com/latticexyz/mud/issues/411) and this will only affect canary releases anyway, this seems like a decent quick fix?


holic:
gonna merge to unlock myself, will revert if this causes trouble

PR #781, opened 2023-05-10
fix(dev-tools): publish dist

holic:
whoops, forgot an `.npmignore` file and so `dist` wasn't getting published


holic:
going to self-review since this is just a copy paste from our other packages

PR #775, opened 2023-05-10
feat(create-mud): upgrade templates to v2

holic:
closes #708 
closes #701

TODO:
- [x] use singleton pattern instead of singleton key
- [x] can we move to `mud.config.ts`?
- [x] remove `Table` suffix from tables
- [x] figure out if we need `.env` defaults?
- [x] remove localhost chain (wagmi's foundry chain has WS enabled now)
- [x] should we check in types or not?
- [x] should we check in deploys or not?
- [x] add react eslint presets
- [x] add `.eslintrc` to project root with `"root": true`
- [x] add dev tools

PR #784, opened 2023-05-10
build(network): clean up deps

holic:
related to #411 

Saw issues in v2 templates (#775) related to `store` being a peer dep here, which caused `dev-tools` to fail to build inside a client app, so I figured I'd clean this up.

I did this by going through each peer dep, looking at import usages, and deciding if it should be a regular dep (if used within the app bundle) or dev dep (if only used in tests, type checking, etc.). I also removed a couple extraneous deps.


packages/network/package.json
```diff
@@ -34,66 +34,40 @@
...
+  "devDependencies": {
     "@types/debug": "^4.1.7",
     "@types/google-protobuf": "^3.15.6",
     "@types/jest": "^27.4.1",
     "@types/lodash": "^4.14.182",
     "@types/node": "^18.15.11",
-    "async-mutex": "^0.3.2",
-    "ethers": "^5.7.2",
     "fake-indexeddb": "^4.0.0",
-    "google-protobuf": "^3.21.0",
```
holic, line 61: I couldn't find any usages of this and the example app seems to run locally without issue.

packages/network/package.json
```diff
@@ -34,66 +34,40 @@
...
     "@ethersproject/providers": "^5.7.2",
     "@improbable-eng/grpc-web": "^0.15.0",
+    "@latticexyz/common": "workspace:*",
     "@latticexyz/recs": "workspace:*",
     "@latticexyz/schema-type": "workspace:*",
     "@latticexyz/services": "workspace:*",
     "@latticexyz/solecs": "workspace:*",
     "@latticexyz/store": "workspace:*",
     "@latticexyz/utils": "workspace:*",
-    "@protobuf-ts/grpcweb-transport": "^2.7.0",
```
holic, line 52: I couldn't find any usages of this and the example app seems to run locally without issue.

packages/network/package.json
alvrs: `google-protobuf` is a peer dependency of `@improbable-eng/grpc-web`, which is a dependency of `nice-grpc-web`. It doesn't look like `nice-grpc-web` is installing `google-protobuf`, but it also seems like it works without it (and pnpm doesn't complain about missing indirect peer dependencies), so down to remove it for now.

packages/network/package.json
alvrs: yeah we just use the web transport exports of `@improbable-eng/grpc-web` (which is installed by `nice-grpc-web`)

alvrs (approved):

PR #785, opened 2023-05-10
build(std-client): remove peer deps

holic:
related to #411 


packages/std-client/package.json
```diff
@@ -50,43 +45,24 @@
...
-    "react-collapse": "^5.1.1",
-    "react-dom": "^18.2.0",
-    "rimraf": "^3.0.2",
-    "rxjs": "^7.5.5",
-    "tslib": "^2.5.0",
     "tsup": "^6.7.0",
     "tsx": "^3.12.6",
     "typedoc": "0.23.21",
     "typedoc-plugin-markdown": "^3.13.6",
-    "typescript": "^4.9.5",
```
alvrs, line 71: do we not need typescript as a dev dependency? (it seems to work without but i'm surprised)

holic: It's installed at the root since ~all packages need it and need to be aligned to the same version.

PR #789, opened 2023-05-10
build(solecs): clean up deps

holic:


packages/solecs/package.json
```diff
@@ -21,32 +21,23 @@
...
-    "@solidstate/contracts": "^0.0.52",
-    "@typechain/ethers-v5": "^10.2.0",
-    "@types/mocha": "^9.1.1",
     "ds-test": "https://github.com/dapphub/ds-test.git#c9ce3f25bde29fc5eb9901842bf02850dfd2d084",
     "ethers": "^5.7.2",
+    "memmove": "https://github.com/dk1a/memmove.git#ffd71cd77b1708574ef46a667b23ca3a5cc9fa27",
+    "solmate": "https://github.com/transmissions11/solmate.git#9cf1428245074e39090dceacb0c28b1f684f584c"
+  },
+  "devDependencies": {
+    "@solidstate/contracts": "^0.0.52",
```
alvrs, line 33: I think `@solidstate/contracts` should be a dependency so it's automatically installed by consumers

packages/solecs/package.json
```diff
@@ -21,32 +21,23 @@
...
     "typechain": "^8.1.1",
     "typedoc": "0.23.21"
   },
-  "peerDependencies": {
-    "@ethersproject/abi": "^5.7.0",
-    "@ethersproject/providers": "^5.7.2",
-    "@solidstate/contracts": "^0.0.52",
-    "ds-test": "https://github.com/dapphub/ds-test.git#c9ce3f25bde29fc5eb9901842bf02850dfd2d084",
-    "ethers": "^5.7.2",
-    "forge-std": "https://github.com/foundry-rs/forge-std.git#b4f121555729b3afb3c5ffccb62ff4b6e2818fd3",
```
alvrs, line 48: moving `forge-std` to dependencies so it's installed automatically by consumers

packages/solecs/package.json
holic: oops, good catch

packages/solecs/package.json
holic: just a dev dep should be fine for forge-std

PR #791, opened 2023-05-10
build(noise): clean up deps

holic:


packages/noise/package.json
```diff
@@ -17,39 +17,24 @@
...
-    "@types/mocha": "^9.1.1",
     "@types/node-fetch": "^2.6.2",
-    "abdk-libraries-solidity": "^3.0.0",
     "assemblyscript": "^0.21.2",
-    "base64-sol": "^1.1.0",
-    "big-integer": "^1.6.51",
     "chai": "^4.3.6",
     "ds-test": "https://github.com/dapphub/ds-test.git#c9ce3f25bde29fc5eb9901842bf02850dfd2d084",
     "ethers": "^5.7.2",
-    "forge-std": "https://github.com/foundry-rs/forge-std.git#b4f121555729b3afb3c5ffccb62ff4b6e2818fd3",
```
alvrs, line 41: i'd prefer to keep `forge-std` installed (and use it for `forge-std/Test.sol` in tests instead of `ds-test/test.sol`), because it makes it easier to use `console.log` if needed without having to first install `forge-std` (and figuring out the right commit hash)

alvrs (approved):

alvrs (approved):

PR #792, opened 2023-05-10
build(phaserx): move to tsup, clean up deps

holic:
related to https://github.com/latticexyz/mud/issues/411 and https://github.com/latticexyz/mud/issues/612

For peer deps, I did this by going through each dep, looking at import usages, and deciding if it should be a regular dep (if used within the app bundle) or dev dep (if only used in tests, type checking, etc.). I also removed a couple extraneous deps.


packages/phaserx/package.json
```diff
@@ -7,48 +7,33 @@
...
   "dependencies": {
-    "@use-gesture/vanilla": "10.2.9"
-  },
-  "devDependencies": {
     "@latticexyz/utils": "workspace:*",
-    "@rollup/plugin-commonjs": "^21.0.2",
-    "@rollup/plugin-node-resolve": "^13.1.3",
-    "@rollup/plugin-typescript": "^11.1.0",
-    "@types/jest": "^27.4.1",
-    "jest": "^29.3.1",
```
holic, line 33: There were no tests, so I removed all the jest stuff.

packages/phaserx/package.json
```diff
@@ -7,48 +7,33 @@
...
-    "phaser": "3.60.0-beta.14",
-    "rimraf": "^3.0.2",
-    "rollup": "^2.70.0",
-    "rollup-plugin-peer-deps-external": "^2.2.4",
-    "rxjs": "^7.5.5",
-    "ts-jest": "^29.0.5",
-    "tslib": "^2.5.0",
-    "typedoc": "0.23.21",
-    "typedoc-plugin-markdown": "^3.13.6",
-    "typescript": "^4.9.5"
```
alvrs, line 45: everything seems to work without `typescript` as a dev dependency, but I'm confused why (would have expected eg. `tsc` to require `typescript` as a dep)

alvrs (approved):

packages/phaserx/package.json
holic: typescript is installed at the root (since we need it ~everywhere and its version consistent across packages)

PR #793, opened 2023-05-10
build(ecs-browser): clean up deps

holic:

### DIFF
--- a/packages/ecs-browser/tsup.config.js
+++ b/packages/ecs-browser/tsup.config.js
@@ -4,8 +4,8 @@ export default defineConfig({
   entry: ["src/index.ts"],
   outDir: "dist",
   format: ["esm"],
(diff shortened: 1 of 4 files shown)


packages/ecs-browser/tsup.config.js
```diff
@@ -4,8 +4,8 @@ export default defineConfig({
   entry: ["src/index.ts"],
   outDir: "dist",
   format: ["esm"],
+  dts: false,
   sourcemap: true,
   clean: true,
-  dts: true,
-  noExternal: ["goober", "shiki"],
```
holic, line 10: not sure if it's necessary anymore to force these into the bundle, so I'm trying without them for now

but we'll likely remove this package from v2 anyway

alvrs (approved):

PR #797, opened 2023-05-11
test: setting 10000 records in a Store array

yonadaaa:
This minimal repro shows that arrays in MUD have an implicit max size which is < 10000 elements.

The length returned here is 170 instead of 10000, indicating some overflow error.


yonadaaa:
Closing this as #804 fixed the array length issue

PR #801, opened 2023-05-11
feat(dev-tools): bundle react deps

holic:
tsup by default [doesn't bundle deps](https://tsup.egoist.dev/#excluding-packages):
> By default tsup bundles all import-ed modules but dependencies and peerDependencies in your packages.json are always excluded

So when we use this package in a non-react project, it will fail to mount because of missing react deps.

We could declare these deps as peer deps, but we've had bad experiences with that, and it feels awkward for non-react projects to have to have react deps to use the dev tools.

We're gonna try bundling react deps, even though it may cause issues if multiple versions of react are detected on the page. We'll address that as it comes up.

### DIFF
--- a/packages/dev-tools/src/mount.tsx
+++ b/packages/dev-tools/src/mount.tsx
@@ -5,18 +5,18 @@ export async function mount() {
   }
 
   try {
-    const { StrictMode } = await import("react");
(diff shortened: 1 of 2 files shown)


holic:
gonna merge to unblock my work on #775 (this package isn't really in use yet anyway)

PR #803, opened 2023-05-11
feat(cli,world): deploy writes to worlds.json

holic:
related to #680 

Puts all the latest world addresses by chain into a `worlds.json` file (by default, but is configurable) so that we can have one file to import/read from rather than an import per chain.

This intentionally doesn't include the block number for local chains so the file can remain somewhat consistent between local deploys, since the file is designed to be checked in. We'll lean on [the new `emit HelloWorld()`](#742) to find the block number by querying for the event at the start of the sync.


ludns:
The issue with the event approach is I am not sure RPCs will let you query for the event over the whole chain; or at the very least you will need to partition that into multiple requests (and have some code to determine from the RPC what an acceptable event range is).

Checking for an event still requires you to go through the bloom filter for every block. It's a linear operation.

We could check for an Erigon node, and then use [`ots_getContractCreator` ](https://github.com/otterscan/otterscan/blob/develop/docs/custom-jsonrpc.md) to find the TX where the contract got deployed.

holic:
> Checking for an event still requires you to go through the bloom filter for every block. It's a linear operation.

TIL!

holic:
> The issue with the event approach is I am not sure RPCs will let you query for the event over the whole chain; or at the very least you will need to partition that into multiple requests (and have some code to determine from the RPC what an acceptable event range is).

Alchemy let's me:

image

(for event `hello()` from OKPC contract deployed at block 14434985)

holic:
> Hopefully Anvil can do it too :)
Anvil has no trouble with it either!

holic:
gonna wait on this until v2 templates land, then carry over the client changes there

examples/minimal/packages/contracts/worlds.json
```diff
@@ -0,0 +1,9 @@
+{
+  "31337": {
+    "address": "0x5FbDB2315678afecb367f032d93F642f64180aa3"
+  },
+  "4242": {
+    "address": "0xf0F5e9b00b92f3999021fD8B88aC75c351D93fc7",
+    "blockNumber": 10591501
+  }
```
holic, line 8: another potential structure for this is

```
[{
  chainId: 31337,
  address: "0x...",
  blockNumber: 123456,
}],
```
but I figured that'd be a little more painful to "splice" if we need to replace a world address/block number for a particular chain

templates/vanilla/packages/contracts/deploys/31337/latest.json
```diff
@@ -1,4 +1,4 @@
 {
   "worldAddress": "0x5FbDB2315678afecb367f032d93F642f64180aa3",
-  "blockNumber": 9
+  "blockNumber": 10
```
alvrs, line 3: should we git ignore this file?

alvrs (approved):

templates/vanilla/packages/contracts/deploys/31337/latest.json
holic: whoops, it was ignored but forgot to delete

Issue #798, opened 2023-05-11
bug: weird behaviour when pushing many elements to a Store array 


yonadaaa:
See #797 for a minimal repro.

Consider this simple MUD table with a single `address[]` field.

```
AddressArray: {
  directory: "../test/tables",
  schema: "address[]",
  tableIdArgument: true,
},
```

In the test example, the result of `AddressArray.length(...)` after pushing 1000 keys is, as expected, 1000. However, after setting 10000 keys, the length returned is **170**.

In addition, pushing the 1000th element costs a large **455801** gas, but pushing the 10000th element costs less at **93081** gas.

PR #804, opened 2023-05-11
feat(store): resize PackedCounter to use 5 bytes per length

dk1a:
fixes https://github.com/latticexyz/mud/issues/798


packages/network/src/v2/schemas/decodeData.ts
```diff
@@ -29,13 +29,20 @@ export const decodeData = (schema: TableSchema, hexData: string): Record<number,
...
+    const packedCounterCounterType = SchemaType.UINT40;
+    const dynamicDataLength = decodeStaticField(packedCounterAccumulatorType, dynamicDataLayout, 0) as number;
 
     schema.dynamicFields.forEach((fieldType, i) => {
-      const dataLength = dynamicDataLayout.getUint16(4 + i * 2);
+      const dataLength = decodeStaticField(
+        packedCounterCounterType,
+        dynamicDataLayout,
+        getStaticByteLength(packedCounterAccumulatorType) + i * getStaticByteLength(packedCounterCounterType)
+      ) as number;
```
holic, line 45: haven't actually tested this yet! please test before merging!

I assume that before these client changes, array values wouldn't decode properly (they'd be sliced in the wrong places). These changes should address that.

packages/cli/src/commands/deploy.ts
```diff
@@ -51,64 +26,8 @@ export const yDeployOptions = {
...
-      console.log(
-        chalk.bgGreen(
-          chalk.whiteBright(`\n Deployment result (written to ${mudConfig.worldsFile} and ${outputDir}): \n`)
-        )
-      );
-    }
-
-    console.log(deploymentInfo);
-    return deploymentInfo;
+    deployHandler(args);
```
alvrs, line 30: this seems to be an infinite recursion now? I guess we want to import `deployHandler` from `deployHandler`?

packages/config/src/node/loadConfig.ts
```diff
@@ -15,7 +15,7 @@ export async function loadConfig(configPath?: string): Promise<unknown> {
     configPath = await resolveConfigPath(TEMP_CONFIG);
     return (await import(configPath)).default;
   } finally {
-    rmSync(TEMP_CONFIG);
+    rmSync(TEMP_CONFIG, { force: true });
```
alvrs, line 18: why is `force: true` needed here?

packages/cli/src/commands/deploy.ts
dk1a: looks like I broke it while merging, tried to fix, can test some time later today/tomorrow

packages/config/src/node/loadConfig.ts
dk1a: if something within `try` throws, the TEMP_CONFIG may not exist, which will cause rmSync to throw, which will swallow the original error. force makes rmSync do nothing if the file doesnt exist

packages/cli/src/commands/deploy.ts
```diff
@@ -125,7 +35,12 @@ const commandModule: CommandModule<DeployOptions, DeployOptions> = {
   },
 
   async handler(args) {
-    await deployHandler(args);
+    try {
+      deployHandler(args);
```
holic, line 39: should this `await`?

alvrs: it should, fixed it

alvrs (approved):

alvrs (dismissed):

alvrs (approved):

PR #805, opened 2023-05-12
feat(store-cache): add typescript datastore with support for composite keys

alvrs:
This PR adds a new client data store with support for composite keys, typed based on a given `StoreConfig`. The intention is for this to be used in combination with with the network package to mirror the state of a `Store` contract to the client. (Previously we used `recs` for this purpose, but `recs` does not support composite keys).

Todos in this PR:
- [x] add default values for each abi type
- [x] add mapping to primitive type for every abi type

Followup:
- [ ] add `recs` wrapper around a store cache (to be able to interact with contract ecs data via `recs` as before)
- [ ] update `setupMUDV2Network` to pipe contract data into store core rather than `recs`


packages/schema-type/src/typescript/utils/getAbiTypeDefaultValue.ts
```diff
@@ -0,0 +1,11 @@
+import { AbiTypeToPrimitiveType } from "../mappings";
+import { AbiTypeToDefaultValue } from "../mappings/AbiTypeToDefaultValue";
+
+/**
+ * Return a value corresponding to the Solidity default value for a given abi type
+ */
+export function getAbiTypeDefaultValue<T extends string>(abiType: T) {
+  // Remove fixed array lengths (eg map uint256[10] to uint256[])
+  const normalizedType = abiType.replace(/(\w+)\[\d+\]/g, "$1[]");
```
holic, line 9: let's hope no one names their array `$items` 🤓

packages/store-cache/package.json
```diff
@@ -0,0 +1,39 @@
...
+  },
+  "license": "MIT",
+  "type": "module",
+  "exports": {
+    ".": "./dist/index.js"
+  },
+  "types": "src/index.ts",
+  "scripts": {
+    "build": "tsup",
+    "clean": "rimraf dist",
```
holic, line 19: not blocking and not at all critical but I've been trying to keep the build scripts somewhat aligned so if we ever want to revisit turbo build steps, etc., we don't have to go through and audit all the build scripts again

holic (approved):
🎉🎉🎉🎉

holic (approved):
🎉🎉🎉🎉

dk1a (dismissed):

PR #807, opened 2023-05-12
fix(recs,cli,std-client): fix recs types for bytes arrays

holic:
I noticed my `bytes32[]` field was coming through as `BigIntArray`

image
image


Kooshaba (approved):
lgtm

PR #808, opened 2023-05-12
feat: sync data to store cache, decode keys

alvrs:
- Adds decoding of key tuples and passing them through the `NetworkEvent` stream
- Adds an optional (enabled by default) connection to `store-cache` as client data backend that allows interacting with the data with typed key tuples (data is still also stored to `recs`, `store-cache` is just an add-on for now)
- Removes some v1 networking code that's not used anymore (would have had to update it with the changes of this PR, which seemed really unnecessary)
- [x] pulled out from this PR and should be merged first:
  - #809
  - #810
  - #811 
  - #815 
  - #817 
  - #818


packages/std-client/package.json
```diff
@@ -42,6 +42,7 @@
     "@latticexyz/recs": "workspace:*",
     "@latticexyz/solecs": "workspace:*",
     "@latticexyz/store": "workspace:*",
+    "@latticexyz/store-cache": "workspace:^",
```
holic, line 45: what command are you using to add this? wondering if we should just make them all `^`

alvrs: this happens when i do `pnpm add @latticexyz/store-cache`

examples/minimal/packages/client-react/src/mud/setup.ts
```diff
@@ -7,17 +7,19 @@ import { world } from "./world";
 import { Contract, Signer, utils } from "ethers";
 import { JsonRpcProvider } from "@ethersproject/providers";
 import { IWorld__factory } from "contracts/types/ethers-contracts/factories/IWorld__factory";
+import storeConfig from "../../../contracts/mud.config";
```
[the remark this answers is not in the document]
holic: can you `import storeConfig` from `contracts/mud.config`?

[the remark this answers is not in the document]
alvrs: that worked!

examples/minimal/packages/contracts/mud.config.ts
```diff
@@ -1,6 +1,8 @@
-import { mudConfig, resolveTableId } from "@latticexyz/world/register";
+import { mudConfig } from "@latticexyz/world/register";
+import { resolveTableId } from "@latticexyz/config";
 
 export default mudConfig({
+  namespace: "",
```
holic, line 5: isn't this the default?

examples/minimal/packages/contracts/mud.config.ts
```diff
@@ -22,6 +23,14 @@ export default mudConfig({
       },
       ephemeral: true,
     },
+    Inventory: {
+      primaryKeys: {
+        user: "address",
+        item: "bytes32",
+        variant: "uint32",
```
holic, line 30: what is variant?

packages/network/src/v2/schemas/tableSchemas.ts
```diff
@@ -6,7 +6,7 @@ import { IStore } from "@latticexyz/store/types/ethers-contracts/IStore.sol/ISto
 
 // worldAddress:tableId => schema
 // TODO: add chain ID to namespace?
-const schemaCache: Partial<Record<`${string}:${string}`, Promise<TableSchema>>> = {};
+const schemaCache: Record<`${string}:${string}`, Promise<TableSchema>> = {};
```
holic, line 9: any reason this changed?

I tend to use `Partial<Record<...>>` for any object that I'm filling in over time, otherwise the types don't reflect that a key can sometimes not have a value

examples/minimal/packages/contracts/mud.config.ts
alvrs: just a random example of a `uint32` key tuple (though of a variant of some item, like "red cup" vs "green cup", lemme know if you can think of a better example haha)

packages/std-client/src/setup/utils.ts
```diff
@@ -155,6 +159,35 @@ export function applyNetworkUpdates<C extends Components>(
...
+          // Apply network updates to cache store
+          if (!tableConfig || namespace !== storeConfig.namespace) {
+            // console.warn("Ignoring table config outside own mud config", update, storeConfig.namespace);
+          } else {
+            // TODO: key names are not yet part of the on-chain table metadata, so we have to
+            // load them from the local mud.config (and have to ignore all tables that are not
+            // defined in the local mud config)
+            // (see https://github.com/latticexyz/mud/issues/824)
+
+            // `Object.getOwnPropertyNames` guarantees key order, `Object.keys` does not
```
holic, line 177: TIL

holic (dismissed):

examples/minimal/packages/contracts/mud.config.ts
holic: ahh, in that case, `itemVariant` might be a little more descriptive

examples/minimal/packages/contracts/mud.config.ts
alvrs: yes it is, good catch! added this in some intermediate version where importing `mudConfig` didn't work yet

packages/network/src/v2/schemas/tableSchemas.ts
alvrs: yep agree this should be `Partial`!

Issue #746, opened 2023-05-05
Proposal for strongly typed config and plugin system


alvrs:
This proposal is based on explorations for an extendable plugin system by @dk1a, see #679

## Terminology

- “Strongly typed”: In the context of Typescript we call a type “strong”, if it includes specific instances of primitive values instead of the general type. Example: a type `{ field1: "value", field2: 1 }` is strongly typed, `{ field1: string, field2: number }` is not.

## Motivation

- With a strongly typed config, we’re able to dynamically define types for `recs`, the upcoming composite datastore (#641), and more, without the need for code generation. A dynamic type definition is usually easier to maintain than logic for code generation, and errors can be caught more easily.

## Current Situation

- We rely on `zod` for config validation, which unfortunately does not support strong types (only weak primitive types like `string`, `number` etc are returned)
- Another requirement next to strong types is an extendable plugin system for the config, which got started in #679.
    - It is important that plugins are able to define their own config types, and expand/transform the existing config. This makes it much more complex to achieve strong config types.
    - Currently we solve the requirement of plugins extending the config types and transforming the existing config with an approach similar to Hardhat's, which involves overriding global type declarations and a global context, but unfortunately doesn’t support strong types (yet).

## Proposal

[...]

PR #809, opened 2023-05-13
fix: codesplit cli code to allow mud.config.ts to be imported in the browser

alvrs:
We need to be able to import `mud.config.ts` in browser code to infer types for various client libraries (eg `store-cache`).
This PR fixes a couple issues that prevented this:
* Importing `mudConfig` also imported code intended for node environments, which caused `vite` in consumer projects to error (example: 
image)
* `global` (used for mud plugins extending the config) was not defined in browser contexts: 
image


alvrs:
Unclear why tests are failing, all are passing locally and there were no changes to the store package that would influence the forge tests. Trying `foundryup` to see if it's caused by a new foundry version.

holic:
maybe time for? https://github.com/latticexyz/mud/issues/746

holic:
> Unclear why tests are failing, all are passing locally and there were no changes to the store package that would influence the forge tests. Trying `foundryup` to see if it's caused by a new foundry version.

Yeah, it might be this: https://github.com/foundry-rs/foundry/pull/4920

> The intended behavior is that it only works for the next immediate call. This new behavior only allows to:
>
> - Call the cheatcode AND fill the expected emit right after calling the cheatcode. Declaring the cheatcode N times and then filling N events will fail.
> - match events after the next CALL is triggered. Events not triggered by a call will be ignored. Events nested in the triggered CALL subtree will be matched. If the events defined are not matched, the execution will revert.

It's a little unclear by the above "nested in the CALL subtree" description if this should succeed or fail in our case, but I noticed that `StoreCore.pushToField` is doing more calls internally before emitting, so maybe it's not matching on the new criteria for `expectEmit`?

This feels like a bummer constraint if so, and wonder if there are ways around this or alternative vm helpers.

holic:
Another option is to use Vulcan test utils, which has some nicer ergonomics: https://github.com/nomoixyz/vulcan

```solidity
mc.doSomethingElse();

// Check event emissions
expect(address(mc).lastCall()).toHaveEmitted(
    "SomeEvent(address,bytes32,uint256)",
    [address(1).topic(), any()], // Event topics
    abi.encode(123) // Event data
);
```

(I also have an open issue and they have a pending PR for better gas reporting)

holic:
opened an issue for this since its unrelated to this PR: https://github.com/latticexyz/mud/issues/814

.github/workflows/templates.yml
```diff
@@ -35,5 +35,5 @@ jobs:
         uses: ./.github/actions/require-empty-diff
 
       - name: Run tests
-        working-directory: ./examples/${{ matrix.example }}
+        working-directory: ./templates/${{ matrix.template }}
```
holic, line 38: nice catch

packages/config/package.json
```diff
@@ -11,7 +11,8 @@
   "type": "module",
   "exports": {
     ".": "./dist/library/index.js",
-    "./register": "./dist/register/index.js"
+    "./register": "./dist/register/index.js",
+    "./cli": "./dist/cli/index.js"
```
holic, line 15: should this be `/node`?

holic (approved):
Looks good! Only question is if we should call this export `node` rather than `cli`, if node is the intended environment. `cli` sounds too use case specific and very internally oriented, and if that's what we're aiming for, maybe we name it `cli-for-internal-use-only` haha

Separately wondering if we should just bite the bullet on #746 (maybe not before the hackathon but quickly following?)

holic (approved):
Looks good! Only question is if we should call this export `node` rather than `cli`, if node is the intended environment. `cli` sounds too use case specific and very internally oriented, and if that's what we're aiming for, maybe we name it `cli-for-internal-use-only` haha

Separately wondering if we should just bite the bullet on #746 (maybe not before the hackathon but quickly following?)

holic (dismissed):
Looks good! Only question is if we should call this export `node` rather than `cli`, if node is the intended environment. `cli` sounds too use case specific and very internally oriented, and if that's what we're aiming for, maybe we name it `cli-for-internal-use-only` haha

Separately wondering if we should just bite the bullet on #746 (maybe not before the hackathon but quickly following?)

packages/config/package.json
alvrs: yeah I was torn between the two, but now agree `node` fits better - changed it for both config and world exports

alvrs:
> Separately wondering if we should just bite the bullet on https://github.com/latticexyz/mud/issues/746 (maybe not before the hackathon but quickly following?)

Unfortunately I think it's too close to the hackathon now to refactor this, but agree with shortly after!

dk1a (approved):
I like `node`

holic (approved):

PR #810, opened 2023-05-13
fix(store-cache): deserialise bigint keys when turning them back from tuples to records

alvrs:
We're serialising `bigint` keys since `tuple-database` does not support them in tuples yet (see https://github.com/ccorcos/tuple-database/issues/25). 

This PR adds deserialising of those serialised bigint keys so that the key includes the proper bigint value when subscribing to table updates.

### DIFF
--- a/packages/store-cache/src/createDatabaseClient.test.ts
+++ b/packages/store-cache/src/createDatabaseClient.test.ts
@@ -11,6 +11,7 @@ const config = mudConfig({
     MultiKey: { primaryKeys: { first: "bytes32", second: "uint32" }, schema: "int32" },
     EnumTable: { primaryKeys: { first: "Enum1" }, schema: "Enum2" },
     MultiTable: { schema: { arr: "int32[]", str: "string", bts: "bytes" } },
+    BigInt: { primaryKeys: { first: "uint256" }, schema: "uint256" },
   },
   enums: {
     Enum1: ["A1", "A2"],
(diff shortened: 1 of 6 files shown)


packages/store-cache/src/createDatabaseClient.test.ts
```diff
@@ -361,4 +362,30 @@ describe("createDatabaseClient", () => {
     expect(mock).toHaveBeenNthCalledWith(1, [{ set: [{ key, value }], remove: [], table, namespace }]);
     expect(mock).toHaveBeenNthCalledWith(2, [{ set: [], remove: [{ key }], table, namespace }]);
   });
+
+  // Requires bigint keys to be supported by `tuple-database`, see https://github.com/ccorcos/tuple-database/issues/2
+  // For now we serialize bigints to use them as keys, which changes the comparison (eg. 2n < 10n but "2n" > "10n")
+  it.skip("should be possible to filter based on bigint keys", () => {
```
alvrs, line 368: this test is skipped for now because this doesn't work yet without support for bigints in `tuple-database` (it is possible to use the bigint values as filter, but the comparison is based on strings rather than bigints)

packages/store-cache/src/utils.ts
```diff
@@ -201,18 +201,34 @@ function tupleToRecord(tuple: Tuple): Record<string, any> {
...
 }
 
 /**
  * Helper to serialize values that are not natively supported in keys by tuple-database.
+ * (see https://github.com/ccorcos/tuple-database/issues/25)
  * For now only `bigint` needs serialization.
  */
 function serializeKey(key: unknown) {
-  if (typeof key === "bigint") return String(key);
+  if (typeof key === "bigint") return `${key.toString()}n`;
```
holic, line 216: do you know if there are any pending specs for serializing bigints? thinking about JSON etc. where it may not have landed but maybe coming?

ultimately wondering if the `n` suffix is aligned with whatever the JS ecosystem is thinking as the standard for encoding/serializing these kinds of values (vs a plain number string, but just much larger than what JS can support in terms of a number)

and brings up another question in my mind around caching - do we have or will we have a "version" stamp for a given store cache so that, if we need to change the internal serialization, we can bump the version to "invalidate" the cache and get clients to rebuild from scratch with the new serialization method?

packages/store-cache/src/createDatabaseClient.test.ts
holic: can we add a TODO and/or issue to remind us to come back to this?

holic (approved):

packages/store-cache/src/utils.ts
alvrs: > do you know if there are any pending specs for serializing bigints? thinking about JSON etc. where it may not have landed but maybe coming?

From https://github.com/GoogleChromeLabs/jsbi/issues/30: 
> Working as intended/spec'ed: https://tc39.es/proposal-bigint/#sec-serializejsonproperty (step 10).
>
> In short, the background is that JSON is so commonly used among all sorts of different systems and programming languages that it is effectively impossible to add anything to the format without breaking compatibility in some crucial case.

So it seems like won't be support for `bigint` in JSON and it doesn't seem like there is a standard in progress.

> and brings up another question in my mind around caching - do we have or will we have a "version" stamp for a given store cache so that, if we need to change the internal serialization, we can bump the version to "invalidate" the cache and get clients to rebuild from scratch with the new serialization method?

Agree with versioning! In the current version `store-cache` doesn't persist the data anywhere yet, but that's definitely planned. I'd probably add the versioning in the PR where we add functionality to persist the cache so it can be informed by the needs we have there.

holic (approved):

PR #811, opened 2023-05-13
feat(cli): wait for deploy transactions to be included before executing post deploy script

alvrs:
Context:
* in #606 we started waiting for transactions to be included before executing the next batch of transactions to resolve some race condition issues on non-local chains. This made the deploy script much slower, also during local development where anvil would be fast enough to handle the deploys script without waiting for txs to be included.
* in #730 we added the default behavior for also waiting for contract deployment (before we were just waiting for inclusion of refular txs), and also added the `disableTxWait` cli flag to skip waiting for transactions (to speed up deployment on local chains where we don't need to wait)

As it turns out in some cases there are issues when the `PostDeploy` script is executed before all deploy transactions are included, because the `PostDeploy` script's nonce is managed by forge and not the deploy script.

This PR adds a check before executing the `PostDeploy` script to wait for all previous transactions to be included (by waiting for the remote nonce to be equal to the local nonce).

With this change it is possible to add `--disableTxWait` to the `dev` commands of our templates/projects to speed up the deployment to local development chains.

### DIFF
--- a/packages/cli/src/utils/deploy.ts
+++ b/packages/cli/src/utils/deploy.ts
@@ -27,7 +27,7 @@ export interface DeployConfig {
   debug?: boolean;
   worldAddress?: string;
   disableTxWait: boolean;
-  pollInterval: 1000;
+  pollInterval: number;
 }
 
 export interface DeploymentInfo {
--- a/packages/cli/src/utils/deploy.ts
+++ b/packages/cli/src/utils/deploy.ts
@@ -27,6 +27,7 @@ export interface DeployConfig {
   debug?: boolean;
   worldAddress?: string;
   disableTxWait: boolean;
+  pollInterval: 1000;
 }
 
 export interface DeploymentInfo {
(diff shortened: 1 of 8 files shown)


packages/cli/src/utils/deploy.ts
```diff
@@ -301,6 +301,19 @@ export async function deploy(
...
+  let remoteNonce = await signer.getTransactionCount();
+  while (remoteNonce !== nonce) {
+    console.log(
+      chalk.gray(
+        `Waiting for transactions to be included before executing ${postDeployScript} (local nonce: ${nonce}, remote nonce: ${remoteNonce})`
+      )
+    );
+    await new Promise((resolve) => setTimeout(resolve, 1000));
+    remoteNonce = await signer.getTransactionCount();
+  }
```
holic, line 315: should we have a timeout to fail after a period of time?

packages/cli/src/utils/deploy.ts
```diff
@@ -301,6 +301,19 @@ export async function deploy(
...
+
+  // Confirm the current nonce is the expected nonce to make sure all transactions have been included
+  let remoteNonce = await signer.getTransactionCount();
+  while (remoteNonce !== nonce) {
+    console.log(
+      chalk.gray(
+        `Waiting for transactions to be included before executing ${postDeployScript} (local nonce: ${nonce}, remote nonce: ${remoteNonce})`
+      )
+    );
+    await new Promise((resolve) => setTimeout(resolve, 1000));
```
holic, line 313: should this poll interval be configurable?

holic (approved):

holic (approved):

holic:
(sorry about these double approvals, airplane wifi is finicky)

holic (approved):

PR #812, opened 2023-05-14
fix(common): publish source files for types

holic:
I noticed when creating a new project from a template using `pnpm create mud`, I wasn't getting a strongly typed config:

image

When I dug into the `mudConfig` reference, I noticed there was a broken/missing TS definition for our common package:

image

Building this package locally shows it's missing `src` files that we usually have in our packages and expect with our `typesVersions` definition:

image

I found that this is because the `files` definition here is taking precedence over `.npmignore` in `npm pack` and thus excluding `src`. By removing `files` definition and packing, I can see `src` files in there, so this should fix it.

### DIFF
--- a/packages/common/package.json
+++ b/packages/common/package.json
@@ -35,9 +35,6 @@
       ]
     }
   },
-  "files": [
-    "dist"
-  ],
   "scripts": {
     "build": "pnpm run build:js",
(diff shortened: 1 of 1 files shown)


holic:
merging to unblock myself on emojimon work while I fly home

PR #815, opened 2023-05-14
test(store): fix tests after foundry update, move gas reports to separate tests

### LINKED ISSUE #814: figure out failing emits with latest foundry
main branch's tests started failing. I think it might be this: https://github.com/foundry-rs/foundry/pull/4920

> The intended behavior is that it only works for the next immediate call. This new behavior only allows to:
>
> - Call the cheatcode AND fill the expected emit right after calling the cheatcode. Declaring the cheatcode N times and then filling N events will fail.
> - match events after the next CALL is triggered. Events not triggered by a call will be ignored. Events nested in the triggered CALL subtree will be matched. If the events defined are not matched, the execution will revert.

It's a little unclear by the above "nested in the CALL subtree" description if this should succeed or fail in our case, but I noticed that `StoreCore.pushToField` is doing more calls internally before emitting, so maybe it's not matching on the new criteria for `expectEmit`?

This feels like a bummer constraint if so, and wonder if there are ways around this or alternative vm helpers.

I asked about this in the Foundry Support Telegram group. If there's not a quick/clear way to test for emits in the same way with the latest foundry version, we may need to pin our version for now.

Another option is to use Vulcan test utils, which has some nicer ergonomics: https://github.com/nomoixyz/vulcan

```solidity
mc.doSomethingElse();

// Check event emissions
expect(address(mc).lastCall()).toHaveEmitted(
    "SomeEvent(address,bytes32,uint256)",
    [address(1).topic(), any()], // Event topics
    abi.encode(123) // Event data
);
```

(I also have an open issue and they have a pending PR for better gas reporting)

alvrs:
Fixes #814 

* Integrates the new `expectEmit` behavior introduced in https://github.com/foundry-rs/foundry/pull/4920 by calling functions on a wrapper contract (`IStore(this)`) instead of the library directly (`StoreCore`) to trigger a `CALL`.

* Moves the gas reports to a separate test which still uses `StoreCore`, since calling the methods on a wrapper contract increases gas cost (but for the Store library gas reports we're interested in the library cost, not the contract call cost)

### DIFF
--- a/packages/store/gas-report.json
+++ b/packages/store/gas-report.json
@@ -276,264 +276,6 @@
     "gasUsed": 1108
   },
   {
-    "source": "test/StoreCore.t.sol",
-    "name": "access non-existing record",
-    "functionCall": "bytes memory data1 = StoreCore.getRecord(table, key)",
-    "gasUsed": 7316
-  },
-  {
-    "source": "test/StoreCore.t.sol",
-    "name": "access static field of non-existing record",
-    "functionCall": "bytes memory data2 = StoreCore.getField(table, key, 0)",
-    "gasUsed": 2988
-  },
-  {
-    "source": "test/StoreCore.t.sol",
-    "name": "access dynamic field of non-existing record",
-    "functionCall": "bytes memory data3 = StoreCore.getField(table, key, 1)",
-    "gasUsed": 3599
-  },
-  {
-    "source": "test/StoreCore.t.sol",
-    "name": "access length of dynamic field of non-existing record",
-    "functionCall": "uint256 data3Length = StoreCore.getFieldLength(table, key, 1, schema)",
-    "gasUsed": 1342
-  },
-  {
-    "source": "test/StoreCore.t.sol",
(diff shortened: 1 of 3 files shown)


packages/store/test/StoreCore.t.sol
```diff
@@ -80,21 +41,18 @@ contract StoreCoreTest is Test, StoreReadWithStubs {
     vm.expectEmit(true, true, true, true);
     emit StoreSetRecord(StoreCoreInternal.SCHEMA_TABLE, key, abi.encodePacked(schema.unwrap(), keySchema.unwrap()));
 
-    // !gasreport StoreCore: register schema
-    StoreCore.registerSchema(table, schema, keySchema);
+    IStore(this).registerSchema(table, schema, keySchema);
```
alvrs, line 44: We don't have to cast to `IStore` here since `this` already extends the `StoreMock` contract, but I figured it makes it a bit more explicit to see what's going on here (and we don't care about gas cost in here because gas reports are moved to `StoreCoreGas`)

holic (approved):

PR #817, opened 2023-05-14
fix(store): encode `address` like `abi.encode` does (left padded)

alvrs:


dk1a (approved):
Good catch! Right-padding was not intentional

holic (dismissed):
does this fix the entity ID thing that @Kooshaba was running into?

PR #828, opened 2023-05-15
docs: add hackathon idea pages and splash plage

biscaryn:
This PR adds two things to the docs page for the ETHGlobal hackathon:
- Adds hackathon ideas as .mdx pages
- Adds a AW Hackathon splash/pop-up page that only shows up the first time someone lands on the website. It contains a sign-up link to the hackathon and information about the hackathon ideas page (with a link)


docs/components/ComponentsWrapper.tsx
```diff
@@ -0,0 +1,26 @@
+import { useTheme } from "nextra-theme-docs";
+import Logo from "./Logo";
+import Splash from "./Splash";
+import WarningBox from "./WarningBox";
+
+export default function ComponentsWrapper() {
+  // This component wraps the Splash and Logo components, to make sure that they're only loaded once light/dark mode has been confirmed.
+  // This makes sure that the correct MUD logo (light/dark-mode) is displayed
```
holic, line 8: 🙏

docs/components/ComponentsWrapper.tsx
```diff
@@ -0,0 +1,26 @@
+import { useTheme } from "nextra-theme-docs";
+import Logo from "./Logo";
+import Splash from "./Splash";
+import WarningBox from "./WarningBox";
+
+export default function ComponentsWrapper() {
```
holic, line 6: is there a better name for this? `IndexIntro` or something

docs/components/Splash.tsx
```diff
@@ -0,0 +1,70 @@
+import React, { useState, useEffect } from "react";
+import styles from "./Splash.module.css";
+
+const Splash = ({ theme }) => {
+  const [isSplashOpen, setSplashOpen] = useState(false);
+  const [display, setDisplay] = useState(false);
```
holic, line 6: not blocking since we're in a rush prepping for eth global, but as a general note, we should lean on things like [Radix's dialog](https://www.radix-ui.com/docs/primitives/components/dialog) for accessibility, otherwise we're missing out on support for things like keyboard/tab navigation, screen readers, etc.

docs/components/Splash.tsx
```diff
@@ -0,0 +1,70 @@
...
+  };
+
+  if (!isSplashOpen) return null;
+
+  return (
+    display && (
+      <div className={styles.splashOverlay} onClick={closeSplash}>
+        <div className={styles.splashContent} onClick={stopPropagation}>
+          <button className={styles.splashClose} onClick={closeSplash}>
+            X
```
holic, line 42: tiny thing but I find `&times;` to be slightly better for rendering close icons

holic (approved):

docs/components/ComponentsWrapper.tsx
biscaryn: Done, updated to IndexIntro

ludns:
Open in a new tab instead of same tab? Usually a good default for links on a different domain.

docs/components/Splash.tsx
biscaryn: Done, thanks for the suggestion!

docs/components/Splash.tsx
biscaryn: Noted, but this should all get deleted post-hackathon so should be fine for now

holic (approved):

PR #837, opened 2023-05-16
docs: update imports, remove name from example configs, fix hooks link

yonadaa:


docs/pages/world/modules.mdx
```diff
@@ -55,7 +55,7 @@ keyTuple[0] = "bob";
 bool memory isPlayer = hasKey(world, PlayerTableId, keyTuple);
 ```
 
-Internally, it works by installing a [hook](http://localhost:3000/store/advanced-features#storage-hooks) that maintains:
+Internally, it works by installing a [hook](https://v2.mud.dev/store/advanced-features#storage-hooks) that maintains:
```
ludns, line 58: Can you use a relative path instead? It will work both on localhost and when published:
`/store/advanced-features#storage-hooks` will work

yonadaaa: ooh, duh. thanks - means we don't point to `v2.` either.

yonadaaa:
Marking these fixes as Ready now - writing the new querying page will take a little longer.

docs/pages/world/modules.mdx
```diff
@@ -83,7 +83,7 @@ import { Owners, OwnersId } from "./tables/Owners.sol";
 bytes32[] memory keysWithValue = getKeysWithValue(world, OwnersId, Owners.encode(address(42)));
```
ludns, line 83: Imports pls, where do folks get `getKeysWithValue` from

docs/pages/world/modules.mdx
```diff
@@ -55,7 +55,7 @@ keyTuple[0] = "bob";
 bool memory isPlayer = hasKey(world, PlayerTableId, keyTuple);
```
ludns, line 55: Import

ludns: also for the `getKeysInTable` above

docs/pages/world/modules.mdx
```diff
@@ -132,13 +134,11 @@ import { mudConfig } from "@latticexyz/config";
 export default mudConfig({
   tables: {
     MyTable: {
-      name: "mytable",
       schema: "uint32",
     },
   },
   modules: [
     {
-      name: "KeysWithValueModule",
```
ludns, line 141: Modules still have names.
Pls typecheck stuff in your editor before changing the doc. We need to be careful to not put stuff that doesn't work.

ludns (changes requested):

docs/pages/world/modules.mdx
yonadaaa: Typechecking in editor is a good idea thanks. The imports for the config were actually wrong.

ludns (dismissed):

alvrs (approved):

PR #840, opened 2023-05-16
docs: add querying tutorial page

yonadaa:


docs/pages/world/querying.mdx
```diff
@@ -0,0 +1,90 @@
+## Querying
+
+The `query` module allows Systems to ask questions about keys in the World. These queries are evaluated left to right and can include fragments like "Give me all keys in the `CanFly` table" and "Then keep keys in the Position table with value `{x: 1, y: 43}`".
+
+`query` requires the [`KeysInTable`](/world/modules#keysintablemodule) and [`KeysWithValue`](/world/modules#keyswithvaluemodule) modules to be installed.
```
ludns, line 5: On all tables that are part of query fragments?
Installed is a bit unclear here given modules can be installed multiple time

yonadaaa: Yes, I wasn't sure what to expose here, as the internals might change. I will clarify it.

Actually, you only need the module on the tables you use the fragment with. For example, you only need to install `KeysInTable` on tables you will use `Has` with.

PR #841, opened 2023-05-16
feat(store-cache): add `scan` to get all values of a table/database

alvrs:


holic (approved):
not blocking, but do we want any tests for scan filters?

alvrs:
> not blocking, but do we want any tests for scan filters?

There is one in the `utils.test.ts`:
https://github.com/latticexyz/mud/pull/841/files#diff-f5a91d2797891954fe2a6a87ce97e6532b3628472203f8ad1b9e5c89bb7a0520R79-R87

but tbh `tuple-database`'s approach to filtering is kind of cursed

holic:
> but tbh `tuple-database`'s approach to filtering is kind of cursed

say more!

PR #847, opened 2023-05-16
feat(react): add `useRows`, `useRow` hooks for `store-cache`

alvrs:
* replaces jest with vite in react package
* adds two hooks: `useRows`, `useRow` to interact with `storeCache`


packages/react/src/store-cache/useRow.test.ts
```diff
@@ -0,0 +1,122 @@
...
+
+    // Update the position table
+    act(() => {
+      client.tables.Position.remove({ key: "0x00" });
+    });
+    expect(result.all.length).toBe(5);
+    expect(result.current).toEqual(undefined);
+  });
+
+  it.todo("should re-render when the filter changes");
```
alvrs, line 121: @holic somehow the tests get stuck if I add `filter` to the hook's dependency array. Any idea why? Right now the hook just doesn't change if `filter` changes.

ludns:
can you update my readme for the store cache and add these? 
alternatively can rebase this on the store-cache pr

packages/react/src/store-cache/useRows.ts
```diff
@@ -0,0 +1,27 @@
...
+    setRows(storeCache.scan(filter));
+
+    const unsubscribe = storeCache.subscribe(() => {
+      // very naive implementation for now, but easier and probably more efficient than
+      // manually looping through the `rows` array for every update event
+      setRows(storeCache.scan(filter));
+    }, filter);
+
+    return unsubscribe;
+  }, []);
```
holic, line 24: This empty dependency array is the reason things aren't updating when args change.

See `useEntityQuery` for how I work around this to create a stable reference to an arbitrary object:
https://github.com/latticexyz/mud/blob/91b37a4414fb9730be1886bfbdf4352c37b70d1a/packages/react/src/useEntityQuery.ts#L21

So we'd prob want something like
```ts
const stableStoreCache = useMemo(storeCache); // not deep memo since it should be roughly stable anyway and unlikely inner values would change without an entirely different client
const stableFilter = useDeepMemo(filter); // deep memo since this is more like an argument/set of options

useEffect(() => {
  setRows(stableStoreCache.scan(stableFilter));

  const unsubscribe = stableStoreCache.subscribe(() => {
    // very naive implementation for now, but easier and probably more efficient than
    // manually looping through the `rows` array for every update event
    setRows(stableStoreCache.scan(stableFilter));
  }, stableFilter);

  return unsubscribe;
}, [stableStoreCache, stableFilter]);
```
(I don't love the "stable" term here but also don't love "memoized")

packages/react/src/store-cache/useRow.test.ts
holic: added some details below!

holic:
🎉

PR #848, opened 2023-05-16
feat(world): SnapSync config plugin

Kooshaba:


packages/world/src/modules/snapsync/ts/typeExtensions.ts
```diff
@@ -0,0 +1,13 @@
...
+
+interface SnapSyncConfig {
+  snapSync: boolean;
+}
+
+declare module "@latticexyz/config" {
+  // eslint-disable-next-line @typescript-eslint/no-empty-interface
+  export interface MUDCoreUserConfig extends WorldUserConfig, SnapSyncConfig {}
+  // eslint-disable-next-line @typescript-eslint/no-empty-interface
+  export interface MUDCoreConfig extends WorldConfig, SnapSyncConfig {}
```
dk1a, line 12: I don't think you need to repeat `extends WorldUserConfig` and `WorldConfig`, `SnapSyncConfig` should be enough

examples/minimal/packages/contracts/mud.config.ts
```diff
@@ -1,7 +1,8 @@
-import { mudConfig } from "@latticexyz/world/register";
+import { mudConfig } from "@latticexyz/world/snapSync";
 import { resolveTableId } from "@latticexyz/config";
 
 export default mudConfig({
+  snapSync: true,
```
holic, line 5: just as a general note, I think this is the first time the mud config has had an option that could effect network code

since we're passing in the mud config to the network config, maybe should just use this key to determine if snap sync should be used?

examples/minimal/packages/contracts/mud.config.ts
```diff
@@ -1,7 +1,8 @@
-import { mudConfig } from "@latticexyz/world/register";
+import { mudConfig } from "@latticexyz/world/snapSync";
```
holic, line 1: this approach looks like it wouldn't allow for composability of plugins

@dk1a is this how we're expected to create/use plugins?

packages/world/src/modules/snapsync/ts/index.ts
```diff
@@ -0,0 +1,6 @@
+import "@latticexyz/store/register";
```
dk1a, line 1: should the plugin code be in `src/ts/plugins`, or https://github.com/latticexyz/mud/pull/584? 
modules are more like things that can be created/used by plugins

alvrs: for now probably in `src/ts/plugins`, and later in its own package (or 584)

examples/minimal/packages/contracts/mud.config.ts
dk1a: this should work well, and it's the approach I use for `world` the plugin. It just means this plugin can't be used without `store` and `world`, and implicitly registers those so you can skip em (or you can still import em).
Here I guess it just looks a bit weird/misleading

dk1a: maybe keep `import { mudConfig } from "@latticexyz/world/register";` and have snapSync be 2nd?

packages/world/src/modules/snapsync/ts/index.ts
Kooshaba: is there another plugins dir i don't know about or do you just mean one here?

dk1a: currently there is no plugins dir at all, I just meant to add one in `packages/world/src/ts/plugins` (temporarily, until we separate it as its own plugin)

packages/world/package.json
```diff
@@ -27,6 +28,9 @@
       ],
       "node": [
         "./ts/node/index.ts"
+      ],
+      "snapSync": [
+        "./src/ts/plugins/snapsync/index.ts"
```
alvrs, line 33:
```suggestion
        "./ts/plugins/snapsync/index.ts"
```

PR #853, opened 2023-05-16
fix(config): only normalize to file url for node imports

alvrs:
* follow up to #849: turns out `esbuild` does't like a `file:///` url for entry points; this PR changes the `resolveConfigPath` function to only add the `file:///` prefix if needed (for node.js esm imports on windows)

### DIFF
--- a/packages/config/src/node/loadConfig.ts
+++ b/packages/config/src/node/loadConfig.ts
@@ -14,14 +14,14 @@ export async function loadConfig(configPath?: string): Promise<unknown> {
   configPath = await resolveConfigPath(configPath);
   try {
     await esbuild.build({ entryPoints: [configPath], format: "esm", outfile: TEMP_CONFIG });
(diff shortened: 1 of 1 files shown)


packages/config/src/node/loadConfig.ts
```diff
@@ -14,14 +14,14 @@ export async function loadConfig(configPath?: string): Promise<unknown> {
   configPath = await resolveConfigPath(configPath);
   try {
     await esbuild.build({ entryPoints: [configPath], format: "esm", outfile: TEMP_CONFIG });
-    configPath = await resolveConfigPath(TEMP_CONFIG);
+    configPath = await resolveConfigPath(TEMP_CONFIG, true);
```
alvrs, line 17: we were adding `file:///` also in line 14, so `esbuild` in line 16 complained

alvrs:
self approving to unblock myself and since this is only impacting windows environments (which are broken without this)

PR #854, opened 2023-05-16
fix(store): replace windows path separator with posix path separator

alvrs:
followup to #851 - unfortunately I added the regex at the wrong place


alvrs:
self-approving to unblock myself and this change is not impacting any other platform than windows (which is broken without this change)

PR #865, opened 2023-05-17
feat(cli): add dev-contracts

alvrs:
* Adds a `dev-contracts` script to the CLI to:
- start a dev node in the background (or use one provided via the `--rpc` arg)
- watch config and file changes, run all necessary codegen steps and redeploy contracts when stuff changes


docs/pages/quick-start.mdx
```diff
@@ -24,13 +24,9 @@ You'll be prompted to pick the type of template to use (vanilla JS or React).
...
 
-Jump into your new project's directory. In one terminal, run the anvil node with:
+Jump into your new project's directory.
 
-```
-pnpm run dev:node
-```
-
-And in another terminal, run the app with:
+Run the development server with:
```
holic, line 29: Can't do a suggestion over these lines because of the deletion, but this can probably be one line

> Jump into your new project's directory and run the app with:

examples/minimal/package.json
```diff
@@ -3,19 +3,17 @@
   "private": true,
   "scripts": {
     "build": "pnpm recursive run build",
-    "dev": "run-pty % pnpm dev:node % pnpm dev:client-react % pnpm dev:contracts",
+    "dev": "run-pty % pnpm dev:client-react % pnpm dev:contracts",
```
holic, line 6: `run-pty` is still here but removed from deps below?

templates/react/packages/client/mud/contractComponents.ts
```diff
@@ -0,0 +1,24 @@
+/* Autogenerated file. Do not edit manually. */
+
+import { TableId } from "@latticexyz/utils";
+import { defineComponent, Type as RecsType, World } from "@latticexyz/recs";
+
+export function defineContractComponents(world: World) {
```
holic, line 6: weird that this was missing before and not caught by CI

templates/react/package.json
```diff
@@ -3,10 +3,9 @@
   "private": true,
   "scripts": {
     "build": "pnpm recursive run build",
-    "dev": "pnpm run initialize && pnpm recursive run dev",
+    "dev": "concurrently \"cd packages/contracts && pnpm run dev\" \"cd packages/client && pnpm run dev\"",
```
holic, line 6: this is to get around the `pnpm --filter` issues on windows right?

holic: also single quotes in these might be a little easier to read rather than escaped quotes?

examples/minimal/package.json
alvrs: good catch! fixed

holic (approved):

templates/react/package.json
alvrs: and `pnpm run recursive` doesn't like non-exiting dev scripts

holic:
does this impact our need for `initialize`? been kinda unhappy with that pattern and I hope its replaced by all of this

alvrs:
> does this impact our need for initialize? been kinda unhappy with that pattern and I hope its replaced by all of this
> there is still some bare-minimum to do before the client starts up so we don't get client import errors (e.g. world types)

Yep, at least for `dev` there is no more need for `initialize`! Still kept the script around because it's still needed for `deploy` for now if you didn't run `dev` before

holic:
one thing we could do in the future is split `dev-contracts` with a 1) deploy prerequisites and 2) deploy, then `pnpm dev` runs prerequisites then runs client + contracts dev (but with a flag in contracts dev to skip prereqs)

ludns:
can you add docs to the cli page? with the new command + args

alvrs:
> actually, on the first ever run of dev it might be that the client doesn't have the world types yet. Wonder how we can get around that without having to run initialize before every start...

Fixed this by running the codegen step once before anvil starts, and making the clients wait for anvil before starting their dev server

templates/react/package.json
holic: can use `pnpm recursive --no-bail`

docs/pages/cli.mdx
```diff
@@ -65,7 +94,7 @@ This command also wipes the Anvil cache. Anvil cache blow-up problems won’t ha
 Connects to a MUD faucet service to fund an address.
 
 ```bash
-mud faucet --faucetUrl <faucetService> --adress <address>
+pnpm mud faucet --faucetUrl <faucetService> --adress <address>
```
holic, line 97:
```suggestion
pnpm mud faucet --faucetUrl <faucetService> --address <address>
```

examples/minimal/packages/client-react/package.json
```diff
@@ -6,7 +6,7 @@
   "type": "module",
   "scripts": {
     "build": "vite build",
-    "dev": "vite",
+    "dev": "wait-port -t 10000 localhost:8545 && vite",
```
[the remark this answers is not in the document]
alvrs: yeah probably, what would be a good place for this? Also added a 10s timeout just in case

holic (approved):

docs/pages/cli.mdx
```diff
@@ -32,11 +33,39 @@ The valid templates are:
...
+
+To run the dev server with a custom development node:
+
+```bash
+# Terminal 1:
+anvil --port 8545
+
+# Terminal 2:
+pnpm mud dev-contracts --tsgenOutput ../client/src/mud --rpc localhost:8545
+```
```
holic, line 62: Maybe a note at the end of this section about the anvil port needing to be adjusted in client if customized?

examples/minimal/packages/client-react/package.json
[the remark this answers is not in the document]
holic: ideally we can just detect the port via foundry config or env var or whatever but that seems like a lot more work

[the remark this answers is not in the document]
holic: another option/idea: write out the anvil host+port to a hidden/temp file at the start of dev-contracts, load from that file to wait on the proper host+port

docs/pages/cli.mdx
alvrs: this doc is technically unrelated to the templates, wonder if it would be best to couple it with the templates? But also fine to mention it here too

examples/minimal/packages/client-react/package.json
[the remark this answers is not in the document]
alvrs: let's add an issue to track that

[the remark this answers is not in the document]
alvrs: https://github.com/latticexyz/mud/issues/866

templates/react/packages/client/mud/contractComponents.ts
holic: ahh, because it wasn't supposed to be here: https://github.com/latticexyz/mud/pull/877

PR #869, opened 2023-05-17
feat(create-mud,common): separate boilerplate from user code, move mud chain to common

holic:
Currently the templates include a bunch of MUD setup boilerplate where user code is buried inside. Not a great user experience and hard to communicate in tutorials.

This does a quick pass to create stronger boundaries between code we expect users to edit vs code that is boilerplate and should probably be moved into mud packages (but prob not before hackathon).


packages/common/src/chains/index.ts
```diff
@@ -0,0 +1,5 @@
+// re-export wagmi chains to make them easier to reach
+export * from "@wagmi/chains";
```
holic, line 2: in the future, we might consider either not doing this (users include `@wagmi/chains` in their own deps) or make it a peer dep so that users can upgrade the chains version to get new chains without us having to constantly chase version updates here

holic: nevermind, gonna just remove the re-export now

alvrs (approved):

alvrs (dismissed):

PR #876, opened 2023-05-17
fix(dev-tools): simulate result at last block

holic:
By using the block number of the confirmed tx, a simulation for `increment` would show the value incrementing again (one more than what it should be) because the simulation happens at the end of the block (after the tx we're simulating)

image
image


holic:
landing this so I can make sure it's included in emojimon

PR #884, opened 2023-05-17
fix(network): connect to mud chains before public chains

holic:
Fixes the following error

image
image

### DIFF
--- a/packages/network/src/createNetwork.ts
+++ b/packages/network/src/createNetwork.ts
@@ -10,7 +10,8 @@ import { Signer, Wallet } from "ethers";
 import { computedToStream } from "@latticexyz/utils";
 import { privateKeyToAccount } from "viem/accounts";
 import { Address, fallback, webSocket, http, createPublicClient, createWalletClient } from "viem";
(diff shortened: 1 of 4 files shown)


packages/network/src/createNetwork.ts
```diff
@@ -10,7 +10,8 @@ import { Signer, Wallet } from "ethers";
 import { computedToStream } from "@latticexyz/utils";
 import { privateKeyToAccount } from "viem/accounts";
 import { Address, fallback, webSocket, http, createPublicClient, createWalletClient } from "viem";
-import * as chains from "viem/chains";
```
alvrs, line 13: why wagmi chains instead of viem chains?

alvrs (approved):

packages/network/src/createNetwork.ts
holic: viem chains are pinned to a specific version (I thought it was a peer dep allowing any version), wagmi chains at least aligns with everything else using them

alvrs (approved):

PR #885, opened 2023-05-17
feat(solhint-plugin-mud): create package

### LINKED ISSUE #711: [CLI] Linter to prevent use of `msg.sender` in Systems (and other antipatterns)
V1: a script that can be run in the console and is run before deployment
V2: vscode extension

dk1a:
fixes #711

seems to work well via cli (like via `solhint` in `package.json` of `examples`)

caveats:
- does not work in vscode
- has to be cjs-only, but it's fine because it should be imported only by solhint
- solhint and its plugin examples are all in js, I may have made the first plugin with typescript. It is a bit incomplete (some `any` because solhint doesn't export any types) and relies on `@solidity-parser/parser` for some types (which solhint uses internally)

To make it work in https://github.com/juanfranblanco/vscode-solidity these 2 issues would need to be resolved (stale, but I could probably do this myself if we wanna prioritize this):
https://github.com/protofire/solhint/issues/206
https://github.com/juanfranblanco/vscode-solidity/issues/169

I've also tried hardhat extension but immediately got remapping problems before even getting to solhint plugins, but maybe worth exploring alternatives

Haven't looked into solium yet, that may also be worth exploring


examples/minimal/packages/contracts/package.json
```diff
@@ -40,7 +40,8 @@
     "prettier-plugin-solidity": "^1.0.0-beta.19",
     "rimraf": "^3.0.2",
     "run-pty": "^4.0.3",
-    "solhint": "^3.3.7",
+    "solhint": "^3.4.1",
+    "solhint-plugin-mud": "file:../../../../packages/solhint-plugin-mud",
```
holic, line 44: this should prob use `link` rather than `file`?

examples/minimal/packages/contracts/.solhint.json
```diff
@@ -1,11 +1,13 @@
...
+  "plugins": ["mud"],
   "rules": {
     "compiler-version": ["error", ">=0.8.0"],
     "avoid-low-level-calls": "off",
     "no-inline-assembly": "off",
     "func-visibility": ["warn", { "ignoreConstructors": true }],
     "no-empty-blocks": "off",
-    "no-complex-fallback": "off"
+    "no-complex-fallback": "off",
+    "mud/no-msg-sender": "error"
```
holic, line 11: any way to make this the default if you enable/use the plugin?

holic:
> these 2 issues would need to be resolved (stale, but I could probably do this myself if we wanna prioritize this)

Probably worth figuring this out? cc @alvrs @ludns 

@juanfranblanco was quite responsive with the foundry monorepo support once I got his attention :)

examples/minimal/packages/contracts/package.json
dk1a: Oh I was thinking I wanted a comment there, but afaik I can't, right?
This is intentional, how pnpm does links conflicts with how solhint requires plugins apparently (my guess is something about flat node_modules, took a while to narrow this down)

examples/minimal/packages/contracts/.solhint.json
dk1a: a mud extension with defaults, that would be another package in a followup PR

dk1a:
regarding the issues - from what I understood the root one is on solhint side, they'd need to significantly change how plugins are loaded
vscode extension would come after

examples/minimal/packages/contracts/package.json
holic: if we add this to templates, we need to adjust `copy-templates.sh` to also parse `file:` deps (not just `link:`)

juanfranblanco:
The issue with plugins was due to not being able to run them separately (IE providing a path) the package runs in the context of the extension not on the users workspace so it cannot find them. If this is changed it could be configured to look for plugins in a specific folder i.e user's node modules in her project.

packages/solhint-plugin-mud/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "solhint-plugin-mud",
+  "version": "1.42.0",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/latticexyz/mud.git",
+    "directory": "packages/solhint-plugin"
```
alvrs, line 7:
```suggestion
    "directory": "packages/solhint-plugin-mud"
```

alvrs (approved):
lgtm! VSCode support would be amazing but not blocking for this PR

alvrs (approved):

PR #891, opened 2023-05-18
fix(cli): account for arrays of tuples in deploy, add test

yonadaaa:
Passing an array of a structs with reference type fails with a custom error. This is probably similar to #658.


dk1a:
I'll probably want to move the tests to integration-sandbox, when I fix its flakiness. But they work in examples very well for now

alvrs (approved):

PR #899, opened 2023-05-18
feat(create-mud): use solhint-plugin-mud in templates

dk1a:
needs the plugin to be published to work correctly

### DIFF
--- a/packages/create-mud/scripts/copy-templates.sh
+++ b/packages/create-mud/scripts/copy-templates.sh
@@ -11,6 +11,7 @@ find ./dist/templates/* -name "package.json" -type f | while read -r file; do
   cat "$file"
   # GPT-4 recommended perl to edit-in-place rather than sed, because sed wasn't working on CI
   perl -pi -e 's|"(?=@latticexyz)([^"]+)":\s*"link:[^"]+"|"\1": "{{mud-version}}"|g' "$file"
+  perl -pi -e 's|"(solhint-plugin-mud)":\s*"file:[^"]+"|"\1": "{{mud-version}}"|g' "$file"
   echo "After replacement in $file:"
   cat "$file"
   echo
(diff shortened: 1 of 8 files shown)


holic:
if you add this, it should release automatically via CI
https://github.com/latticexyz/mud/blob/d38232032b86d3dbccc00ffc23f6eee4e93c7732/packages/dev-tools/package.json#L55-L57

packages/solhint-plugin-mud/package.json
```diff
@@ -23,5 +23,8 @@
   "devDependencies": {
     "tsup": "^6.7.0"
   },
+  "publishConfig": {
+    "access": "public"
+  },
```
dk1a, line 28: @holic did you mean here?

holic: yep! though our linter will prob complain about exact placement in this, might wanna run `pnpm sort-package-json` from root

packages/create-mud/scripts/copy-templates.sh
```diff
@@ -11,6 +11,7 @@ find ./dist/templates/* -name "package.json" -type f | while read -r file; do
   cat "$file"
   # GPT-4 recommended perl to edit-in-place rather than sed, because sed wasn't working on CI
   perl -pi -e 's|"(?=@latticexyz)([^"]+)":\s*"link:[^"]+"|"\1": "{{mud-version}}"|g' "$file"
+  perl -pi -e 's|"(solhint-plugin-mud)":\s*"file:[^"]+"|"\1": "{{mud-version}}"|g' "$file"
```
holic, line 14: just to confirm this works as expected, you can run `pnpm build` in `create-mud` and look at `dist/templates` to see if the solhint plugin version got swapped with the latest canary?

dk1a: Capture

packages/solhint-plugin-mud/package.json
dk1a: linter seems to think this is the correct order

holic (approved):

PR #905, opened 2023-05-18
fix(cli): account for arrays of tuples in deploy

dk1a:
fixes the issue from #891


holic:
can we combine this with the PR that has the tests?

dk1a:
good point, I'll just drop the fix there

PR #909, opened 2023-05-18
docs: update MUD 101 video

alvrs:
updates to the reuploaded MUD 101 video without audio bugs


holic (approved):
I was also thinking of moving the Talks heading up above the Packages heading, what do you think?

alvrs:
yeah I think that would make sense!

holic (approved):

PR #919, opened 2023-05-19
fix(network,std-client): make nodejs compatible

alvrs:
NOTE: some of these fixes feel very hacky and I'm not sure we should merge this PR.

If you're a hacker at the [Autonomous World hackathon](https://ethglobal.com/events/autonomous) and need to run the network sync in nodejs, here is how you can use this branch to make it work: 

1. Clone this repository, checkout the `alvrs/nodejs` branch
2. Run `pnpm install && pnpm build` at the root
3. Run `pnpm mud set-version --link <relative/path/to/mud/root>` from your nodejs project root to link your project to your local MUD setup
4. Run `pnpm install` at your nodejs project root to install the local MUD packages

Example to changes in a project to get this running: https://github.com/oostvoort/imaigine/pull/1


packages/network/src/initCache.ts
```diff
@@ -1,6 +1,9 @@
 import { arrayToIterator, deferred, mergeIterators, transformIterator } from "@latticexyz/utils";
+import fakeIndexedDb from "fake-indexeddb";
```
holic, line 2: I wonder if there's a way to update the methods below to use an async import so we don't have to import this into browser bundles.

```
const indexedDB = typeof self === 'undefined' ? await import('fake-indexeddb') : self.indexedDB.
```

PR #922, opened 2023-05-19
fix(network): fix key schema sync from MODE

alvrs:
`decodeSchema` expects the schema to be a single hex string of both value and key schema, but when syncing via MODE we were only passing in the value schema


Kooshaba (approved):
tested against local MODE it is working 👍🏼

PR #928, opened 2023-05-20
build(dev-tools): replace local import with react package

alvrs:


holic:
whoooops, must have been one of those IDE "fix missing imports" but did it badly

PR #930, opened 2023-05-20
fix(cli): clean anvil cache/logs at start of dev-contracts

alvrs:
* Prune the anvil cache before every start of `mud dev-contracts` to avoid issues with large cache files

### DIFF
--- a/packages/cli/src/commands/dev-contracts.ts
+++ b/packages/cli/src/commands/dev-contracts.ts
@@ -10,6 +10,8 @@ import { debounce } from "throttle-debounce";
 import { worldgenHandler } from "./worldgen";
 import { WorldConfig } from "@latticexyz/world";
 import { deployHandler, printMUD, worldtypes } from "../utils";
+import { homedir } from "os";
+import { rmSync } from "fs";
 
 type Options = {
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/dev-contracts.ts
```diff
@@ -57,6 +59,10 @@ const commandModule: CommandModule<Options, Options> = {
 
     // Start an anvil instance in the background if no RPC url is provided
     if (!args.rpc) {
+      console.log(chalk.gray("Cleaning devnode cache"));
+      const userHomeDir = homedir();
+      rmSync(path.join(userHomeDir, ".foundry", "anvil", "tmp"), { recursive: true, force: true });
```
alvrs, line 64: this corresponds to anvil's cache location as configured here: https://github.com/foundry-rs/foundry/blob/c331b5eeee1b4151ef7354a081667e2d770b37f5/anvil/src/config.rs#L1085-L1086

alvrs: opened an issue on anvil's side here: https://github.com/foundry-rs/foundry/issues/4989

PR #934, opened 2023-05-20
feat(perlin): change function visibility to internal

### LINKED ISSUE #935: Perlin Lib not compatible with MUD2
#### Description

Currently, MUD2 does not play nicely with public libraries. The functions in the [Perlin.sol](https://github.com/latticexyz/mud/blob/main/packages/noise/contracts/Perlin.sol) file all have `public` modifiers. 

#### Motivation

Inability to use Perlin noise in systems.

#### Minimal Repro

```solidity
pragma solidity >=0.8.0;

import { System } from "@latticexyz/world/src/System.sol";
import { Perlin } from "@latticexyz/noise/contracts/Perlin.sol";

contract MySystem is System {

    function callsPerlin(int256 positionX, int256 positionY) public pure returns (uint256) {
        int128 perlinNoise = Perlin.noise2d(positionX, positionY, 0, 0);
    }
}
```

alvrs:
fixes #935 

* tooling for deploying public libraries is bad and with MUD2 using (private) systems is preferable to public libraries

### DIFF
--- a/packages/noise/contracts/Perlin.sol
+++ b/packages/noise/contracts/Perlin.sol
@@ -155,32 +155,32 @@ library Perlin {
     return Math.div(Math.add(h.r, _1), _2) >> (64 - precision);
   }
 
-  function dec(int128 x) public pure returns (int128) {
+  function dec(int128 x) internal pure returns (int128) {
     return Math.sub(x, _1);
   }
 
-  function floor(int128 x) public pure returns (int128) {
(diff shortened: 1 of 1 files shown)


dk1a (commented):
the other funcs (dec, floor etc) should also be internal

alvrs:
> the other funcs (dec, floor etc) should also be internal

ahh good catch

dk1a (approved):

PR #936, opened 2023-05-20
fix(world): fix config keySchema

dk1a:
the issue didn't show up in CI because config isn't typechecked in CI, zod silently ignores unknown keys, and keySchema is optional
that config override also doesn't actually change the default key type, just adds a name to it

### DIFF
--- a/packages/world/mud.config.ts
+++ b/packages/world/mud.config.ts
@@ -69,7 +69,7 @@ export default mudConfig({
     },
     SystemHooks: {
       directory: "modules/core/tables",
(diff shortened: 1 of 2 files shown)


holic:
> because config isn't typechecked in CI

can we fix this?

holic (approved):

dk1a:
> > because config isn't typechecked in CI
> 
> can we fix this?

yep, just wanted to check some things so separated the PRs https://github.com/latticexyz/mud/pull/938

PR #941, opened 2023-05-22
feat: make worldAddress as the caller and address of mud tests

### LINKED ISSUE #940: Allow tables to be used in tests without providing explicit store argument
This is especially a problem for libraries that use tables - having to manually pass worldAddress in all custom lib methods is bad devex
Making systems specifically for tests is an option of course, but also rather inconvenient and unnecessary in many cases

1. One solution is to call tests with msg.sender equal to worldAddress (as if they're systems), and also startPrank(worldAddress) in tests to bypass AccessControl (as if whenever tests call world, it's like world calls itself, e.g. a root system).

2. Another solution would be to change how StoreSwitch detects store, for example instead of `isStore` have a method that actually returns storeAddress

3. Or we can double down on systems for tests, and not do this. I'm not entirely certain spoofing world in tests is the best idea

dk1a:
fixes #940
Here I try the 1st solution, as outlined in #940

Note that while this is a slightly breaking change (for tests that use vm.startPrank, since they'd need to call vm.stopPrank 1st),
the changes to examples/templates aren't necessary, the old versions would work fine

### DIFF
--- a/packages/std-contracts/src/test/MudV2Test.t.sol
+++ b/packages/std-contracts/src/test/MudV2Test.t.sol
@@ -8,5 +8,8 @@ contract MudV2Test is Test {
 
   function setUp() public virtual {
     worldAddress = vm.parseAddress(vm.readFile(".mudtest"));
+    // allows passing AccessControl when writing tables directly through libraries
+    // (this should be combined with `--sender` flag when calling `forge test`)
+    vm.startPrank(worldAddress);
   }
 }
--- a/examples/minimal/packages/contracts/test/CounterTest.t.sol
+++ b/examples/minimal/packages/contracts/test/CounterTest.t.sol
@@ -30,19 +30,19 @@ contract CounterTest is MudV2Test {
   function testCounter() public {
     // Expect the counter to be 1 because it was incremented in the PostDeploy script.
     bytes32 key = SingletonKey;
-    uint32 counter = CounterTable.get(world, key);
+    uint32 counter = CounterTable.get(key);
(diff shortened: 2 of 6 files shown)


packages/std-contracts/src/test/MudV2Test.t.sol
```diff
@@ -8,5 +8,8 @@ contract MudV2Test is Test {
 
   function setUp() public virtual {
     worldAddress = vm.parseAddress(vm.readFile(".mudtest"));
+    // allows passing AccessControl when writing tables directly through libraries
+    // (this should be combined with `--sender` flag when calling `forge test`)
+    vm.startPrank(worldAddress);
```
holic, line 13: this does feel like a bit of a "gotcha" in that anyone calling `startPrank` later will get errors

Is there another way to run tests in some sort of custom "context" in which we'd could specify where the store calls go?

dk1a: yeah that would be 2nd solution from https://github.com/latticexyz/mud/issues/940, which allows tests to explicitly specify worldAddress for tables called directly from the test contract.
I'd like @alvrs's opinion on that one, it's a pretty big low-level change

holic: wouldn't this approach also mean that regular system calls via the world have the wrong `_msgSender()`?

alvrs:
closing this, as we've discussed to go with a different approach (approach n.2 in #940)

PR #951, opened 2023-05-22
feat(mode): handle singleton table updates

authcall:
Handles singleton table updates, where a setField only updates fields without specifying what the key is (because there are no keys)

### DIFF
--- a/packages/services/pkg/mode/db/api.go
+++ b/packages/services/pkg/mode/db/api.go
@@ -87,7 +87,14 @@ func (dl *DatabaseLayer) Create(table string, record interface{}) error {
 
 // Updates updates a record in the specified table using the provided record struct and filter criteria.
 func (dl *DatabaseLayer) Updates(table string, record interface{}, filter map[string]interface{}) (*gorm.DB, error) {
-	updates := dl.gorm__db.Table(table).Where(filter).Updates(record)
+	var updates *gorm.DB
+	if len(filter) == 0 {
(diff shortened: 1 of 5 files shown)


alvrs (approved):

alvrs (approved):

packages/services/pkg/mode/db/api.go
```diff
@@ -87,7 +87,14 @@ func (dl *DatabaseLayer) Create(table string, record interface{}) error {
 
 // Updates updates a record in the specified table using the provided record struct and filter criteria.
 func (dl *DatabaseLayer) Updates(table string, record interface{}, filter map[string]interface{}) (*gorm.DB, error) {
-	updates := dl.gorm__db.Table(table).Where(filter).Updates(record)
+	var updates *gorm.DB
+	if len(filter) == 0 {
+		// If the WHERE clause is empty, then update all records in the table.
+		// This is to handle the case where the filter is not provided due to a singleton table.
+		updates = dl.gorm__db.Table(table).Where("true").Updates(record)
```
holic, line 94: ooc why is `.Where("true")` needed?

authcall: where true acts as a no-op when there are no WHERE conditions, which happens in the singleton table case. see comment above the line

authcall: it is

PR #954, opened 2023-05-22
feat(services): docker compose setup for MODE

Kooshaba:


packages/services/config.mode-docker.yaml
```diff
@@ -0,0 +1,19 @@
+---
+chains:
+  - name: "localhost"
+    id: "31337"
+    rpc:
+      http: "http://host.docker.internal:8545"
```
Kooshaba, line 6: this was a workaround i found for Mac specifically. not sure if this works on any other platform.

https://github.com/docker/compose/issues/3800#issuecomment-496480571

packages/services/docker-compose.mode.yml
```diff
@@ -0,0 +1,33 @@
+version: '3.8'
+services:
+  db:
+    image: postgres:latest
```
alvrs, line 4: I think we could use the custom image I created for the github runner here (which extending `postgres:latest` by setting the WAL config like mode needs it)
```suggestion
    image: ghcr.io/latticexyz/postgres-wal-logical:latest
```

alvrs: although, actually it might require additional steps like authorising to the github docker registry. maybe just using the init script would be easier for devs.

packages/services/docker-compose.mode.yml
```diff
@@ -0,0 +1,33 @@
...
+    restart: always
+    volumes: 
+      - ./data:/var/lib/postgresql/data
+      - ./init-scripts:/docker-entrypoint-initdb.d
+    environment:
+      - POSTGRES_DB=mode_ephemeral
+      - POSTGRES_USER=postgres
+      - POSTGRES_PASSWORD=postgres
```
alvrs, line 12: I'm getting some errors when booting this up:

```
 2023-06-16T12:23:01Z FATAL   db/core.go:129  failed to connect to database       {"error": "dial tcp 172.18.0.2:5432: connect: connection refused"}
services-mode-1  | latticexyz/mud/packages/services/pkg/mode/db.NewDatabaseLayer
services-mode-1  |      /service/pkg/mode/db/core.go:129
services-mode-1  | main.main
services-mode-1  |      /service/cmd/mode/main.go:81
services-mode-1  | runtime.main
services-mode-1  |      /usr/local/go/src/runtime/proc.go:250
```

for the github action I just used `POSTGRES_HOST_AUTH_METHOD: trust`, which allows us to use almost the same mode config as for running it outside of docker.

alvrs (approved):

PR #955, opened 2023-05-22
feat(examples): test ephemeral table within namespaced system

dk1a:
test for https://github.com/latticexyz/mud/pull/952

### DIFF
--- a/examples/minimal/packages/contracts/test/ChatNamespaced.t.sol
+++ b/examples/minimal/packages/contracts/test/ChatNamespaced.t.sol
@@ -16,8 +16,8 @@ contract ChatNamespacedTest is MudV2Test {
   }
 
   function testEmitEphemeral() public {
-    vm.expectEmit(true, true, true, true);
     bytes32[] memory keyTuple;
+    vm.expectEmit(true, true, true, true);
     emit StoreCore.StoreEphemeralRecord(MessageTableTableId, keyTuple, MessageTable.encode("test"));
     IChatNamespacedSystem(worldAddress).namespace_ChatNamespaced_sendMessage("test");
   }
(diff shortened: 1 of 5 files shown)


examples/minimal/packages/contracts/test/ChatNamespaced.t.sol
```diff
@@ -0,0 +1,24 @@
...
+import { IChatNamespacedSystem } from "../src/interfaces/IChatNamespacedSystem.sol";
+
+contract ChatNamespacedTest is MudV2Test {
+  function setUp() public override {
+    super.setUp();
+  }
+
+  function testEmitEphemeral() public {
+    vm.expectEmit(true, true, true, true);
+    bytes32[] memory keyTuple;
```
holic, line 20:
```suggestion
    bytes32[] memory keyTuple;
    vm.expectEmit(true, true, true, true);
```
just adjusting the order here because `expectEmit` has gotten stricter and makes it a little clearer to me when `emit Event` exactly follows `expectEmit`

examples/minimal/packages/contracts/test/ChatNamespaced.t.sol
```diff
@@ -0,0 +1,24 @@
...
+import { StoreCore } from "@latticexyz/store/src/StoreCore.sol";
+
+import { IWorld } from "../src/codegen/world/IWorld.sol";
+import { MessageTable, MessageTableTableId } from "../src/codegen/Tables.sol";
+import { IChatNamespacedSystem } from "../src/interfaces/IChatNamespacedSystem.sol";
+
+contract ChatNamespacedTest is MudV2Test {
+  function setUp() public override {
+    super.setUp();
+  }
```
holic, line 16: Is this needed?

holic (dismissed):

examples/minimal/packages/contracts/test/ChatNamespaced.t.sol
dk1a: nope, good catch

PR #956, opened 2023-05-22
fix(network): pad keys for entity IDs

holic:


packages/network/src/v2/keyTupleToEntityID.ts
```diff
@@ -1,12 +1,13 @@
...
-import { normalizeEntityID } from "../utils";
+import { toHex, pad, isHex, Hex } from "viem";
 import { SingletonID } from "../workers";
 
 // TODO: revisit key tuple format?
 export function keyTupleToEntityID(keyTuple: any[]): Entity {
   // v2 uses an empty key tuple as the singleton ID, so we'll return the corresponding v1 singleton entity ID to normalize this for now
   if (keyTuple.length === 0) {
-    return SingletonID;
+    return pad(SingletonID as Hex, { size: 32 }) as Entity;
```
holic, line 9: It'll take a little extra work (mostly loading states and in templates + tutorials) but I think we can probably move this to `return ""` and nix the `singletonEntity` stuff. cc @ludns

packages/network/src/v2/keyTupleToEntityID.spec.ts
```diff
@@ -0,0 +1,17 @@
...
+  it("should pad addresses", () => {
+    expect(keyTupleToEntityID(["0xc0035952298729c7086058EaA13921626e22C070"])).toBe(
+      "0x000000000000000000000000c0035952298729c7086058EaA13921626e22C070"
+    );
+  });
+
+  it("convert number key to padded entity ID", () => {
+    expect(keyTupleToEntityID([1])).toBe("0x0000000000000000000000000000000000000000000000000000000000000001");
+  });
+});
```
holic, line 17: any other cases I should test for?

packages/std-client/src/setup/setupMUDV2Network.ts
```diff
@@ -143,8 +144,7 @@ export async function setupMUDV2Network<C extends ContractComponents, S extends
   const singletonEntity = world.registerEntity({ id: SingletonID });
   // Register player entity
   const address = network.connectedAddress.get();
-  const playerEntityId = address;
-  const playerEntity = playerEntityId ? world.registerEntity({ id: playerEntityId }) : undefined;
+  const playerEntity = address ? world.registerEntity({ id: keyTupleToEntityID([address]) }) : undefined;
```
holic, line 147: this was still registering the unpadded version, which means if you used this anywhere in a v2 app, you'd get the wrong results for a lookup/query

packages/std-client/src/setup/setupMUDV2Network.ts
```diff
@@ -203,8 +203,6 @@ export async function setupMUDV2Network<C extends ContractComponents, S extends
     components,
     singletonEntityId: SingletonID,
     singletonEntity,
-    /* @deprecated playerEntityId is equivalent to playerEntity */
-    playerEntityId,
```
holic, line 207: now that we have `Entity` and symbols, we don't really need this

packages/network/src/v2/keyTupleToEntityID.ts
holic: since I'm here and already modifying `playerEntity`, should I just rip out `singletonEntity` too?

ludns: I'd rip all of those, unless we are thinking of moving straight to refactoring all of this and killing; and making this a non breaking change that just fixes things instead of cleaning the API

packages/std-client/src/setup/setupMUDV2Network.ts
```diff
@@ -143,8 +144,7 @@ export async function setupMUDV2Network<C extends ContractComponents, S extends
   const singletonEntity = world.registerEntity({ id: SingletonID });
```
alvrs, line 144: I think this one needs to be normalized too?

alvrs (approved):

alvrs (approved):

PR #967, opened 2023-05-23
test: add failing for KeysInTable with singleton

yonadaaa:


packages/world/src/modules/keysintable/KeysInTableHook.sol
```diff
@@ -45,19 +58,47 @@ contract KeysInTableHook is IStoreHook {
...
 
-      uint32 length = KeysInTable.getLength(tableId);
+      uint256 length = KeysInTable.lengthKeys0(tableId);
 
       if (length == 1) {
         // Delete the list of keys in this table
         KeysInTable.deleteRecord(tableId);
       } else {
-        bytes32 lastKey = KeysInTable.getKeys(tableId)[length - 1];
+        if (key.length > 0) {
```
yonadaaa, line 67: I wish we could avoid this switch statement, but we can't index the key fields programmatically

packages/world/src/modules/keysintable/getKeysInTable.sol
```diff
@@ -11,34 +13,69 @@ import { KeysInTable } from "./tables/KeysInTable.sol";
...
+function getKeysInTable(bytes32 tableId) view returns (bytes32[][] memory keyTuples) {
   /**
-   * Note: this module does not yet support composite keys.
-   * Therefore we create an array of key tuples but only fill the first entry.
+   * Note: this module only supports up to 5 composite keys.
    */
-  bytes32[] memory keysInTableRaw = KeysInTable.getKeys(tableId);
 
-  keysInTable = new bytes32[][](keysInTableRaw.length);
+  Schema schema;
```
yonadaaa, line 21: FIXME

packages/world/src/modules/keysintable/KeysInTableHook.sol
alvrs: technically we could just use the lower level `StoreCore` methods directly (and programatically provide the `fieldIndex` param), but we'd lose the type helpers so probably not worth it

packages/world/src/modules/keysintable/getKeysInTable.sol
```diff
@@ -11,34 +13,69 @@ import { KeysInTable } from "./tables/KeysInTable.sol";
...
+   * Note: this module only supports up to 5 composite keys.
+   */
+
+  Schema schema = store.getKeySchema(tableId);
+
+  uint256 length = KeysInTable.lengthKeys0(store, tableId);
+  keyTuples = new bytes32[][](length);
 
-  keysInTable = new bytes32[][](keysInTableRaw.length);
+  for (uint256 i; i < keyTuples.length; i++) {
```
alvrs, line 62:
```suggestion
  for (uint256 i; i < length; i++) {
```

packages/world/src/modules/keysintable/getKeysInTable.sol
```diff
@@ -11,34 +13,69 @@ import { KeysInTable } from "./tables/KeysInTable.sol";
...
+    keyTuples[i] = new bytes32[](schema.numFields()); // the length of the key tuple depends on the schema
 
-  for (uint256 i; i < keysInTableRaw.length; i++) {
-    keysInTable[i] = new bytes32[](1);
-    keysInTable[i][0] = keysInTableRaw[i];
+    if (schema.numFields() > 0) {
```
alvrs, line 65: I think it's worth assigning `schema.numFields()` to a memory variable

packages/world/src/modules/keysintable/KeysInTableHook.sol
```diff
@@ -15,14 +15,27 @@ contract KeysInTableHook is IStoreHook {
...
+      }
+      if (key.length > 2) {
+        KeysInTable.pushKeys2(tableId, key[2]);
+      }
+      if (key.length > 3) {
+        KeysInTable.pushKeys3(tableId, key[3]);
+      }
+      if (key.length > 4) {
+        KeysInTable.pushKeys4(tableId, key[4]);
+      }
```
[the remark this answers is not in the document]
yonadaaa: great point, i will nest them

packages/world/src/modules/keysintable/getKeysInTable.sol
```diff
@@ -11,34 +13,69 @@ import { KeysInTable } from "./tables/KeysInTable.sol";
...
+          keyTuples[i][2] = KeysInTable.getItemKeys2(tableId, i);
+          if (numFields > 3) {
+            keyTuples[i][3] = KeysInTable.getItemKeys3(tableId, i);
+            if (numFields > 4) {
+              keyTuples[i][4] = KeysInTable.getItemKeys4(tableId, i);
+            }
+          }
+        }
+      }
+    }
```
alvrs, line 44: since we're interested in the full array, but just need to change the format of the result data, it might be cheaper to first load the full arrays and then change the format (instead of going though the table library for every item, which has some overhead)

alvrs: not blocking though, we can add an issue to track and come back to gas optimizing this

yonadaaa: True. Either way the gas impact this has had is pretty crazy. I am on the side of leaving it for now and refactoring this whole module in future to support any number of keys, codegen etc.

packages/world/test/KeysInTableModule.t.sol
```diff
@@ -22,64 +22,106 @@ contract KeysInTableModuleTest is Test {
...
     // -> requires `callFrom` for the module to be able to register a hook in the name of the original caller
     // !gasreport install keys in table module
-    world.installRootModule(keysInTableModule, abi.encode(sourceTableId));
+    world.installRootModule(keysInTableModule, abi.encode(tableId));
+    world.installRootModule(keysInTableModule, abi.encode(singletonTableId));
+    world.installRootModule(keysInTableModule, abi.encode(compositeTableId));
   }
 
-  function testInstall() public {
+  function testFailInstallSingleton() public {
```
alvrs, line 75: should we add a quick comment here about that this is currently failing and why but we should get back to it in the future?

PR #971, opened 2023-05-23
feat: namespace overrides for tables and systems

dk1a:


packages/store-cache/src/createDatabaseClient.ts
```diff
@@ -23,6 +22,7 @@ export function createDatabaseClient<C extends StoreConfig>(database: TupleDatab
 
   // Create utils with client, namespace and table argument prefilled
   for (const table in config.tables) {
+    const namespace = config.tables?.[table].namespace ?? config.namespace;
```
alvrs, line 25: would the table config's namespace ever be undefined? since put the default namespace there while expanding the config?

dk1a: no, namespace is always set by zod.
`config.tables?.[table]` the ? confused me a bit into thinking the whole thing could be undefined

dk1a:
deprecated by the preferred approach from https://github.com/latticexyz/mud/issues/994

PR #973, opened 2023-05-23
feat: add keyNames to table metadata

### LINKED ISSUE #824: Add `keyNames` to table metadata
For now `setMetadata` only supports `fieldNames` but not `keyNames`. We should add support for `keyNames` to allow decoding them in the network package and use them in `store-cache`

yonadaaa:
Solves #824.

### DIFF
--- a/packages/network/src/v2/mode/syncTablesFromMode.ts
+++ b/packages/network/src/v2/mode/syncTablesFromMode.ts
@@ -48,6 +48,7 @@ export async function syncTablesFromMode(
     const keySchemaTypes = keyAbiTypes.map((abiType) => AbiTypeToSchemaType[abiType]);
     const keySchemaHex = arrayToHex(encodeSchema(keySchemaTypes));
 
+    const keyNames: string[] = []; // FIXME
     const fieldNames = cols.slice(keyLength);
     // TODO: remove this hack once MODE is fixed (https://github.com/latticexyz/mud/issues/444)
     const fieldAbiTypes = types.slice(keyLength).map((modeType) => modeType.match(/tuple\((.*)\[]\)/)?.[1] ?? modeType);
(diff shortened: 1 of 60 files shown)


packages/store/mud.config.ts
```diff
@@ -12,6 +12,7 @@ export default mudConfig({
       },
       schema: {
         tableName: "string",
+        abiEncodedKeyNames: "bytes",
         abiEncodedFieldNames: "bytes",
       },
```
holic, line 17: I think we'll need to update network decoding to handle this:
https://github.com/latticexyz/mud/blob/dd8413340c429ec28f909bbb1b96fd117cf23a5e/packages/network/src/v2/schemas/tableMetadata.ts#L62

yonadaaa: good spot. with MODE too this is quite involved.

packages/network/src/v2/mode/syncTablesFromMode.ts
```diff
@@ -48,6 +48,7 @@ export async function syncTablesFromMode(
     const keySchemaTypes = keyAbiTypes.map((abiType) => AbiTypeToSchemaType[abiType]);
     const keySchemaHex = arrayToHex(encodeSchema(keySchemaTypes));
 
+    const keyNames: string[] = []; // FIXME
```
yonadaaa, line 51: not 100% sure how to decode this in MODE

alvrs:
this generally looks great, but we recently decided to go the route of combining the metadata registration with the schema registration to tighten the constraints and drastically simplify lots of the indexer logic (https://github.com/latticexyz/mud/issues/1043), so i'm gonna close this (as it wouldn't be worth integrating this in MODE if we're gonna change it shortly after)

PR #980, opened 2023-05-25
fix(store): `tableId` should be bytes32

holic:
found this when digging around for reasons I'm not seeing key schema for `mudstore.schema` records propagate to client

this doesn't fix anything but seems like it was meant to be `bytes32` to match the rest of table IDs?


dk1a (approved):
yeah it should be bytes32

PR #981, opened 2023-05-25
add failing test for key schema

holic:


holic:
@dk1a Do you have a sense for why this happens?

dk1a:
looking into this, seems like a bug, which just doesn't affect things much (unless u specifically query keySchema, which isn't really used anywhere, and client has events for it, which are correct)

holic:
> unless u specifically query keySchema, which isn't really used anywhere, and client has events for it, which are correct

I am trying to use this, but had to make an ugly workaround because of this bug.

dk1a:
ok so it was stored, but at the wrong slot, so if u query the whole record it breaks (the few existing use-cases seem to query fields separately)
the problem is that unlike memory, storage slots use whole words, so if u add 0x20 to storage location, u get offset by 1024 bytes

alvrs (approved):

PR #982, opened 2023-05-26
fix(network): decode negative keys

smallbraingames:
- viem's decodeAbiParameters doesn't expect 0 padding for negatives, but expects f padding
- this causes negative keys to be decoded as large positive numbers in decodeKeyTuple
- fix is to use decodeStaticField instead of viem
- see conversation here: https://discord.com/channels/865335009915961364/1111501157077094533/1111503609058828319

### DIFF
--- a/packages/network/src/v2/schemas/decodeKeyTuple.ts
+++ b/packages/network/src/v2/schemas/decodeKeyTuple.ts
@@ -1,11 +1,17 @@
-import { SchemaTypeToAbiType } from "@latticexyz/schema-type";
+import { StaticSchemaType, getStaticByteLength } from "@latticexyz/schema-type";
+
 import { Schema } from "../common";
-import { decodeAbiParameters } from "viem";
+import { decodeStaticField } from "./decodeStaticField";
+import { hexToArray } from "@latticexyz/utils";
 
 export function decodeKeyTuple(keySchema: Schema, keyTuple: unknown[]) {
-  const abiTypes = keySchema.staticFields.map((schemaType) => SchemaTypeToAbiType[schemaType]);
-  const decodedKeys = keyTuple.map(
-    (key, index) => decodeAbiParameters([{ type: abiTypes[index] }], key as `0x${string}`)[0]
-  );
+  const decodedKeys = keyTuple.map((key, index) => {
+    const schemaType = keySchema.staticFields[index] as StaticSchemaType;
+    const byteLength = getStaticByteLength(schemaType);
+    const bytes = hexToArray(key as string);
+    // Key data is padded, so need offset
+    const offset = bytes.length - byteLength;
+    return decodeStaticField(keySchema.staticFields[index] as StaticSchemaType, bytes, offset);
+  });
   return decodedKeys.reduce<Record<number, unknown>>((acc, curr, i) => ({ ...acc, [i]: curr }), {});
 }


packages/network/src/v2/schemas/decodeKeyTuple.ts
```diff
@@ -1,11 +1,16 @@
...
 export function decodeKeyTuple(keySchema: Schema, keyTuple: unknown[]) {
-  const abiTypes = keySchema.staticFields.map((schemaType) => SchemaTypeToAbiType[schemaType]);
-  const decodedKeys = keyTuple.map(
-    (key, index) => decodeAbiParameters([{ type: abiTypes[index] }], key as `0x${string}`)[0]
-  );
+  const decodedKeys = keyTuple.map((key, index) => {
+    const schemaType = keySchema.staticFields[index] as StaticSchemaType;
+    const byteLength = getStaticByteLength(schemaType);
+    // Key data is padded, so need offset
+    const offset = 32 - byteLength;
```
holic, line 12: this is a bit dangerous because a bytes8 value will be padded differently than an int32 value

smallbraingames: @holic okay just made an extremely small brain fix to that, lmk what you think

smallbraingames: @holic just a polite lil bump on this (locally linking for now but would prefer merged)

holic: I am still worried about the approach here using offsets to account for padding, since the direction of the padding will be different depending on the data type.

I've taken a light week this week after the hackathon but can dig into this next week.

smallbraingames: Okay, thanks!

alvrs:
I wonder if we should just encode keys onchain via `bytes32(abi.encode(key))` instead of the custom casting we do right now, so that we can use abi.decode to decode them. That would allow us to use `abi.decode` here instead of having to use this custom function (here and in all other languages we eventually want to support). Need to benchmark the gas impact though.

holic:
> I wonder if we should just encode keys onchain via `bytes32(abi.encode(key))` instead of the custom casting we do right now, so that we can use abi.decode to decode them. That would allow us to use `abi.decode` here instead of having to use this custom function (here and in all other languages we eventually want to support). Need to benchmark the gas impact though.

I just wrote a bunch of tests for key schema encoding on the Solidity side so I could get some confidence/clarity on how we we're doing it. Since I'm already in here, I'll see what `abi.encode` does to gas.

holic:
`abi.encode` seems to add a non-trivial amount of gas to each operation: https://github.com/latticexyz/mud/commit/7f38230070d54e0c4c090d987c723f60e79ea2c6

holic:
I figured out why we'd need to work around negative signed integers: our codegen's type casting approach doesn't match `abi.encode`

```
➜ int32 signed = -42
➜ bytes32(uint256(uint32(signed)))
Type: bytes32
└ Data: 0x00000000000000000000000000000000000000000000000000000000ffffffd6
➜ bytes32(abi.encode(signed))
Type: bytes32
└ Data: 0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffd6
```

Trying to get gas reports working for the integration sandbox so I can test before/after changing, since there'll prob be some slight gas cost to `f`-padded vs. zero-padded values?

holic:
> If it's not too costly gas-wise

It's a whopping 4 gas more, so going to go with the `abi.encode`-compatible approach!

holic:
@smallbraingames Can you give https://github.com/latticexyz/mud/pull/1001 a try instead of this PR?

PR #984, opened 2023-05-26
fix(store): getKeySchema shouldn't revert for singleton tables

holic:
`keySchema` can be empty for singleton tables so thus method reverts on singleton tables.

I considered removing the check all together, but thought it might be nice to know when a table doesn't exist vs. key schema being empty.

TODO:
- [x] add regression tests


dk1a:
I think the check can be removed, because `hasTable` uses `schema`, and `keySchema` always comes together with `schema`, and `schema` can never be empty, so you always retain an existence check regardless of `keySchema`

holic:
This changes means this test no longer fails (which means tests pass when its expected to fail):
https://github.com/latticexyz/mud/blob/2908c80c37f09e430991a50ded40d1abcef964f0/packages/world/test/KeysInTableModule.t.sol#L79-L93

@yonadaaa Is it safe to remove this test?

yonadaaa:
@holic using the `KeysIn` modules on singleton tables is kinda undefined behaviour (should it return no keys or some singleton key?). In practice it returns no keys - so instead of deleting it, you could remove the `TestFail` annotation so the tests can document this.

packages/store/test/StoreCore.t.sol
```diff
@@ -68,12 +69,21 @@ contract StoreCoreTest is Test, StoreMock {
...
+
+    vm.expectRevert(
+      abi.encodeWithSelector(IStoreErrors.StoreCore_TableNotFound.selector, table2, TableId.toString(table2))
+    );
+    StoreCore.getSchema(table2);
+
+    vm.expectRevert(
+      abi.encodeWithSelector(IStoreErrors.StoreCore_TableNotFound.selector, table2, TableId.toString(table2))
+    );
+    StoreCore.getKeySchema(table2);
```
dk1a, line 86: I think this might have weird behaviour because StoreCore is a lib, not a contract, and `expectRevert` might pass even if only 1 of them actually reverts. Might have changed since last time I came across this though

holic: this test seems to pass as-is 🤷‍♂️

dk1a: yeah that should work

before this change, to see the problem I meant:
remove the revert from `getKeySchema`,
see that tests still pass (they shouldn't),
remove
```
vm.expectRevert(
   abi.encodeWithSelector(IStoreErrors.StoreCore_TableNotFound.selector, table2, TableId.toString(table2))
 );
StoreCore.getSchema(table2);
```
(so that there's only 1 expectRevert in the test),
see that test doesn't pass now

dk1a: basically you wouldn't notice this problem by tests failing, but rather by tests not failing when they should

holic:
@yonadaaa is this what you mean? https://github.com/latticexyz/mud/pull/984/commits/09558e16973978001ab58972ba5e057b3b8fac8d

PR #987, opened 2023-05-30
refactor(world): use dynamic fieldIndex in KeysInTableHook

dk1a:
https://github.com/latticexyz/mud/pull/985 should be merged 1st

KeysInTableHook's hardcode can all be replaced with custom methods that accept `fieldIndex`, instead of `fieldIndex` being part of a method name
This refactor increases gas by ~0.1%, but greatly increases readability and makes dynamic field limit increase painless

Keeping `KeysInTableDynamicFieldIndex` in sync with `KeysInTable` shouldn't be too troublesome, since it's only a few methods that wouldn't change much, and can be mostly copy-pasted


packages/world/src/modules/keysintable/KeysInTableHook.sol
```diff
@@ -64,56 +53,20 @@ contract KeysInTableHook is IStoreHook {
         // Delete the list of keys in this table
         KeysInTable.deleteRecord(tableId);
       } else {
-        if (key.length > 0) {
-          bytes32[] memory lastKeyTuple = new bytes32[](key.length);
+        bytes32[] memory lastKeyTuple = new bytes32[](key.length);
 
-          bytes32 lastKey = KeysInTable.getItemKeys0(tableId, length - 1);
-          lastKeyTuple[0] = lastKey;
+        for (uint8 fieldIndex = 0; fieldIndex < key.length; fieldIndex++) {
```
alvrs, line 58: We should throw an explicit error if a the key has more than 5 elements, otherwise it would be very hard to debug if someone runs into it

packages/world/src/modules/keysintable/KeysInTableDynamicFieldIndex.sol
```diff
@@ -0,0 +1,49 @@
...
+import { Bytes } from "@latticexyz/store/src/Bytes.sol";
+
+import { KeysInTable, KeysInTableTableId } from "./tables/KeysInTable.sol";
+
+/**
+ * Based on `KeysInTable`, but allows dynamically passing fieldIndex.
+ * Dynamic fieldIndex relies on all fields having the same schema type.
+ */
+library KeysInTableDynamicFieldIndex {
+  function pushKeys(uint8 fieldIndex, bytes32 sourceTable, bytes32 _element) internal {
```
alvrs, line 14: nit - why keys, not key?
```suggestion
  function pushKey(uint8 fieldIndex, bytes32 sourceTable, bytes32 _element) internal {
```

packages/world/src/modules/keysintable/KeysInTableDynamicFieldIndex.sol
```diff
@@ -0,0 +1,49 @@
...
+pragma solidity >=0.8.0;
+
+import { StoreSwitch } from "@latticexyz/store/src/StoreSwitch.sol";
+import { Bytes } from "@latticexyz/store/src/Bytes.sol";
+
+import { KeysInTable, KeysInTableTableId } from "./tables/KeysInTable.sol";
+
+/**
+ * Based on `KeysInTable`, but allows dynamically passing fieldIndex.
+ * Dynamic fieldIndex relies on all fields having the same schema type.
```
alvrs, line 11: Can you elaborate on this comment a bit? Do you mean the field types of `KeysInTable`? If so, since they all have the same type (bytes32 array since it's all keys), do we need this comment?

dk1a: yeah, the point is this trick shouldn't be used for other tables unless all their fields have the same type. I just didn't think it was obvious what's going on there so explicitly pointed it out

packages/world/src/modules/keysintable/KeysInTableDynamicFieldIndex.sol
dk1a: tbf it's neither key nor keys, but keyParts - renamed it
it's keyParts (not keyPart) because the fields are arrays
it's keyParts (not keys) because the fields are each a single element of a keyTuple
pushKeys is slightly misleading and maybe I should make it `push* -> pushTo*` and `pop* -> popFrom*` (I didn't because it's more verbose)

dk1a: oh also this table is mostly copypasted from `KeysInTable` and I don't think they should diverge more than necessary, so I changed `mud.config.ts` too

packages/world/src/modules/keysintable/KeysInTableHook.sol
dk1a: added a check during module installation, then the hook can assume it's correct and avoid using more gas

PR #990, opened 2023-06-01
build: use older foundry version in CI

dk1a:
fixes the recent examples/templates CI failures
@Kooshaba this is based on ur comments from discord, do u have more context on the root issue?


holic:
Possibly related to this? https://github.com/foundry-rs/foundry/pull/5067

PR #915, opened 2023-05-19
[WIP] debugging flaky CI

dk1a:

PR #991, opened 2023-06-01
build: re-enable integration sandbox CI

dk1a:
did a lot of CI reruns and it's reliably passing now, as long as I wait for setup to finish

### DIFF
--- a/integration-sandbox/packages/contracts/test/index.test.ts
+++ b/integration-sandbox/packages/contracts/test/index.test.ts
@@ -75,17 +75,17 @@ describe("arrays", async () => {
     await expect(page.getByTitle("Setup status")).toHaveText("finished");
 
     await resetButton.click();
-    await expect(listLength).toHaveText("0", { timeout: 120_000 });
+    await expect(listLength).toHaveText("0");
     await expect(lastItem).toHaveText("unset");
 
     await pushManyButton.click();
-    await expect(listLength).toHaveText("5000", { timeout: 60_000 });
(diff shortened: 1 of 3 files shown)


integration-sandbox/packages/contracts/test/index.test.ts
```diff
@@ -71,21 +71,21 @@ describe("arrays", async () => {
     await expect(pushManyButton).toBeVisible();
     await expect(pushOneButton).toBeVisible();
 
-    // make sure scripts are loaded
-    await page.waitForLoadState("domcontentloaded");
```
holic, line 75: Did you try `load` instead of `domcontentloaded`?

afaik `domcontentloaded` is when the DOM itself is ready (you can manipulate the DOM in scripts), but `load` waits for all resources to load/execute (scripts, styles, images, etc.)
https://developer.mozilla.org/en-US/docs/Web/API/Window/load_event

dk1a: tried some events in https://github.com/latticexyz/mud/pull/915 a couple weeks ago, don't remember which at this point. I can check tomorrow

holic: `load` seems to have worked! feels like this will scale better as we add more tests, too

holic (approved):

holic:
oof, my thing didn't work: https://github.com/latticexyz/mud/actions/runs/5150179780/jobs/9274027367

PR #993, opened 2023-06-05
fix(integration-sandbox): use custom setup status

dk1a:
trying the original fix from https://github.com/latticexyz/mud/pull/991 (which at the end got replaced by `load` event, which unfortunately didn't fix flakiness)

### DIFF
--- a/integration-sandbox/packages/contracts/test/index.test.ts
+++ b/integration-sandbox/packages/contracts/test/index.test.ts
@@ -71,8 +71,8 @@ describe("arrays", async () => {
     await expect(pushManyButton).toBeVisible();
     await expect(pushOneButton).toBeVisible();
 
-    // make sure scripts are loaded
-    await page.waitForLoadState("load");
+    // make sure setup is finished before clicking buttons
+    await expect(page.getByTitle("Setup status")).toHaveText("finished");
 
     await resetButton.click();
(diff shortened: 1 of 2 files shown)


integration-sandbox/packages/contracts/test/index.test.ts
```diff
@@ -71,8 +71,8 @@ describe("arrays", async () => {
     await expect(pushManyButton).toBeVisible();
     await expect(pushOneButton).toBeVisible();
 
-    // make sure scripts are loaded
-    await page.waitForLoadState("load");
+    // make sure setup is finished before clicking buttons
+    await expect(page.getByTitle("Setup status")).toHaveText("finished");
```
alvrs, line 75: Where does the title "Setup status" come from? Should we use the element id (`setup-status`) instead of the title here?

holic (approved):

integration-sandbox/packages/contracts/test/index.test.ts
dk1a: btw I couldn't strictly use `id` itself, that's why some elements there have `data-testid`, which playwright uses instead

dk1a:
merging since it should be a strict improvement, but @alvrs if u wanna change title or something I can do a follow-up

PR #995, opened 2023-06-05
feat(protocol-parser): add protocol parser package

holic:
I intentionally left out "parse set record/field using schema" because they're stateful (they need a schema store) and/or require an RPC client (to lazily fetch schemas/metadata) and I wanna explore them being part of the "glue" that ties together chain sync + decoding + data store (potentially stored within store-cache)


packages/protocol-parser/package.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "@latticexyz/protocol-parser",
```
holic, line 2: I don't love this name and open to suggestions!

packages/noise/package.json
```diff
@@ -26,7 +26,7 @@
   "devDependencies": {
     "@nomicfoundation/hardhat-chai-matchers": "^1.0.3",
     "@nomiclabs/hardhat-ethers": "^2.1.1",
-    "@types/chai": "^4.3.3",
+    "@types/chai": "^4.3.5",
```
holic, line 29: fun fact: vitest had a bug until ~a month ago where it would display the wrong color for expected/received data in failing tests, and just discovered it when writing the tests for this PR

bumping vitest to at least 0.31.1 fixes it, but includes/requires a `@types/chai` bump, which conflicts with the one here so I bumped this too

I'll prob pull this out into its own PR once I get closer to this PR being ready and take a clean up pass

packages/protocol-parser/src/staticAbiTypes.ts
```diff
@@ -0,0 +1,221 @@
+import { Hex } from "viem";
```
roninjin10, line 1: This file is generally useful I kinda wish it were in viem or abitype

roninjin10:
@holic noticed you have a stacked pr here.   You should check out [graphite](https://graphite.dev/) sometime.   It's so useful for handling stacks of prs.

Also if you make the tracking branch the previous pr, as long as you have the "delete branches on merge" checked github will be smart enough to make the tracking branch main after the previous pr merges

packages/protocol-parser/src/abiTypesToSchemaData.ts
```diff
@@ -0,0 +1,21 @@
+import { Hex } from "viem";
+import { DynamicAbiType } from "./dynamicAbiTypes";
+import { schemaAbiTypes } from "./schemaAbiTypes";
+import { StaticAbiType, staticAbiTypeToByteLength } from "./staticAbiTypes";
+
+export function abiTypesToSchemaData(staticFields: StaticAbiType[], dynamicFields: DynamicAbiType[] = []): Hex {
```
alvrs, line 6: thoughts on calling this `EncodedSchema` instead of `SchemaData`?

packages/protocol-parser/src/schemaAbiTypes.ts
```diff
@@ -0,0 +1,211 @@
+import { AbiType } from "abitype";
```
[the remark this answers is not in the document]
holic: happy to move em over! It was just cleaner while prototyping to start these here from a clean slate

we should probably decide: do we like packages with multiple languages (i.e. schema-type having both solidity and TS) or should we split these/colocate them with their language-specific library (e.g. store and protocol-parser)

packages/protocol-parser/src/hexToPackedCounter.ts
```diff
@@ -0,0 +1,28 @@
...
+} {
+  if (data.length !== 66) {
+    throw new InvalidHexLengthForPackedCounterError(data);
+  }
+  const totalByteLength = decodeStaticField("uint56", sliceHex(data, 0, 7));
+
+  // TODO: use schema to make sure we only parse as many as we need (rather than zeroes at the end)?
+  const fieldByteLengths = decodeDynamicField("uint40[]", sliceHex(data, 7));
+
+  // TODO: assert that the sum of the field byte lengths is equal to the total byte length
```
alvrs, line 25: should we do that here? seems easy enough

packages/protocol-parser/src/abiTypesToSchemaData.ts
holic: I originally had this as `rawSchema` but using `data` seemed to align better with everything else (each helper in this package that accepts hex arg named "data", viem calls hex "data" in many helpers)

could also see this being `abiTypesToSchemaHex ` instead and using `hex` everywhere

since each helper is pretty well contained, having consistent naming made it a little easier to move logic around into different places, otherwise we'd have hex with a bunch of different names: `encodedSchema`, `encodedTableSchema`, `encodedFieldValue`, `encodedKeyTuple`, etc.

packages/protocol-parser/src/hexToPackedCounter.ts
holic: right now this util doesn't accept/need a schema, but those validations would, so I wasn't sure if it was useful to have this as a pure parser vs. something that needed a schema to work properly

packages/protocol-parser/src/schemaAbiTypes.ts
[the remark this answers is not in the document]
holic: if you don't mind, I'd prefer to move it over in a follow up as part of a larger refactor, because once I move it over and align naming, its likely a bunch of other downstream packages (e.g. network) will need updating too, and that feels outside the scope of this PR

alvrs (approved):
so clean! ✨✨✨
(none of my comments are blocking)

packages/protocol-parser/src/abiTypesToSchemaData.ts
holic: re: naming, I was also stuck on whether things like `decodeDynamicField` should be named differently, since it's not a pure "hex to something" but rather "abi type + hex to its JS primitive type"

packages/protocol-parser/src/schemaAbiTypes.ts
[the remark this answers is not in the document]
holic: moved schema type stuff to schema-type in https://github.com/latticexyz/mud/pull/1029

PR #996, opened 2023-06-06
ERC721

0xhank:
This PR allows users to create fully compatible ERC721 tokens in their MUD worlds.
<a href="https://github.com/0xhank/mud-tokens/tree/erc721-example">An example can be seen here.</a>
<br>
### Tasks:
- [x] implement standard ERC721 proxy contract
- [x] implement ERC721 base system that proxies events and is extensible by a user's token contract 
- [x] implement base erc registration contract that is extensible by a user's custom token module
- [x] implement module framework that allows users to create custom tables and functions for their tokens
- [ ] implement plugin that autodeploys modules with the user's parameters (custom tables and function selectors)
- [ ] implement codegen script that autogenerates modules based on the tables and function selectors within token contracts 
<br>
### Design concerns
- The base `ERC721System` contract requires namespace to be passed in. The benefit is this contract doesn't need to be generated for each token contract, and the downside is the devex/gas costs of passing a namespace through each function
- Creating tables for tokens requires adding them to the namespace of the token. It is necessary for each table to have its own name within the namespace. I would recommend adding a tableId to the autogenerated table solidity files, with in this format:
```bytes16 constant tableName = bytes16('MyTokenTable.t');```

### CODE REVIEW

**Comment by 0xhank** [comment]
I have moved the functionality from being a module to a script. To make your namespace ERC721 compatible, you call ERC721Registration.install in the PostDeploy script: 
```
ERC721Registration.install(world, namespace, "Test", "TST");
```
A namespace can now take on full functionality of an ERC721 by accessing functions in the LibERC721 function. For example, to mint, you call
```
    LibERC721._mint(namespace, _msgSender(), tokenId);
```

alvrs:
Amazing initiative, highly appreciated! Unfortunately I don't have the capacity to review this at the moment because we're fully focussed on a stable release of v2. We're also gonna move in the direction of putting modules/plugins in separate packages from the core packages, maybe even a separate repository, once the plugin system is fully done and stable. We can already promote this project on the https://community.mud.dev/ page though!

0xhank:
🫡

I can turn it into an npm package? Or should I hold off until there is a more clear roadmap?

dk1a:
it's best to hold off on plugin implementation specifics, we might significantly change the API there soon

0xhank:
To clarify, there is no module or plugin. It's just a script and a library

alvrs:
i think ideally it would be a plugin/module, but for that we need to finalize the plugin foundation, so probably best to hold off for now

alvrs:
gonna close this for now to keep the PR list lean, but happy to reopen and move to the plugin package once it's ready

PR #999, opened 2023-06-07
feat(cli): add trace

dk1a:


holic:
> also updates pnpm (this wasn't very intentional, do u want it in a separate PR? )

yeah if it's not needed for this PR, it'd be great to remove this or separate it

packages/cli/src/commands/trace.ts
```diff
@@ -0,0 +1,105 @@
+import type { CommandModule } from "yargs";
+import { ethers } from "ethers";
+import { loadConfig } from "@latticexyz/config/node";
+import { StoreConfig } from "@latticexyz/store";
+import { cast, getSrcDirectory } from "@latticexyz/common/foundry";
+import { resolveWorldConfig, WorldConfig } from "@latticexyz/world";
+import { TableId } from "@latticexyz/utils";
+import { IBaseWorld } from "@latticexyz/world/types/ethers-contracts/IBaseWorld";
+import { getChainId, getExistingContracts } from "../utils";
+
```
alvrs, line 10:
```suggestion
```

packages/cli/src/commands/trace.ts
```diff
@@ -0,0 +1,105 @@
...
+    process.exit(0);
+  },
+};
+
+export default commandModule;
+
+async function getWorldAddress(worldsFile: string, worldAddress?: string, rpc?: string) {
+  if (worldAddress) {
+    return worldAddress;
```
alvrs, line 93: feels weird for this util to take an optional `worldAddress` param and then early return if it's provided; we should just do 
```ts
const worldAddress = args.worldAddress ?? getWorldAddress(worldFile, rpc);
```

packages/cli/src/commands/trace.ts
```diff
@@ -0,0 +1,105 @@
...
+  builder(yargs) {
+    return yargs.options({
+      tx: { type: "string", required: true, description: "Transaction hash to replay" },
+      worldAddress: {
+        type: "string",
+        description: "World contract address. Defaults to the value from worlds.json, based on rpc port",
+      },
+      configPath: { type: "string", desc: "Path to the config file" },
+      srcDir: { type: "string", desc: "Source directory. Defaults to foundry src directory." },
+      rpc: { type: "string", description: "json rpc endpoint, defaults to http://localhost:8545" },
```
alvrs, line 39: we should make this explicitly default to `localhost:8545` with the `default` option here. (Then we also don't need to name the default in the description because it will be displayed by default).

packages/cli/src/commands/trace.ts
```diff
@@ -0,0 +1,105 @@
...
+  },
+};
+
+export default commandModule;
+
+async function getWorldAddress(worldsFile: string, worldAddress?: string, rpc?: string) {
+  if (worldAddress) {
+    return worldAddress;
+  } else if (existsSync(worldsFile)) {
+    const chainId = rpc ? await getChainId(rpc) : 8545;
```
alvrs, line 95: if we make the rpc default to localhost (see comment above) we don't need to make `rpc` optional here either and can always use `getChainId` here

packages/cli/src/commands/trace.ts
```diff
@@ -0,0 +1,105 @@
...
+    const deploys = JSON.parse(readFileSync(worldsFile, "utf-8"));
+
+    if (!deploys[chainId]) {
+      throw new MUDError(`worldsFile is missing chainId ${chainId}`);
```
alvrs, line 99: might be good to include the path of the worlds file here, or at least call it `worlds.json` so people know where to look

packages/cli/src/commands/worldgen.ts
```diff
@@ -4,9 +4,9 @@ import { StoreConfig } from "@latticexyz/store";
 import { WorldConfig } from "@latticexyz/world";
 import { worldgen } from "@latticexyz/world/node";
 import { getSrcDirectory } from "@latticexyz/common/foundry";
-import glob from "glob";
-import path, { basename } from "path";
+import path from "path";
```
alvrs, line 7: looks like we don't need this anymore
```suggestion
```

packages/cli/src/commands/worldgen.ts
```diff
@@ -4,9 +4,9 @@ import { StoreConfig } from "@latticexyz/store";
...
 import { rmSync } from "fs";
+import { getExistingContracts } from "../utils/getExistingContracts";
```
alvrs, line 9: nit
```suggestion
import { getExistingContracts } from "../utils";
```

packages/cli/src/commands/trace.ts
alvrs: actually we should probably default to the rpc of the foundry config, and just fall back to localhost if it doesn't exist

packages/cli/src/commands/trace.ts
```diff
@@ -0,0 +1,109 @@
...
+import { MUDError } from "@latticexyz/common/errors";
+import { cast, getRpcUrl, getSrcDirectory } from "@latticexyz/common/foundry";
+import { TableId } from "@latticexyz/utils";
+import { StoreConfig } from "@latticexyz/store";
+import { resolveWorldConfig, WorldConfig } from "@latticexyz/world";
+import { IBaseWorld } from "@latticexyz/world/types/ethers-contracts/IBaseWorld";
+import { getChainId, getExistingContracts } from "../utils";
+
+import IBaseWorldData from "@latticexyz/world/abi/IBaseWorld.sol/IBaseWorld.json" assert { type: "json" };
+
```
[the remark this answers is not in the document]
alvrs: i'm not sure why spacing is like this in `deploy.ts` either haha, if we put spacing between imports i'd expect to have a group for external dependencies, _maybe_ a separate one for mud dependencies, and _maybe_ a third one for local imports. So in this case i'd expect the `IBaseWorldData` import to be in the same group as the other mud imports, and then maybe a space between that group and `../utils`, but no strong opinion about that.

PR #1000, opened 2023-06-07
build: update pnpm

dk1a:
I just updated pnpm and recreated all the lockfiles
templates are unchanged because they already have 6.1 lockfileVersion apparently


alvrs (approved):

holic:
ooc what version did you upgrade to? we should make sure we're all on the same version so the lockfile doesn't keep jumping around

dk1a:
> ooc what version did you upgrade to? we should make sure we're all on the same version so the lockfile doesn't keep jumping around

8.6.1 (latest)

PR #1001, opened 2023-06-07
fix(common): int keys should encode like `abi.encode`

holic:
closes #982

- adjusts our key encoding codegen to encode ints like `abi.encode`
- ~~small tweak to codegen to 1) not wrap in parens if we don't need to and 2) not type cast `bytes32` if it's already there~~ (gonna move this out to not pollute the diff)
- adds tests to assert our key encoding is as we expect and follows `abi.encode` for the same types


alvrs (approved):

dk1a (approved):
I agree matching abi.encode is better, and the renderer func is a bit simpler too,
this was more a mistake than an intentional tradeoff on my part

PR #1007, opened 2023-06-08
WIP global hooks exploration

dk1a:
https://www.notion.so/latticexyz/MUD-Plugins-0941930781384485abb56f64f8d5e33d


dk1a:
we decided on local in #1025

PR #1008, opened 2023-06-09
fix: updated template allow_paths

### LINKED ISSUE #988: update template `allow_paths` with mud repo location
https://github.com/latticexyz/mud/blob/5e6cc0f5e39caa23c7751398ca8af0afb02f4652/templates/vanilla/packages/contracts/foundry.toml#L11

For all templates, this can be updated to just
```
allow_paths = ["../../../mud/packages"]
```
(which means mud monorepo would be a sibling dir of the template/project dir)

Tuhinm2002:
fixes #988

 `allow_paths = ["../../node_modules", "../../../../packages"]`
chaged to
`allow_paths = ["../../../mud/packages"]`

### DIFF
--- a/templates/vanilla/packages/contracts/foundry.toml
+++ b/templates/vanilla/packages/contracts/foundry.toml
@@ -8,7 +8,7 @@ verbosity = 2
 src = "src"
 test = "test"
 out = "out"
-allow_paths = ["../../node_modules", "../../../../packages"]
+allow_paths = ["../../../mud/packages"]
 extra_output_files = [
   "abi",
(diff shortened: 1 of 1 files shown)


alvrs:
thank you! while you're here, could you update this in the other templates as well?

alvrs:
also it seems like we do need the `node_modules` path here: https://github.com/latticexyz/mud/actions/runs/5218627855/jobs/9421909757?pr=1008#step:5:264

Tuhinm2002:
@alvrs ok

PR #1011, opened 2023-06-09
test: add MODE integration test

alvrs:
* refactor some of the integration test setup
* add initial setup for running integration tests with MODE

Followup:
* Add more tests both for syncing from RPC and MODE. Currently we're mostly testing that the initial connection works and a single table can be synced, but we should add test cases for all kinds of different key and value schema types


.github/workflows/integration-sandbox.yml
```diff
@@ -10,8 +10,28 @@ jobs:
   main:
     name: Build/test integration sandbox
     runs-on: ubuntu-latest
-
+    services:
+      postgres:
+        image: ghcr.io/latticexyz/postgres-wal-logical:latest
```
alvrs, line 15: need a custom postgres image here that adds this to the postgres config:
```
wal_level = logical
max_replication_slots = 1
max_wal_senders = 1
```

postgres needs to restart after changing these config params, so the only way i found to make it work in the CI was to create a new docker image that has this config from the start

holic (approved):
👏

PR #1020, opened 2023-06-12
test: bump vitest

holic:
pulled out of https://github.com/latticexyz/mud/pull/995

### DIFF
--- a/integration-sandbox/pnpm-lock.yaml
+++ b/integration-sandbox/pnpm-lock.yaml
@@ -1,4 +1,4 @@
-lockfileVersion: '6.0'
+lockfileVersion: '6.1'
 
 settings:
   autoInstallPeers: true
--- a/integration-sandbox/pnpm-lock.yaml
+++ b/integration-sandbox/pnpm-lock.yaml
@@ -1,4 +1,4 @@
-lockfileVersion: '6.1'
+lockfileVersion: '6.0'
 
 settings:
   autoInstallPeers: true
@@ -90,8 +90,8 @@ importers:
         specifier: ^4.2.1
         version: 4.3.5(@types/node@20.1.3)
       vitest:
-        specifier: ^0.31.0
-        version: 0.31.0(jsdom@22.0.0)
+        specifier: 0.31.4
+        version: 0.31.4(jsdom@22.0.0)
 
   packages/contracts:
     devDependencies:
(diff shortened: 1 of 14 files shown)


pnpm-lock.yaml
```diff
@@ -1,4 +1,4 @@
-lockfileVersion: '6.1'
+lockfileVersion: '6.0'
```
holic, line 1: @dk1a did you do anything in particular to force the lockfile version? when I try to `pnpm install` or `pnpm install -- force`, I still get the old version, even though I'm on pnpm 8.6.2

holic: even clearing out `node_modules` and reinstalling doesn't seem to help

dk1a: no, pnpm sometimes suggests upgrading itself, so I just did `pnpm add -g pnpm`,
I actually had a reverse problem where I couldn't really get '6.0'
you could try removing pnpm and installing it again I guess

@alvrs @roninjin10 curious about your experience with lockfileVersion

roninjin10: Npm uninstall and reinstall pnpm is another thing to try.   Also if you use nvm or anything where you have more than one node version might need to check specific node versions global store

holic: I tried `pnpm store prune`, removing `pnpm-lock.yaml` and reinstalling, uninstalling `pnpm` and reinstalling, and still no luck.

integration-sandbox/pnpm-lock.yaml
```diff
@@ -1,4 +1,4 @@
-lockfileVersion: '6.1'
+lockfileVersion: '6.0'
```
holic, line 1:
```suggestion
lockfileVersion: '6.1'
```

pnpm-lock.yaml
holic, line 1:
```suggestion
lockfileVersion: '6.1'
```

alvrs (approved):

pnpm-lock.yaml
holic: turns out it was this: 
image

roninjin10: Ah glad it was something that actually makes sense.   I was about to quit using pnpm when you mentioned reinstalling pnpm doesn't fix it.  

As an aside it's interesting how often pnpm changes their schema version compared to yarn which never changes it

PR #1021, opened 2023-06-12
Upgrade rxjs

curtischong:
upgrade rxjs so it is the same version as the one in voxel-aw. This fixes sooooo many type errors and makes curtis happy.


curtischong:
wrong repo lol

PR #1022, opened 2023-06-12
build(store,world): check in ABIs and types

### LINKED ISSUE #1017: check in abi, types
By checking in ABI files, types, etc., they are immediately usable by other dependencies (i.e. network) without a build step. We can use CI to enforce that they're up-to-date.

Draft PR: https://github.com/latticexyz/mud/pull/448

holic:
closes #1017 
replaces #448 

I want to import/use these as a source of truth for store event ABIs in protocol parsing (#995)


alvrs:
can we add an action to verify these are up to date?

alvrs (approved):

PR #1029, opened 2023-06-13
refactor: move protocol schema type stuff into schema-type, deprecate old stuff

holic:
follows #995 

I moved older `schema-type` things under `/deprecated` to indicate we should migrate away them, and most will get ~deleted once we have new networking code to replace what's in `network` package. These are all leaning on an enum intermediary when we can just use `AbiType`s directly.

One thing I noticed is that `common` is becoming a catch-all and at risk of creating a circular dependency. I can't, for example, move `TupleSplit` into a type util under `common` because `common` already imports `schema-type` for codegen. We should probably consider moving out codegen, etc. into its own package?

One follow-up that might be nice to do is add a test to make sure our ABI type -> primitive types match abitype: https://abitype.dev/api/utilities.html#abitypetoprimitivetype


packages/schema-type/src/typescript/schemaAbiTypes.ts
```diff
@@ -1,5 +1,5 @@
 import { AbiType } from "abitype";
-import { TupleSplit } from "./utils";
+import { TupleSplit } from "../../../protocol-parser/src/utils";
```
alvrs, line 2: seems like we need to move TupleSplit here as well to avoid a circular dependency?

alvrs (approved):

packages/schema-type/src/typescript/schemaAbiTypes.ts
holic: whoops, yep! thanks

PR #1034, opened 2023-06-15
feat(store): add TS store events/ABI, move around exports

holic:
(pulled out of #1033)

Moves codegen into its own build/export, because we can't import from `@latticexyz/store` in the browser due to a node-only dependency, and I want to colocate TS definition of store events near the Solidity where they're defined.

Once codegen got moved out, it felt a little weird that config was in "library" still, so I moved that out too, and then there wasn't much left, so I just got rid of the notion of "library" (which won't make much sense anyway once #1025 is done and "register" is removed)


packages/store/ts/storeEvents.ts
```diff
@@ -0,0 +1,6 @@
+export const storeEvents = [
+  "event StoreDeleteRecord(bytes32 table, bytes32[] key)",
+  "event StoreSetField(bytes32 table, bytes32[] key, uint8 schemaIndex, bytes data)",
+  "event StoreSetRecord(bytes32 table, bytes32[] key, bytes data)",
+  "event StoreEphemeralRecord(bytes32 table, bytes32[] key, bytes data)",
+] as const;
```
holic, line 6: this is the main thing I wanted to import from another package into the browser bundle

packages/store/ts/storeEventsAbi.test.ts
```diff
@@ -0,0 +1,27 @@
...
+        inputs: [
+          ...item.inputs.map((input) => ({
+            name: input.name,
+            type: input.type,
+          })),
+        ],
+      }))
+      .sort((a, b) => a.name.localeCompare(b.name));
+
+    const sortedStoreEventsAbi = storeEventsAbi.slice().sort((a, b) => a.name.localeCompare(b.name));
```
dk1a, line 24: `slice()` feels unintuitive as a way to copy an array, `[...storeEventsAbi]` would be immediately clear, to me at least

dk1a (approved):
nice restructure, I prefer this regardless of #1025

PR #1036, opened 2023-06-15
feat(ci): add changesets

roninjin10:
- Add changesets
- Changesets will open a pr if you merge any changesets that will keep updating as you merge more
- When you merge that pr it will trigger a release
- A github action can allow you to preview a release. This is similar to canary releases

- Start by copy pasting from https://github.com/evmts/evmts-monorepo

- TODO Can/should we remove lerna?
- TODO this has not been tested yet at all
  - kinda hard to test tbh other than merging and just trying it out and fixing issues as they arise
  - current setup working in evmts repo though


roninjin10:
* main
  * **PR #1036** <a href="https://app.graphite.dev/github/pr/latticexyz/mud/1036" target="_blank">Graphite</a>  👈
    * **PR #1037** <a href="https://app.graphite.dev/github/pr/latticexyz/mud/1037" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/latticexyz/mud/1036?utm_source=stack-comment).

.changeset/config.json
```diff
@@ -0,0 +1,12 @@
+{
+  "$schema": "https://unpkg.com/@changesets/config@2.3.0/schema.json",
+  "changelog": ["@changesets/changelog-github", { "repo": "latticexyz/mud" }],
+  "commit": false,
+  "fixed": [],
+  "linked": [],
+  "access": "public",
+  "baseBranch": "main",
+  "updateInternalDependencies": "patch",
+  "ignore": []
```
roninjin10, line 10: If there are any packages included in the pnpm-workspace.yml at the root we don't want to publish we should include them here.

If there are any packages NOT included in the pnpm-workspace.yml that we do want to publish I'm not sure immediately how other than publishing it seperately as a one off or adding it to workspace.yml

.changeset/config.json
```diff
@@ -0,0 +1,12 @@
+{
+  "$schema": "https://unpkg.com/@changesets/config@2.3.0/schema.json",
+  "changelog": ["@changesets/changelog-github", { "repo": "latticexyz/mud" }],
```
roninjin10, line 3: I honestly have no idea what the difference is between this and the default changelog but wagmi used it and they are goated so copy pasta ftw

roninjin10: I think maybe the default publishes to packages/my-package/CHANGELOG instead of github?

package.json
```diff
@@ -8,24 +8,22 @@
     "url": "https://github.com/latticexyz/mud.git"
   },
   "scripts": {
+    "// TODO make this use turob instead?  an we get rid of lerna?  link:packages": "lerna exec 'pnpm link --global'",
```
alvrs, line 11: no need for this anymore since we have `mud set-version --link path/to/mud` to link to a local version of MUD now

package.json
```diff
@@ -8,24 +8,22 @@
...
+    "// TODO what is this one?   Can we get rid of lerna?  entry:dist": "lerna run prepack",
+    "// TODO what is this one?   Can we get rid of lerna?  entry:src": "lerna run postpack",
+    "// TODO what is this one?release:yalc": "(pnpm entry:dist && lerna exec npx yalc push); pnpm entry:src",
```
alvrs, line 14: these are all out of date and can be removed

package.json
```diff
@@ -34,6 +32,10 @@
     "*.{ts,tsx,css,md,mdx,sol}": "prettier --write",
     "package.json": "pnpm sort-package-json"
   },
+  "dependencies": {
+    "@changesets/changelog-github": "^0.4.8",
+    "@changesets/cli": "^2.26.1"
+  },
```
alvrs, line 38: should these be dependencies or devDependencies?

package.json
```diff
@@ -8,24 +8,22 @@
...
     "lint": "pnpm prettier:check && eslint . --ext .ts --ext .tsx",
     "prepare": "husky install && (forge --version || pnpm foundryup)",
     "prettier": "prettier --write '**/*.{ts,tsx,css,md,mdx,sol}'",
     "prettier:check": "prettier --check '**/*.{ts,tsx,css,md,sol}'",
-    "release": "pnpm install && lerna publish --no-private --force-publish",
-    "release:canary": "lerna publish premajor --preid alpha.1 --canary --no-private --force-publish",
-    "release:ci": "lerna publish --no-private --force-publish --yes",
-    "release:manual": "pnpm --recursive --no-bail run release",
-    "release:yalc": "(pnpm entry:dist && lerna exec npx yalc push); pnpm entry:src",
+    "release": "pnpm install && changeset publish",
```
alvrs, line 25: do we need to run `pnpm build` here too (to create the dist files that are not on github but should be on npm) or does changeset do that internally by default?

roninjin10: Yes meant to add it here.   can do it in the GitHub action too with ‘with: build’ but feels better here just in case this gets ran manually

.changeset/config.json
alvrs: the package doesn't have a readme but the google preview makes it sounds like this makes the changelog include links to github commits, PRs and users, which sounds useful
CleanShot 2023-06-20 at 10 46 06

.changeset/hungry-fireants-buy.md
alvrs, line 1: should we remove this file before merging?

.changeset/pre.json
```diff
@@ -0,0 +1,30 @@
+{
+  "mode": "pre",
+  "tag": "next",
+  "initialVersions": {
+    "@latticexyz/cli": "1.42.0",
```
alvrs, line 5: is this the version the prerelease will be prefixed with? if so we should change it to 2.0.0

.github/workflows/prerelease-exit.yml
```diff
@@ -0,0 +1,43 @@
...
+        uses: ./.github/actions/setup
+        
+      - name: Generate gas reports
+        run: pnpm gas-report
+
+      - name: Outdated files, run `pnpm gas-report` and commit them
+        uses: ./.github/actions/require-empty-diff
+
+      - name: Remove pre.json
+        run: npx rimraf .changeset/pre.json
```
alvrs, line 36: what's the difference between removing the `pre.json` file to running `pnpm changeset pre exit` (from https://github.com/changesets/changesets/blob/main/docs/prereleases.md)?

.github/workflows/prerelease.yml
```diff
@@ -0,0 +1,62 @@
...
+          submodules: recursive
+
+      - name: "Setup"
+        uses: ./.github/actions/setup
+        
+      - name: Generate gas reports
+        run: pnpm gas-report
+
+      - name: Outdated files, run `pnpm gas-report` and commit them
+        uses: ./.github/actions/require-empty-diff
```
alvrs, line 36: i think we can skip this for the release action (since this only runs on main, where checks should already have passed in the PRs that got merged into main, and even if the gas reports are out of date it's not blocking for the release)

package.json
```diff
@@ -8,24 +8,22 @@
...
-    "entry:src": "lerna run postpack",
     "foundryup": "curl -L https://foundry.paradigm.xyz | bash && bash ~/.foundry/bin/foundryup",
     "gas-report": "pnpm recursive run gas-report",
-    "link:packages": "lerna exec 'pnpm link --global'",
     "lint": "pnpm prettier:check && eslint . --ext .ts --ext .tsx",
     "prepare": "husky install && (forge --version || pnpm foundryup)",
     "prettier": "prettier --write '**/*.{ts,tsx,css,md,mdx,sol}'",
     "prettier:check": "prettier --check '**/*.{ts,tsx,css,md,sol}'",
-    "release": "pnpm install && lerna publish --no-private --force-publish",
-    "release:canary": "lerna publish premajor --preid alpha.1 --canary --no-private --force-publish",
```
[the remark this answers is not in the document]
alvrs: i think we can get rid of the canary releases if we have prereleases (i'd prefer a regular prerelease that includes a changelog over a canary release with every push to main, and this also allows us to get rid of lerna)

alvrs:
Okay so let me try to summarize the approach to check my understanding:
- Merging to `main` will always enter pre release mode, unless we are already in pre release mode or the previous commit was explicitly exiting pre release mode
  - Merging a normal PR will open/update the changesets PR
  - Merging a changesets PR will create a prerelease (and update the changelog?)
- Dispatching the exit pre release mode action will exit the pre release mode
- Merging from `main` to `v1` (this should probably later be `v2` or `stable`) will create a regular release if we're not in pre release mode

If we're going with this two branch model, could we just always be in pre release mode on main, and always be in regular release mode on the `stable` branch (by exiting the pre release mode in the action that is dispatched when merging to the `stable` branch)

roninjin10:
> Merging to main will always enter pre release mode, unless we are already in pre release mode or the previous commit was explicitly exiting pre release mode

Not quite

#### Repo is in normal mode
Normal mode means you don't have a prerelease

- Merging to main will make a versioning pr compiling all the changesets moving the repo into preset mode
- Until you merge this pr main is still in normal mode 
- In normal mode you can easily keep making patches to the release branch from main still without needing to cherry-pick manually publish
- As prs keep being added this pr will keep taking on commits
- When you choose to release, you enter prerelease mode via merging

#### Repo is in prerelease mode

- Merging to main will make a versioning pr just like before
- When you merge this versioning pr it will make another prerelease but increasing the nonce by 1 each time

#### Time to actually publish

- Running `changeset pre exit` will exit prerelease mode
- For convenience there is a github action to do it
- Once out of prerelease mode you can now publish normally do a normal changeset flow

> could we just always be in pre release mode on main

This actually is exactly how it works. At least this is how it works based on the intention of what you meant.  

By "always be in pre release mode" if I reframe that as "every publish from main is a prerelease" this is exactly what is happening here. 

The thing to consider though, I made it so every release is intentional. You always merge a pr to make a release. Now, the first release is special though. When one is in prerelease mode this means they already have a prerelease published. So to literally always be in prerelease means we have already published your first prerelease.  This means we published prerelease with nonce 0 with whatever the code from the latest stable release was. That is not what you want.

A simple way to think about it is not being in prerelease mode just happens to be the time between publishing your latest stable and publishing the first (nonce 0) prerelease of the next release.

I do think you want to make every release intentional. It's an extra step (you gotta intentionally merge a pr to create any prerelease or any release) but I believe it's worth it.

#### Random aside

One may ask why not just have this all happen on one branch.  Basically this was in the docs and it makes sense

```
Warning: If you decide to do prereleases from the main branch of your repository, without having a branch for your last stable release without the prerelease changes, you will block other changes until you are ready to exit prerelease mode. We thoroughly recommend only running prereleases from a branch other than the main branch.
```

.changeset/hungry-fireants-buy.md
roninjin10: yea definitely should put a do-not-merge on this too

.changeset/config.json
roninjin10: Oh nice! I think I was confused because I thought the default did that but looking now it doesn't.

.changeset/hungry-fireants-buy.md
roninjin10: yes also the pre.json is gone after we remove this too.  pre.json is generated from this file via the github action. I generated it manually here to test it and didn't mean to git add it actually

.changeset/pre.json
roninjin10: Yes this generated file will be the correct version via adding major changeset versions to the packages

.changeset/README.md
```diff
@@ -0,0 +1,8 @@
+# Changesets
+
+Hello and welcome! This folder has been automatically generated by `@changesets/cli`, a build tool that works
+with multi-package repos, or single-package repos to help you version and publish your code. You can
+find the full documentation for it [in our repository](https://github.com/changesets/changesets)
+
+We have a quick list of common questions to get you started engaging with this project in
+[our documentation](https://github.com/changesets/changesets/blob/main/docs/common-questions.md)
```
roninjin10, line 8: We should update this to explain how this release process works before landing

.changeset/config.json
```diff
@@ -0,0 +1,12 @@
...
+  "commit": false,
+  "fixed": [],
+  "linked": [],
+  "access": "public",
+  "baseBranch": "main",
+  "updateInternalDependencies": "patch",
+  "ignore": []
+}
```
roninjin10, line 11: see wagmi config they use some experimental options that I didn't use here but maybe we want?

.changeset/hungry-fireants-buy.md
```diff
@@ -0,0 +1,27 @@
+---
```
roninjin10, line 1: remove me

.changeset/pre.json
roninjin10: remove me

.github/workflows/canary.yml
```diff
@@ -0,0 +1,40 @@
+name: Release
+
+# Releases a canary release when new commits merge to main
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
```
roninjin10, line 8:
```suggestion
```
Don't auto publish canary.

We should include this as a workflow_dispatch because needing to publish something to test is something that devs need to be able to do without interupting the normal release process

.github/workflows/canary.yml
```diff
@@ -0,0 +1,40 @@
...
+
+jobs:
+  release-canary:
+    name: Publish canary release to npm
+    runs-on: ubuntu-latest
+    if: github.event_name != 'workflow_dispatch'
```
roninjin10, line 14:
```suggestion
```
This is copy pasta bug

.github/workflows/prerelease-exit.yml
roninjin10: this is a bug I forgot to change this from the original action I copy pastad

roninjin10: We should use the action the one I copy pastad was not clean and doing this for pretty hacky reasons

.github/workflows/prerelease-exit.yml
```diff
@@ -0,0 +1,43 @@
...
+
+      - uses: stefanzweifel/git-auto-commit-action@v4
+        with:
+          commit_message: Exit prerelease mode
+          # Commit these changes to the branch workflow is running against
+          branch: ${{ github.event.inputs.branch }}
```
roninjin10, line 42: This is likely cleaner if we made this a pull request but don't have a strong preference either way.  The pull request could be automatically opened if it detects a pre.json instead of an action like here. THen we could delete this action file completely

.github/workflows/prerelease.yml
```diff
@@ -0,0 +1,62 @@
...
+        uses: andstor/file-existence-action@v2.0.0
+        with:
+          files: ".changeset/pre.json"
+
+      - name: Enter next prerelease mode
+        # If .changeset/pre.json does not exist and we did not recently exit
+        # prerelease mode, enter prerelease mode with tag next
+        if: steps.check_files.outputs.files_exists == 'false' && !contains(github.event.head_commit.message, 'Exit prerelease')
+        run: npx changeset pre enter next
+
```
roninjin10, line 54: Note this will end up getting thrown away if the next action does not detect a changeset and make a pr

roninjin10: It can be confusing because at first it can look like it happens directly to the main branch but no commit is made

roninjin10: we should comment more heavily

.github/workflows/release.yml
```diff
@@ -1,73 +1,57 @@
...
 
-# Releases a canary release when new commits merge to main
-# Releases an official release when workflow dispactch is
-# manually triggered
 on:
-  workflow_dispatch:
   push:
     branches:
-      - main
+      - v1
```
roninjin10, line 6: This should not be v1 branch.  Do we suspect we will be making any more prs to v1 branch?  If so we can just use the old publishing workflow for v1

.github/workflows/release.yml
```diff
@@ -1,73 +1,57 @@
...
-    if: github.event_name == 'workflow_dispatch'
+    permissions:
+      contents: write
+      pull-requests: write
+      id-token: write
     steps:
-      - name: Checkout
-        uses: actions/checkout@v3
+      - uses: actions/checkout@v3
+        name: Checkout Repo
```
roninjin10, line 21: This changed from copy pasta didn't need to change

package.json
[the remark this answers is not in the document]
roninjin10: I don't have a strong preference here but let me give you my reasoning why I included a canary release.  Basically it's the idea of continous delivery. My default is try to publish (or dry-run publish) every single pr so there is confidence the deployment process is working on all changes. I could not figure out a way to do a dry run with changesets which is why I included this.

Though this is a general principle I follow by default, changesets are pretty bullet proof so I don't think it actually matters in this case. Especially since I expect ya'll will not be waterfalling releases but publishing prereleases often.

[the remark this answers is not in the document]
roninjin10: One more piece of context, I shouldn't call this canary because that's not what they really are.  So a prerelease if you install `pnpm install @lattice-xyz/cli@*` or @next it will install the latest prerelease even if one of these previews got published.  `pnpm install @lattice-xyz/cli` or `@latest` will install only the latest stable release.  This release is kinda a one off without an actual tag I believe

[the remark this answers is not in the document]
roninjin10: Looks like changesets calls them snapshot releases not canary releases.  https://github.com/changesets/changesets/blob/main/docs/snapshot-releases.md.

Should name them snapshots for consistency if we add them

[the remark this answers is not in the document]
alvrs: ohh okay gotscha! yeah so if these snapshots are just a way to explicitly install a commit version, but `@next` going to the latest prerelease and `@latest` going to the stable release, then I agree with this approach!

.github/workflows/release.yml
alvrs: don't suspect to be making any more PRs to the v1 branch!

.changeset/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+## How changesets work in this repo
```
roninjin10, line 10: copied this from the pr comment can clean it up if necessary

.github/workflows/prerelease-exit.yml
roninjin10: Pull request thinking about this now makes it more confusing since we would have pull requests doing all types of things

package.json
```diff
@@ -34,13 +29,16 @@
...
   "devDependencies": {
+    "@changesets/changelog-github": "^0.4.8",
+    "@changesets/cli": "^2.26.1",
     "@typescript-eslint/eslint-plugin": "5.46.1",
     "@typescript-eslint/parser": "5.46.1",
     "chalk": "^5.2.0",
     "eslint": "8.29.0",
     "husky": ">=6",
-    "lerna": "^6.6.1",
+    "jq-cli-wrapper": "^1.6.1",
```
roninjin10, line 41:
```suggestion
```
Looks like a bad rebase

package.json
```diff
@@ -34,13 +29,15 @@
     "*.{ts,tsx,css,md,mdx,sol}": "prettier --write",
     "package.json": "pnpm sort-package-json"
   },
+  "dependencies": {},
```
roninjin10, line 32:
```suggestion
```

package.json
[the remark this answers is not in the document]
roninjin10: Oops I just nuked the canary action

.github/workflows/release.yml
```diff
@@ -1,60 +1,30 @@
...
-  workflow_dispatch:
   push:
     branches:
-      - main
+      - v1
+
+concurrency: ${{ github.workflow }}-${{ github.ref }}
 
 jobs:
   release:
```
roninjin10, line 11:
```suggestion
  version:
```

package.json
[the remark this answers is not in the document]
roninjin10: brought it back

.github/workflows/release.yml
```diff
@@ -1,60 +1,30 @@
...
-    if: github.event_name != 'workflow_dispatch'
+    permissions:
+      contents: write
+      pull-requests: write
+      id-token: write
     steps:
-      - name: Checkout
-        uses: actions/checkout@v3
+      - uses: actions/checkout@v3
+        name: Checkout Repo
```
roninjin10, line 21:
```suggestion
      - name: Checkout
        uses: actions/checkout@v3
```

.github/workflows/prerelease.yml
roninjin10: added comments to the .changeset readme

.github/workflows/prerelease-exit.yml
alvrs: agree with making it an action to avoid cluttering the PR

.github/workflows/prerelease.yml
```diff
@@ -0,0 +1,56 @@
...
+        id: check_files
+        uses: andstor/file-existence-action@v2.0.0
+        with:
+          files: ".changeset/pre.json"
+
+      - name: Enter next prerelease mode
+        # If .changeset/pre.json does not exist and we did not recently exit
+        # prerelease mode, enter prerelease mode with tag next
+        if: steps.check_files.outputs.files_exists == 'false' && !contains(github.event.head_commit.message, 'Exit prerelease')
+        run: npx changeset pre enter next
```
alvrs, line 47: do we need to commit after entering pre release mode like when exiting pre release mode? or does this automatically happen with the next step (release)?

.github/workflows/release.yml
```diff
@@ -67,10 +37,26 @@ jobs:
...
         run: npm config set '//registry.npmjs.org/:_authToken' "${NPM_TOKEN}"
+
+      # https://github.com/changesets/action#with-publishing
+      - name: Handle Release Pull Request or Publish to npm
+        id: changesets
+        if: steps.check_files.outputs.files_exists == 'false'
+        uses: changesets/action@v1
+        with:
+          title: "chore: Version packages 🔖"
+          commit: "chore: Version packages 🔖"
```
alvrs, line 56:
```suggestion
          title: "chore: version packages 🔖"
          commit: "chore: version packages 🔖"
```

ocd requires me to lowercase this

alvrs (approved):

PR #1039, opened 2023-06-15
fix(cli): temporary workaround to reset anvil base fee to 0

alvrs:
Anvil is currently increasing the base fee to 1 after a MUD deploy, this temporary workaround is resetting the base fee to 0 after a deployment. We can remove this workaround once the issue is fixed upstream, see https://github.com/foundry-rs/foundry/issues/5161.


holic (approved):
🙏

PR #1041, opened 2023-06-16
ci: remove pinned foundry version, use nightly instead

alvrs:
the issue should be fixed in foundry (https://github.com/foundry-rs/foundry/pull/5163), so it should be fine to remove the pinned version and use the latest nightly


alvrs:
merging this to unblock https://github.com/latticexyz/mud/pull/691

PR #1045, opened 2023-06-16
ci: clean rebuild in build action, use turbo cache in all other actions

alvrs:
* In #1027 we added a `clean` step before the `build` step. The `clean` step removes build artifacts, so afterwards a full rebuild is necessary. This caused an issue if the turbo cache was hit and the rebuild was skipped, eg https://github.com/latticexyz/mud/actions/runs/5293787981/jobs/9582365505?pr=1011
* This PR changes pulls out the `build` step from the the `Setup` composite action and instead creates a new `Build` composite action that uses the turbo cache and does not clean the artifacts
* This new `Build` composite action is used in all actions, except the `Build and validate artifacts` action, which first cleans the artifacts and then performs a full rebuild without the turbo cache.

### DIFF
--- a/.github/actions/build/action.yml
+++ b/.github/actions/build/action.yml
@@ -1,0 +1,21 @@
+name: Build
+description: Build with turbo cache
+runs:
+  using: composite
+  steps:
+    - name: Cache turbo build
+      uses: actions/cache@v3
+      with:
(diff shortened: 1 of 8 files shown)


alvrs:
self-approving to unblock #1011

PR #1047, opened 2023-06-17
build: fix turbo clean

holic:
reverts #1045 #1046 in favor of turbo clean

not sure how to repro the original issue to verify this approach works

if this doesn't work, another possible option is to have the mud codegen step (`build:mud`) do a `rimraf src/codegen` before running, so its part of the turbo build cache


alvrs:
Conceptually, if we clear the artifacts in the setup step, wouldn't we always have to rebuild, so we could never benefit from the turbo cache in CI? If so, I feel like it would be sufficient to do a full clean+rebuild in one action (eg. "Build and validate artifacts") and use the cache in all other actions

holic:
> Conceptually, if we clear the artifacts in the setup step, wouldn't we always have to rebuild, so we could never benefit from the turbo cache in CI? If so, I feel like it would be sufficient to do a full clean+rebuild in one action (eg. "Build and validate artifacts") and use the cache in all other actions

My understanding when running this locally is that the build is still cached. I can run `pnpm clean` and `pnpm build` and the 2nd or 3rd time the build is run, it'll be fully cached. I think the difference is it starts the build from a clean output, so the cached output should be a little more accurate/what is intended? I can't say for sure without a repro case.

I can also see it being useful to have an uncached build in CI to ensure we get fully accurate results, and it should probably be tied to the same job that does releases so we can ensure a clean and complete set of artifacts for npm, etc.?

(I don't feel strongly about any of this, just thought I'd take a stab at trying to teach turbo a bit more about our intentions to see what the results were.)

holic:
Another thing to note about the composite steps is that the get combined into a single entry in the CI logs, so it's a bit hard to figure out how much time is spent on the "sub-steps" of each composite action.

holic:
after #1057 is merged, I could potentially just use this PR as a revert for adding the "clean" steps in CI, since we shouldn't really need em anymore (pretty much all builds are either gitignored or clean themselves)

PR #1048, opened 2023-06-17
feat(mode): more cleanup

authcall:
among other things
- update block number while syncing
- do not build unnecessarily large indexes


packages/services/pkg/mode/common.go
```diff
@@ -20,16 +20,16 @@ func FieldToString(field *pb_mode.Field) string {
 // Namespace returns the namespace string for the specified chain ID and world address.
 //
 // Parameters:
-// - chainId (string): The chain ID to include in the namespace.
+// - chainID (string): The chain ID to include in the namespace.
```
alvrs, line 23: why the change from `chainId` to `chainID`? `ID` vs `Id` in camel case code is an endless discussion, but it seems like major libraries (ethers, viem) converged on `chainId`, so I'd prefer us to use that too

authcall: google's Go style uses ID and so is used as a sort of reference. some more references here https://google.github.io/styleguide/go/decisions. I don't have strong preferences though

packages/services/pkg/mode/storecore/encoding.go
```diff
@@ -743,52 +748,60 @@ func (schemaType SchemaType) ToSolidityType() string {
...
 func (schemaType SchemaType) ToPostgresType() string {
+	const text = "text"
+	const jsonb = "jsonb"
+	const boolean = "boolean"
+
+	//nolint:gocritic,nestif // no switch, simplify
 	if (schemaType >= UINT8 && schemaType <= UINT32) || (schemaType >= INT8 && schemaType <= INT32) {
 		// Integer.
-		return "integer"
+		return text
```
ludns, line 770: does this mean we can't query with numbers anymore?
If I want to query for an "area" (ie: X >= X_START & X < X_END & Y >= Y_START & Y < Y_END), does this mean this change would make this query impossible (or impossibly slow)?

alvrs (approved):

Issue #667, opened 2023-04-24
move gas report out of cli?


holic:
see https://github.com/latticexyz/mud/pull/633#discussion_r1173166556

PR #1049, opened 2023-06-19
feat(cli): replace gas report

holic:
Replaces our slow and error prone approach of generating new test files from magic comments with a Solidity-native approach, inspired by the approach in https://github.com/nomoixyz/vulcan/pull/104.

This hopefully means we can re-open https://github.com/latticexyz/mud/pull/771 and easier to do #667 (since gas report is now ~just a test output parser)


.github/workflows/test.yml
```diff
@@ -23,6 +23,12 @@ jobs:
       - name: Run tests
         run: pnpm test
 
+      - name: Generate gas reports
+        run: pnpm gas-report
```
holic, line 27: This is moved here since the gas report runs best if the tests are already compiled/cached.

alvrs (approved):
lgtm!

PR #1060, opened 2023-06-20
refactor(e2e): rename integration-sandbox to e2e

alvrs:
Normally i would advocate to avoid abbreviations, but `e2e` is such a common alias for "end-to-end-tests" that I think it's fine here. (Googling for `e2e` shows better results that googling for `end to end`)


holic (dismissed):
lgtm!

One small thing: can we rename the `test` package within this dir to something more meaningful? Like `playwrite-tests` or `mode-tests` or whatever the intention is.

dk1a (dismissed):

alvrs:
> One small thing: can we rename the test package within this dir to something more meaningful? Like playwrite-tests or mode-tests or whatever the intention is.

How about `sync-test`? The intention for this is to make sure syncing data from chain to client works on via all available methods

holic:
> How about `sync-test`? The intention for this is to make sure syncing data from chain to client works on via all available methods

works for me!

PR #1061, opened 2023-06-20
refactor(store,world): replace isStore with storeAddress

dk1a:
fixes https://github.com/latticexyz/mud/issues/940
the previous approach used the absence of `isStore` to indicate that the contract isn't the Store
I change it so all store consumers ~~need `storeAddress` method, and its absence is an error~~ use a storage variable to detect storeAddress
Moved MudV2Test (otherwise circular dependency) and took the opportunity to rename it to MudTest


packages/store/src/StoreSwitch.sol
```diff
@@ -1,42 +1,41 @@
...
 
-    // Check whether this contract implements the IStore interface
-    try IStore(address(this)).isStore() {
-      success = true;
-    } catch {
-      success = false;
+    try IStoreConsumer(address(this)).storeAddress(msg.sender) returns (address _storeAddress) {
+      return _storeAddress;
+    } catch (bytes memory lowLevelData) {
+      // catch and rename the error, otherwise it's usually "EvmError: Revert" which is very unhelpful
```
holic, line 34:
```suggestion
      // catch and rename the error, otherwise it's usually "EvmError: Revert" which isn't very unhelpful
```

holic (commented):
looks like this upped the gas everywhere?

packages/store/src/StoreSwitch.sol
dk1a: I assume you missed "un", I can change to `isn't very helpful` if you prefer

dk1a:
> looks like this upped the gas everywhere?

pretty much. It increases StoreSwitch calls by ~300 gas (which means any table interactions); relatively it's not big, but affects everything
I don't try to optimize it here, that'd be a separate PR. I'd want to compare using a method to using storage, and maybe rely on a magic return value instead of passing `msg.sender` arg.
But the best way to do this would probably be some combo of code data location and transient storage, neither of which are out yet

packages/store/src/StoreSwitch.sol
holic: oops, yes, misread

holic:
is this to help us with the "what context am I running in" issue where tests don't work well with the previous approach?

dk1a:
> is this to help us with the "what context am I running in" issue where tests don't work well with the previous approach?

yep

packages/world/src/WorldConsumer.sol
```diff
@@ -1,11 +1,12 @@
 // SPDX-License-Identifier: MIT
```
dk1a, line 1: I think Consumer is clearer than Context (which makes me think it's something World inherits, not systems), and it matches StoreConsumer

packages/store/src/StoreSwitch.sol
```diff
@@ -1,111 +1,112 @@
...
+      revert StoreSwitch_MissingOrInvalidStoreAddressFunction(lowLevelData);
```
dk1a, line 28: an alternative to enforcing `IStoreConsumer` would be to fall back on `msg.sender` if it's absent. That might help with gas too.
But I like this explicit approach since it gives a nice error instead of confusing behavior down the line, making things hopefully easy to debug (or answer questions about on `mud-help`)

packages/world/src/WorldConsumer.sol
alvrs: Context was leaning on [ERC2771Context](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/metatx/ERC2771Context.sol) where this pattern originates

packages/store/src/StoreConsumer.sol
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
+
+import { IStoreConsumer } from "./IStore.sol";
+
+/**
+ * Default Store consumers consider their caller to be the Store, thus delegating access control to the Store.
+ * The Store contract considers itself to be the Store (so StoreRead has a different storeAddress definition).
+ * Tests/scripts may override this value for convenience.
+ * WARNING: World Systems rely on storeAddress being msg.sender for access control!
```
alvrs, line 10: Not sure if we should put this warning here - the Store package should be independent of the World package, and the warning itself is not detailed enough to be helpful without the full context. Maybe we should a more detailed explanation in the docs and link it somewhere in the world package.

alvrs:
I wonder where exactly the gas increase comes from. It seems like before we were calling a method on the `msg.sender`, now instead we're calling a method on `self`, why increase by 300?

I'm still a little bit confused how this works -
* `StoreSwitch` calls `storeAddress(msg.sender)` on `address(this)` -> this expects `storeAddress(address)` to be implemented in the System.
* `StoreConsumer`, which the `System` base contract inherits from, implements `storeAddress` by returning the `msgSender` input to the function.
* `MudTest` implements a `storeAddress` function that returns `worldAddress`, but when is this called? When a System is called in a test, only the system's `storeAddress` function would be called, so the final `storeAddress` would be the test contract?

dk1a:
> I wonder where exactly the gas increase comes from. It seems like before we were calling a method on the msg.sender, now instead we're calling a method on self, why increase by 300?

I think it's decoding calldata and return value and internal checks around it. Pushing to stack is very cheap but solidity does more than that

> MudTest implements a storeAddress function that returns worldAddress, but when is this called?

Only when you call a table directly within the test, it has no effect on systems

> When a System is called in a test, only the system's storeAddress function would be called, so the final storeAddress would be the test contract?

In tests a system would still be called through the world, so msg.sender would be world as usual (or `address(this)` for root systems, which would also be world)

packages/world/src/WorldConsumer.sol
holic: I feel like Context still makes sense for this?

packages/store/src/StoreConsumer.sol
dk1a: added a comment to `WorldConsumer` instead
also agree with docs, should it be in this PR or separate?

packages/world/src/modules/keyswithvalue/KeysWithValueHook.sol
```diff
@@ -20,15 +20,19 @@ import { getTargetTableSelector } from "../utils/getTargetTableSelector.sol";
...
  */
-contract KeysWithValueHook is IStoreHook {
+contract KeysWithValueHook is StoreHook {
   using ArrayLib for bytes32[];
   using ResourceSelector for bytes32;
 
+  function _world() internal view returns (IBaseWorld) {
+    return IBaseWorld(storeAddress(msg.sender));
+  }
+
```
dk1a, line 30: this is just for consistency, I'm not sure how hooks should handle this, but it wouldn't break things to just use `msg.sender` and optimize `storeAddress` away

packages/world/src/WorldConsumer.sol
alvrs: yeah i see how `Consumer` would also make sense but I'd prefer to keep `Context` to draw the analogy to `ERC2771Context`

alvrs (commented):
I'd be curious to see how this compares in gas to an approach where we read the store address from a known storage slot:
- if storage slot is unset, use `msg.sender` as store (via external call)
- if storage slot is set to `0x01`, use self as store (use StoreCore lib directly) -> World contract sets it to `0x01` so systems use StoreCore when delegatecalled
- if storage slot is set to an address, use that address as store (via external call)

I would have a slight preference for that approach since we could keep all the logic to figure out where to write to in `StoreSwitch` instead of spread out across multiple interfaces/contracts (eg `StoreConsumer`, `StoreHook`)

If the storage slot approach turns out to be more expensive, one alternative solution could be to make the `storeAddress` function optional (similar to how we currently attempt to call `isStore`) to then still keep the logic mostly contained in `StoreSwitch` (eg instead of passing in `msgSender`, use the same decision tree as described above)

examples/minimal/pnpm-lock.yaml
```diff
@@ -1,4 +1,4 @@
-lockfileVersion: '6.1'
+lockfileVersion: '6.0'
```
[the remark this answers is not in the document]
dk1a: 8.6.7

packages/store/abi/StdAssertions.sol/StdAssertions.abi.json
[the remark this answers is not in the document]
dk1a: yeah I could try putting `MudTest` into `test` instead of `src`

packages/store/src/IStore.sol
```diff
@@ -4,7 +4,17 @@ pragma solidity >=0.8.0;
...
+  // TODO: benchmark this vs. using a known storage slot to determine the Store address
+  // (see https://github.com/latticexyz/mud/issues/444)
+  /**
+   * Returns the Store address used by this consumer (like Hooks, Store itself, World Systems, or tests/scripts)
+   * for StoreSwitch to determine how to interact with the Store.
+   */
+  function storeAddress(address msgSender) external view returns (address storeAddress);
+}
+
+interface IStoreRead is IStoreConsumer {
```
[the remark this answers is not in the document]
dk1a: Store is also its own consumer, which is relevant for delegatecalled systems

packages/store/src/IStore.sol
```diff
@@ -4,7 +4,17 @@ pragma solidity >=0.8.0;
...
 import { Schema } from "./Schema.sol";
 
-interface IStoreRead {
+interface IStoreConsumer {
+  // TODO: benchmark this vs. using a known storage slot to determine the Store address
+  // (see https://github.com/latticexyz/mud/issues/444)
+  /**
+   * Returns the Store address used by this consumer (like Hooks, Store itself, World Systems, or tests/scripts)
+   * for StoreSwitch to determine how to interact with the Store.
+   */
```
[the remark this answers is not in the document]
dk1a: https://github.com/latticexyz/mud/pull/1158

packages/store/abi/StdAssertions.sol/StdAssertions.abi.json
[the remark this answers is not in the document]
dk1a: decided to skip it during `forge build` instead

dk1a:
storage slot is mostly better (see https://github.com/latticexyz/mud/pull/1158), rewrote this PR to use storage instead of a method

packages/store/src/StoreSwitch.sol
```diff
@@ -9,103 +9,120 @@ import { Schema } from "./Schema.sol";
...
+      return _storeAddress;
     }
   }
 
-  function inferStoreAddress() internal view returns (address) {
-    return isDelegateCall() ? address(this) : msg.sender;
+  /**
+   * Set the Store address for use by other StoreSwitch functions.
+   * 0x00 is a magic number for msg.sender
+   * (which means that uninitialized storeAddress is msg.sender by default)
```
alvrs, line 42:
```suggestion
   * If it stays uninitialized, StoreSwitch falls back to calling store methods on msg.sender.
```

packages/store/src/StoreCore.sol
```diff
@@ -29,6 +30,10 @@ library StoreCore {
    * (see https://github.com/latticexyz/mud/issues/444)
    */
   function initialize() internal {
+    // Use address(this) instead of msg.sender (the default) when using StoreSwitch
+    // from within the Store contract itself (e.g. during delegatecall)
```
alvrs, line 34:
```suggestion
    // StoreSwitch uses the storeAddress to decide where to write data to.
    // If StoreSwitch is called in the context of a Store contract (storeAddress == address(this)),
    // StoreSwitch uses internal methods to write data instead of external calls.
```

alvrs (commented):
This is great! Just two minor proposed changes for comments. Also let's add a changeset to describe the change.

.changeset/selfish-cycles-retire.md
```diff
@@ -0,0 +1,16 @@
+---
+"@latticexyz/std-contracts": minor
+"@latticexyz/store": minor
+"@latticexyz/world": minor
+---
+
+`MudV2Test` has been renamed to `MudTest` and moved to another package.
```
alvrs, line 7: Nit: let's use present tense for consistency

```suggestion
Rename `MudV2Test` to `MudTest` and move from `@latticexyz/std-contracts` to `@latticexyz/store`.
```

.changeset/selfish-cycles-retire.md
```diff
@@ -0,0 +1,16 @@
...
+
+```solidity
+// old import
+import { MudV2Test } from "@latticexyz/std-contracts/src/test/MudV2Test.t.sol";
+// new import
+import { MudTest } from "@latticexyz/store/src/MudTest.sol";
+```
+
+StoreSwitch has been rewritten to use a storage slot instead of `function isStore()` to determine which contract is Store.
```
alvrs, line 16:
```suggestion
Refactor `StoreSwitch` to use a storage slot instead of `function isStore()` to determine which contract is Store:

- Previously `StoreSwitch` called `isStore()` on `msg.sender` to determine if `msg.sender` is a `Store` contract. If the call succeeded, the `Store` methods were called on `msg.sender`, otherwise the data was written to the own storage.
- With this change `StoreSwitch` instead checks for an `address` in a known storage slot. If the address equals the own address, data is written to the own storage. If it is an external address, `Store` methods are called on this address. If it is unset (`address(0)`), store methods are called on `msg.sender`.
- In practice this has the same effect as before: By default the `World` contracts sets its own address in `StoreSwitch`, while `System` contracts keep the Store address undefined, so `Systems` write to their caller (`World`) if they are executed via `call` or directly to the `World` storage if they are executed via `delegatecall`.
- Besides gas savings, this change has two additional benefits:
  1. it is now possible for `Systems` to explicitly set a `Store` address to make them exclusive to that `Store` and
  2. table libraries can now be used in tests without having to provide an explicit `Store` argument, because the `MudTest` base contract redirects reads and writes to the internal `World` contract.
```

alvrs (approved):

PR #1063, opened 2023-06-21
feat(schema-type): add schemaTypeToPrimitiveType

alvrs:

PR #1064, opened 2023-06-21
test(e2e): make e2e infrastructure more extendable

alvrs:
This PR changes the end-to-end test setup to allow us to add a lot of test cases for many different schemas without having to touch test files or write custom systems.

* Adding utils to write contract data (by calling `world.setRecord` on the client with the test data)
* Adding utils to read client state to confirm the state sync was successful (by reading directly from the client's datastore which is exposed on `window`)

Note: this doesn't include tests for dynamic schema types yet, because we're missing typescript utils to encode dynamic data to the bytes blob expected by `world.setRecord`. (So far we never needed to encode data on the client, since this usually happens in the autogenerated Solidity table libraries, but in this case we want to call `world.setRecord` directly from the client, so we need to encode the data on the client).


alvrs:
* main
  * **PR #1063** <a href="https://app.graphite.dev/github/pr/latticexyz/mud/1063" target="_blank">Graphite</a> 
    * **PR #1064** <a href="https://app.graphite.dev/github/pr/latticexyz/mud/1064" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/latticexyz/mud/1064?utm_source=stack-comment).

dk1a:
> So far we never needed to encode data on the client, since this usually happens in the autogenerated Solidity table libraries, but in this case we want to call world.setRecord directly from the client, so we need to encode the data on the client

I'm not sure we want test-only ways to call tables that bypass systems, since that doesn't reflect how MUD is actually used. An alternative would be to dynamically make tables at the config level and autogenerate systems that wrap them (a postTablegen hook?)
Though testing the protocol (setRecord etc) from the client isn't exactly bad either

alvrs:
> I'm not sure we want test-only ways to call tables that bypass systems, since that doesn't reflect how MUD is actually used. An alternative would be to dynamically make tables at the config level and autogenerate systems that wrap them (a postTablegen hook?)
Though testing the protocol (setRecord etc) from the client isn't exactly bad either

Yeah I think we should have both! This set of integration tests aims to make sure the sync pipeline works as expected - i think for this it would even be fine to set data directly via a contract call in the test, but setting it via the client was equally straight forward and goes one step further. In addition to that we should have a test for calling contract systems and checking the expected state updates are received by the client.

The latter was basically what the previous test case did, maybe we should just keep it in addition to the new state sync tests.

e2e/packages/client-vanilla/src/mud/getNetworkConfig.ts
```diff
@@ -41,7 +41,7 @@ export async function getNetworkConfig(): Promise<NetworkConfig> {
       jsonRpcUrl: params.get("rpc") ?? chain.rpcUrls.default.http[0],
       wsRpcUrl: params.get("wsRpc") ?? chain.rpcUrls.default.webSocket?.[0],
     },
-    privateKey: getBurnerWallet().value,
+    privateKey: params.get("privateKey") ?? getBurnerWallet().value,
```
alvrs, line 44: this is so that we can provide a known private key in the tests (eg the deployer private key to be able to call `world.setRecord` from the client)

holic (approved):
didn't look too too deep but at a high level looks good!

PR #1068, opened 2023-06-22
refactor: move some utils to common, clean up

holic:
- Moves `TableId` and `isDefined`/`isNotNull` from `utils` to `common`
- Defines return values of exported functions in `common` in enables linter for it


examples/minimal/packages/client-phaser/src/layers/network/contractComponents.ts
```diff
@@ -1,24 +0,0 @@
-/* Autogenerated file. Do not edit manually. */
```
holic, line 1: this is an unused file and was likely leftover from copy+pasting phaser things (the phaser template removed this file a while back)

holic:
* main
  * **PR #1068** <a href="https://app.graphite.dev/github/pr/latticexyz/mud/1068" target="_blank">Graphite</a>  👈
    * **PR #1069** <a href="https://app.graphite.dev/github/pr/latticexyz/mud/1069" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/latticexyz/mud/1068?utm_source=stack-comment).

packages/common/src/TableId.ts
```diff
@@ -0,0 +1,37 @@
+import { Hex, stringToHex, hexToString, sliceHex, concatHex } from "viem";
+
+export class TableId {
```
alvrs, line 3: should we rename this to `ResourceSelector`? Or have a base class `ResourceSelector` this inherits from? (See https://github.com/latticexyz/mud/pull/999#discussion_r1222208895)

alvrs (approved):
comment above is not blocking

packages/common/src/TableId.ts
holic: ohh I like that better, will do

holic: hmm, gonna save this for later! in searching the codebase for "tableid", we have a lot of instances and so might wanna rethink the boundaries of this

PR #1070, opened 2023-06-22
feat(block-logs-stream): add block logs stream package

holic:
pulled out of #1033


packages/block-events-stream/src/createBlockEventsStream.ts
```diff
@@ -0,0 +1,125 @@
...
+        events,
+      });
+
+      // TODO: do something other than just throwing out pending logs
+      const nonPendingLogs = logs.filter(isNonPendingLog);
+
+      if (logs.length !== nonPendingLogs.length) {
+        // TODO: better error
+        console.warn("pending logs discarded");
+      }
```
alvrs, line 73: for context, what are pending logs and how would we ideally handle them?

packages/block-events-stream/src/createBlockNumberStream.ts
```diff
@@ -0,0 +1,44 @@
...
+      blockTag?: never;
+      block$: ReadonlyBehaviorSubject<Block>;
+    };
+
+export async function createBlockNumberStream({
+  publicClient,
+  blockTag,
+  block$: initialBlock$,
+}: CreateBlockNumberStreamOptions): Promise<ReadonlyBehaviorSubject<BlockNumber>> {
+  const block$ = initialBlock$ ?? (await createBlockStream({ publicClient, blockTag: blockTag as BlockTag }));
```
alvrs, line 26: is it necessary to request the full block from the rpc here if we're only interested in the block number?

packages/block-events-stream/src/createBlockNumberStream.ts
```diff
@@ -0,0 +1,44 @@
...
+  const block = block$.value;
+
+  if (!block.number) {
+    // TODO: better error
+    throw new Error(`${blockTag} block missing or pending`);
```
alvrs, line 31: what does it mean if a block we receive here is pending? Is this error just for if the initial block of our stream is not available?

packages/block-events-stream/src/createBlockStream.ts
```diff
@@ -0,0 +1,38 @@
+import { BehaviorSubject } from "rxjs";
+import type { Block, BlockTag, PublicClient } from "viem";
+import { ReadonlyBehaviorSubject } from "./common";
+
+// TODO: pass through viem's types, e.g. WatchBlocksParameters -> GetBlockReturnType
+// TODO: make stream closeable to make use of unwatch?
```
alvrs, line 6: this is a good point, i think we have to provide a way to unsubscribe all the internal subscriptions (ie everytime we set up a BehaviorSubject to return and do `.subscribe` somewhere), otherwise we might end up with a lot of subscriptions when the client hot reloads or sth

packages/block-events-stream/src/getLogs.ts
```diff
@@ -0,0 +1,71 @@
...
+        const { eventName, args } = decodeEventLog({
+          abi: events,
+          data: log.data,
+          topics: log.topics,
+          strict: true,
+        });
+        return formatLog(log, { args, eventName });
+      } catch (err) {
+        // We're using strict mode, so just skip if there is an error decoding.
+        return;
```
alvrs, line 67: for context what does using strict mode here entail?

alvrs (commented):
Mostly non blocking nits/questions. One thing I think we should think about sooner rather than later is how to handle unsubscribing from internal subscriptions, maybe by already including a `unsubscribe` method in this PR? Lmk what you think.

For the TODOs left in this PR, are you expecting to get to them soon in a follow up as part of the sync package work or is this something we should track to remember to get back to later?

packages/block-events-stream/src/createBlockEventsStream.ts
```diff
@@ -0,0 +1,126 @@
...
+import { getLogs } from "./getLogs";
+import { storeEventsAbi } from "@latticexyz/store";
+
+export type CreateBlockEventsStreamOptions<TAbiEvent extends AbiEvent> = {
+  publicClient: PublicClient;
+  fromBlock?: BlockNumber;
+  toBlock?: BlockNumber | ReadonlyBehaviorSubject<BlockNumber>;
+  address?: Hex;
+  events: readonly TAbiEvent[];
+  maxBlockRange?: number; // defaults to 1000
```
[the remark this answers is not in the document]
holic: :chefkiss:

packages/block-events-stream/src/createBlockEventsStream.ts
```diff
@@ -0,0 +1,126 @@
...
+      if (logs.length !== nonPendingLogs.length) {
+        // TODO: better error
+        console.warn("pending logs discarded");
+      }
+
+      // TODO: handle RPC block range errors
+      // TODO: handle RPC rate limit errors (hopefully via client retry policy)
+
+      const blockNumbers = Array.from(new Set(nonPendingLogs.map((log) => log.blockNumber)));
+      blockNumbers.sort((a, b) => (a < b ? -1 : a > b ? 1 : 0));
```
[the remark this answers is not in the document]
holic: I think this was leftover after some refactoring where `a` and `b` were previously something like `a.blockNumber` and `b.blockNumber` - will simplify!

packages/block-events-stream/src/createBlockNumberStream.ts
holic: `eth_blockNumber` doesn't support block tags and just returns the latest block number, so we have to request the block by block tag to get the number
(I already tried to refactor this to just use the block number RPC methods because they're cheaper, but ran into this issue)

packages/block-events-stream/src/createBlockNumberStream.ts
holic: yep! it's more edge-casey and to help TS infer the proper types and bypass the "block number can be null for pending blocks"

packages/block-events-stream/src/createBlockStream.ts
holic: yeah I wasn't sure how best to handle this given we're in the context of a stream

we could create our own type/wrapper around `BehaviorSubject` that has a `close` method or something, which will unsubscribe, but felt weird and error prone to extend rxjs types

we could return both the stream and unsubscribe/close methods, but that also felt weird

packages/block-events-stream/src/getLogs.ts
holic: https://viem.sh/docs/actions/public/getLogs.html#strict-mode

packages/block-events-stream/src/createBlockStream.ts
alvrs: Agree it's tricky, wonder what others do in this situation. I think we can't get around doing one of those to allow cleaning up old subscriptions (we ran into a bunch of issues with duplicate subscriptions in the past due to not cleaning them up properly, that was one of the reasons `recs`'s `world` had the concept of `disposers`). Returning both the stream and an `unsubscribe` method feels a bit better than extending the `BehaviorSubject` class because it makes it more explicit that this is something that should be unsubscribed. 

Btw I think if all of these were just `Observables` instead of `Subjects` only the last subscriber would have to unsubscribe, but it comes with its own set of issues (no concept of "current" value, issues with multiple subscribers to the same observable, and we would still need to pass through things like the `watchBlocks` `close` method)

holic: worth also deciding if rxjs is what we want as the underlying primitive vs. a native JS stream or event emitter or some other thing

packages/block-events-stream/src/createBlockEventsStream.ts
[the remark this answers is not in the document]
holic: oh I remember why: it's because these are bigints and native `.sort` doesn't handle bigints 🤦 

image

packages/block-events-stream/src/createBlockStream.ts
holic: After our chat, I did some digging and native streams are quite clunky. They only support byte buffers (so that the streams can move between threads), so we'd be JSON serializing and deserializing a bunch. They're kind of a pain to construct and read/write from. Since it's a pure byte stream, I don't think we can guarantee JSON blobs would come in meaningful chunks that we could immediately parse, so we'd need some separator between each JSON object and a stream parser that is JSON aware to correctly split on those separators (and not if the separator exists in a JSON string inside an object).

I couldn't find any new libs that wrap native streams to make them easier to use. A bunch of the libs I found are meant for Node.js streams, which deviate a bit from the browser stream spec (like `objectMode`).

packages/block-events-stream/src/blockRangeToLogs.test.ts
```diff
@@ -0,0 +1,112 @@
...
+        },
+        {
+          "fromBlock": 1001n,
+          "logs": [],
+          "toBlock": 1004n,
+        },
+        {
+          "fromBlock": 1005n,
+          "logs": [],
+          "toBlock": 1008n,
```
holic, line 107: with the loop above, this should loop until `toBlock` is `1010n`, so I'm not sure why it stopped here

packages/block-events-stream/README.md
```diff
@@ -0,0 +1,34 @@
...
+latestBlockNumber$
+  .pipe(
+    map((latestBlockNumber) => ({ startBlock: 0n, endBlock: latestBlockNumber })),
+    blockRangeToLogs({
+      publicClient,
+      address,
+      events,
+    }),
+    map(({ logs }) => from(groupLogsByBlockNumber(logs))),
+    mergeAll()
```
alvrs, line 29: should we add `import { mergeAll } from "rxjs"` for this example?

alvrs: also what's the difference between `map + mergeAll` to `mergeMap`?

packages/block-events-stream/README.md
```diff
@@ -0,0 +1,34 @@
...
+  map((block) => block.number)
+);
+
+latestBlockNumber$
+  .pipe(
+    map((latestBlockNumber) => ({ startBlock: 0n, endBlock: latestBlockNumber })),
+    blockRangeToLogs({
+      publicClient,
+      address,
+      events,
```
alvrs, line 26: is it worth adding the definitions of these to the example? or a comment about how to define them?

packages/block-events-stream/src/blockRangeToLogs.test.ts
alvrs: for context: turns out `exhaustMap` discards all block numbers that arrive before the inner observable completes. we switched the approach to `concatMap` to solve it.

packages/block-events-stream/src/blockRangeToLogs.ts
```diff
@@ -0,0 +1,52 @@
...
+  events: TAbiEvents;
+  maxBlockRange?: bigint;
+};
+
+export type BlockRangeToLogsResult<TAbiEvents extends readonly AbiEvent[]> = OperatorFunction<
+  { startBlock: BlockNumber; endBlock: BlockNumber },
+  FetchLogsResult<TAbiEvents>
+>;
+
+export function blockRangeToLogs<TAbiEvents extends readonly AbiEvent[]>({
```
alvrs, line 18: should we add a code documentation comment above here that explains the purpose / behavior?

packages/block-events-stream/src/fetchLogs.test.ts
```diff
@@ -0,0 +1,440 @@
...
+    const requests: any[] = [];
+    mockedTransportRequest.mockImplementation(async ({ method, params }): Promise<RpcLog[]> => {
+      if (method !== "eth_getLogs") throw new Error("not implemented");
+      requests.push(params);
+
+      if (hexToNumber((params as any)[0].toBlock) - hexToNumber((params as any)[0].fromBlock) > 500) {
+        throw new LimitExceededRpcError(
+          new RpcRequestError({
+            body: (params as any)[0],
+            url: "https://viem.sh",
```
alvrs, line 148:
```suggestion
            url: "https://mud.dev/",
```
🙈

alvrs (dismissed):
minor comments, not blocking

packages/block-events-stream/README.md
holic: no real difference! just got here iteratively, can do `mergeMap` instead

alvrs:
just merged #1036 - can we add a changeset to this PR to try out the new release pipeline?

alvrs (approved):

Issue #700, opened 2023-04-10
[Plugins] General purpose plugin system


alvrs:
- We need the core of MUD to be as small and as possible to make it easier to understand and easier to maintain, and eventually audit
- At the same time we want to be able to support arbitrary complex use cases

→ We need a rock solid plugin system!

Plugins can:
- modify the MUD config + types
- modify the codegen / build / deploy pipeline
- provide client libraries to interact with whatever the plugin provides
- provide systems / tables / modules to add contract functionality

alvrs:
part of #1668 / #1561

PR #1077, opened 2023-06-23
feat: local plugins

dk1a:
replaces https://github.com/latticexyz/mud/pull/1025
fixes https://github.com/latticexyz/mud/issues/746
fixes https://github.com/latticexyz/mud/issues/700

new issues we should add if this PR is merged as-is (the first two to replace a general #700):
- plugin hooks
- plugin cli
- mudup command to handle `.mud`
- tsgen should be part of mudup, not in userland with a hardcoded `.mud/expandedConfig.ts`
- reconsider `AsDependent` and its need in StoreConfig
- better errors when something gets an incorrect config (especially if it's an unexpanded config, this can probably be generalized with an extra option `expandConfig` injects)
- docs

Should `mudConfig` be in `world` and include `storePlugin, worldPlugin` by default?

(not very related to this PR) I really dislike how `as const` forces incorrect type highlighting to spread onto the entire config, rather than a specific line


packages/config/src/node/loadConfig.ts
```diff
@@ -7,8 +7,8 @@ import { pathToFileURL } from "url";
 import os from "os";
 
 // In order of preference files are checked
-const configFiles = ["mud.config.js", "mud.config.mjs", "mud.config.ts", "mud.config.mts"];
-const TEMP_CONFIG = "mud.config.temp.mjs";
+const CONFIG_FILE = path.join(".mud", "expandedConfig.ts");
+const TEMP_CONFIG = path.join(".mud", "expandedConfig.temp.mjs");
```
dk1a, line 11: in this PR I expect `.mud` to just exist (which it would because templates have it, and it doesn't need to change)
I think we should have a `mudup` as an actual command rather than a script wrapper in templates, and it can then take care of `.mud` properly

packages/react/src/store-cache/useRow.test.ts
```diff
@@ -1,15 +1,24 @@
...
+    tables: {
+      MultiKey: { keySchema: { first: "bytes32", second: "uint32" }, schema: "int32" },
+      Position: { schema: { x: "int32", y: "int32" } },
+    },
+  } as const);
+
+  const _typedExpandConfig = expandConfig as ExpandConfig<typeof config>;
+  type ExpandedConfig = MergeReturnType<typeof _typedExpandConfig<typeof config>>;
+  return expandConfig(config) as ExpandedConfig;
+})();
```
dk1a, line 21: this mess in each test is the biggest downside of this PR imo

holic:
Closing for now since this is very outdated and we'll likely start fresh when we get to the config overhaul.

PR #1080, opened 2023-06-26
feat(store-cache): make everything async

### LINKED ISSUE #1055: move store-cache to AsyncTupleDatabase
we'll need everything to be async to support IndexedDB, SQLite, etc.

frolic:
closes #1055

### DIFF
--- a/packages/react/src/store-cache/useRows.ts
+++ b/packages/react/src/store-cache/useRows.ts
@@ -2,6 +2,7 @@ import { DatabaseClient, FilterOptions, ScanResult } from "@latticexyz/store-cac
 import { StoreConfig } from "@latticexyz/store";
 import { useEffect, useState } from "react";
 import { useDeepMemo } from "../utils/useDeepMemo";
+import { useMountedState } from "../utils/useMountedState";
 
 /**
  * Returns an array of all rows matching the provided filter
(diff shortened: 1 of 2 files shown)


packages/store-cache/src/utils.ts
```diff
@@ -26,43 +34,42 @@ import { getAbiTypeDefaultValue } from "@latticexyz/schema-type/deprecated";
...
   config: C,
-  client: TupleDatabaseClient,
+  client: AsyncTupleDatabaseClient,
   namespace: C["namespace"],
   table: T & string,
   key: Key<C, T>,
   value: Partial<Value<C, T>>,
   options?: SetOptions
-) {
+): Promise<void> {
```
holic, line 38: I couldn't get this to work well with the conditional transaction. I tried something like:
```ts
SetOptions['transaction'] extends undefined ? Promise<void> : Promise<AsyncTupleRootTransactionApi>
```
but no luck. 

Also seems fine to not return the transaction, since you'll already have a reference to it if passed into `options`.

packages/store-cache/src/utils.ts
```diff
@@ -188,15 +185,17 @@ function getScanArgsFromFilter<C extends StoreConfig, T extends keyof C["tables"
...
-    scanArgs.gte = key?.gte && recordToTuple(key.gte, getKeyOrder(config, table));
-    scanArgs.gt = key?.gt && recordToTuple(key.gt, getKeyOrder(config, table));
-    scanArgs.lte = key?.lte && recordToTuple(key.lte, getKeyOrder(config, table));
-    scanArgs.lt = key?.lt && recordToTuple(key.lt, getKeyOrder(config, table));
+    scanArgs.gte =
+      key?.gte && recordToTuple(key.gte as Record<string, StaticPrimitiveType>, getKeyOrder(config, table));
+    scanArgs.gt = key?.gt && recordToTuple(key.gt as Record<string, StaticPrimitiveType>, getKeyOrder(config, table));
+    scanArgs.lte =
+      key?.lte && recordToTuple(key.lte as Record<string, StaticPrimitiveType>, getKeyOrder(config, table));
+    scanArgs.lt = key?.lt && recordToTuple(key.lt as Record<string, StaticPrimitiveType>, getKeyOrder(config, table));
```
holic, line 193: TS didn't like using `StaticPrimitiveType` alongside `Key<C, T>` (I think because a table's key schema types aren't specific to just static ABI types?)

packages/store-cache/src/utils.ts
```diff
@@ -257,7 +260,7 @@ function tupleToRecord(tuple: Tuple): Record<string, any> {
  * (see https://github.com/ccorcos/tuple-database/issues/25)
  * For now only `bigint` needs serialization.
  */
-function serializeKey(key: unknown) {
+function serializeKey(key: string | number | bigint | boolean): string | number | boolean {
```
holic, line 263: The casting above was mostly so I could make these utils more specific in terms of types. Happy to roll this back if you prefer how it was.

packages/store-cache/src/utils.ts
alvrs: yep i think that makes sense - if you want to stack transactions you have to pass it in, otherwise the transaction created in the function call is immediately committed and not useful outside the function

packages/store-cache/src/utils.ts
alvrs: ohh interesting, yeah i think you're right and fundamentally caused by the `keySchema` not explicitly being typed as `StaticAbiType` (we have validations for this but seems like somewhere the strong type gets lost). This is something we can revisit after #1077 is done.

alvrs (approved):
lgtm! this is a breaking change so we should add the `BREAKING CHANGE:` footer on the commit. For the first "snapshot" release after #1036 we should probably write some manual release note (coincides well with the first external core dev call on thursday) mentioning this change.

holic:
> lgtm! this is a breaking change so we should add the `BREAKING CHANGE:` footer on the commit. For the first "snapshot" release after #1036 we should probably write some manual release note (coincides well with the first external core dev call on thursday) mentioning this change.

Can do. This should only be breaking for folks using store-cache directly instead of via v1 networking and React hooks.

That said, there might be some timing issues with v1 networking code - we just fire off store-cache `set`/`remove` without awaiting or combining into a tx. Since we're only using `InMemoryTupleStorage` right now, this should be fine?

PR #1083, opened 2023-06-26
fix(templates): make chain id case insensitive in templates

### LINKED ISSUE #1015: make `chainId` case insensitive in templates
* template projects currently expect a `chainId` param, `chainid` doesn't work. We should make this case insensitive.

0xnaylor:
resolves #1015


alvrs (commented):
Thank you for the PR! I wonder if it would be worth adding a small util to the utils package that does something like this to reduce boilerplate code in the templates:

```typescript
function getLowercaseSeachParams() {
  const params = new URLSearchParams(window.location.search);
   return new Map([...params.entries()].map(([key, value]) => [key.toLowerCase(), value]))
}
```

One could also argue this is a total overkill for the use case and it would be sufficient to just do 

```typescript
const chainId = Number(params.get("chainId") || params.get("chainid") || import.meta.env.VITE_CHAIN_ID || 31337);
```

I'm tending towards the second options since it seems like an edge case

0xnaylor:
Thanks for the review and yeah, totally agree this is overkill. Have implemented your second suggestion.

alvrs (approved):

PR #1089, opened 2023-06-29
fix: resolved a minor bug in the empty array validation check

0xrhsmt:


packages/network/src/workers/SyncWorker.ts
```diff
@@ -187,7 +187,7 @@ export class SyncWorker<C extends Components> implements DoWork<Input, NetworkEv
 
     let initialState = createCacheStore();
 
-    if (initialRecords) {
```
0xrhsmt, line 190: fix__resolved_a_minor_bug_in_the_empty_array_validation_check_by_0xrhsmt_·_Pull_Request__1089_·_latticexyz_mud

empty array is truthy.

holic:
ooc what bug is this solving for?

PR #1096, opened 2023-07-03
feat(schema-type): add isDynamicAbiType

alvrs:
* adds a util function for `isDynamicAbiType` to `schema-type`


holic (approved):

alvrs:
rebased this after it was approved, but nothing changed, so gonna go ahead with merging

PR #1097, opened 2023-07-03
fix(protocol-parser): only append packed counter if there is dynamic data

alvrs:
* The packed counter is only appended to the encoded data if there is dynamic data: https://github.com/latticexyz/mud/blob/main/packages/store/src/StoreCore.sol#L434-L442

### DIFF
--- a/packages/protocol-parser/src/encodeRecord.ts
+++ b/packages/protocol-parser/src/encodeRecord.ts
@@ -20,9 +20,12 @@ export function encodeRecord(schema: Schema, values: readonly (StaticPrimitiveTy
 
   const dynamicData = dynamicDataItems.join("");
 
-  const packedCounter = `${encodeField("uint56", dynamicTotalByteLength).replace(/^0x/, "")}${dynamicFieldByteLengths
-    .map((length) => encodeField("uint40", length).replace(/^0x/, ""))
-    .join("")}`.padEnd(64, "0");
+  const packedCounter =
+    schema.dynamicFields.length > 0
+      ? `${encodeField("uint56", dynamicTotalByteLength).replace(/^0x/, "")}${dynamicFieldByteLengths
+          .map((length) => encodeField("uint40", length).replace(/^0x/, ""))
+          .join("")}`.padEnd(64, "0")
+      : "";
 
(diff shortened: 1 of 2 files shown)


alvrs:
* main
  * **PR #1096** <a href="https://app.graphite.dev/github/pr/latticexyz/mud/1096" target="_blank">Graphite</a> 
    * **PR #1097** <a href="https://app.graphite.dev/github/pr/latticexyz/mud/1097" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/latticexyz/mud/1097?utm_source=stack-comment).

holic (dismissed):
do we want a test for this?

alvrs:
> do we want a test for this?

good call, that would be good, also as a baseline for folks who want to port this

packages/protocol-parser/src/encodeRecord.ts
```diff
@@ -20,9 +20,12 @@ export function encodeRecord(schema: Schema, values: readonly (StaticPrimitiveTy
...
 
-  const packedCounter = `${encodeField("uint56", dynamicTotalByteLength).replace(/^0x/, "")}${dynamicFieldByteLengths
-    .map((length) => encodeField("uint40", length).replace(/^0x/, ""))
-    .join("")}`.padEnd(64, "0");
+  const packedCounter =
+    schema.dynamicFields.length > 0
+      ? `${encodeField("uint56", dynamicTotalByteLength).replace(/^0x/, "")}${dynamicFieldByteLengths
+          .map((length) => encodeField("uint40", length).replace(/^0x/, ""))
+          .join("")}`.padEnd(64, "0")
+      : "";
```
holic, line 28: sorry, one more thing

it might make sense to do this check and return early so we avoid all the other work above

shortly after `const staticData = ...`, do something like
```ts
if (!dynamicValues.length) {
  return `0x${staticData}`;
}
```

alvrs: agree, this also mirrors the on-chain logic better

holic (approved):

PR #1102, opened 2023-07-04
chore: synchronize versioning across all packages

alvrs:
- we want all mud packages to have the same version number to make it easier for consumers to upgrade all of them to compatible versions
- more info: https://github.com/changesets/changesets/blob/main/docs/fixed-packages.md


holic (approved):

holic:
changesets wants you to add a changeset for your changesets change

PR #1108, opened 2023-07-05
fix(services): revert switch from text to numeric

alvrs:


alvrs:
closing this since it's wrong

PR #1109, opened 2023-07-05
fix: remove devEmit when sending events from SyncWorker

Kooshaba:
big hack that just removes `devEmit` when we are sending events from the SyncWorker

lmk if this is too hacky and you have a less terrible way of doing it


holic (approved):
image

PR #1112, opened 2023-07-06
refactor(phaserx): update phaserx scene function signatures

vasa-develop:
Updated the Scene function signatures according to the documentation shown [here](https://rexrainbow.github.io/phaser3-rex-notes/docs/site/scene/#es6-class).

License: BUSL-1.1


alvrs:
what's the context / use case for this change? Will close this for now to declutter the PR list, since our current priority is more fundamental contract and state sync changes, and we don't have capacity to review implications of this change in depth.

PR #1113, opened 2023-07-06
WIP sync stack

alvrs:
- minimal POC implementation of a nodejs indexer using the new sync stack
```
/**
 * Idea:
 * - Users read data via views (or directly from tables), but not dynamic sql
 * - client defines views
 *   - if remote view: use materialized views on the indexer, propagate updates; later: only admin can create views, so views from the client are rejected if they don't exist yet
 *   - if local view: define view when starting up the client, rerun the view if a dependent table changes
 *
 * - might need a wrapper around insert (1. to match the interface of the sync stack and 2. to rerun views if tables change)
 */
```

TODO:
- [ ] normalize types across different packages (table, schema, etc.) #1296 
- [x] name -> tableName
- [x] make sure all schemas have a "schema" suffix e.g. keyTuple -> keyTupleSchema
- [x] keyTuple -> key
- [x] loading state
- [ ] catch errors in parsing events and do something with it #1294 
- [ ] refactor `_isDeleted` approach #1292 
- [x] handle errors when trying to fetch from indexer, skip
- [ ] add streaming of raw events, storage ops from trpc #1297
- [ ] detect and clear indexer, etc. when the chain restarts (look at first block hash? deploy block hash?) #1293 
- [ ] detect if offline and pause rpc requests (can this be done with rxjs?) #1295 
- [x] move `encodeEntity`/`decodeEntity` to recs package


packages/store-sync/src/recs/syncToRecs.ts
```diff
@@ -0,0 +1,136 @@
...
+        contractId: string;
+        keySchema: KeySchema;
+        valueSchema: ValueSchema;
+      }
+    >
+  >;
+  initialState?: {
+    blockNumber: bigint | null;
+    tables: (Table & { records: TableRecord[] })[];
+  };
```
holic, line 48: I initially tried this with `indexerUrl` and did the indexer fetch here, but that creates a circular dependency between store-sync and store-indexer that I couldn't figure out how to resolve.

(Mostly related to the fact that tRPC expects you to export `typeof appRouter` rather than giving you type helpers to construct the expected type, so defining the app router is coupled to the type you have to export.)

packages/store-sync/src/recs/defineInternalComponents.ts
```diff
@@ -0,0 +1,30 @@
+import { World, defineComponent, Type, Component } from "@latticexyz/recs";
+import { Table } from "../common";
+
+function typedComponent<TType>(
+  component: Component
+): typeof component extends Component<infer TSchema, infer TMetadata> ? Component<TSchema, TMetadata, TType> : never {
+  return component;
+}
```
holic, line 8: I am trying to change the position of the custom type to allow `Type.T` to resolve as a strong type, while inferring the schema/metadata without having to define them twice in `defineComponent` (once in the type, once in the object).

Without wrapping, I get the inferred schema/metadata as expected:
image

But once I wrap, it looks like:
image

I also tried a separate `defineTypedComponent` that takes in different order of type parameters, using the same inference as `defineComponent`, but that did the same thing.
image

PR #1116, opened 2023-07-08
feat: protocol-parser in go

authcall:
Implements the same api in the most consistent way possible (including file structure) as https://github.com/latticexyz/mud/tree/main/packages/protocol-parser


.github/workflows/services.yml
```diff
@@ -0,0 +1,26 @@
...
+        uses: actions/setup-go@v4
+        with:
+          go-version: '1.20'
+
+      - name: Build
+        working-directory: ./packages/services
+        run: go build -v ./...
+
+      - name: Test
+        working-directory: ./packages/services
```
alvrs, line 25: if we add this to the `build` and `test` scripts in the `package.json` it will automatically run when running `pnpm build` / `pnpm test` at the root, so we don't need this separate action

packages/services/pkg/protocol-parser/decodeDynamicField.go
```diff
@@ -0,0 +1,137 @@
...
+		BYTES25_ARRAY,
+		BYTES26_ARRAY,
+		BYTES27_ARRAY,
+		BYTES28_ARRAY,
+		BYTES29_ARRAY,
+		BYTES30_ARRAY,
+		BYTES31_ARRAY,
+		BYTES32_ARRAY,
+		ADDRESS_ARRAY,
+		BOOL_ARRAY:
```
alvrs, line 121: ultra tiny nit but my OCD brain wants these to be switched (first BOOL_ARRAY, then ADDRESS_ARRAY) to match the schema type enum

packages/services/pkg/protocol-parser/decodeStaticField.go
```diff
@@ -0,0 +1,133 @@
...
+		BYTES24,
+		BYTES25,
+		BYTES26,
+		BYTES27,
+		BYTES28,
+		BYTES29,
+		BYTES30,
+		BYTES31,
+		BYTES32:
+		// TODO: should we return a fixed size array per bytes size or a slice?
```
alvrs, line 124: good q, in the ts version we return a hex string padded to the size of the bytes. In go, is it possible to pass a fixed size byte array to a function that accepts variable size byte arrays? if so, I think returning a fixed size byte array here would be a nice touch

packages/services/pkg/schema-type/arrayAbiTypes.go
```diff
@@ -0,0 +1,5 @@
+package schematype
+
+func ArrayAbiTypeToStaticAbiType(schemaType SchemaType) SchemaType {
+	return (schemaType - UINT8_ARRAY)
```
alvrs, line 4: should we add a check here that the input type is actually an array type and throw an error else? thinking of cases where users incorrectly input `BYTES`, `STRING` or a static type

alvrs (commented):
this is great! just added some minor comments. for completeness, can we add a port for `decodeKeyTuple`?

alvrs:
also let's add a changeset! (run `pnpm changeset` locally and describe this change)

dk1a (commented):
would it be better to move schema-type stuff from `packages/services/pkg/schema-type` to `packages/schema-type/go`?

packages/services/pkg/protocol-parser/decodeDynamicField.go
```diff
@@ -0,0 +1,137 @@
+package protocolparser
+
+import (
+	. "latticexyz/mud/packages/services/pkg/schema-type"
+
+	"github.com/andriidski/abiencode-go/convert"
```
alvrs, line 6: this seems to be a private package? would it be possible to vendor it in this repo?

alvrs (approved):

alvrs:
> would it be better to move schema-type stuff from packages/services/pkg/schema-type to packages/schema-type/go?

Agree! not blocking for this PR though imo, can also do that in a follow up

PR #1117, opened 2023-07-08
docs: update snap sync example conf

noyyyy:
To use snap sync, you need install snapsync module first. This PR update the config in snap sync example.


noyyyy:
Update: Just found there is auto install description in docs:

```This will automatically install the necessary [smart contract modules](https://v2.mud.dev/world/modules#snapsyncmodule).
```

May I didn't use the latest mud version, I use `2.0.0-alpha.1.251+deda0235`

PR #1138, opened 2023-07-10
SQL backend for indexer

holic:


holic:
After fiddling with this, Kysely feels a little clunky compared to Drizzle, with less deeply integrated types. I think all we'd end up using it for creating/altering schemas, but it's not giving us any guard rails there for things like SQLite-specific column types, etc. @alvrs and I agree that it's probably worth starting with template strings for now instead of pulling in this lib just for schema creation.

PR #1147, opened 2023-07-11
feat(gas-report): create package, move relevant files to it

dk1a:
fixes https://github.com/latticexyz/mud/issues/667


dk1a:
> thoughts on refactoring gas report TS to be more of a output parser than a test runner?

if you mean to separate it into a parser and a runner that seems nice if not particularly important;
either way I'd rather keep this PR simple, and move on to gas optimizations

pnpm-lock.yaml
```diff
@@ -1,4 +1,4 @@
-lockfileVersion: '6.1'
+lockfileVersion: '6.0'
```
alvrs, line 1: which version of pnpm are you currently on?

alvrs (commented):
looking good! would you mind adding a simple example for how to use the gas report to the readme and adding a changeset?

pnpm-lock.yaml
dk1a: 8.6.7

dk1a:
> adding a changeset

I'm assuming that during we don't want to strictly follow semver for now and avoid `major`

holic (approved):

Issue #1086, opened 2023-06-28
Clean up solidity warnings


0xhank:
![image](https://github.com/latticexyz/mud/assets/71292860/f97b31f7-ccf9-4ad5-ba1b-65e4f1bfd75b)

There are lots of solidity warnings coming from mud files. Would be nice to get these cleaned so I can have an easier time resolving compiler warnings in my own code.

Issue #1133, opened 2023-07-10
Store gas metrics and optimizations


dk1a:
Go through StoreCore and its dependencies, see what needs metrics and what can be optimized

(in approximate order of less to more dependencies)
- [ ] schema-type
- [ ] Utils, TableId
- [ ] Memory
- [ ] Storage
- [ ] Slice
- [ ] TightCoder https://github.com/latticexyz/mud/issues/1132
- [ ] Bytes
- [ ] Schema
- [ ] PackedCounter
- [ ] StoreCore https://github.com/latticexyz/mud/issues/1131 and more
- [ ] StoreRead
- [ ] StoreSwitch
- [ ] autogenerated tables

dk1a:
this issue kind of depends on having store finalized, like
https://github.com/latticexyz/mud/pull/1061 
https://github.com/latticexyz/mud/issues/1121
https://github.com/latticexyz/mud/issues/1130

holic:
As part of this, do we wanna also do some kind of benchmark/comparison with native solidity?

dk1a:
> As part of this, do we wanna also do some kind of benchmark/comparison with native solidity?

would be nice, but it's not as useful for Store (since usually people don't use Store by itself), so this applies much more to World (TBD another issue)

dk1a:
(move 2 world-related items from @alvrs's comment into https://github.com/latticexyz/mud/issues/1316)

PR #1153, opened 2023-07-13
refactor(store): clean up Memory, make mcopy pure

dk1a:
fixes https://github.com/latticexyz/mud/issues/1086
fixes part of https://github.com/latticexyz/mud/issues/1133

this significantly ups the gas cost of mcopy and consequently `Slice.toBytes()`, but we shouldn't need this memory copying in the first place, which I'll address in a later PR

(it doesn't affect much, only some dynamic fields)

### DIFF
--- a/packages/store/src/StoreCore.sol
+++ b/packages/store/src/StoreCore.sol
@@ -436,7 +436,9 @@ library StoreCore {
     memoryPointer += staticLength;
 
     // Append the encoded dynamic length
-    Memory.store(memoryPointer, dynamicDataLength.unwrap());
+    assembly {
+      mstore(memoryPointer, dynamicDataLength)
+    }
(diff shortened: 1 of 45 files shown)


packages/store/src/StoreCore.sol
```diff
@@ -436,7 +436,9 @@ library StoreCore {
     memoryPointer += staticLength;
 
     // Append the encoded dynamic length
-    Memory.store(memoryPointer, dynamicDataLength.unwrap());
+    assembly {
+      mstore(memoryPointer, dynamicDataLength)
+    }
```
dk1a, line 441: for just mstore I think using assembly directly is clearer, and this method was only used in this 1 place

holic (approved):

PR #1155, opened 2023-07-13
refactor(schema-type): add gas-report, optimize

dk1a:
fixes a part of https://github.com/latticexyz/mud/issues/1133

### DIFF
--- a/packages/schema-type/src/solidity/SchemaType.sol
+++ b/packages/schema-type/src/solidity/SchemaType.sol
@@ -211,7 +211,7 @@ enum SchemaType {
  * Get the length of the data for the given schema type
  * (Because Solidity doesn't support constant arrays, we need to use a function)
  */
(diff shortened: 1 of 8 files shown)


packages/schema-type/src/solidity/SchemaType.sol
```diff
@@ -211,27 +211,20 @@ enum SchemaType {
  * Get the length of the data for the given schema type
  * (Because Solidity doesn't support constant arrays, we need to use a function)
  */
-function getStaticByteLength(SchemaType schemaType) pure returns (uint256) {
+function getStaticByteLength(SchemaType schemaType) pure returns (uint256 r) {
```
holic, line 214: for free functions, do we not need/get any benefit from `internal`? (I assume not)

dk1a: free functions can't have visibility modifiers at all
effectively they're all always internal

PR #1157, opened 2023-07-13
feat(cli): update set-version to match new release structure, add `--tag`, `--commit`

alvrs:
- updates the `set-version` cli command to work with the new release structure by adding two new options:
  - `--tag`: install the latest version of the given tag. For snapshot releases tags correspond to the branch name, commits to `main` result in an automatic snapshot release, so `--tag main` is equivalent to what used to be `-v canary`
  - `--commit`: install a version based on a given commit hash. Since commits from `main` result in an automatic snapshot release it works for all commits on main, and it works for manual snapshot releases from branches
- `set-version` now updates all `package.json` nested below the current working directory (expect `node_modules`), so no need for running it each workspace of a monorepo separately.

### DIFF
--- a/.changeset/thin-buses-reply.md
+++ b/.changeset/thin-buses-reply.md
@@ -1,0 +1,15 @@
+---
+"@latticexyz/cli": minor
+---
+
+- update the `set-version` cli command to work with the new release process by adding two new options:
+  - `--tag`: install the latest version of the given tag. For snapshot releases tags correspond to the branch name, commits to `main` result in an automatic snapshot release, so `--tag main` is equivalent to what used to be `-v canary`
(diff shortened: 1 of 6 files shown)


dk1a (commented):
I think you wanna update templates, they still use `set-version -v canary`

have you seen https://yargs.js.org/docs/#api-reference-conflictsx-y ? not sure it fits this case though

alvrs:
> I think you wanna update templates, they still use set-version -v canary

good catch, updated

> have you seen https://yargs.js.org/docs/#api-reference-conflictsx-y ? not sure it fits this case though

We could use it, but would have to add an array of the other 4 elements to each of `["mudVersion", "link", "tag", "commit", "restore"];` so I think in this case the non-yargs solution is easier

dk1a (approved):

PR #1158, opened 2023-07-14
Gas examples: storeAddress via storage

dk1a:
https://github.com/latticexyz/mud/pull/1061#pullrequestreview-1528903639 I should've started with this

This is comparing storeAddress as a method (optimized to not pass in `msgSender` for a better comparison)
to storeAddress as a storage variable

storage seems almost universally better. We pay ~1000-1500 up front, but get ~500-1000 less gas on each subsequent table interaction. So as long as you read/write more than like twice, it evens out and then keeps getting cheaper

the code is a mess, this is just a gas comparison. I'll continue the PR in https://github.com/latticexyz/mud/pull/1061


dk1a:
I should note that the method version is not fully optimized, whereas the storage version doesn't need as much optimization, so the comparison is still possibly unfair, this may be worth exploring further

PR #1161, opened 2023-07-14
fix(world): fix SnapSync index for records

dhvanipa:
The function `getRecords()` in `SnapSyncSystem` populates the `records` array with elements from the specified table, starting from a specified offset.

The for loop's index (`i`) is for the table's records (ie with an offset), but the index to populate `records` should be from 0 (ie without an offset). Currently, it's including the offset. This causes an out of bounds error.

Found the bug when running [getSnapSyncRecords](https://github.com/latticexyz/mud/blob/main/packages/network/src/v2/snapSync/getSnapSyncRecords.ts#L15) from the client which uses an offset after a table has more than 100 keys.

Also, added a test that uses an offset.


alvrs (dismissed):
thank you!

alvrs:
small request: could you add a changeset describing this change? (run `pnpm changeset` locally)

dhvanipa:
> small request: could you add a changeset describing this change? (run `pnpm changeset` locally)

Patch or minor bump to world?

holic:
> > small request: could you add a changeset describing this change? (run `pnpm changeset` locally)
> 
> Patch or minor bump to world?

patch is probably fine!

dhvanipa:
> > Patch or minor bump to world?
> 
> patch is probably fine!

added!

holic (approved):

alvrs:
we removed snap sync in #1311 since it wasn't compatible with the v2 networking stack and the recommended way to load a MUD app is via `store-indexer` (the snap sync module added a lot of gas to each store operation on-chain, and syncing via the RPC node is computationally heavy, can't be cached, and is expensive in terms of RPC cost, so not suitable for production use cases)

PR #1163, opened 2023-07-17
feat(world): move access control to module

alvrs:


packages/world/gas-report.json
```diff
@@ -81,240 +81,240 @@
...
     "name": "Set metadata",
-    "gasUsed": 282542
+    "gasUsed": 307546
   },
   {
     "file": "test/World.t.sol",
     "test": "testSetRecord",
     "name": "Write data to the table",
-    "gasUsed": 42502
+    "gasUsed": 67036
```
alvrs, line 294: Unfortunately this change significantly increases gas cost, to the point where I don't think it's a good tradeoff to have modular access control

holic, line 294: ouch, seems like gas bumped quite a bit on write operations?

dk1a: ouch, I wonder if there's a way to optimize it

PR #1164, opened 2023-07-17
build: bump viem, abitype

holic:
bump viem to 1.3.0 and abitype to 0.8.11 everywhere
and move to `viem/chains` instead of `@wagmi/chains` (since viem depends on `@wagmi/chains` and re-exports it)

PR #1165, opened 2023-07-18
build: bump TS

holic:
In prep for bumping some other deps (#1164)


docs/pnpm-lock.yaml
```diff
@@ -1,4 +1,4 @@
-lockfileVersion: '6.1'
+lockfileVersion: '6.0'
```
holic, line 1: I dunno why this keeps flip flopping

dk1a: I had a similar thing in https://github.com/latticexyz/mud/pull/1061 on pnpm 8.6.7, assume it's just some of our lockfiles being outdated since pnpm rolled back their version to 6.0

alvrs (approved):

PR #1166, opened 2023-07-18
docs: throw out typedoc, solidity-docgen

holic:
typedoc had some conflicts with #1165 and since we're not using it anymore, decided to rip it out

unclear if we're still using solidity-docgen, so I ripped that out too


packages/create-mud/package.json
```diff
@@ -26,8 +26,7 @@
   },
   "devDependencies": {
     "@types/node": "^18.15.11",
-    "tsup": "^6.7.0",
-    "typescript": "^4.9.5"
```
holic, line 30: we use a TS dep at the monorepo/workspace root

packages/gas-report/package.json
```diff
@@ -41,7 +41,6 @@
     "@types/yargs": "^17.0.10",
     "ds-test": "https://github.com/dapphub/ds-test.git#c9ce3f25bde29fc5eb9901842bf02850dfd2d084",
     "forge-std": "https://github.com/foundry-rs/forge-std.git#b4f121555729b3afb3c5ffccb62ff4b6e2818fd3",
-    "rimraf": "^3.0.2",
```
holic, line 44: same as above, we use rimraf at workspace root

alvrs (approved):

PR #1168, opened 2023-07-18
test: bump forge-std and ds-test

dk1a:


holic (approved):

dk1a:
btw regarding changesets - do we add them even for devDeps / test stuff?

holic:
> btw regarding changesets - do we add them even for devDeps / test stuff?

good question! I have been defaulting to adding them, mostly to quiet the warning by the changeset-bot, but that's probably not a good metric 🙃

PR #1172, opened 2023-07-18
docs: bump nextra

holic:
fixes issue with copying code selection (https://github.com/shuding/nextra/pull/1947)


holic:
(confirmed via preview deploy that this did what I wanted)

PR #1174, opened 2023-07-18
feat(world, store): stop loading schema from storage, require schema as an argument

alvrs:


alvrs:
We should probably wait for v2 networking to be merged and v1 networking to be removed to avoid double work in integrating this in the old stack.

alvrs:
* main
  * **PR #1174** <a href="https://app.graphite.dev/github/pr/latticexyz/mud/1174" target="_blank">Graphite</a>  👈
    * **PR #1182** <a href="https://app.graphite.dev/github/pr/latticexyz/mud/1182" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/latticexyz/mud/1174?utm_source=stack-comment).

packages/store/gas-report.json
```diff
@@ -261,13 +261,13 @@
     "file": "test/Mixed.t.sol",
     "test": "testSetAndGet",
     "name": "set record in Mixed",
-    "gasUsed": 110021
+    "gasUsed": 111113
```
alvrs, line 264: very interesting that the gas cost increased for this one (while decreasing for the other tests). @dk1a do you have an intuition around why?

dk1a: pretty sure it's cause `registerSchema` is called in that test (before gas report), which warms the slot, so there's no gas reduction, but there is a call to `getSchema` which is very suboptimal (should be better after https://github.com/latticexyz/mud/pull/1252 and SchemaType->staticLength)

e2e/pnpm-lock.yaml
```diff
@@ -1,9 +1,5 @@
 lockfileVersion: '6.0'
 
-settings:
-  autoInstallPeers: true
-  excludeLinksFromLockfile: false
-
```
alvrs, line 6: not sure why this changed, what's your guys' pnpm version?

packages/schema-type/src/typescript/deprecated/utils/encodeSchema.ts
```diff
@@ -6,6 +6,7 @@ import { SchemaType } from "../SchemaType";
  * Port of `Schema.sol` from `@latticexyz/store`
  * @param schema The schema to encode SchemaType[]
  * @returns The encoded schema as a 32 byte hex string
+ * @deprecated use `schemaToHex` from `protocol-parser` instead
```
alvrs, line 9: added this small note because i keep forgetting how the new method is called and the first thing i try is `encodeSchema`; now vscode gives a hint for which method to use as replacement of `encodeSchema`

e2e/packages/sync-test/data/encodeTestData.ts
```diff
@@ -2,19 +2,31 @@ import { mapObject } from "@latticexyz/utils";
...
+
+      const encodedValueSchema = schemaToHex(abiTypesToSchema(Object.values(config.tables[table].schema)));
+
+      return {
+        key: encodedKey,
+        value: encodedValue,
+        valueSchema: encodedValueSchema,
+      };
+    });
+  }) as EncodedData<typeof testData>;
```
alvrs, line 31: this change is fixing a type issue, adding the `valueSchema` to the encoded data and slightly refactoring because the nested functions became hard to read

packages/cli/src/commands/trace.ts
```diff
@@ -74,7 +76,12 @@ const commandModule: CommandModule<Options, Options> = {
     for (const name of names) {
       const systemSelector = new TableId(namespace, name);
       // Get the first field of `Systems` table (the table maps system name to its address and other data)
-      const address = await WorldContract.getField(systemsTableId.toHex(), [systemSelector.toHex()], 0);
+      const address = await WorldContract.getField(
+        systemsTableId.toHex(),
+        [systemSelector.toHex()],
+        0,
+        systemTableSchema
+      );
```
alvrs, line 84: `getField` now requires the value schema to be passed in

packages/store/src/StoreCore.sol
alvrs, line 1: This file contains the most significant changes (reading the schema from args instead of storage and renaming `schema` to `valueSchema`). The interface changes are then propagated through all the interfaces (`IStore`, `World`, etc)

packages/store/ts/codegen/field.ts
alvrs, line 1: passing the schema in all methods interacting with the store

alvrs:
This looks like a massive PR but it's mostly small refactors (passing the schema as arg to all methods interacting with the store). Added some PR comments to point out the relevant areas to make reviewing a bit easier.

e2e/pnpm-lock.yaml
dk1a: 8.6.12, `pnpm install --force` doesn't change lockfile for me

.changeset/few-mirrors-reflect.md
```diff
@@ -0,0 +1,23 @@
...
+
+All `Store` methods now require the table's value schema to be passed in as an argument instead of loading it from storage.
+This decreases gas cost and removes circular dependencies of the Schema table (where it was not possible to write to the Schema table before the Schema table was registered).
+
+```diff
+  function setRecord(
+    bytes32 table,
+    bytes32[] calldata key,
+    bytes calldata data,
++   Schema valueSchema
```
holic, line 16: Does the argument ordering here make sense? I usually order args by specificity and would kind of expect `valueSchema` to be further ahead in the list of arguments (before or after `table` perhaps).

Unless it's an optional arg, then it being at the end makes a bit more sense.

e2e/packages/contracts/worlds.json
```diff
@@ -1,5 +1,5 @@
 {
   "31337": {
-    "address": "0x5FbDB2315678afecb367f032d93F642f64180aa3"
+    "address": "0xA7c59f010700930003b33aB25a7a0679C860f29c"
```
holic, line 3: wonder if this will muck with e2e tests or if this will get overwritten in CI after deploy

e2e/pnpm-lock.yaml
holic: 8.6.8

packages/cli/src/commands/trace.ts
holic: another approach that might be a little less error prone if we change this schema later and forget to update this is to fetch the schema from the chain before we get field

packages/store/src/codegen/tables/KeyEncoding.sol
```diff
@@ -90,7 +90,7 @@ library KeyEncoding {
     _keyTuple[4] = _boolToBytes32(k5);
     _keyTuple[5] = bytes32(uint256(uint8(k6)));
 
-    bytes memory _blob = StoreSwitch.getField(_tableId, _keyTuple, 0);
+    bytes memory _blob = StoreSwitch.getField(_tableId, _keyTuple, 0, getSchema());
```
holic, line 93: I don't know the gas implications but I wonder if it would make sense to provide an alternative code path for these cases where we can generate a byte offset rather than a schema index + schema definition.

holic (dismissed):
Looks great! Mostly just a question about where we want to put the schema arg.

dk1a (approved):

packages/store/src/codegen/tables/KeyEncoding.sol
alvrs: yeah could add this, but would defer the decision until after #1222

.changeset/few-mirrors-reflect.md
alvrs: i agree with ordering by specificity but would defer this refactor until after this current stack of PRs is merged since they touch lots of the same places in the code and changing it here would likely lead to massive merge conflicts in the upstream branches

e2e/pnpm-lock.yaml
alvrs: upgraded to `8.6.12` and the diff is gone 👍

packages/cli/src/commands/trace.ts
```diff
@@ -11,8 +11,10 @@ import { resolveWorldConfig, WorldConfig } from "@latticexyz/world";
 import { IBaseWorld } from "@latticexyz/world/types/ethers-contracts/IBaseWorld";
 import IBaseWorldData from "@latticexyz/world/abi/IBaseWorld.sol/IBaseWorld.json" assert { type: "json" };
 import { getChainId, getExistingContracts } from "../utils";
+import { schemaToHex } from "@latticexyz/protocol-parser";
```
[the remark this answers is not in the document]
holic: one day I'll add a linter/autofixer for this: https://github.com/latticexyz/mud/issues/1277

holic (approved):

PR #1176, opened 2023-07-18
feat(store-sync): rework blockLogsToStorage

holic:
Pulled out of https://github.com/latticexyz/mud/pull/1113


alvrs (commented):
to double check my understanding, instead of passing in an object with handlers like like `{ setRecord, setField, deleteRecord }` etc., users pass in a single handler `storeOperations` correct? for context why did you end up going this route instead of a handler per store operation?

packages/store-sync/src/blockLogsToStorage.ts
```diff
@@ -0,0 +1,305 @@
...
+
+// TODO: change table schema/metadata APIs once we get both schema and field names in the same event
+
+// TODO: export these from store or world
+export const schemaTableId = new TableId("mudstore", "schema");
+export const metadataTableId = new TableId("mudstore", "StoreMetadata");
+
+// I don't love carrying all these types through. Ideally this should be the shape of the thing we want, rather than the specific return type from a function.
+export type StoreEventsLog = GetLogsResult<StoreEventsAbi>[number];
+export type BlockLogs = GroupLogsByBlockNumberResult<StoreEventsLog>[number];
```
alvrs, line 27: could you elaborate on this comment? having trouble following

packages/store-sync/src/blockLogsToStorage.ts
```diff
@@ -0,0 +1,305 @@
...
+  blockNumber: BlockLogs["blockNumber"];
+  operations: StorageOperation<TConfig>[];
+}>;
+
+type TableNamespace = string;
+type TableName = string;
+type TableKey = `${Address}:${TableNamespace}:${TableName}`;
+
+// hacky fix for schema registration + metadata events spanning multiple blocks
+// TODO: remove this once schema registration+metadata is one event or tx
```
alvrs, line 103:
```suggestion
// TODO: remove this once schema registration+metadata is one event or tx
// (https://github.com/latticexyz/mud/pull/1182)
```

packages/store-sync/src/blockLogsToStorage.ts
```diff
@@ -0,0 +1,305 @@
...
+      const schema = hexToTableSchema(log.args.data);
+
+      const key: TableKey = `${log.address}:${tableId.namespace}:${tableId.name}`;
+      if (!visitedSchemas.has(key)) {
+        visitedSchemas.set(key, { address: log.address, tableId, schema });
+        newTableKeys.add(key);
+      }
+    });
+
+    // Then find all metadata events. These should follow schema registration events and be in the same block (since they're in the same tx).
```
alvrs, line 137: seems like there is a comment missing above like "First find all schema events..." or something

packages/store-sync/src/blockLogsToStorage.ts
```diff
@@ -0,0 +1,305 @@
...
+    // TODO: rework contracts so schemas+tables are combined and immutable
```
alvrs, line 138:
```suggestion
    // TODO: rework contracts so schemas+tables are combined and immutable
    // (https://github.com/latticexyz/mud/pull/1182)
```

packages/store-sync/src/blockLogsToStorage.ts
```diff
@@ -0,0 +1,305 @@
...
+      }
+
+      const tableId = TableId.fromHex(tableForSchema);
+      const [tableName, abiEncodedFieldNames] = decodeRecord(
+        // TODO: this is hardcoded for now while metadata is separate from table registration
+        { staticFields: [], dynamicFields: ["string", "bytes"] },
+        log.args.data
+      );
+      const valueNames = decodeAbiParameters(parseAbiParameters("string[]"), abiEncodedFieldNames as Hex)[0];
+      // TODO: add key names to table registration when we refactor it
```
alvrs, line 155:
```suggestion
      // TODO: add key names to table registration when we refactor it
      // (https://github.com/latticexyz/mud/pull/1182)
```

packages/store-sync/src/blockLogsToStorage.ts
holic: I was thinking that using the `GetLogsResult` import type ties it the shape defined in the other module (and its package version) and was worried it make it harder to pass in an object from another location (rather from this specific function in this specific module). But maybe that's not how these types actually work and they just compare shapes.

This worry comes from seeing weird type errors when certain package versions get out of alignment and is why we have most deps pinned to a specific version rather than a version range.

alvrs (approved):
lgtm!

PR #1178, opened 2023-07-19
fix(cli): add support for legacy transactions in deploy script

TheGreatAxios:
The Problem
---
Currently if you try to deploy to networks that do not support EIP-1559 then you need to use the --legacy flag within foundry. The custom deploy script as part of the @latticexyz/cli package throws an error on deploy. This blocks the usage of MUD on SKALE (https://skale.space) which is arguably a top network choice for MUD due to the fast speed and zero gas fees.

The Solution
---
I have made a minor update to the script so that it uses EIP-1159 if the values exist, otherwise it falls back to just gasPrice which works on legacy networks i.e SKALE Chains. 

If all values are undefined then it throws an error which is a catch all meaning that there should not be any new issues that arise. 

Note: I have not done any changeset or bumped version as the documentation on both is limited.

### DIFF
--- a/packages/cli/src/utils/deploy.ts
+++ b/packages/cli/src/utils/deploy.ts
@@ -573,7 +573,7 @@ export async function deploy(
         : Math.floor(1_500_000_000 * multiplier)
       : undefined;
     if (maxPriorityFeePerGas) {
-      maxFeePerGas = feeData.lastBaseFeePerGas ? feeData.lastBaseFeePerGas.mul(2).add(maxPriorityFeePerGas) : undefined;
+      maxFeePerGas = feeData.lastBaseFeePerGas?.mul(2).add(maxPriorityFeePerGas);
     } else {
       gasPrice = feeData.gasPrice;
     }
--- a/packages/cli/src/utils/deploy.ts
+++ b/packages/cli/src/utils/deploy.ts
@@ -572,9 +572,11 @@ export async function deploy(
         ? 0
         : Math.floor(1_500_000_000 * multiplier)
       : undefined;
-    if (maxPriorityFeePerGas)
+    if (maxPriorityFeePerGas) {
       maxFeePerGas = feeData.lastBaseFeePerGas ? feeData.lastBaseFeePerGas.mul(2).add(maxPriorityFeePerGas) : undefined;
-    gasPrice = feeData.gasPrice;
+    } else {
+      gasPrice = feeData.gasPrice;
+    }
   }
 }
 
--- a/packages/cli/src/utils/deploy.ts
+++ b/packages/cli/src/utils/deploy.ts
@@ -399,12 +399,12 @@ export async function deploy(
     try {
       const factory = new ethers.ContractFactory(abi, bytecode, signer);
       console.log(chalk.gray(`executing deployment of ${contractName} with nonce ${nonce}`));
- const deployPromise = factory
+      const deployPromise = factory
         .deploy({
           nonce: nonce++,
           maxPriorityFeePerGas,
           maxFeePerGas,
-          gasPrice
+          gasPrice,
         })
         .then((c) => (disableTxWait ? c : c.deployed()));
 
@@ -546,21 +546,26 @@ export async function deploy(
     // Compute maxFeePerGas and maxPriorityFeePerGas like ethers, but allow for a multiplier to allow replacing pending transactions
     const feeData = await provider.getFeeData();
 
-    // Set the priority fee to 0 for development chains with no base fee, to allow transactions from unfunded wallets
-    if(!feeData.lastBaseFeePerGas && !feeData.gasPrice) throw new MUDError("Can not fetch lastBaseFeePerGas from RPC");
-    
-    if(feeData.lastBaseFeePerGas) {
+    if (feeData.lastBaseFeePerGas) {
       if (!feeData.lastBaseFeePerGas.eq(0) && (await signer.getBalance()).eq(0)) {
         throw new MUDError(`Attempting to deploy to a chain with non-zero base fee with an account that has no balance.
-If you're deploying to the Lattice testnet, you can fund your account by running 'pnpm mud faucet --address ${await signer.getAddress()}'`);
+        If you're deploying to the Lattice testnet, you can fund your account by running 'pnpm mud faucet --address ${await signer.getAddress()}'`);
       }
+
+      // Set the priority fee to 0 for development chains with no base fee, to allow transactions from unfunded wallets
       maxPriorityFeePerGas = feeData.lastBaseFeePerGas.eq(0) ? 0 : Math.floor(1_500_000_000 * multiplier);
       maxFeePerGas = feeData.lastBaseFeePerGas.mul(2).add(maxPriorityFeePerGas);
-    } else {
+    } else if (feeData.gasPrice) {
+      // Legacy chains with gasPrice instead of maxFeePerGas
       if (!feeData.gasPrice.eq(0) && (await signer.getBalance()).eq(0)) {
-        throw new MUDError(`Attempting to deploy to a chain with non-zero gas price with an account that has no balance.`);
+        throw new MUDError(
+          `Attempting to deploy to a chain with non-zero gas price with an account that has no balance.`
+        );
       }
+
       gasPrice = feeData.gasPrice;
+    } else {
+      throw new MUDError("Can not fetch fee data from RPC");
     }
   }
 }
--- a/packages/cli/src/utils/deploy.ts
+++ b/packages/cli/src/utils/deploy.ts
@@ -401,7 +401,7 @@ export async function deploy(
       console.log(chalk.gray(`executing deployment of ${contractName} with nonce ${nonce}`));
       let deployPromise;
 
-      if (maxFeePerGas && maxPriorityFeePerGas) {
+      if (maxFeePerGas !== undefined && maxPriorityFeePerGas !== undefined) {
         deployPromise = factory
           .deploy({
             nonce: nonce++,
@@ -410,7 +410,7 @@ export async function deploy(
           })
           .then((c) => (disableTxWait ? c : c.deployed()));
         /// Gas Price === Legacy
-      } else if (gasPrice) {
+      } else if (gasPrice !== null) {
         deployPromise = factory
           .deploy({
             nonce: nonce++,
--- a/packages/cli/src/utils/deploy.ts
+++ b/packages/cli/src/utils/deploy.ts
@@ -512,8 +512,15 @@ export async function deploy(
     try {
       const gasLimit = await contract.estimateGas[func].apply(null, args);
       console.log(chalk.gray(`executing transaction: ${functionName} with nonce ${nonce}`));
+      const _baseTxConfig = { gasLimit, nonce: nonce++ };
+      const txConfig = Object.assign(
+        _baseTxConfig,
+        maxPriorityFeePerGas !== undefined && maxFeePerGas !== undefined
+          ? { maxPriorityFeePerGas, maxFeePerGas }
+          : { gasPrice }
+      );
       const txPromise = contract[func]
-        .apply(null, [...args, { gasLimit, nonce: nonce++, maxPriorityFeePerGas, maxFeePerGas }])
+        .apply(null, [...args, txConfig])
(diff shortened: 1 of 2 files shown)


packages/cli/src/utils/deploy.ts
```diff
@@ -397,13 +399,26 @@ export async function deploy(
...
-      const deployPromise = factory
-        .deploy({
-          nonce: nonce++,
-          maxPriorityFeePerGas,
-          maxFeePerGas,
-        })
-        .then((c) => (disableTxWait ? c : c.deployed()));
+      let deployPromise;
+
+      if (maxFeePerGas && maxPriorityFeePerGas) {
```
holic, line 404: I think we might need an explicit `== null` check or similar here because we set anvil to use `0` for `maxFeePerGas`?

packages/cli/src/utils/deploy.ts
```diff
@@ -397,13 +399,26 @@ export async function deploy(
...
+            nonce: nonce++,
+            maxPriorityFeePerGas,
+            maxFeePerGas,
+          })
+          .then((c) => (disableTxWait ? c : c.deployed()));
+        /// Gas Price === Legacy
+      } else if (gasPrice) {
+        deployPromise = factory
+          .deploy({
+            nonce: nonce++,
```
alvrs, line 416: do we need to pass `gasPrice` as an argument here?

packages/cli/src/utils/deploy.ts
```diff
@@ -397,13 +399,26 @@ export async function deploy(
...
+      if (maxFeePerGas && maxPriorityFeePerGas) {
+        deployPromise = factory
+          .deploy({
+            nonce: nonce++,
+            maxPriorityFeePerGas,
+            maxFeePerGas,
+          })
+          .then((c) => (disableTxWait ? c : c.deployed()));
+        /// Gas Price === Legacy
+      } else if (gasPrice) {
```
alvrs, line 413: similar to above we should check for `null` here, otherwise there might be issues with gas price 0

packages/cli/src/utils/deploy.ts
TheGreatAxios: Thanks for the feedback. 
For anvil is that the default setting in order to make it EIP-1559 compliant? If yes - then I think it should be fine as is but I will test.

packages/cli/src/utils/deploy.ts
```diff
@@ -397,13 +399,27 @@ export async function deploy(
...
-          nonce: nonce++,
-          maxPriorityFeePerGas,
-          maxFeePerGas,
-        })
-        .then((c) => (disableTxWait ? c : c.deployed()));
+      let deployPromise;
+
+      if (maxFeePerGas !== undefined && maxPriorityFeePerGas !== undefined) {
+        deployPromise = factory
+          .deploy({
```
alvrs, line 406: Since we only set either (`maxFeePerGas` and `maxPriorityFeePerGas`) or (`gasPrice`), with the other one staying `undefined`, we don't need this `if/else` statement but can just pass all three variables to the function.

packages/cli/src/utils/deploy.ts
```diff
@@ -496,8 +512,15 @@ export async function deploy(
     try {
       const gasLimit = await contract.estimateGas[func].apply(null, args);
       console.log(chalk.gray(`executing transaction: ${functionName} with nonce ${nonce}`));
+      const _baseTxConfig = { gasLimit, nonce: nonce++ };
+      const txConfig = Object.assign(
+        _baseTxConfig,
+        maxPriorityFeePerGas !== undefined && maxFeePerGas !== undefined
+          ? { maxPriorityFeePerGas, maxFeePerGas }
+          : { gasPrice }
+      );
```
alvrs, line 521: same here, since we only set either (maxFeePerGas and maxPriorityFeePerGas) or (gasPrice), with the other one staying undefined, we don't need this conditional and can just pass all three variables to the function

packages/cli/src/utils/deploy.ts
```diff
@@ -542,15 +565,18 @@ export async function deploy(
...
     // Set the priority fee to 0 for development chains with no base fee, to allow transactions from unfunded wallets
-    maxPriorityFeePerGas = feeData.lastBaseFeePerGas.eq(0) ? 0 : Math.floor(1_500_000_000 * multiplier);
-    maxFeePerGas = feeData.lastBaseFeePerGas.mul(2).add(maxPriorityFeePerGas);
+    maxPriorityFeePerGas = feeData.lastBaseFeePerGas
+      ? feeData.lastBaseFeePerGas.eq(0)
+        ? 0
+        : Math.floor(1_500_000_000 * multiplier)
+      : undefined;
+    if (maxPriorityFeePerGas) {
+      maxFeePerGas = feeData.lastBaseFeePerGas ? feeData.lastBaseFeePerGas.mul(2).add(maxPriorityFeePerGas) : undefined;
```
alvrs, line 576:
```suggestion
      maxFeePerGas = feeData.lastBaseFeePerGas?.mul(2).add(maxPriorityFeePerGas);
```

packages/cli/src/utils/deploy.ts
```diff
@@ -59,8 +59,10 @@ export async function deploy(
   console.log("Initial nonce", nonce);
 
   // Compute maxFeePerGas and maxPriorityFeePerGas like ethers, but allow for a multiplier to allow replacing pending transactions
-  let maxPriorityFeePerGas: number;
-  let maxFeePerGas: BigNumber;
+  let maxPriorityFeePerGas: number | undefined;
+  let maxFeePerGas: BigNumber | undefined;
+  let gasPrice: BigNumber | null;
```
alvrs, line 64: why `null` instead of `undefined` in this case?

TheGreatAxios:
Hey @alvrs 

Are the above blocking you from moving this forward? 
The functionality works and was tested and I cannot get the EIP-1559 values to work on a legacy transaction. 
Can we move this forward and re-visit in another PR related to optimizing this legacy flow and possibly adding some test suites around the flow?

alvrs:
> Are the above blocking you from moving this forward?

Yes, I'm only comfortable merging this once it matches the code style of this project. The requested changes are pretty small though, I could just do it if you don't have capacity so we can move forward with this.

packages/cli/src/utils/deploy.ts
alvrs, line 64:
```suggestion
  let gasPrice: BigNumber | undefined;
```

packages/cli/src/utils/deploy.ts
```diff
@@ -397,13 +399,27 @@ export async function deploy(
...
+      } else if (gasPrice !== null) {
+        deployPromise = factory
+          .deploy({
+            nonce: nonce++,
+            gasPrice,
+          })
+          .then((c) => (disableTxWait ? c : c.deployed()));
+      } else {
+        throw new Error("Error: No Fee Data Found");
+      }
```
alvrs, line 422:
```suggestion
 const deployPromise = factory
        .deploy({
          nonce: nonce++,
          maxPriorityFeePerGas,
          maxFeePerGas,
          gasPrice
        })
        .then((c) => (disableTxWait ? c : c.deployed()));
```

packages/cli/src/utils/deploy.ts
alvrs, line 521:
```suggestion
```

packages/cli/src/utils/deploy.ts
```diff
@@ -496,8 +512,15 @@ export async function deploy(
...
+      const _baseTxConfig = { gasLimit, nonce: nonce++ };
+      const txConfig = Object.assign(
+        _baseTxConfig,
+        maxPriorityFeePerGas !== undefined && maxFeePerGas !== undefined
+          ? { maxPriorityFeePerGas, maxFeePerGas }
+          : { gasPrice }
+      );
       const txPromise = contract[func]
-        .apply(null, [...args, { gasLimit, nonce: nonce++, maxPriorityFeePerGas, maxFeePerGas }])
+        .apply(null, [...args, txConfig])
```
alvrs, line 523:
```suggestion
        .apply(null, [...args, { gasLimit, nonce: nonce++, maxPriorityFeePerGas, maxFeePerGas, gasPrice }])
```

packages/cli/src/utils/deploy.ts
```diff
@@ -542,15 +565,18 @@ export async function deploy(
...
 
     // Set the priority fee to 0 for development chains with no base fee, to allow transactions from unfunded wallets
-    maxPriorityFeePerGas = feeData.lastBaseFeePerGas.eq(0) ? 0 : Math.floor(1_500_000_000 * multiplier);
-    maxFeePerGas = feeData.lastBaseFeePerGas.mul(2).add(maxPriorityFeePerGas);
+    maxPriorityFeePerGas = feeData.lastBaseFeePerGas
+      ? feeData.lastBaseFeePerGas.eq(0)
+        ? 0
+        : Math.floor(1_500_000_000 * multiplier)
+      : undefined;
+    if (maxPriorityFeePerGas) {
```
alvrs, line 575: this would fail for `maxPriorityFeePerGas === 0`

TheGreatAxios:
Hey @alvrs 

Makes sense :) 

Please feel free to make the changes and move forward if you are able, I'm currently tied up with a number of other things and would appreciate your help on that. 

Thanks again as well for your teams great communication and work on this!

packages/cli/src/utils/deploy.ts
```diff
@@ -542,15 +565,18 @@ export async function deploy(
...
+      maxFeePerGas = feeData.lastBaseFeePerGas?.mul(2).add(maxPriorityFeePerGas);
+    } else {
+      gasPrice = feeData.gasPrice;
+    }
```
alvrs, line 579:
```suggestion
    if(!feeData.lastBaseFeePerGas && !feeData.gasPrice) throw new MUDError("Can not fetch lastBaseFeePerGas from RPC");
    
    if(feeData.lastBaseFeePerGas) {
      if (!feeData.lastBaseFeePerGas.eq(0) && (await signer.getBalance()).eq(0)) {
        throw new MUDError(`Attempting to deploy to a chain with non-zero base fee with an account that has no balance.
If you're deploying to the Lattice testnet, you can fund your account by running 'pnpm mud faucet --address ${await signer.getAddress()}'`);
      }
      maxPriorityFeePerGas = feeData.lastBaseFeePerGas.eq(0) ? 0 : Math.floor(1_500_000_000 * multiplier);
      maxFeePerGas = feeData.lastBaseFeePerGas.mul(2).add(maxPriorityFeePerGas);
    } else {
      if (!feeData.gasPrice.eq(0) && (await signer.getBalance()).eq(0)) {
        throw new MUDError(`Attempting to deploy to a chain with non-zero gas price with an account that has no balance.`);
      }
      gasPrice = feeData.gasPrice;
    }
```

alvrs (commented):
I added some suggested changes, could you please accept them and test if everything works as expected? (I did it via the github interface and didn't test yet)

TheGreatAxios:
@alvrs Just invited you to the forked repo so you can push up directly :) that is fastest otherwise won't be able to get back to this until much later today, thanks!

TheGreatAxios:
@alvrs Sorry - also will try and sneak a quick test in as well just ping me here or in the MUD discord channel (it's TheGreatAxios) there as well and I'll try to run a quick test. If you would prefer to test on your end with SKALE it's fairly simple. 

use the RPC endpoint https://staging-v3.skalenodes.com/v1/staging-fast-active-bellatrix and you can get the gas token (no economic value) called sFUEL from https://sfuel.dirtroad.dev/staging/chaos

Note this rpc and endpoint to get sFUEL is the public testnet endpoint for SKALE that any developer can use to test out the power of SKALE :)

alvrs:
Added the changes, would appreciate if you could test if it works for SKALE!

TheGreatAxios:
Thanks! Will do in the next couple of hours. Appreciate your support.

TheGreatAxios:
@alvrs https://app.warp.dev/block/RO4zpxMOM1CXtmEBFvdhwq

Successful deployment 🎉 

Works on my end! Please do whatever other cleanup or updating to this branch you need and merge when fully approved and ready!

Thank you!!

alvrs (dismissed):

PR #1182, opened 2023-07-19
feat(store,world): combine schema and metadata registration, rename getSchema to getValueSchema, change Schema table id

### LINKED ISSUE #1043: Combine registration of schema with registration of metadata to have tighter constraints and simplify indexer logic
pulled out from #760
> Can we tighten constraints around schema + metadata? i.e. always register key schema + value schema + metadata together (like a regular table), so we can reliably expect them to exist together, and not allow for them to be written more than once. -> yes

### LINKED ISSUE #1122: Combine registration of schema and metadata, add keySchema
* Currently `registerSchema` and `registerMetadata` are two separate methods, making metadata optional and mutable. This leads to a lot of complication in the networking stack and typing. Instead we should combine the two methods, and make metadata immutable.
* Also change the order of parameters to `keySchema`, `valueSchema` instead of the other way around
* Add `keySchema`

### LINKED ISSUE #1134: Turn `SchemaTable` into an codegen table instead of a special case
See see https://github.com/latticexyz/mud/pull/422#discussion_r1119463350 for context

alvrs:
Fixes #1122 
Fixes #1043 
Fixes #1134

This PR includes a few breaking changes:
- Store's registerSchema and setMetadata are combined into a single registerTable method. This means metadata (key names, field names) is immutable and indexers can create tables with this metadata when a new table is registered on-chain.
- World's registerTable method is updated to also include key names and field names, setMetadata is removed
- The getSchema method is renamed to getValueSchema on all interfaces
- The internal Schema table is now an actual table instead of receiving special treatment. It is renamed to Tables, and the table id chagned from "mudstore:schema" to "mudstore:Tables"

We need to integrate these changes into our networking stack before merging, so this PR is blocked on replacing v1 networking code with v2 (otherwise we'd have to integrate these chagnes into the deprecated v1 networking code which will be removed very soon).

Consumers who are using default MUD table libraries don't need to do anything to integrate these breaking changes except update to the latest version and regenerate the table libraries (which happens automatically during the build step of projects started from MUD templates)


packages/store/src/codegen/tables/TableData.sol
```diff
@@ -0,0 +1,536 @@
...
+import { Memory } from "../../Memory.sol";
+import { SliceLib } from "../../Slice.sol";
+import { EncodeArray } from "../../tightcoder/EncodeArray.sol";
+import { Schema, SchemaLib } from "../../Schema.sol";
+import { PackedCounter, PackedCounterLib } from "../../PackedCounter.sol";
+
+bytes32 constant _tableId = bytes32(abi.encodePacked(bytes16("mudstore"), bytes16("TableData")));
+bytes32 constant TableDataTableId = _tableId;
+
+struct TableDataData {
```
holic, line 23: not blocking but "data data" is a little weird

would calling this table `Table` be too weird/confusing?

packages/store/mud.config.ts
```diff
@@ -9,6 +9,17 @@ export default mudConfig({
...
+        tableId: "bytes32",
+      },
+      schema: {
+        keySchema: "bytes32",
+        valueSchema: "bytes32",
+        abiEncodedKeyNames: "bytes",
+        abiEncodedValueNames: "bytes",
+      },
+    },
     StoreMetadata: {
```
holic, line 23: this can go away now, right?

packages/store/src/codegen/tables/TableData.sol
dk1a: we'd get `TableTableId`

dk1a: what about `TableMetadata` or `StoreMetadata`? (since the old table can go away, and the new one can get an extra field - `tableName`)

alvrs:
* main
  * **PR #1174** <a href="https://app.graphite.dev/github/pr/latticexyz/mud/1174" target="_blank">Graphite</a> 
    * **PR #1182** <a href="https://app.graphite.dev/github/pr/latticexyz/mud/1182" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/latticexyz/mud/1182?utm_source=stack-comment).

alvrs:
note to self: rename all `valueNames` to `fieldNames`

packages/store-sync/src/recs/defineInternalComponents.ts
```diff
@@ -4,10 +4,10 @@ import { Table } from "../common";
 // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
 export function defineInternalComponents(world: World) {
   return {
-    TableMetadata: defineComponent<{ table: Type.T }, Metadata, Table>(
+    Tables: defineComponent<{ table: Type.T }, Metadata, Table>(
       world,
       { table: Type.T },
-      { metadata: { componentName: "TableMetadata" } }
+      { metadata: { componentName: "Tables" } }
```
holic, line 10: I think this got renamed by mistake!

holic: Might be worth adjusting this name anyway to like `RegisteredClientTables` or something

holic: That said, I wonder if the now-built-in `Tables` component (derived from mud config) would have everything I need (minus some nice-to-haves like an already parsed table schema).

alvrs: good catch!

.changeset/modern-hornets-jam.md
```diff
@@ -0,0 +1,32 @@
+---
+"@latticexyz/cli": major
+"@latticexyz/store": major
+"@latticexyz/world": major
+"@latticexyz/store-sync": major
+"create-mud": patch
+---
+
+- BREAKING: `Store`'s internal Schema table is now a normal table instead of using special code paths. It is renamed to Tables, and the table id chagned from `mudstore:schema` to `:TableMetaata`
```
holic, line 9:
```suggestion
- BREAKING: `Store`'s internal schema table is now a normal table instead of using special code paths. It is renamed to Tables, and the table ID changed from `mudstore:schema` to `mudstore:Tables`
```

holic: I haven't been prefixing anything with "BREAKING", assuming that the `major` tag does this. Should we be marking changesets like this?

.changeset/modern-hornets-jam.md
```diff
@@ -0,0 +1,32 @@
...
+- BREAKING: `Store`'s registerSchema and setMetadata are combined into a single registerTable method. This means metadata (key names, field names) is immutable and indexers can create tables with this metadata when a new table is registered on-chain.
```
holic, line 10:
```suggestion
- BREAKING: `Store`'s `registerSchema` and `setMetadata` are combined into a single `registerTable` method. This means metadata (key names, field names) is immutable and indexers can create tables with this metadata when a new table is registered on-chain.
```

.changeset/modern-hornets-jam.md
```diff
@@ -0,0 +1,32 @@
...
+  +  function registerTable(
+  +    bytes32 table,
+  +    Schema keySchema,
+  +    Schema valueSchema,
+  +    string[] calldata keyNames,
+  +    string[] calldata fieldNames
+  +  ) external;
+  ```
+
+- BREAKING: `World`'s registerTable method is updated to match the `Store` interface, setMetadata is removed
```
holic, line 26:
```suggestion
- BREAKING: `World`'s `registerTable` method is updated to match the `Store` interface, `setMetadata` is removed
```

.changeset/modern-hornets-jam.md
```diff
@@ -0,0 +1,32 @@
...
+- BREAKING: The `getSchema` method is renamed to `getValueSchema` on all interfaces
+  ```diff
+  - function getSchema(bytes32 table) external view returns (Schema schema);
+  + function getValueSchema(bytes32 table) external view returns (Schema valueSchema);
+  ```
+- The `store-sync` and `cli` packages was updated to integrate the breaking protocol changes. Downstream projects only need to manually integrate these changes if they access low level `Store` or `World` functions.
```
holic, line 32:
```suggestion
- The `store-sync` and `cli` packages are updated to integrate the breaking protocol changes. Downstream projects only need to manually integrate these changes if they access low level `Store` or `World` functions. Otherwise, a fresh deploy with the latest MUD will get you these changes.
```

docs/pages/store/internals.mdx
```diff
@@ -5,9 +5,5 @@ It is called within `StoreRead`'s constructor (which is implemented by `World`).
 
 ### Tables
 
-- `SCHEMA_TABLE` - a very low-level table with no library for it.
-  All table schemas are stored here.
-  Directly using it is not recommended, table libraries generated via `tablegen` have schema methods like `registerSchema`.
-  Alternatively, `StoreCore` exposes `getSchema`, `getKeySchema`, `hasTable`, `registerSchema`, `setMetadata`, most of which are used by table libraries internally.
+- `Tables` - maps tableId to the table's key schema, value schema, key names and field names.
 - `Hooks` - maps tableId to an array of table hooks
```
holic, line 9:
```suggestion
- `Tables` - maps table ID to the table's key schema, value schema, key names, and field names.
- `Hooks` - maps table ID to an array of table hooks
```

packages/store/src/StoreCore.sol
```diff
@@ -32,34 +32,9 @@ library StoreCore {
...
-    StoreMetadata.setMetadata();
-
-    // Set metadata for the schema table
-    string[] memory fieldNames = new string[](2);
-    fieldNames[0] = "valueSchema";
-    fieldNames[1] = "keySchema";
-    StoreMetadata.set(StoreCoreInternal.SCHEMA_TABLE, "schema", abi.encode(fieldNames));
+    // Register internal tables
+    Tables.register();
+    Hooks.register();
```
holic, line 37: :chefkiss:

packages/store/src/StoreCore.sol
```diff
@@ -93,39 +68,46 @@ library StoreCore {
    * Check if a table with the given tableId exists
    */
   function hasTable(bytes32 tableId) internal view returns (bool) {
-    return !StoreCoreInternal._getSchema(tableId).isEmpty();
+    return Tables.getValueSchema(tableId) != bytes32(0);
   }
 
   /**
-   * Register a new tableId schema
+   * Register a new table with key schema, value schema, key names and field names
```
holic, line 75:
```suggestion
   * Register a new table with key schema, value schema, key names, and field names
```

.changeset/modern-hornets-jam.md
alvrs: I saw this in the viem changelog but actually agree that major changes are by definition breaking so this is redundant. (Especially if we change the headlines from patch/minor/major to fix/feature/breaking or something like that)

alvrs:
merging since only comment changes since last approval

Issue #1132, opened 2023-07-10
Tracker: TightCoder


dk1a:
- [x] Revisit `TightCoder` naming - decided to keep it
- [x] ~consider moving it to a separate package (TightCoder doesn't really depend on `store`, except `Slice.sol` which is complicated)~ not worth it - it's useless outside of `store`, and separating it would be complicated (requires Slice and Memory as yet another package)
- [x] ~add a typescript rendering utility in addition to pre-rendered solidity stuff~ making a generalized tight encoder is needlessly complicated; codegen was refactored to use typescript though
- [x] ~(using the utility) try preallocate for all, encodePacked statics, and direct encode dynamics in 
 `store/ts/codegen/record.ts` (see https://github.com/latticexyz/mud/pull/422#discussion_r1119458406)~ making a generalized tight encoder is needlessly complicated; gas implications likely not significant; may be worth exploring again in the future
- [x] Revisit `_encodeToLocation` in `TightCoder` (see https://github.com/latticexyz/mud/pull/422#discussion_r1119465262) https://github.com/latticexyz/mud/pull/1210
- [x] Remove sanity checks in `TightCoder` (see https://github.com/latticexyz/mud/pull/422#discussion_r1119465344)
- [x] ~consider removing solidity stuff, keeping only the js utility (not sure, seems bad)~ making a generalized tight encoder is needlessly complicated

dk1a:
> Revisit TightCoder naming

some proposed alternatives from https://github.com/latticexyz/mud/pull/387:
DataPack, DataPacker, BytePacker, BytesPacker, DataCoder, ByteCoder, Serializer

alvrs:
closing this as since all todos are done

PR #1183, opened 2023-07-19
refactor(store): refactor tightcoder, optimize record encoder

dk1a:
This addresses the major structural issue with tightcoder and full record encoding. It's not everything and I'll PR minor tightcoder changes later so they don't obfuscate gas effects here.
Depends on https://github.com/latticexyz/mud/pull/1190 for a correct gas diff
Fixes parts of https://github.com/latticexyz/mud/issues/1132


dk1a:
why BytePack? that sounds like we are doing new custom byte packing
These libs are hacky workarounds for missing abi functions
The decoder part isn't going away any time soon, but for the encoder ideally I'd PR `abi.encodeTightlyPacked` when I have a couple weeks to understand solidity sources

packages/store/package.json
```diff
@@ -34,7 +34,7 @@
     }
   },
   "scripts": {
-    "build": "pnpm run build:mud && pnpm run build:abi && pnpm run build:typechain && pnpm run build:js",
+    "build": "pnpm run generate-tightcoder && pnpm run build:mud && pnpm run build:abi && pnpm run build:typechain && pnpm run build:js",
```
dk1a, line 37: should this be done differently? (the only other precedent of custom codegen is in `cli`, but that is test-only)

holic:
> why BytePack? that sounds like we are doing new custom byte packing These libs are hacky workarounds for missing abi functions The decoder part isn't going away any time soon, but for the encoder ideally I'd PR `abi.encodeTightlyPacked` when I have a couple weeks to understand solidity sources

Sorry, meant to come back around to this! @alvrs and I had been doing a bunch of brainstorming/chatting around naming (both in terms of self-describing names and "branding") and both liked BytePack for this collection of tools/libraries.

e2e/packages/contracts/src/codegen/tables/NumberList.sol
```diff
@@ -172,9 +172,30 @@ library NumberList {
...
+      _resultPointer := add(_result, 0x20)
+      mstore(0x40, add(_resultPointer, and(add(_resultLength, 31), not(31))))
+      mstore(_result, _resultLength)
+
+      mstore(add(_resultPointer, 0), shl(0, _encodedLengths))
+
+      _resultPointer := add(_resultPointer, 32)
+    }
+    EncodeArray.encodeToLocation((value), _resultPointer);
+    return _result;
```
alvrs, line 198: one general thought: having complex / low level logic like this in code that is produced by codegeneration (even if it's literally just a template string) is gonna be a red flag for folks during the audit. Is there any way you can think of how we could encapsulate this in a general function outside of codegen that can be audited separately, so that codegen code looks as simple as possible?

dk1a: yeah this is an interesting/unusual situation. Essentially I'm doing what solidity does with its IR codegen, but it's scuffed because I don't have access to all the solidity's yul utils. This part of MUD is more a language compiler / addition to solidity, than just a contract.
The encapsulated version would be what we had before this PR. This is not strictly better than before and I have considered this as a reason to reject this PR if you don't think we should do compiler-like yul codegen in MUD

dk1a: TLDR no, this is a critical part of the PR. You can just reject it, I'll make a much simpler PR refactoring ejs to typescript and some other minor things

alvrs: I'd say if this is causing dramatic improvements in gas we should consider it, if the improvements are modest we should optimize for simplicity in the generated code

dk1a:
I think it's not worth it. The gas effects are certainly not dramatic. For now I'll split this up and PR the less contentious stuff, and consider the scary assembly after some more gas tests

PR #1185, opened 2023-07-20
feat(store-sync): sync to sqlite

holic:
pulled out of #1113


packages/store-sync/src/sqlite/sqliteStorage.ts
```diff
@@ -0,0 +1,186 @@
...
+              })
+              .where(eq(sqliteTable.__key, key))
+              .run();
+          }
+        }
+
+        tx.insert(chainState)
+          .values({
+            schemaVersion,
+            chainId: publicClient.chain.id,
```
alvrs, line 173: it seems like the only reason we need a viem `publicClient` in here is for the chain id - would it be sufficient to just pass in the chain id?

packages/store-sync/src/sqlite/sqliteTableToSql.ts
```diff
@@ -0,0 +1,38 @@
...
+import { getTableColumns, getTableName } from "drizzle-orm";
+
+const db = new Kysely<any>({
+  dialect: new SqliteDialect({ database: new SqliteDatabase(":memory:") }),
+});
+
+export function sqliteTableToSql(table: SQLiteTableWithColumns<any>): string {
+  const tableName = getTableName(table);
+
+  // db.schema.dropTable(tableName).ifExists().compile();
```
alvrs, line 13: do we need this or should we remove the comment?

packages/store-sync/src/sqlite/sqliteTableToSql.ts
```diff
@@ -0,0 +1,38 @@
...
+  let query = db.schema.createTable(tableName).ifNotExists();
+
+  const columns = Object.values(getTableColumns(table)) as AnySQLiteColumn[];
+  for (const column of columns) {
+    query = query.addColumn(column.name, column.getSQLType() as ColumnDataType, (col) => {
+      if (column.notNull) {
+        col = col.notNull();
+      }
+      if (column.hasDefault && typeof column.default !== "undefined") {
+        // col = col.defaultTo(column.mapToDriverValue(column.default));
```
alvrs, line 24: to we need this or should we remove the comment?

packages/store-sync/src/sqlite/sqliteStorage.ts
alvrs: gotcha, makes sense

alvrs (approved):

PR #1186, opened 2023-07-20
chore: bump to viem 1.3.1

### LINKED ISSUE #1180: remove abitype version overrides
introduced in https://github.com/latticexyz/mud/pull/1179 to fix a bug in latest viem/abitype

once viem upgrades to abitype >= 0.9.3, we should be able to remove all the overrides in https://github.com/latticexyz/mud/pull/1179

holic:
closes #1180

viem 1.3.1 bumps to the same abitype version we're on, so we no longer need overrides


holic:
gonna land this since it's a ~no-op and I wanna avoid folks making mud apps from templates with these overrides

PR #1189, opened 2023-07-23
feat(world): Add support for composite keys in KeysWithValue module

dhvanipa:
- Added support for up to 5 composite keys in the `KeysWithValue` module. Followed the same approach currently being used in the `KeysInTable` module.
- Updated query in `KeysInTable` to no longer need the helper function `valuesToTuples`
- Added test for composite keys
- Updated doc

### DIFF
--- a/packages/world/src/modules/keyswithvalue/KeysWithValueHook.sol
+++ b/packages/world/src/modules/keyswithvalue/KeysWithValueHook.sol
@@ -9,7 +9,7 @@ import { IBaseWorld } from "../../interfaces/IBaseWorld.sol";
 import { ResourceSelector } from "../../ResourceSelector.sol";
 
 import { MODULE_NAMESPACE } from "./constants.sol";
-import { KeysWithValue } from "./tables/KeysWithValue.sol";
+import { KeysWithValue, KeysWithValueData } from "./tables/KeysWithValue.sol";
 import { ArrayLib } from "../utils/ArrayLib.sol";
 import { getTargetTableSelector } from "../utils/getTargetTableSelector.sol";
 
@@ -19,7 +19,7 @@ import { getTargetTableSelector } from "../utils/getTargetTableSelector.sol";
  * and then replicate logic from solecs's Set.sol.
  * (See https://github.com/latticexyz/mud/issues/444)
  *
- * Note: if a table with composite keys is used, only the first key is indexed
+ * Note: this module only supports up to 5 composite keys.
  */
 contract KeysWithValueHook is IStoreHook {
   using ArrayLib for bytes32[];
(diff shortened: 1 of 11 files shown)


packages/world/src/modules/keyswithvalue/KeysWithValueHook.sol
```diff
@@ -29,45 +29,78 @@ contract KeysWithValueHook is IStoreHook {
     return IBaseWorld(StoreSwitch.getStoreAddress());
   }
 
+  function handleSet(bytes32 tableId, bytes32 valueHash, bytes32[] memory key) internal {
```
alvrs, line 32: nit: should we call this `_addKeyToList`, move to next to `_removeKeyFromList` and align the function signatures?

packages/world/test/KeysWithValueModule.t.sol
```diff
@@ -202,22 +321,62 @@ contract KeysWithValueModuleTest is Test, GasReporter {
...
     endGasReport();
 
     // Assert that the list is correct
     assertEq(keysWithValue.length, 1);
-    assertEq(keysWithValue[0], key1);
+    assertEq(keysWithValue[0].length, 1);
+    assertEq(keysWithValue[0][0], key1);
 
-    // Set a another key with the same value
+    // // Set a another key with the same value
```
alvrs, line 332:
```suggestion
    // Set a another key with the same value
```

packages/world/test/KeysWithValueModule.t.sol
```diff
@@ -202,22 +321,62 @@ contract KeysWithValueModuleTest is Test, GasReporter {
...
     world.setRecord(namespace, sourceName, keyTuple2, abi.encodePacked(value));
 
-    // Get the list of keys with value from the target table
+    // // Get the list of keys with value from the target table
```
alvrs, line 335:
```suggestion
    // Get the list of keys with value from the target table
```

packages/world/test/KeysWithValueModule.t.sol
```diff
@@ -202,22 +321,62 @@ contract KeysWithValueModuleTest is Test, GasReporter {
...
     keysWithValue = getKeysWithValue(world, sourceTableId, abi.encode(value));
 
+    // // Assert that the list is correct
```
alvrs, line 338:
```suggestion
    // Assert that the list is correct
```

alvrs (commented):
Amazing, thank you! Just a couple minor nits, good to merge after those. Could you add a changeset describing this change? (run `pnpm changeset` locally)

alvrs (commented):
Amazing, thank you! Just a couple minor nits, good to merge after those. Could you add a changeset describing this change? (run `pnpm changeset` locally)

packages/world/test/KeysWithValueModule.t.sol
dhvanipa: updated

packages/world/test/KeysWithValueModule.t.sol
dhvanipa: updated

packages/world/test/KeysWithValueModule.t.sol
dhvanipa: updated

dhvanipa:
> Member

changeset added!

alvrs (approved):
great!

alvrs:
oh seems like there are some outdated gas reports in here, would you mind running `pnpm build && pnpm gas-report` at the mud root and commit the results? Also would be good to merge in the latest main, there was a small change to fix the end-to-end test

dhvanipa:
> oh seems like there are some outdated gas reports in here, would you mind running `pnpm build && pnpm gas-report` at the mud root and commit the results? Also would be good to merge in the latest main, there was a small change to fix the end-to-end test

ah updated!

PR #1191, opened 2023-07-25
docs: add a 'view roadmap' button to callout

nickcernera:
- new cover photo, updated style on callout
CleanShot 2023-07-28 at 11 12 05@2x
- new community section
CleanShot 2023-07-28 at 11 12 31@2x
- updated header links
CleanShot 2023-07-28 at 11 12 37@2x


holic:
the button looks weird, almost unstyled:
https://mud-docs-c3zh83nlw-latticexyz.vercel.app/

what if we instead put a link in the top nav for roadmap? seems like we'll want this link around ~forever, not just in the current "unstable" alert

holic (approved):

ludns:
I actually think it's nice to have a big button.
People don't read the top level links. It's important to put it in their face right now. This is the biggest source of confusion with MUD. And we can have it on the "unstable" alert.

holic:
> I actually think it's nice to have a big button. People don't read the top level links. It's important to put it in their face right now. This is the biggest source of confusion with MUD. And we can have it on the "unstable" alert.

It's in three spots now, the top nav, the callout, and the community support section.

ludns:
oh i am referring to the status, which is what i think -- from talking to
users -- is what is creating confusion. i think it's only in the list of
links end of readme.

PR #1194, opened 2023-07-26
refactor(store): optimize Storage library

dk1a:
| | Solidity  | MUD |
| -- | -- | -- |
| write (cold, 1 word) | 22107 | 22339 |
| write (cold, 1 word, partial) | 22136 | 22406 |
| write (cold, 10 words) | 199902 | 199795 |
| write (warm, 1 word) | 107 | 339 |
| write (warm, 1 word, partial) | 236 | 506 |
| write (warm, 10 words) | 1988 | 1795 |
| load (cold, 1 word) | 2109 | 2411 |
| load (cold, 1 word, partial) | 2126 | 2460 |
| load (cold, 10 words) | 19894 | 19911 |
| load (warm, 1 word) | 109 | 414 |
| load (warm, 1 word, partial) | 126 | 462 |
| load (warm, 10 words) | 1897 | 1935 |

surprised I got `storage write (warm, 10 words)` to be more efficient than solidity
but also it doesn't mean anything, this is apples to oranges: https://github.com/latticexyz/mud/blob/b4a68649acdc1f84467f63df150a56750f5cbe72/packages/store/test/Gas.t.sol#L107-L110

Not a fan of how big the relative difference in `load (warm, 1 word)` is but it's kinda unavoidable. Rough estimate of reasons:
- internal function - 40 each, and we need 2 (not counting leftMask, optimizer can inline it) for the code to be readable/auditable
- ~40 for jagged head/tail checks (which solidity can do at compile time); more if they actually exist, see `(warm, 1 word, partial)` for jagged head
- ~145 for a cycle (which solidity can hardcode)

### DIFF
--- a/packages/store/src/Storage.sol
+++ b/packages/store/src/Storage.sol
@@ -9,10 +9,6 @@ import { Memory } from "./Memory.sol";
  * (see https://github.com/latticexyz/mud/issues/444)
  */
 library Storage {
-  function store(uint256 storagePointer, bytes memory data) internal {
-    store(storagePointer, 0, data);
-  }
-
   function store(uint256 storagePointer, bytes32 data) internal {
     assembly {
       sstore(storagePointer, data)
--- a/packages/store/src/Storage.sol
+++ b/packages/store/src/Storage.sol
@@ -27,48 +27,50 @@ library Storage {
    * @notice Stores raw bytes to storage at the given storagePointer and offset (keeping the rest of the word intact)
    */
   function store(uint256 storagePointer, uint256 offset, uint256 memoryPointer, uint256 length) internal {
-    // Support offsets that are greater than 32 bytes by incrementing the storagePointer and decrementing the offset
-    unchecked {
-      storagePointer += offset / 32;
-      offset %= 32;
-    }
-
-    // For the first word, if there is an offset, apply a mask to beginning
     if (offset > 0) {
-      // Get the word's remaining length after the offset
-      uint256 wordRemainder;
-      // (safe because of `offset %= 32` at the start)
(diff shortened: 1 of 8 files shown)


packages/store/src/Storage.sol
```diff
@@ -113,69 +105,77 @@ library Storage {
...
-    // TODO this will probably use less gas via manual memory allocation
-    // (see https://github.com/latticexyz/mud/issues/444)
-    result = new bytes(length);
     uint256 memoryPointer;
+    /// @solidity memory-safe-assembly
     assembly {
+      // Solidity's YulUtilFunctions::roundUpFunction
+      function round_up_to_mul_of_32(value) -> _result {
+        _result := and(add(value, 31), not(31))
+      }
```
dk1a, line 118: can't really make a solidity function for this so figured I'd just copy the one solidity uses internally

dk1a:
> do we have a measurement for "load cold"?

no, that's intentional, the cold margin is similar for read/write and didn't feel necessary to overdo gas metrics; I can add it if u're interested

alvrs (dismissed):
this is great!

alvrs:
> no, that's intentional, the cold margin is similar for read/write and didn't feel necessary to overdo gas metrics; I can add it if u're interested

just feels like something that would be good for completeness, i'm sure people will ask about it

dk1a:
> just feels like something that would be good for completeness, i'm sure people will ask about it

sure will add in a few mins; we shouldn't make these metrics too visible for end users though, that'd be the role of table metrics (which I have yet to add). This stuff is kinda misleading and concerns an internal library most people don't need to understand

PR #1195, opened 2023-07-27
feat(cli,recs,std-client): update RECS components with v2 key/value schemas

holic:
and move custom RECS schema type default from `undefined` to `unknown`

pulled out of #1113


packages/cli/src/render-ts/renderRecsV1Tables.ts
```diff
@@ -18,14 +18,20 @@ export function defineContractComponents(world: World) {
...
       return defineComponent(world, {
-        ${table.fields.map(({ name, recsTypeString }) => `${name}: ${recsTypeString},`).join("")}
+        ${table.fields.map(({ name, recsTypeString }) => `${name}: ${recsTypeString}`).join(",")}
       }, {
-        metadata: { contractId: tableId.toHex(), tableId: tableId.toString() },
-      });
+        id: ${JSON.stringify(tableId.toHex())},
+        metadata: {
+          componentName: ${JSON.stringify(name)},
+          tableName: ${JSON.stringify([namespace, name].join(":"))},
```
alvrs, line 30: why do we need to run `JSON.stringify` on these? Aren't they already strings?

packages/std-client/src/setup/setupMUDV2Network.ts
```diff
@@ -107,11 +107,12 @@ export async function setupMUDV2Network<C extends ContractComponents, S extends
...
   function registerComponent(key: string, component: ContractComponent) {
-    const { contractId, tableId } = component.metadata;
-    if (tableId) {
-      mappings[tableId] = key;
+    if (typeof component.metadata?.tableId === "string") {
+      mappings[component.metadata.tableId] = key;
     } else {
-      mappings[keccak256(contractId)] = key;
+      mappings[
+        keccak256(typeof component.metadata?.contractId === "string" ? component.metadata.contractId : component.id)
```
alvrs, line 114: would this be equivalent?
```suggestion
        keccak256(component.metadata?.contractId ?? component.id)
```

packages/std-client/src/setup/types.ts
```diff
@@ -26,7 +26,7 @@ export type DecodedSystemCall<
   updates: DecodedNetworkComponentUpdate[];
 };
 
-export type ContractComponent = Component<Schema, { contractId: string; tableId?: string }>;
+export type ContractComponent = Component<Schema>;
```
alvrs, line 29: does this break any assumptions of existing code? maybe the ecs browser, i believe it's still in use by sky strife? (cc @Kooshaba)

alvrs (approved):
Lgtm, just small nits/questions. Is this a breaking change in any way?

packages/cli/src/render-ts/renderRecsV1Tables.ts
holic: this is generating JS code and namespace/name are technically user generated, so it's a bit safer to encode to string with `JSON.stringify` rather than wrapping it in quotes

packages/std-client/src/setup/setupMUDV2Network.ts
holic: I had this before, but TS complained about the `contractId` value being `{}` or something? Might be related to `Metadata` default being `Record<string, unknown>`

holic:
> Is this a breaking change in any way?

I don't _think_ so? If you're using plain RECS, this shouldn't have any impact. It was only Sky Strife (they overloaded `component.id` and other things) that would break.

alvrs (approved):

PR #1197, opened 2023-07-27
feat(store-sync): sync to RECS

holic:
pulled out of #1113


packages/store-sync/src/recs/syncToRecs.ts
```diff
@@ -0,0 +1,217 @@
...
+    receipt: TransactionReceipt;
+    operations: StorageOperation<TConfig>[];
+  }> {
+    // Wait for tx to be mined, then find the resulting block storage operations.
+    // We could just do this based on the blockStorageOperations$, but for txs that have no storage operations, we'd never get a value.
+    const receipt = await publicClient.waitForTransactionReceipt({ hash: tx });
+    const operationsForTx$ = blockStorageOperations$.pipe(
+      filter(({ blockNumber }) => blockNumber === receipt.blockNumber),
+      map(({ operations }) => operations.filter((op) => op.log.transactionHash === tx))
+    );
```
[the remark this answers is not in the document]
holic: I think you're right, hmmm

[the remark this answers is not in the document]
holic: updated!

packages/store-sync/src/recs/encodeEntity.ts
```diff
@@ -0,0 +1,19 @@
+import { Entity } from "@latticexyz/recs";
+import { StaticAbiType } from "@latticexyz/schema-type";
+import { encodeAbiParameters } from "viem";
+import { SchemaToPrimitives } from "../common";
+import { hexKeyTupleToEntity } from "./hexKeyTupleToEntity";
+
+export function encodeEntity<TKeySchema extends Record<string, StaticAbiType>>(
```
holic, line 7: I was gonna move these into RECS but it's a bigger refactor and we have to figure out where to draw lines in terms of type definitions (where should `KeySchema` and `ValueSchema` live and, if not centralized, is it okay to duplicate them in a few packages?)

packages/store-sync/src/recs/syncToRecs.ts
```diff
@@ -0,0 +1,227 @@
...
+  config: TConfig;
+  address: Address;
+  // TODO: make this optional and return one if none provided (but will need chain ID at least)
+  publicClient: PublicClient<Transport, Chain>;
+  // TODO: generate these from config and return instead?
+  components: TComponents;
+  initialState?: {
+    blockNumber: bigint | null;
+    tables: (Table & { records: TableRecord[] })[];
+  };
```
holic, line 46: I originally built this using `indexerUrl` that would fetch from the indexer here, but it caused a circular dependency (because of tRPC types).

@alvrs and I talked about moving the indexer code into `store-sync` package and reducing the `store-indexer` package to just a "indexer runner", but now that I am looking back at this, I quite like the flexibility of being able to provide the initial state from ~anywhere, rather than strictly our indexer.

Maybe a good middle ground is exposing a helper in `store-indexer` that is `getInitialState(indexerUrl, chainId, address)` that abstracts over the raw tRPC call:

image

But honestly, this isn't much code, so maybe it's fine how it is?

alvrs: >  I quite like the flexibility of being able to provide the initial state from ~anywhere, rather than strictly our indexer.

agree

> It would be nice to remove the need to know about our transformer, AppRouter type, etc. so maybe an abstraction layer over it is nice.

agree

What if we move it to one package but keep it modular (so users could do it as shown in your example) but also have a lean wrapper around it with simple options for the "default setup"?

holic: okay, took another pass and got it down to an option in `syncToRecs`

image

internally in `syncToRecs`, it does this:

image

holic: updated store-indexer package branch so it's built on top of this one: https://github.com/latticexyz/mud/pull/1198

packages/store-sync/src/recs/encodeEntity.ts
alvrs: After thinking more about this I think it makes sense to keep these utils in here instead of in `recs`. In `recs` land the entity is just a string, while the encoding and decoding based on key schema is a concept on top of it.

packages/store-sync/src/recs/syncToRecs.ts
```diff
@@ -0,0 +1,227 @@
...
+    ...defineInternalComponents(world),
+  };
+
+  const singletonEntity = world.registerEntity({ id: hexKeyTupleToEntity([]) });
+
+  let startBlock = 0n;
+
+  if (initialState != null && initialState.blockNumber != null) {
+    debug("hydrating from initial state to block", initialState.blockNumber);
+    startBlock = initialState.blockNumber + 1n;
```
alvrs, line 92: why `+ 1n`?

packages/store-sync/src/recs/syncToRecs.ts
```diff
@@ -0,0 +1,227 @@
...
+      for (const record of table.records) {
+        const entity = encodeEntity(table.keySchema, record.key);
+        setComponent(component, entity, record.value as ComponentValue);
+
+        recordsProcessed++;
+        setComponent(components.SyncProgress, singletonEntity, {
+          step: SyncStep.SNAPSHOT,
+          message: `Hydrating from snapshot to block ${initialState.blockNumber}`,
+          percentage: (recordsProcessed / numRecords) * 100,
+        });
```
alvrs, line 121: I think we should only update the sync progress every x records (eg 100 or 1000), otherwise every other update event is a progress update event, which might unnecessarily overwhelm the client if there are a lot of records in the initial state (think OPCraft)

packages/store-sync/src/recs/syncToRecs.ts
```diff
@@ -0,0 +1,227 @@
...
+  const latestBlockNumber = new BehaviorSubject<bigint | null>(null);
+  {
+    const sub = latestBlockNumber$.subscribe(latestBlockNumber);
+    world.registerDisposer(() => sub.unsubscribe());
+  }
+
+  const blockLogs$ = latestBlockNumber$.pipe(
+    tap((blockNumber) => {
+      debug("latest block number", blockNumber);
+    }),
```
alvrs, line 148: why does `latestBlockNumber` need to be a `BehaviorSubject`? Could we just update a simple variable in this tap? (Would also save us from having to manage the subscription / registering the disposer in the World)

alvrs (approved):
lgtm! just some minor non-blocking nits

packages/store-sync/src/recs/syncToRecs.ts
holic: I originally had wired up like that, but it felt a little dirty. I think the BehaviorSubject approach is effectively the same thing, minus the additional subscriber, and felt cleaner to me.

packages/store-sync/src/recs/syncToRecs.ts
holic: I did a naive thing for now, but would like to move a bunch of the sync stuff over to a less-blocking approach with `requestIdleCallback` or similar.

packages/store-sync/src/recs/syncToRecs.ts
alvrs: Can you elaborate on what was harder to read / felt dirtier?

To me it seems like the BehaviorSubject is just unnecessary overhead since we only ever read the current value via `.value`. (Like it just feels weird to have all this code for a variable https://github.com/ReactiveX/rxjs/blob/master/src/internal/BehaviorSubject.ts, https://github.com/ReactiveX/rxjs/blob/master/src/internal/Subject.ts, https://github.com/ReactiveX/rxjs/blob/master/src/internal/Observable.ts even if there is little overhead at runtime). The other thing that feels a bit weird is that we're breaking the convention of using the `$` suffix for streams here.

For the point of easier to read, we could still update the `latestBlockNumber` as a variable in a subscription like here instead of in `tap`.

packages/store-sync/src/recs/syncToRecs.ts
```diff
@@ -115,7 +116,7 @@ export async function syncToRecs<
         setComponent(component, entity, record.value as ComponentValue);
 
         recordsProcessed++;
-        if (recordsProcessed % 1000 === 0) {
+        if (recordsProcessed % recordsPerSyncProgressUpdate === 0) {
```
alvrs, line 119: nice

packages/store-sync/src/recs/syncToRecs.ts
holic: chatted in person and ended up moving off of behavior subject

PR #1200, opened 2023-07-27
chore: pin node to 18.16.1

holic:
18.17.0 is breaking our end-to-end tests with some internal URL changes

> Node.js v18.17.0 comes with the latest version of the URL parser, Ada. This update brings significant performance improvements to URL parsing, including enhancements to the url.domainToASCII and url.domainToUnicode functions in node:url.
>
> Ada 2.0 has been integrated into the Node.js codebase, ensuring that all parts of the application can benefit from the
improved performance. Additionally, Ada 2.0 features a significant performance boost over its predecessor, Ada 1.0.4,
while also eliminating the need for the ICU requirement for URL hostname parsing.

image


holic:
I assume we can get back on the latest v18 changes once Playwright updates?

alvrs:
> I assume we can get back on the latest v18 changes once Playwright updates?

Or vite? looks like the function that causes the type error is in vite? (should we open an issue on their end?)

alvrs (approved):

holic:
> Or vite? looks like the function that causes the type error is in vite? (should we open an issue on their end?)

I checked vite CI and looks like their node 18 tests ran against 18.17.0 just fine so I'm not sure it's there

PR #1206, opened 2023-07-27
fix(cli): explicit import of world as type

### LINKED ISSUE #1204: World type incorrectly imported in contractComponents.ts
Client fails to load with the following error in the console:

`Uncaught SyntaxError: The requested module '/node_modules/.vite/deps/@latticexyz_recs.js?v=e822f7f3' does not provide an export named 'World'`

Caused by line 3 in the autogenerated contractComponents.ts:
`import { defineComponent, Type as RecsType, World } from "@latticexyz/recs";`

Changing to the following solves it, but is reverted on each build:
`import { defineComponent, Type as RecsType, type World } from "@latticexyz/recs";`

I have `verbatimModuleSyntax` set in tsconfig.

Version: 2.0.0-main-70e4d8eb

holic:
fixes #1204 
closes #1205
(tried to commit to original PR at #1205 but didn't have repo access)


holic:
cc @pwrstudio

PR #1208, opened 2023-07-28
refactor: combine name and namespace to resource selector in World methods

alvrs:
* remove separate function selectors for namespace/name in World and instead align with Store interfaces on only using bytes32 resourceSelector

* side effect: saving some byte code in the world contract:

| before (contract/size/margin) | after (contract/size/margin) |
| --- | --- |
| CleanShot 2023-08-11 at 16 34 12@2x | CleanShot 2023-08-17 at 00 02 16@2x |


alvrs:
* main
  * **PR #1174** <a href="https://app.graphite.dev/github/pr/latticexyz/mud/1174" target="_blank">Graphite</a> 
    * **PR #1182** <a href="https://app.graphite.dev/github/pr/latticexyz/mud/1182" target="_blank">Graphite</a> 
      * **PR #1208** <a href="https://app.graphite.dev/github/pr/latticexyz/mud/1208" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/latticexyz/mud/1208?utm_source=stack-comment).

packages/common/src/tableIdToHex.ts
```diff
@@ -1,5 +1,6 @@
 import { Hex, stringToHex, concatHex } from "viem";
 
+// TODO: rename to `resourceIdToHex` or `resourceSelectorToHex` since it can be used with other resources than tables
```
holic, line 3: but this is specific to tables (namespace+name), not all resources follow that same naming scheme, right?

.changeset/grumpy-geckos-raise.md
```diff
@@ -0,0 +1,14 @@
+---
+"@latticexyz/cli": major
+"@latticexyz/common": patch
```
holic, line 3:
```suggestion
```

packages/common/src/tableIdToHex.ts
holic: I see now that we've sprinkled this throughout our deploy process etc.

Maybe worth renaming this but still exposing a table ID variant/wrapper that does ~the same thing for clarity when working with tables.

dk1a (approved):
lgtm, just some nits
virtual/override look inconsistent in World (u're not using virtual in some places, and using override for interfaces)

PR #1209, opened 2023-07-28
feat(recs): recs components with built in getters and setters

0xhank:
Introduces Recs components that have built in setters and getters.

The problem is getComponentValue, setComponentValue, etc are pains to write.

This PR allows you to set up components (that are fully backwards compatible) like this:
export const Position = defineCoordComponent(world, { id: "Position" });

And usage is now

```
Position.get(entity)
Position.set(entity, value)
Position.use(entity)
```
### Features:
 - Adds overridable as a field in the options parameter
 - Adds get, getAll, getAllWith, getAllWithout
 - Adds set, update
 - Adds remove, clear
 - Adds has, equals
 - Removes proxy logic from overridable components
 - Adds a React version that includes a `use` hook to replace useComponentValue


alvrs:
I like the improved dev ex of this approach, but would like to avoid using javascript classes. (We're made the decision to move away from them for a couple reasons, here are a couple links for context: https://dev.to/giantmachines/stop-using-javascript-classes-33ij).

I think we can achieve the same improved dev ex with a simple wrapper function like this: 

```ts

function defineComponentClosure<TSchema extends Schema>(component: Component<Schema>) {
  return {
    set: (entity: Entity, value: ComponentValue<TSchema>) {
      return setComponent(component, entity, value);
    }
   // ...
  }
}

```

That way we stay consistent with using functions over classes, and also give users the choice of which approach to use.
(We could even add this to the `defineComponent` function to make it available by default).

Can you elaborate on the advantages of the `react` part of this PR? Is it mainly less characters to type or a different set of features?

0xhank:
> 

This pr uses objects, not classes.

I agree we can add wrappers, but it feels cleaner to add the additional functionality using the Options object instead of nesting wrappers. An override component with getters and setters becomes `defineComponentClosure(defineOverrideComponent(defineComponent(world, schema, options)`, which is much less elegant than `defineComponent(world, schema, {...options, override: true, closure: true})` or even just include the getters and setters by default.

Note that the original component is still exposed as `defineRawComponent` if users still want to use it.

The reason `createReactComponent` exists is to add the pattern of `Component._method()` for the `useComponentValue` hook, exposed as `Component.use(entity)`. This is a separate function so Recs doesn't depend on `@latticexyz/react`.

holic:
closing for now as we're rethinking our ECS abstractions on top of our new config resolver and query API

PR #1210, opened 2023-07-31
refactor(store): update tightcoder codegen, optimize TightCoder library

dk1a:
Extracted from https://github.com/latticexyz/mud/pull/1183
Fixes parts of https://github.com/latticexyz/mud/issues/1132

lumping codegen update and optimizations together cause `TightCoder.sol` gets different arguments and codegen is for the new methods

PR #1213, opened 2023-07-31
docs: prettier

dk1a:
wrong formatting is causing problems in CI


holic (approved):

holic:
We should add a linter/CI check for these things!

dk1a:
we have CI for this, made the PR cause it was failing in https://github.com/latticexyz/mud/pull/1210 because of `world-101.mdx`
dunno why it wasn't caught earlier though, #1210 has nothing to do with docs

dk1a:
https://github.com/latticexyz/mud/actions/runs/5713702115/job/15480350248

PR #1216, opened 2023-07-31
refactor(store): separate SchemaLib into specialized libraries

dk1a:


packages/store/test/SchemaEncodeHelper.sol
```diff
@@ -0,0 +1,68 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
+
+import { SchemaType } from "@latticexyz/schema-type/src/solidity/SchemaType.sol";
+import { Schema, SchemaLib } from "../src/Schema.sol";
+
+/**
+ * Overrides for encode function to simplify tests
+ */
+library SchemaEncodeHelper {
```
dk1a, line 10: these hardcoded methods should only be used in tests; otherwise you'd use a table's autogenerated schema.
so to avoid confusion it seems better to explicitly put them in a `test` dir

alvrs (approved):
lgtm

PR #1219, opened 2023-07-31
feat: docker monorepo build

ludns:


Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+COPY . /app
+WORKDIR /app
+
+ENV PATH="${PATH}:/usr/local/go/bin"
+ENV PATH="${PATH}:/root/.foundry/bin"
+RUN --mount=type=cache,id=pnpm,target=/pnpm/store pnpm install --frozen-lockfile
+RUN pnpm run -r build
+
+FROM builder AS store-indexer
+WORKDIR /app/packages/store-indexer
```
holic, line 35: since this dockerfile is specific to store-indexer, should we put this in `packages/store-indexer/bin`? or does that make the rest of the infra complicated?

Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+EXPOSE 3001
+ENV DEBUG=*
+ENV NODE_ENV=production
```
holic, line 38: you should now be able to add `RPC_HTTP_URL` and `RPC_WS_URL` here

Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+FROM base AS builder
+COPY . /app
+WORKDIR /app
+
+ENV PATH="${PATH}:/usr/local/go/bin"
+ENV PATH="${PATH}:/root/.foundry/bin"
+RUN --mount=type=cache,id=pnpm,target=/pnpm/store pnpm install --frozen-lockfile
+RUN pnpm run -r build
+
+FROM builder AS store-indexer
```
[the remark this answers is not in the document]
roninjin10: Actually looking at this now you could likely Copy in just the buld artifacts and then run `pnpm i --production` to not install dev deps and have a smaller image. This is a nit though if the image is not too large don't need to optimize now

Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+WORKDIR /app/packages/store-indexer
+EXPOSE 3001
+ENV DEBUG=*
+ENV NODE_ENV=production
+CMD [ "pnpm", "start:testnet" ]
```
roninjin10, line 39: You might need to set HOST to 0.0.0.0 ymmv

roninjin10 (changes requested):
additional nit: should add a docker-compose file.   Some of my comments you should definitely do is why I'm requesting changes

Dockerfile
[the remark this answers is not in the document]
roninjin10: If you want to make it super tiny you would use a smaller image than bullseye slim. You don't need very much to run just a node process usually requires more extra tooling to build an app than to run an app.

As an example I build with bullseye-slim. but then run the application with the much tinier `gcr.io/distroless/nodejs18 as example-server-runner` image here https://github.com/roninjin10/stax/blob/main/Dockerfile#L242

I use distroless specifically because it's debian based

Dockerfile
ludns: we are having one top level docker file that builds the entire monorepo, then takes specific artifacts and put them in different images. This dockerfile actually describes two images right now: a global `builder`, and a specific `store-indexer`. Over time we will add more of them, one for each TS service that requires a Docker image.
having one per package is actually worse, given we would need to build the entire monorepo for each of them (instead of looping through all targets -- currently only `store-indexer`, and building + tagging them individually which will leverage the cache of the build stage).

holic (approved):
going to save the other dockerfile optimizations for a later pass once we get this wired into CI

holic (approved):

PR #1227, opened 2023-08-02
feat: Add trpc-panel

roninjin10:
Adds trpc panel which creates a dev playground with documentation to use the server. Useful for devs and creates some automatic documentation for anyones trpc server

https://github.com/iway1/trpc-panel
https://fastify.dev/

Needs to run on it's on route so added fastify. Express is mostly unmaintained and fastify is a faster version of express with pretty much the same API and built by old express maintainers who no longer maintain express.

I looked into adding open API plugin. It would be interesting to add but difficult to add because it would require typing the entire return type with zod which is non trivial since there are generics involved based on shape of the store

TODO

- [ ] Remove the old AppRouter
- [ ] Better descriptions
- [ ] Haven't actually ran the code yet


roninjin10:
* main
  * **PR #1227** <a href="https://app.graphite.dev/github/pr/latticexyz/mud/1227" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/latticexyz/mud/1227?utm_source=stack-comment).

holic:
Can we add this to store-indexer instead of the store-sync package? https://github.com/latticexyz/mud/blob/main/packages/store-indexer/bin/sqlite-indexer.ts#L80-L86

roninjin10:
Makes sense then the only thing that needs to change in this package is adding the zod descriptions and meta

holic:
closing this because it's fairly outdated now and we're moving away from trpc

PR #1232, opened 2023-08-03
feat(store-indexer): use fastify, move trpc to /trpc

holic:
cc @roninjin10 (since this is a prereq for trpc panel)

puts fastify server in front of trpc and moves trpc endpoints to `/trpc`, which will let us add other top-level endpoints like trpc panel, other frontends (grpc) etc


packages/store-indexer/bin/sqlite-indexer.ts
```diff
@@ -77,13 +77,21 @@ createIndexer({
   maxBlockRange: env.MAX_BLOCK_RANGE,
 });
 
-const server = createHTTPServer({
-  middleware: cors(),
-  router: createAppRouter(),
-  createContext: async () => ({
-    storageAdapter: await createStorageAdapter(database),
-  }),
+const server = fastify({
```
roninjin10, line 80:
```suggestion
// @see https://fastify.dev/docs/latest/
const server = fastify({
```

packages/store-indexer/bin/sqlite-indexer.ts
```diff
@@ -77,13 +77,21 @@ createIndexer({
...
+  maxParamLength: 5000,
 });
 
-const { port } = server.listen(env.PORT);
-console.log(`tRPC listening on http://127.0.0.1:${port}`);
+await server.register(import("@fastify/cors"));
+
+server.register(fastifyTRPCPlugin, {
```
roninjin10, line 86:
```suggestion
// @see https://trpc.io/docs/server/adapters/fastify
server.register(fastifyTRPCPlugin, {
```

roninjin10 (commented):
Holy 6 to 20s?!?! That's nuts

alvrs (approved):
let's add a changeset with context for this change (also curious myself, what's the advantage of fastify?)

holic:
> what's the advantage of fastify?

you can think of it like express, but more modern and faster

alvrs (approved):

PR #1233, opened 2023-08-03
feat(network): don't filter historical ephemeral events

smallbraingames:
- can always filter later, but can't unfilter


smallbraingames:
If this is approved before 8/6 I will be mega happy bc launch coming up soon <3

But if not that's also ok

holic (approved):
I'm fine with this but wanna make sure this is compatible with Sky Strife and they have a good workaround.

alvrs:
reminder to add a changeset describing this change

alvrs:
I wonder if this change is still necessary since it's a change to the v1 network stack that will be removed _very soon_ and the v2 stack already syncs ephemeral table events?

smallbraingames:
> I wonder if this change is still necessary since it's a change to the v1 network stack that will be removed _very soon_ and the v2 stack already syncs ephemeral table events?

Yeah true, I don't need it anymore since I'm on new network stack (should have updated here, ty for reminder)

holic:
might be nice to merge for consistency between v1 and v2 though since this'll be the default in v2

PR #1234, opened 2023-08-03
feat(store-indexer,store-sync): make chain optional, configure indexer with RPC

holic:
Needed to make the indexed configurable with just RPC URLs

### DIFF
--- a/packages/store-indexer/bin/sqlite-indexer.ts
+++ b/packages/store-indexer/bin/sqlite-indexer.ts
@@ -17,6 +17,7 @@ import { isNotNull } from "@latticexyz/common/utils";
 
 const possibleChains = Object.values({ ...mudChains, ...chains }) as Chain[];
 
+// TODO: refine zod type to be either CHAIN_ID or RPC_HTTP_URL/RPC_WS_URL
 const env = z
   .object({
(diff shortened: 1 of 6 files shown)


packages/store-indexer/bin/sqlite-indexer.ts
```diff
@@ -30,24 +33,36 @@ const env = z
...
-const chain = possibleChains.find((c) => c.id === env.CHAIN_ID);
-if (!chain) {
-  throw new Error(`Chain ${env.CHAIN_ID} not found`);
-}
+const chain = env.CHAIN_ID ? possibleChains.find((c) => c.id === env.CHAIN_ID) : undefined;
+
+const transports: Transport[] = [
+  env.RPC_WS_URL ? webSocket(env.RPC_WS_URL) : null,
+  env.RPC_HTTP_URL ? http(env.RPC_HTTP_URL) : null,
+].filter(isNotNull);
```
alvrs, line 41: should we throw an error if neither `CHAIN_ID` nor `RPC_HTTP_URL`/`RPC_WS_URL` are set?

alvrs (dismissed):
comment is not blocking, can do followup

PR #1240, opened 2023-08-03
feat(store-sync,store-indexer): consolidate sync logic, add syncToSqlite

### LINKED ISSUE #1142: Add sql based datastore for table data
- As part of the new sync stack we need a way to store data coming through the Store event stream
- Using a SQL based solution would allow us to reuse the same code for clients (eg via sqlite) and typescript based indexers (eg via postgres)

holic:
closes #1142


packages/dev-tools/src/events/EventIcon.tsx
```diff
@@ -1,21 +1,22 @@
...
-    case "StoreSetField":
+    case "SetField":
       return <span className="text-cyan-500 font-bold">+</span>;
-    case "StoreDeleteRecord":
+    case "DeleteRecord":
       return <span className="text-red-500 font-bold">-</span>;
-    case "StoreEphemeralRecord":
-      return <span className="text-violet-400 font-bold">~</span>;
+    // case "EphemeralRecord":
+    //   return <span className="text-violet-400 font-bold">~</span>;
```
alvrs, line 18: is the plan to add `EphemeralRecord` as a separate `StorageOperation` or abstract it one level higher?

alvrs (approved):
nice! 🚢

packages/dev-tools/src/events/EventIcon.tsx
holic: I haven't decided 🙈 any opinions?

PR #1241, opened 2023-08-03
docs: Explain the root namespace

qbzzt:
As requested by @ludns .

### DIFF
--- a/docs/pages/world/world-101.mdx
+++ b/docs/pages/world/world-101.mdx
@@ -45,8 +45,8 @@ The organization of resources within namespaces is used for two different featur
 1. **Access control:** resources in a namespace have “write” access to the other resources within their namespace.
    Currently, having write access only matters for systems interacting with tables: it means these systems can create and edit records within those tables.
 
-   Note that systems in the ROOT namespace are not subject to access control.
-   They can bypass any hooks and write directly to any table in the `World`.
+   Note that systems in the ROOT namespace are able to bypass access control.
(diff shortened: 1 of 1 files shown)


docs/pages/world/world-101.mdx
```diff
@@ -36,11 +36,18 @@ root
...
 ```
 
 The organization of resources within namespaces is used for two different features of MUD:
 
-1. **Access control:** resources in a namespace have “write” access to the other resources within their namespace. Currently, having write access only matters for systems interacting with tables: it means these systems can create and edit records within those tables.
+1. **Access control:** resources in a namespace have “write” access to the other resources within their namespace.
+   Currently, having write access only matters for systems interacting with tables: it means these systems can create and edit records within those tables.
+
+   Note that systems in the ROOT namespace are not subject to access control.
+   They can bypass any hooks and write directly to any table in the `World`.
```
dk1a, line 49: dunno if it's worth mentioning - altough bypassing hooks is possible, it isn't supposed to happen and there're no tools for it. You'd have to manually make a version of `StoreCore.setField` without the hook calls and rewrite any table libs accordingly

qbzzt: It's a security matter, so I think we should be as clear as possible about potential vulnerabilities.

dk1a: yeah ofc, I meant clarifying that you shouldn't expect this to normally happen, it's just a security nuance

qbzzt: I clarified it would be bypassing security.

alvrs (approved):

PR #1249, opened 2023-08-07
feat(common): allow hooking into mudTransportObserver requests

holic:
Adds an `onRequest` option to `mudTransportObserver`, so we can hook into transport requests in a generic way.


holic:
this might be better suited/easier to use as a hook inside `createContract`

PR #1252, opened 2023-08-07
refactor(store): optimize Schema

dk1a:
- Optimized all hot paths (everything except `validate` - only did a few obvious tweaks there)
- `encode` can still be made much better by replacing `SchemaType` with a static length, but that's another PR
- Another `encode` optimization would be to inline it in table libs, but not sure it's worth it, will think about it when I get to table libs

Case study of `encode`'s gas optimizations, this function is particularly interesting. (the screenshot shows code before this PR, changes are mostly 1:1, but within the loop I use gas per average iteration, which depends on Schema composition, and I used somewhat different schemas before and after, expect it to vary by 5-10 gas. 56->47 could be mostly noise for example.)
Note that gas within the loop is more impactful, as it's multiplied by the number of schema fields.
Screenshot 2023-08-07 201658


ludns:
great work!

packages/store/gas-report.json
```diff
@@ -293,29 +293,41 @@
...
+    "test": "testGetNumFields",
+    "name": "get number of all fields from schema",
+    "gasUsed": 34
   },
   {
     "file": "test/Schema.t.sol",
     "test": "testGetNumStaticFields",
     "name": "get number of static fields from schema",
-    "gasUsed": 85
+    "gasUsed": 79
```
dk1a, line 330: you may note that getting both dynamic and static is 2x cheaper than getting them separately. This is wrong and related to https://github.com/latticexyz/mud/issues/1267
I wonder if these tests should be removed to avoid confusion?

holic (approved):

PR #1254, opened 2023-08-07
docs: add new homepage, create Card and CardGrid components

nickcernera:
Main goal is to improve visibility of our resources (status.mud.dev, contribute.mud.dev, community.mud.dev). Added a card grid to accomplish this and split out the "Introduction" tab into a separate page.

CleanShot 2023-08-07 at 19 32 45@2x

Some unrelated notes:
- "Introduction" and "What is MUD?" pages seem redundant, but I think @qbzzt should be the one to address this, not me
- MODE is still frequently referenced throughout the docs

### DIFF
--- a/docs/components/Card.tsx
+++ b/docs/components/Card.tsx
@@ -1,0 +1,61 @@
+import React from "react";
+import { useState } from "react";
+
+const Card = ({ title, text, iconSVG, link }) => {
+  const [hover, setHover] = useState(false);
+
+  const styles = {
+    cardBackground: {
+      display: "flex",
+      padding: "16px",
+      alignItems: "center",
+      gap: "12px",
+      flex: "1 0 0",
+      borderRadius: "4px",
+      border: "1px solid rgba(0, 0, 0, 0.08)",
+      transition: "background-color 0.3s, color 0.3s",
+      backgroundColor: hover ? "#F2F0EE" : "#FAFAF9",
+    },
+    cardContent: {
+      display: "flex",
+      flexDirection: "column",
+      alignItems: "flex-start",
+      gap: "4px",
+      flex: "1 0 0",
+    },
+    icon: {
+      display: "flex",
+      padding: "4px",
+      alignItems: "center",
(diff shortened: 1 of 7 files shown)


nickcernera:
I'm not sure why the deploy failed on Vercel. I was running locally the whole time and testing as I went along.

nickcernera:
Briefly tried to look into it but I am very confused. frolic or alvarious can you help out with this?

docs/components/Card.tsx
```diff
@@ -0,0 +1,61 @@
...
+      gap: "12px",
+      flex: "1 0 0",
+      borderRadius: "4px",
+      border: "1px solid rgba(0, 0, 0, 0.08)",
+      transition: "background-color 0.3s, color 0.3s",
+      backgroundColor: hover ? "#F2F0EE" : "#FAFAF9",
+    },
+    cardContent: {
+      display: "flex",
+      flexDirection: "column",
```
holic, line 21: image

https://developer.mozilla.org/en-US/docs/Web/CSS/flex-direction

Weird... not sure what's going on here.

holic: Ohh, I think because `FlexDirection` is expecting a string const (one of a set of values) rather than an arbitrary string. Try adding `as const` just before the semicolon, e.g.
```ts
const style = {
  ...
} as const;
```

docs/components/Card.tsx
```diff
@@ -0,0 +1,61 @@
...
+    },
+    text: {
+      color: "rgba(0, 0, 0, 0.70)",
+      fontFamily: "Inter",
+      fontSize: "15px",
+      fontStyle: "normal",
+      fontWeight: "400",
+      lineHeight: "20px",
+    },
+  };
```
holic, line 47:
```suggestion
  } as const;
```

holic:
There's a new error now. Lemme know when you're online and I can give you a quick TS tutorial :)

holic (approved):

PR #1214, opened 2023-07-31
feat(create-mud): use new sync packages

holic:
Updates all templates to use new sync packages

Waiting on https://github.com/wagmi-dev/viem/pull/963 so we don't need zero-fee overrides everywhere

PR #1258, opened 2023-08-09
feat(common): replace TableId with tableIdToHex/hexToTableId

holic:
pulled out of #1214


packages/common/src/deprecated/getTableIds.ts
```diff
@@ -0,0 +1,6 @@
+import { TableId } from "./TableId";
+
+/** @deprecated Use `tableIdToHex` and `hexToTableId` instead. */
```
alvrs, line 3: this comment seems wrong; is `getTableIds` deprecated and if so what is the recommended alternative?

alvrs (dismissed):

packages/common/src/deprecated/getTableIds.ts
holic: `getTableIds` is used by the snap sync module in network, which expects `TableIds[]` so it didn't seem worth it to migrate

PR #1261, opened 2023-08-09
feat(common): add createContract, createNonceManager utils

holic:
pulled out of #1214

### DIFF
--- a/packages/common/src/createContract.ts
+++ b/packages/common/src/createContract.ts
@@ -1,0 +1,125 @@
+import {
+  Abi,
+  Account,
+  Address,
+  Chain,
+  GetContractParameters,
+  GetContractReturnType,
+  Hex,
+  PublicClient,
+  SimulateContractParameters,
+  Transport,
+  WalletClient,
+  WriteContractParameters,
+  getContract,
+} from "viem";
+import pQueue from "p-queue";
+import pRetry from "p-retry";
+import { createNonceManager } from "./createNonceManager";
+import { debug as parentDebug } from "./debug";
+
+const debug = parentDebug.extend("createContract");
+
+// copied from viem because it isn't exported
+// TODO: import from viem?
+function getFunctionParameters(values: [args?: readonly unknown[], options?: object]): {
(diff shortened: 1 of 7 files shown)


packages/common/src/createContract.ts
```diff
@@ -0,0 +1,125 @@
...
+} from "viem";
+import pQueue from "p-queue";
+import pRetry from "p-retry";
+import { createNonceManager } from "./createNonceManager";
+import { debug as parentDebug } from "./debug";
+
+const debug = parentDebug.extend("createContract");
+
+// copied from viem because it isn't exported
+// TODO: import from viem?
```
alvrs, line 24: should we open a viem issue for this?

packages/common/src/createContract.ts
```diff
@@ -0,0 +1,125 @@
...
+                    // On nonce errors, reset the nonce and retry
+                    if (nonceManager.shouldResetNonce(error)) {
+                      debug("got nonce error, retrying", error);
+                      await nonceManager.resetNonce();
+                      return;
+                    }
+                    throw error;
+                  },
+                }),
+              { throwOnTimeout: true }
```
holic, line 116: we don't have a timeout configured but `queue.add` has a `Promise<T | void>` return type unless you toggle this on

packages/common/src/createContract.ts
```diff
@@ -0,0 +1,125 @@
...
+          return async (...parameters) => {
+            const { args, options } = getFunctionParameters(parameters as any);
+
+            async function write(): Promise<Hex> {
+              if (!nonceManager.hasNonce()) {
+                await nonceManager.resetNonce();
+              }
+
+              debug("simulating write", functionName, args, options);
+              const { request } = await publicClient.simulateContract({
```
alvrs, line 83: so the `request` include the estimated gas limit correct?

alvrs (approved):

packages/common/src/createContract.ts
holic: image

🤷‍♂️

PR #1265, opened 2023-08-09
docs: Add debugging

qbzzt:


docs/pages/reference.mdx
```diff
@@ -35,3 +35,58 @@
   ```
   pnpm mud set-version --commit <hash> && pnpm install
   ```
+
+## Debugging
+
+### Onchain components
+
+#### Foundry Debugging
```
holic, line 43:
```suggestion
#### Foundry debugging
```

docs/pages/reference.mdx
```diff
@@ -35,3 +35,58 @@
...
+   - Add `console.log` statements inside your Solidity functions.
+
+     ```solidity
+     console.log("newValue:", newValue);
+     ```
+
+1. Run `anvil` manually.
+
+   ```sh
+   anvil -b 1 --block-base-fee-per-gas 0 | tee anvil.log
```
holic, line 65: could optionally shorten the base fee arg
```suggestion
   anvil -b 1 --base-fee 0 | tee anvil.log
```

docs/pages/reference.mdx
```diff
@@ -35,3 +35,58 @@
...
+
+   You do not have to use `tee anvil.log`, but `anvil` logs so much that it is often useful to be able to look for information in the log file after it has scrolled off the screen.
+
+1. Edit `packages/contracts/package.json` to change the `scripts.dev` definition.
+
+   ```
+   "dev": "pnpm mud dev-contracts --tsgenOutput ../client/src/mud --rpc http://127.0.0.1:8545",
+   ```
+
+   Use `127.0.0.1`, there is a weird bug that makes it a problem to use `localhost` is some circumstances.
```
holic, line 76: I'm not sure this is worth noting. I think anvil's host just defaults to 127.0.0.1 (and printed in the logs), so it makes sense that you'd use that URL to access it.

I believe you can get localhost working if you run anvil with `anvil --host localhost`

docs/pages/reference.mdx
```diff
@@ -35,3 +35,58 @@
...
+
+1. Run `pnpm dev` normally.
+
+   Note that because of the way things work, there might be multiple copies of the same `console.log` message.
```
holic, line 80: Can you clarify this?

docs/pages/reference.mdx
```diff
@@ -35,3 +35,58 @@
...
+
+#### MUD devtools
```
holic, line 82:
```suggestion
#### MUD dev tools
```

docs/pages/reference.mdx
```diff
@@ -35,3 +35,58 @@
...
+
+You can also see what is happening from your application by enabling to MUD devtools.
```
holic, line 84:
```suggestion
You can also see what is happening from your application by enabling to MUD dev tools.
```

holic (commented):
Mostly some style nits

docs/pages/reference.mdx
qbzzt: No, I can't :-(. Do you know why that happens?

docs/pages/reference.mdx
qbzzt: It doesn't work, even after upgrading to the latest `anvil`.

```
qbzzt1@penguin:~$ anvil --version
anvil 0.1.0 (1044042 2023-08-10T13:04:26.276364846Z)
qbzzt1@penguin:~$ anvil -b 1 --base-fee 0 --host localhost
error: invalid value 'localhost' for '--host <IP_ADDR>': invalid IP address syntax

For more information, try '--help'.
qbzzt1@penguin:~$ 
```

While `anvil -b 1 --base-fee 0 --host 0.0.0.0` works, it still doesn't let us connect to `localhost`. I think it's some kind of confusion related to IPv6.

docs/pages/reference.mdx
qbzzt: On my system (`vanilla` template, `anvil` version 0.1.0), it produces one line for the original `increment` call, and then 20-21 on subsequent ones (the ones that come from the UI). I don't know what causes this, and I don't know enough to research. Somebody should do it when there's time, but I think we should already publish because it's a useful tool.

qbzzt: I will prettify the language a bit and add a suggestion to run with `uniq`.

qbzzt: I add language suggesting they use `uniq`, which reduces the issue.

qbzzt: I verified the problem still happens with the latest (`main`) version. It looks like our code calls `IncrementSystem.increment` multiple times in calls that get reverted.

This is ugly, but I think it would be best to publish this technique already, and fix things when we have time to figure it out. It is already a useful technique.

docs/pages/reference.mdx
```diff
@@ -35,3 +35,59 @@
...
+1. Run `pnpm dev` normally.
+
+#### MUD dev tools
+
+You can also see what is happening from your application by enabling to MUD dev tools.
+To do so, add this code to the client side of your application:
+
+```javascript
+import { mount as mountDevTools } from "@latticexyz/dev-tools";
+mountDevTools();
```
holic, line 90: FYI the usage of this changed:
https://github.com/latticexyz/mud/blob/main/packages/dev-tools/README.md

holic (changes requested):
we changed a couple things with templates recently so I've left comments where we need to update things

feel free to ping me on discord for another approval once you make the changes

docs/pages/reference.mdx
```diff
@@ -35,3 +35,59 @@
...
+
+   Note that due to a bug, there might be multiple copies of the same `console.log` message.
+   Piping through [`uniq`](https://man7.org/linux/man-pages/man1/uniq.1.html) helps reduce this issue.
+
+   You do not have to use `tee anvil.log`, but `anvil` logs so much that it is often useful to be able to look for information in the log file after it has scrolled off the screen.
+
+1. Edit `packages/contracts/package.json` to change the `scripts.dev` definition.
+
+   ```
+   "dev": "pnpm mud dev-contracts --tsgenOutput ../client/src/mud --rpc http://127.0.0.1:8545",
```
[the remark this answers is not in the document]
qbzzt: Why? This is for development purposes, don't we want to update the client fields when the data schema is update?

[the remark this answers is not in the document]
holic: it's not needed anymore because we infer RECS components + types from the config: https://github.com/latticexyz/mud/blob/main/.changeset/pretty-hotels-drop.md

holic (approved):

PR #1269, opened 2023-08-09
style: change `localhost` to `127.0.0.1`

qbzzt:
We found `localhost` causing trouble in certain cases (e.g. using IPv6) where a service would listen on `127.0.0.1` but `localhost` resolves to something else.

### DIFF
--- a/e2e/packages/sync-test/setup/openClientWithRootAccount.ts
+++ b/e2e/packages/sync-test/setup/openClientWithRootAccount.ts
@@ -7,5 +7,5 @@ dotenv.config({ path: "../contracts/.env" });
 export async function openClientWithRootAccount(page: Page, options?: { modeUrl?: string }) {
   const modeFragment = options?.modeUrl ? `&mode=${options.modeUrl}` : "";
   const privateKeyFragment = `&privateKey=${process.env.PRIVATE_KEY}`;
-  await page.goto(`http://127.0.0.1:3000?cache=false${modeFragment}${privateKeyFragment}`);
+  await page.goto(`http://localhost:3000?cache=false${modeFragment}${privateKeyFragment}`);
 }
--- a/e2e/packages/sync-test/setup/openClientWithRootAccount.ts
+++ b/e2e/packages/sync-test/setup/openClientWithRootAccount.ts
@@ -7,5 +7,5 @@ dotenv.config({ path: "../contracts/.env" });
 export async function openClientWithRootAccount(page: Page, options?: { modeUrl?: string }) {
   const modeFragment = options?.modeUrl ? `&mode=${options.modeUrl}` : "";
   const privateKeyFragment = `&privateKey=${process.env.PRIVATE_KEY}`;
-  await page.goto(`http://localhost:3000?cache=false${modeFragment}${privateKeyFragment}`);
+  await page.goto(`http://127.0.0.1:3000?cache=false${modeFragment}${privateKeyFragment}`);
 }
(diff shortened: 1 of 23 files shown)


packages/services/README.md
```diff
@@ -70,7 +70,7 @@ metrics:
   port: 6060
 ```
 
-7. If you're running with the default localhost / `31337` chain, make sure there is a local node running for the chain you want to connect to. For example, a hardhat node or an anvil node.
+7. If you're running with the default 127.0.0.1 / `31337` chain, make sure there is a local node running for the chain you want to connect to. For example, a hardhat node or an anvil node.
```
holic, line 73:
```suggestion
7. If you're running with the default `foundry` / `31337` chain, make sure there is a local node running for the chain you want to connect to. For example, a hardhat node or an anvil node.
```

packages/services/README.v1.md
```diff
@@ -6,7 +6,7 @@ Every service is a Go stand-alone binary that can be run individually. Entry-poi
 
 Each service exposes a gRPC server and a wrapper HTPP server (for ability to make gRPC wrapped requests from a web client, e.g. TypeScript MUD client). By default the gRPC server runs at the default `PORT` (specified above and in each `main.go` file) and the HTTP server runs at that `PORT + 1`. For example, the snapshot service has a gRPC server exposed on `50061` and a wrapper server is automatically exposed on `50062`.
 
-Each service has specific command-line arguments. Each service requires a connection to an Ethereum node (for same network where your MUD application is deployed on) via a websocket. By default, all websocket connection URL parameters use a `localhost` instance running at port `8545`, so the full URL is `ws://localhost:8545`.
+Each service has specific command-line arguments. Each service requires a connection to an Ethereum node (for same network where your MUD application is deployed on) via a websocket. By default, all websocket connection URL parameters use a `127.0.0.1` instance running at port `8545`, so the full URL is `ws://127.0.0.1:8545`.
```
holic, line 9:
```suggestion
Each service has specific command-line arguments. Each service requires a connection to an Ethereum node (for same network where your MUD application is deployed on) via a websocket. By default, all websocket connection URL parameters use a local instance running at port `8545`, so the full URL is `ws://127.0.0.1:8545`.
```

holic (approved):
just a couple fixes on some docs that we'll prob delete soon, but otherwise looks great

holic (dismissed):

e2e/packages/sync-test/setup/openClientWithRootAccount.ts
```diff
@@ -7,5 +7,5 @@ dotenv.config({ path: "../contracts/.env" });
 export async function openClientWithRootAccount(page: Page, options?: { modeUrl?: string }) {
   const modeFragment = options?.modeUrl ? `&mode=${options.modeUrl}` : "";
   const privateKeyFragment = `&privateKey=${process.env.PRIVATE_KEY}`;
-  await page.goto(`http://localhost:3000?cache=false${modeFragment}${privateKeyFragment}`);
+  await page.goto(`http://127.0.0.1:3000?cache=false${modeFragment}${privateKeyFragment}`);
```
holic, line 10: looks like e2e tests are failing here, might revert this for now

```suggestion
  await page.goto(`http://localhost:3000?cache=false${modeFragment}${privateKeyFragment}`);
```

holic:
I edited my comment above and don't want it to get buried:

I assume this broke because viem listens on localhost by default, which may mean we need to be a little more nuanced about our approach to swapping URLs around (localhost for referring to viem/client and 127.0.0.1 for referring to anvil). Or we tell viem to listen on 127.0.0.1 but not sure that'll work everywhere? Or we see if we can get anvil to listen on localhost?

PR #1271, opened 2023-08-09
feat(common): remove need for tx queue in `createContract`

### LINKED ISSUE #1263: separate tx simulate queue from tx write queue in `createContract`
might be able to get more tx execution speed and avoid nonce issues

though we'll prob still need to throttle it to like 4-5 txs at a time to avoid gas issues (gas increases per tx and sending lots of txs = more gas = higher chance of failure due to insufficient gas between simulate and write)

holic:
closes #1263

I initially thought we needed a queue to handle nonces. But once I separated the transaction simulation (to determine gas) from the nonce handling (incrementing before writing to the mem pool), I discovered we don't even need the queue anymore.

I tried this locally against our testnet and couldn't get nonce issues to appear.

We could improve this further by handling [other mem pool cases](https://github.com/ethereum/go-ethereum/blob/e13fa32cea3497aa311226c62daf21c5847c8e4f/core/txpool/validation.go) and re-simulating as needed (would be nice for failures due to gas price increases).

### DIFF
--- a/.changeset/fluffy-moles-march.md
+++ b/.changeset/fluffy-moles-march.md
@@ -1,0 +1,5 @@
+---
+"@latticexyz/common": patch
+---
+
+- Remove need for tx queue in `createContract`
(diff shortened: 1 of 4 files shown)


alvrs (approved):
nice!

PR #1274, opened 2023-08-10
feat(world): allow transferring ownership of namespaces

### LINKED ISSUE #1016: allow transferring ownership of namespaces
* The World currently allows granting/revoking access to/from namespaces/resources, but not transferring ownership. We should add this here: https://github.com/latticexyz/mud/blob/main/packages/world/src/modules/core/implementations/AccessManagementSystem.sol

johngrantuk:
fixes #1016 

* Add new transferOwner function to AccessManagementSystem
  * Sets new owner using NamespaceOwner.set
  * Grants access to new owner using ResourceAccess.set
  * Revokes access to previous owner using ResourceAccess.deleteRecord
* Register transferOwner to CoreModule functionSelectors
* Add test in World.t.sol


packages/world/src/modules/core/implementations/AccessManagementSystem.sol
```diff
@@ -35,4 +36,20 @@ contract AccessManagementSystem is System {
     // Revoke access from the given resource
     ResourceAccess.deleteRecord(resourceSelector, grantee);
   }
+
+  /**
+   * Transfer ownership of the given namespace to newOwner.
+   * Revoke ResourceAccess for previous owner and grant to newOwner.
+   * Requires the caller to own the namespace.
+   */
+  function transferOwner(bytes16 namespace, address newOwner) public virtual {
```
holic, line 45: We should probably name this `transferOwnership`, since the ownership, not the owner, is the thing being transferred, and also aligns with e.g. [OpenZeppelin's Ownable](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/access/Ownable.sol).

holic (approved):
Other than naming, looks good to me! Will defer to @alvrs in case he's got other API ideas for this.

packages/world/src/modules/core/implementations/AccessManagementSystem.sol
johngrantuk: Yeah, that makes more sense - thanks!

packages/world/test/World.t.sol
```diff
@@ -230,6 +230,14 @@ contract WorldTest is Test, GasReporter {
     world.registerNamespace("test");
   }
 
+  function testTransferNamespace() public {
+    world.registerNamespace("testTransfer");
+    world.transferOwnership("testTransfer", address(1));
+    assertEq(NamespaceOwner.get(world, "testTransfer"), address(1), "new owner should be namespace owner");
+    // Expect the new owner to have access
+    assertEq(ResourceAccess.get(world, "testTransfer", address(1)), true, "new owner should have access");
```
alvrs, line 238: let's add a sanity check for ownership and access before the transfer, and an assertion for the old owner not having access anymore after the transfer

johngrantuk: Done. Note - I didn't add check for original ownership/permission on registration as its already covered in `testRegisterNamespace`.

alvrs:
Should we add one test for the transfer reverting if the caller is not the owner?

alvrs (dismissed):
thank you!

alvrs (approved):

Issue #994, opened 2023-06-05
Add support for multiple namespaces in one config


dk1a:
I talk about tables because I think what works for them will work for systems, tables are more prevalent in the config, and systems have no sync to worry about. But u can see them both in the examples.
The examples are slightly contrived to show what happens when key and name differ

### Current state

`mudConfig` tables have
- Record key which represents the library name. It's used for most client stuff
- `namespace` and `name`, which are used for on-chain `tableId`
- `tableId` which is composed of `namespace`+`name`. It's used for all on-chain stuff

`namespace` is currently global (but https://github.com/latticexyz/mud/pull/971 aims to change it)

Record key and `name` are sometimes used interchangeably (like in store-cache), but they can currently be different, which can lead to problems.

```typescript
export default mudConfig({
  namespace: "",
  tables: {
    CounterTableN: {
      schema: "uint32",
    },
    // Until namespace overrides, CounterTableNamespaced must be manually deployed in PostDeploy
  },
  systems: {
    IncrementSystem: {
      name: "Increment",
      openAccess: true,
    },
  },
  excludeSystems: [
    // Until namespace overrides, this system must be manually deployed in PostDeploy
    "IncrementNamespacedSystem",
  ],
});
```

### Solution 1

Keep table key and name separate and make the separation stronger

1. Explicitly name table key as `tableKey`, and table name as `name`
2. Never use them interchangeably, because they can differ
3. Most client things would use `tableKey`, but some (like subscriptions) would still have `name` around, since sync works with the whole world and not just what's in the config
4. Per-table `namespace` overrides just work (since `namespace`+`name` has to be unique, not `namespace`+`tableKey`, you won't get ambiguity even with an existing world)
5. All on-chain things use `name` as before

```typescript
export default mudConfig({
  namespace: "",
  tables: {
    CounterTableN: "uint32",
    CounterTableNamespaced: {
      // name can't actually be specified for tables, but it will be effectively different from tableKey because <=16 length limit
      // name: "CounterTableN"
      namespace: "namespace",
      schema: "uint32",
    }
  },
  systems: {
    IncrementSystem: {
      name: "Increment",
      openAccess: true,
    },
    IncrementNamespacedSystem: {
      namespace: "namespace",
      name: "Increment",
      openAccess: true,
    }
  },
});
```

### Solution 2

Unify table key and name

1. Remove `name` from table config
2. `name` and `tableKey` can not differ, so have zod check that `tableKey.length <= 16`
[...]

PR #1278, opened 2023-08-10
feat(create-mud): infer recs components from config

holic:
- creates components from config
- infers TS types for components from config
- removes all related codegen


packages/std-client/tsconfig.json
```diff
@@ -14,5 +14,5 @@
     "skipLibCheck": true,
     "jsx": "react-jsx"
   },
-  "include": ["src"]
+  "include": ["src", "../store-sync/src/recs/common.test-d.ts"]
```
alvrs, line 17: for context, why is this needed?

packages/store-sync/src/recs/common.test-d.ts
```diff
@@ -0,0 +1,22 @@
...
+describe("ConfigToRecsComponents", () => {
+  expectTypeOf<ConfigToRecsComponents<typeof storeConfig>["StoreMetadata"]>().toEqualTypeOf<
+    Component<
+      {
+        tableName: RecsType.String;
+        abiEncodedFieldNames: RecsType.String;
+      },
+      {
+        componentName: "StoreMetadata";
+        // TODO: fix config namespace so it comes back as a const
```
alvrs, line 15: should keep this in mind for #994

packages/store/ts/common.ts
```diff
@@ -14,6 +14,12 @@ export type TableRecord<TKeySchema extends KeySchema = KeySchema, TValueSchema e
   value: SchemaToPrimitives<TValueSchema>;
 };
 
+export type ConfigFieldTypeToSchemaAbiType<T extends FieldData<string>> = T extends SchemaAbiType
+  ? T
+  : T extends `${string}[${string}]`
+  ? "uint8[]"
+  : "uint";
```
alvrs, line 21: is this to turn enums into schema types? should line 21 be `uint8`?
```suggestion
export type ConfigFieldTypeToSchemaAbiType<T extends FieldData<string>> = T extends SchemaAbiType
  ? T
  : T extends `${string}[${string}]`
  ? "uint8[]"
  : "uint8";
```

alvrs (dismissed):

packages/std-client/tsconfig.json
holic: oops, this got added by vscode after a rename

packages/store/ts/common.ts
holic: good catch!

PR #1284, opened 2023-08-11
feat(dev-tools): use new sync stack

holic:
- change dev tools to mount with arguments rather than magical globals
- update dev tools to use new sync stack data
- add basic recs support


examples/minimal/packages/client-react/src/mud/setupNetwork.ts
```diff
@@ -28,6 +29,15 @@ export async function setupNetwork() {
     account: burnerAccount,
   });
 
+  const write$ = new Subject<ContractWrite>();
+  const worldContract = createContract({
+    address: networkConfig.worldAddress as Hex,
+    abi: IWorld__factory.abi,
+    publicClient,
+    walletClient: burnerWalletClient,
+    onWrite: (write) => write$.next(write),
```
holic, line 38: wondering if we wanna export something like `write$` from `createContract` rather than the `onWrite` callback where user land has to convert it to an observable

downside is that common will need rxjs as a dep, which is quite heavy

examples/minimal/packages/client-react/src/index.tsx
```diff
@@ -2,18 +2,32 @@ import ReactDOM from "react-dom/client";
...
+      config: storeConfig,
+      publicClient: result.network.publicClient,
+      walletClient: result.network.walletClient,
+      latestBlock$: result.network.latestBlock$,
+      blockStorageOperations$: result.network.blockStorageOperations$,
+      worldAddress: result.network.worldContract.address,
+      worldAbi: result.network.worldContract.abi,
+      write$: result.network.write$,
+      recsWorld: result.network.world,
+    });
```
holic, line 31: @alvrs thoughts on this?

alvrs: looks fine, maybe add a comment linking to https://vitejs.dev/guide/env-and-mode.html near `import.meta.env.DEV`?

.changeset/soft-dryers-invite.md
alvrs, line 1: Multiple changesets in one PR might look a little odd in the global changelog because the same headline (the PR title) will be used multiple times. Would it be feasible to split these three changes to three PRs or are they too intertwined?

examples/minimal/packages/client-react/src/mud/setupNetwork.ts
alvrs: i like the lightweight `onWrite` callback

alvrs (approved):
nice!

.changeset/soft-dryers-invite.md
holic: > Multiple changesets in one PR might look a little odd in the global changelog because the same headline (the PR title)

I think we should instead adapt how we generate the changelog to not emphasize the PR titles so much. I really like being able to encapsulate things in one PR yet publish multiple changesets.

alvrs: Do you have a proposal for how to separate changesets in the [global changelog](https://mud.dev/changelog)? One alternative i see is manually enforcing a h2 headline in the changeset, but it might cause annoying formatting issues in case we miss it somewhere

holic: I quite like how viem looks: https://github.com/wagmi-dev/viem/blob/main/CHANGELOG.md

alvrs: Yeah that's the changesets default for a single package. Since our changesets tend to be more verbose and include context on the change and how to integrate it, the list format felt a little less clear. We could use the commit hash instead of the PR title though, or just accept the same PR title being mentioned multiple times.

alvrs (approved):

.changeset/soft-dryers-invite.md
holic: gonna land this and revisit the changelog and see if there's a way to make this appear nicely

PR #1286, opened 2023-08-11
refactor(common,store): optimize _keyTuple allocation

dk1a:
Some experiments with keyTuple optimization, not very significant gas reduction, but significant code bloat (codegen'd assembly, though not nearly as bad as https://github.com/latticexyz/mud/pull/1183 imo)
Not sure if it's worth it, defer to @alvrs 

`commont.ts` is the main change, `renderTable.ts` is a minor one that can be extracted

### DIFF
--- a/e2e/packages/contracts/src/codegen/tables/Multi.sol
+++ b/e2e/packages/contracts/src/codegen/tables/Multi.sol
@@ -82,15 +82,37 @@ library Multi {
     assembly {
       // Allocate memory
       _keyTuple := mload(0x40)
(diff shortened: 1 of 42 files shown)


holic:
> Not sure if it's worth it, defer to @alvrs

yeah I feel like this maybe isn't worth it for the code complexity (little harder to audit) for just ~100-200 gas per store call?

alvrs:
Thanks for the experiment! Agree we the tradeoff of savings vs increased complexity in codegenerated code is not worth it.

PR #1287, opened 2023-08-11
feat(cli,world): batch register function selectors

### LINKED ISSUE #1280: Proposal: Reduce the number of transactions during MUD deployment by bulk registering the function selectors
I have noticed that the MUD framework emits a significant number of transactions during deployment to the chain, particularly in the function selectors section. Each function selector currently requires a unique transaction, which is not gas efficient. To improve this, I propose implementing a bulk register optimization that would reduce the number of transactions, resulting in cost savings and a reduced risk of deployment failures.

You can find the relevant code for this proposal at the following link:

https://github.com/latticexyz/mud/blob/788de8de9f360dcc65b6c2469b8855502c23582a/packages/cli/src/utils/deploy.ts#L211

CrazyNorman:
Optimized the process of function registration in the deploy phase by changing from individual transaction registration per function to batch registration per system.

Prior to the optimization, each public function within a system required a separate transaction for registration. This led to a significant number of transactions during actual production deployment, resulting in excessive gas fees and an increased risk of failure. In the testing environment, function registration also constituted a major time-consuming aspect of deploying complex on-chain applications.

After this optimization, functions with the same namespace and system are grouped together for registration, reducing the number of transactions required for function registration to match the number of systems. This decrease in transaction count reduces both the gas fee and deployment time needed.

An assumption here is that the number of functions within a single system will not surpass the gas limit of an individual transaction. Hence, automatic transaction batch splitting hasn't been implemented. Take the root functions as an example, with each additional function, there's an increase in parameter size by 8 bytes (4 + 4), storage writes by 40 bytes (4 + 4 + 16 + 16), two storage reads, and relevant logical operations (including simple type conversions and conditional checks). Gas consumption can be estimated as follows: in a hypothetical extreme scenario where a single system contract contains 400 functions, it would consume approximately 12,000,000 Gas – calldata writes: 51,200 Gas (16 * 8 * 400), storage writes: 10,000,000 Gas (400 * 40 / 32 * 20000), storage reads: 1,680,000 Gas (2100 * 2 * 200). Logical calculations are negligible. This totals to about 12,000,000 Gas. The current Block Gas Limit of mainstream EVM chains, such as 30,000,000, exceeds the consumption of 400 functions within a single system, thus avoiding any issues of exceeding the Gas Limit and preventing transaction submission. Therefore, transaction splitting has not been implemented for now. As for the non root functions, system contracts should not contains that much functions too.

If this implementation approach for optimization is acceptable, the PR for batch optimization related to system and table registration can proceed.

fix #1280

### DIFF
--- a/packages/world/src/modules/core/CoreModule.sol
+++ b/packages/world/src/modules/core/CoreModule.sol
@@ -107,7 +107,7 @@ contract CoreModule is IModule, WorldContext {
    * Register function selectors for all CoreSystem functions in the World
    */
   function _registerFunctionSelectors() internal {
-    bytes4[17] memory functionSelectors = [
+    bytes4[19] memory functionSelectors = [
       // --- WorldRegistrationSystem ---
       WorldRegistrationSystem.registerNamespace.selector,
       WorldRegistrationSystem.registerTable.selector,
@@ -117,7 +117,9 @@ contract CoreModule is IModule, WorldContext {
       WorldRegistrationSystem.registerSystemHook.selector,
       WorldRegistrationSystem.registerSystem.selector,
       WorldRegistrationSystem.registerFunctionSelector.selector,
+      WorldRegistrationSystem.registerFunctionSelectors.selector,
       WorldRegistrationSystem.registerRootFunctionSelector.selector,
+      WorldRegistrationSystem.registerRootFunctionSelectors.selector,
       // --- StoreRegistrationSystem ---
       StoreRegistrationSystem.registerSchema.selector,
       StoreRegistrationSystem.setMetadata.selector,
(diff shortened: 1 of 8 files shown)


packages/world/src/modules/core/CoreModule.sol
```diff
@@ -117,7 +117,9 @@ contract CoreModule is IModule, WorldContext {
       WorldRegistrationSystem.registerSystemHook.selector,
       WorldRegistrationSystem.registerSystem.selector,
       WorldRegistrationSystem.registerFunctionSelector.selector,
+      WorldRegistrationSystem.registerFunctionSelectors.selector,
       WorldRegistrationSystem.registerRootFunctionSelector.selector,
+      WorldRegistrationSystem.registerRootFunctionSelectors.selector,
```
holic, line 122: @dk1a @alvrs What do you think of just using the plural here and not the singular?

dk1a: this isn't a hot path so I think it's fine to ditch the singular, a little extra gas during deployment shouldn't be a big deal

holic:
replaced by https://github.com/latticexyz/mud/pull/1470

PR #1314, opened 2023-08-16
refactor(store): use _keyTupleDefinition in encodeKeyTuple

dk1a:
originally `_keyTupleDefinition` rendered a call to `encodeKeyTuple`, but now `_keyTupleDefinition` is inlined and so `encodeKeyTuple` should use it instead

### DIFF
--- a/packages/store/ts/codegen/renderTable.ts
+++ b/packages/store/ts/codegen/renderTable.ts
@@ -135,9 +135,9 @@ library ${libraryName} {
   }
   
   /** Encode keys as a bytes32 array using this table's schema */
-  function encodeKeyTuple(${renderArguments([_typedKeyArgs])}) internal pure returns (bytes32[] memory _keyTuple) {
-    _keyTuple = new bytes32[](${keyTuple.length});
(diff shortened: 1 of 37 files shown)


packages/store/ts/codegen/renderTable.ts
```diff
@@ -135,9 +135,9 @@ library ${libraryName} {
   }
   
   /** Encode keys as a bytes32 array using this table's schema */
-  function encodeKeyTuple(${renderArguments([_typedKeyArgs])}) internal pure returns (bytes32[] memory _keyTuple) {
-    _keyTuple = new bytes32[](${keyTuple.length});
-    ${renderList(keyTuple, (key, index) => `_keyTuple[${index}] = ${renderValueTypeToBytes32(key.name, key)};`)}
+  function encodeKeyTuple(${renderArguments([_typedKeyArgs])}) internal pure returns (bytes32[] memory) {
+    ${_keyTupleDefinition}
+    return _keyTuple;
```
alvrs, line 140: this is the only change here correct?

alvrs (approved):

packages/store/ts/codegen/renderTable.ts
dk1a: yep

PR #1324, opened 2023-08-17
docs: Hooks tutorial

qbzzt:


dk1a (commented):
table hooks are mostly meant to call other tables (which are always as trusted as World) and maybe trusted systems. I feel like rather than reentrancy, this shows how to hack around hook limitations and avoid infinite loops

PR #1338, opened 2023-08-21
feat(store-sync,store-indexer): add postgres support

### LINKED ISSUE #1193: Add support for Postgres in sync stack

holic:
fixes #1193 

Mostly adapted from (i.e. copy pasted) sqlite -> postgres. The main changes are in handling schemas/namespaces (which sqlite doesn't have). I did a bit of refactoring too.

I left it as mostly duplicate code for now until we can figure out good abstractions. Drizzle makes it tricky because there aren't really shared types between types like `SqliteTable` and `PgTable`. And executing queries is different for each flavor/dialect (sync vs async, `run` vs `execute`).


packages/store-sync/src/postgres/columnTypes.ts
```diff
@@ -0,0 +1,31 @@
...
+  customType<{ data: Address; driverData: ByteArray }>({
+    dataType() {
+      return "bytea";
+    },
+    toDriver(data: Address): ByteArray {
+      return hexToBytes(data);
+    },
+    fromDriver(driverData: ByteArray): Address {
+      return getAddress(bytesToHex(driverData));
+    },
```
holic, line 30: Any feelings on whether to store hex as binary in Postgres vs hex strings?

Binary will be more compact and probably easier to do some operations, but less ergonomic/human readable when interacting with the DB directly (probably need to do a bunch of casting). This is all abstracted away when working with the DB in TS.

image

holic: I went ahead with bytes/binary columns everywhere as it felt more Postgres-native, even though some tooling (e.g. TablePlus) isn't set up well for displaying it.

e2e/packages/sync-test/indexerSync.test.ts
```diff
@@ -54,14 +65,21 @@ describe("Sync from indexer", async () => {
     expect(asyncErrorHandler.getErrors()[0]).toContain("error fetching initial state from indexer");
   });
 
-  describe("indexer online", () => {
+  describe.each([["sqlite"], ["postgres"]] as const)("%s indexer", (indexerType) => {
```
alvrs, line 68: very cool 👍

e2e/packages/sync-test/package.json
```diff
@@ -6,6 +6,9 @@
   "scripts": {
     "test": "vitest --run"
   },
+  "dependencies": {
+    "zod": "^3.22.2"
```
alvrs, line 10: nit: i think this could be a dev dependency?

packages/store-indexer/bin/postgres-indexer.ts
```diff
@@ -0,0 +1,137 @@
...
+  prefix: "/trpc",
+  trpcOptions: {
+    router: createAppRouter(),
+    createContext: async () => ({
+      queryAdapter: await createQueryAdapter(database),
+    }),
+  },
+});
+
+await server.listen({ port: env.PORT });
```
alvrs, line 136: should we make the host configurable here or do you want to update that for both sqlite and postgres indexer in a followup PR? (like https://github.com/latticexyz/mud/pull/1219/commits/e70bd42836557c36f0b97b589a187fe3f55e1e5b)

packages/store-sync/src/postgres/buildInternalTables.ts
```diff
@@ -0,0 +1,32 @@
...
+import { transformSchemaName } from "./transformSchemaName";
+import { asAddress, asBigInt, asJson, asNumber } from "./columnTypes";
+
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+export function buildInternalTables() {
+  const schema = pgSchema(transformSchemaName("__mud_internal"));
+  return {
+    chain: schema.table("chain", {
+      // TODO: change schema version to varchar/text?
+      schemaVersion: integer("schema_version").notNull().primaryKey(),
```
alvrs, line 12: is this already in preparation for a protocol type?

alvrs (approved):
looking great! just a couple non-blocking nits/questions

e2e/packages/sync-test/package.json
holic: oh sure, I guess the whole package is ~dev so I wasn't paying too close attention to it

packages/store-indexer/bin/postgres-indexer.ts
holic: oh yep, can add that here

packages/store-sync/src/postgres/buildInternalTables.ts
holic: it's a reflection of the `schemaVersion.ts` file (we have this in the sqlite indexer already) to help detect when the indexer meta tables change (the shape of the indexer DB as opposed to the shape of the MUD tables) so we can recreate them

I opened an issue for a better approach: https://github.com/latticexyz/mud/issues/1362

packages/store-sync/src/blockLogsToStorage.ts
```diff
@@ -8,25 +8,15 @@ import {
...
-  getTables: (opts: {
-    blockNumber: BlockLogs["blockNumber"];
-    tables: Pick<Table, "address" | "namespace" | "name">[];
-  }) => Promise<Table[]>;
-  storeOperations: (opts: {
-    blockNumber: BlockLogs["blockNumber"];
-    operations: StorageOperation<TConfig>[];
-  }) => Promise<void>;
-};
+export type BlockLogsToStorageOptions<TConfig extends StoreConfig = StoreConfig> = StorageAdapter<TConfig>;
```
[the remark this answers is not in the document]
holic: removing this revealed a lot more clean up work, so good call! feels much better now without it

e2e/packages/sync-test/package.json
alvrs: yeah it's not important, just OCD inducing that one dependency is not marked as dev

alvrs (dismissed):

PR #1342, opened 2023-08-21
chore: delete std-client package

holic:
depends on #1339 

need to leave the `package.json` behind until we have an answer from https://github.com/changesets/changesets/discussions/1215


packages/std-client/src/deprecated/systems/ActionSystem/createActionSystem.ts
alvrs, line 1: since there are still a couple projects using the action queue, i wonder if we should keep this in some other package? can mark it as `deprecated` but i feel bad about fully removing it without an alternative option for easily handling optimistic updated. A satisfying replacement could be as simple as a util to define an optimistic update that is removed once the tx is confirmed / all state updates triggered by the tx are reduced.

holic: Folks can still pin to the old version via npm if they need to stick on the old version for now. And that's an easy to add util, but would be curious to find folks actually using it, how they're using it, and make sure we're providing the right abstractions over it.

Alternatively, we could just move this to the RECS package.

alvrs: We can find projects still using it via https://github.com/search?q=createActionSystem+std-client&type=code. It's true that folks can still pin to the old version, but I feel like we should encourage staying up to date with the latest version and make upgrading as easy as possible. The path of least resistance would be moving it to the recs package (and still marking it as deprecated and adding a note for how if no "requirements" are needed there is no need to use the action system, and _maybe_ adding a simple util like https://github.com/latticexyz/mud/blob/ce7125a1b97efd3db47c5ea1593d5a37ba143f64/docs/pages/tutorials/emojimon/players-and-movement.mdx#L303-L315 as alternative for the use case of managing optimistic updates)

holic: I was able to port most of it over: https://github.com/latticexyz/mud/pull/1351

But realizing that tests are failing and I'm not sure when they last passed (because we haven't had this package running tests in CI for at least a year).

alvrs (dismissed):

alvrs (approved):

PR #1343, opened 2023-08-21
chore: delete store-cache package

holic:
need to leave the `package.json` behind until we have an answer from https://github.com/changesets/changesets/discussions/1215


holic:
sorryyy @alvrs, this was a really nice package

alvrs (approved):

PR #1345, opened 2023-08-21
docs: Add the synchronization strategies (high level description)

qbzzt:


qbzzt:
I was wrong about what was needed

PR #1346, opened 2023-08-21
docs: explain how to write tooling for other environments

qbzzt:


docs/pages/other-env.mdx
```diff
@@ -0,0 +1,193 @@
...
+- `bytes32[] key`, which includes all the key schema values for the record.
+- `uint8 schemaIndex`, the number in the data schema of the relevant field.
+  This parameter only appears is only the field specific event, `StoreSetField`.
+- `bytes data`, the data either in the applicable field, or the entire record.
+  This parameter does not appear in `StoreDeleteRecord`.
+
+There are four data event types:
+
+- `StoreSetRecord(bytes32 table, bytes32[] key, bytes data)`, when a record is changed or created.
+- `StoreSetField(bytes32 table, bytes32[] key, uint8 schemaIndex, bytes data)`, when one of the data fields in the schema is changed.
```
dk1a, line 28: you can create a record by setting a single field

docs/pages/other-env.mdx
```diff
@@ -0,0 +1,193 @@
...
+Once the client is in sync, it can listen to blockchain events emitted by the `World` to update the local copy of the state as needed.
+This is a lot more efficient than reading the entire state of the world from the Indexer repeatedly.
+
+## Reading and interpreting MUD events
+
+### Data events
+
+Data events emitted by a `World` only have a single topic, the event type, and these parameters:
+
+- `bytes32 table`, which is a single 32 byte value that contains both the namespace and the tableID.
```
dk1a, line 18:
```suggestion
- `bytes32 table`, which is a single 32 byte value that contains both the namespace and the tableId.
```
we generally go with `Id`/`id` over `ID`

docs/pages/other-env.mdx
```diff
@@ -0,0 +1,193 @@
...
+- `bytes32[] key`, which includes all the key schema values for the record.
+- `uint8 schemaIndex`, the number in the data schema of the relevant field.
+  This parameter only appears is only the field specific event, `StoreSetField`.
```
dk1a, line 21:
```suggestion
  This parameter only appears in the field specific event, `StoreSetField`.
```

docs/pages/other-env.mdx
```diff
@@ -0,0 +1,193 @@
...
+The first word is the table.
+It is two ASCII strings, one in byte 0 and one in byte 16.
+If a string is not 16 characters long, it is padded with zeros.
+The first string is `6d756473746f72650`, which translates to `mudstore`.
+This is the namespace that contains the metadata that MUD needs to function.
+The second string is `Tables`, so we are creating a record in the `Tables` table (adding a table).
+You can use [this calculator](https://www.duplichecker.com/hex-to-text.php) to translate ASCII into strings.
+
+Static fields such as `bytes32` are encoded directly.
+But for dynamic fields, whose length is unknown, the standard is the encode a 32 byte word with the offset to where the field starts.
```
dk1a, line 87:
```suggestion
But for dynamic fields, whose length is unknown, the standard is to encode a 32 byte word with the offset to where the field starts.
```

docs/pages/other-env.mdx
```diff
@@ -0,0 +1,193 @@
...
+      schema: {
+        keySchema: "bytes32",
+        valueSchema: "bytes32",
+        abiEncodedKeyNames: "bytes",
+        abiEncodedFieldNames: "bytes",
+      },
+}
+```
+
+So the first two fields are the key and value schema, respectable, and they are one 32-byte word each.
```
dk1a, line 143:
```suggestion
So the first two fields are the key and value schema, respectively, and they are one 32-byte word each.
```

docs/pages/other-env.mdx
```diff
@@ -0,0 +1,193 @@
...
+- `uint8 schemaIndex`, the number in the data schema of the relevant field.
+  This parameter only appears is only the field specific event, `StoreSetField`.
+- `bytes data`, the data either in the applicable field, or the entire record.
+  This parameter does not appear in `StoreDeleteRecord`.
+
+There are four data event types:
+
+- `StoreSetRecord(bytes32 table, bytes32[] key, bytes data)`, when a record is changed or created.
+- `StoreSetField(bytes32 table, bytes32[] key, uint8 schemaIndex, bytes data)`, when one of the data fields in the schema is changed.
+- `StoreDeleteRecord(bytes32 table, bytes32[] key)`, when a record is deleted.
```
dk1a, line 29: There is no on-chain verification of the record's existence. So you can delete a deleted/not-created record, which is a no-op (not sure if this is important info, but seems not obvious to me)

docs/pages/other-env.mdx
```diff
@@ -0,0 +1,193 @@
...
+- `StoreEphemeralRecord(bytes32 table, bytes32[] key, bytes data)`, when a record is created in an [ephemeral table](/store/advanced-features#ephemeral-tables).
```
dk1a, line 30:
```suggestion
- `StoreEphemeralRecord(bytes32 table, bytes32[] key, bytes data)`, when a record is emitted in an [ephemeral table](/store/advanced-features#ephemeral-tables).
```
I think we're going with an analogy of ephemeral records are a MUD-native alternative to events, so maybe saying `emitted` is better

dk1a (commented):
some nits as I was reading through

docs/pages/other-env.mdx
holic: since this is prose and not code, I would change this to "table ID"

docs/pages/other-env.mdx
```diff
@@ -0,0 +1,193 @@
+# Tooling for other environments
+
+The main client-side feature of MUD is _synchronization_, having a local on-client copy of the up-to-date state of the `World` object.
+To support other environments, such as [Unity](https://unity.com/) or [Unreal Engine](https://www.unrealengine.com/), this is the main functionality that needs to be ported.
+
+The initial synchronization is best done using the [Indexer](indexer) to get the starting state.
+The Indexer is written in TypeScript, but it can be [accessed through HTTP](./indexer#using-http)
```
ludns, line 7: The language the indexer is written in doesn't matter for the consumer.
There exist a library in Typescript to access the indexer, but -- as you mentioned -- under the hood it's JSON over HTTP using GET (for caching maybe?)

docs/pages/other-env.mdx
```diff
@@ -0,0 +1,193 @@
...
+This is the begining of the data, divided into 32 byte words:
+
+| Word number | Data                                                             |
+| ----------: | ---------------------------------------------------------------- |
+|           0 | 6d756473746f726500000000000000005461626c657300000000000000000000 |
+|           1 | 0000000000000000000000000000000000000000000000000000000000000060 |
+|           2 | 00000000000000000000000000000000000000000000000000000000000000a0 |
+
+The first word is the table.
+It is two ASCII strings, one in byte 0 and one in byte 16.
```
ludns, line 79: This is not necessarily the case, it is a convention as of now; but the MUD protocol supports any bytes32 table ID.

alvrs:
turning this into a draft for now until we have more capacity to finish it

PR #1352, opened 2023-08-22
docs: Add /trpc subroute to url indexer documentation and read example

gunnigylfa:
When attempting to follow the [installation](https://mud.dev/indexer#installation) instructions on for the indexer and run the indexer. A call to the indexer yields a 404 response. This is because the indexer client url is missing the `/trpc` subroute in order to be able to communicate with the latest indexer.

This commit: 
- Updates the indexer installation documentation
- Updates the indexer client url in the `read-indexer` example


qbzzt (approved):
Thank you.

holic (approved):
🙏

PR #1353, opened 2023-08-22
docs: add instructions for how to debug server components

qbzzt:


docs/pages/reference.mdx
```diff
@@ -105,3 +105,14 @@ if (import.meta.env.DEV) {
 ```
 
 In the minimal application you get from `pnpm create mud@<version> <app>`, the MUD dev tools are already included.
+
+## MUD server components
```
alvrs, line 109: it seems like this should be part of the `## Debugging` category, so I would expect the headline to be `### Server components` (akin to `### Onchain components`)
```suggestion
### Server components
```

alvrs (dismissed):

qbzzt:
Bypassing requirements because @alvrs already approved, except for one change, which *he* asked for. I assume he'd approve of that change ;-) .

Issue #1222, opened 2023-08-01
Proposal: make Store events independent of schema


alvrs:
# Current situation

Store events:

```solidity
event StoreSetRecord(bytes32 tableId, bytes32[] key, bytes data);
event StoreSetField(bytes32 tableId, bytes32[] key, uint8 schemaIndex, bytes data);
event StoreDeleteRecord(bytes32 tableId, bytes32[] key);
event StoreEphemeralRecord(bytes32 table, bytes32[] key, bytes data);
```

- `StoreSetRecord` is independent of schema
- `StoreSetField` only contains data of a single field and needs context of the schema to know how to update the record
- `StoreDeleteRecord` is independent of schema
- `StoreEphemeralRecord` is independent of schema. I will ignore this event for the rest of this proposal because it is handled identical to `StoreSetRecord`.
- We are missing events for updating partial data of a field (eg `StoreSetPartialField`). This is relevant for updating dynamic fields (eg pushing an element to an array). Currently we need to load the entire field from storage on chain, to update it and then to be able to emit it as part of the `StoreSetField` event. Instead we should have an event that accepts partial field data and sufficient context for indexers / clients to be able to update the dynamic field.
- Since `StoreSetRecord` and `StoreDeleteRecord` are independent of the schema, they can be processed in parallel: a simple event indexer can load these events from all blocks in a block range in parallel and insert them into a database including a reference to the block in which they occurred. In a second step this list of events can be filtered to only include the last event corresponding to each record.
- With `StoreSetField` we can still sync all events in parallel, but the filter step becomes much more complex: we only care about the `StoreSetField` events after the last `StoreSetRecord` event for a given record, but to apply these events to compute the final record value we need context on the record’s schema (because the event includes a `schemaIndex` to indicate which field was updated). The table’s schema is stored in the schema table, and is emitted as a `StoreSetRecord` event before the first event of the new table. In other words, processing the `StoreSetField` event depends on other events earlier in the event stream.

# Proposal

### Change the `StoreSetField` event to a more pure `StoreSpliceRecord` event

- If we change the `StoreSetField` event to include sufficient context to update the record without context of the schema, processing `StoreSetField` events becomes much simpler.

```solidity
event StoreSpliceRecord(bytes32 tableId, bytes32[] key, uint40 start, uint40 deleteCount, bytes data)
```

[...]

PR #1354, opened 2023-08-22
feat(store): add splice events

### LINKED ISSUE #1296: protocol-parser: refactor utils to be based around key schemas and value schemas
Sky Strife has a use case where they store schema-encoded data as bytes into tables and needs to parse it back out

Right now our protocol-parsing utils are a little awkward, because their shape mirrors what's on chain rather than what we primarily use in the client (keySchema, valueSchema), so we should rework that or add utils to go from hex data + keySchema/valueSchema -> key/value

https://github.com/latticexyz/mud/pull/1443 added some utils for this but we should refactor away from the old approach (the schema defined with static/dynamic fields + lengths)

dk1a:
fixes https://github.com/latticexyz/mud/issues/1222
fixes #1296

TODO:
- [x] add tests for recs sync
- [ ] add more tests for splice to make sure sync does the right thing
  - [ ] splicing an empty record
  - [ ] splicing on a unset field
  - [ ] decoding default values for unset fields
- [ ] finalize naming/position of packed counter args on splice event
- [ ] should we return `bytes16` from `Schema.staticDataLength()`? since that's what it's actually stored as?
- [ ] explain start/deleteCount types
- [x] ~~figure out if its gas efficient enough to remove total length from encoded lengths~~ moved to https://github.com/latticexyz/mud/issues/1420
- [x] revisit naming (StoreSpliceStaticRecord -> StoreSpliceStaticData)


dk1a:
The on-chain part is ready
Curious if you've already thought about the sync @alvrs @holic 
The way I'm currently doing recs sync is bad, I guess we'd need to either do a middle bytes layer, or a splice parser that figures out which fields to update (with partial updates being the tricky part)

packages/store-sync/src/blockLogsToStorage.ts
```diff
@@ -151,21 +151,17 @@ export function blockLogsToStorage<TConfig extends StoreConfig = StoreConfig>({
...
             namespace: table.namespace,
             name: table.name,
-            type: "SetField",
+            type: "SpliceRecord",
             key,
-            fieldName,
-            fieldValue,
+            data: log.args.data,
+            start: log.args.start,
+            deleteCount: log.args.deleteCount,
```
alvrs, line 164: Is there a way we could perform the steps to combine the new bytes with an existing record in here? Or would that prevent some optimizations we can do downstream (like encode+splice+decode in a single db tx)?

packages/store/gas-report.json
```diff
@@ -693,7 +693,7 @@
     "file": "test/StoreCoreGas.t.sol",
     "test": "testSetAndGetField",
     "name": "set static field (1 slot)",
-    "gasUsed": 33280
+    "gasUsed": 33581
```
alvrs, line 696: +300 in gas; some increase is expected, is there anything we can do to further optimize this or is this as good as it gets?

packages/store/src/StoreCore.sol
```diff
@@ -554,12 +542,32 @@ library StoreCoreInternal {
...
+    unchecked {
+      // (safe because it's a few uint40 values, which can't overflow uint48)
+      start = valueSchema.staticDataLength() + 32;
+      for (uint8 i; i < dynamicSchemaIndex; i++) {
+        start += encodedLengths.atIndex(i);
+      }
+    }
+    uint256 deleteCount = oldFieldLength;
+    // Emit event to notify indexers
+    emit StoreCore.StoreSpliceRecord(tableId, key, uint48(start), uint40(deleteCount), data);
```
alvrs, line 570: Is it worth generalizing this in an internal util function, since very similar logic is used for `_setDynamicField`, `_pushToDynamicField`, `_popFromDynamicField` and `_setDynamicFieldItem`? I feel like might declutter the code a bit, make it easier to audit and easier to write tests for this logic in isolation

packages/store/gas-report.json
dk1a: I didn't intentionally spend time optimizing gas here, probably can do better when I get to StoreCore

packages/store/src/StoreCore.sol
dk1a: I've been thinking about this for a while now. I think the only efficient way to declutter is generalize it in a renderer function, and render StoreCore. Using solidity functions means more gas because the optimizer won't inline any of it and we'll need to return many vars

dk1a: Or we could not worry about gas I guess, it's probably less than 1000 (haven't tested specifics). Though a bunch of return vars may still be ugly compared to a renderer with conditionals in the relevant places

dk1a: @alvrs do you want declutter in this PR? it feels like it'd be better in a separate one for easier reverts and gas comparisons

alvrs: can try to refactor in a separate PR

packages/store/src/StoreCore.sol
```diff
@@ -533,16 +530,25 @@ library StoreCoreInternal {
...
-
-    // Store the provided value in storage
     uint256 location = _getStaticDataLocation(tableId, key);
     uint256 offset = _getStaticDataOffset(valueSchema, schemaIndex);
+
     Storage.store({ storagePointer: location, offset: offset, data: data });
+
+    // Prepare data for the splice event
+    uint256 start = offset;
+    uint256 deleteCount = valueSchema.atIndex(schemaIndex).getStaticByteLength();
```
alvrs, line 540: i think we should use `data.length` here (since onchain all of `data` is written, independent of whether or not it matches the expected static byte length at this index)

packages/store/src/PackedCounter.sol
```diff
@@ -14,6 +14,9 @@ uint256 constant VAL_BITS = 5 * 8;
 // Maximum value of a 5-byte section
 uint256 constant MAX_VAL = type(uint40).max;
 
+// We use a specific, invalid packed counter (total length != sum of lengths) to represent "unchanged" for the purpose of events and off-chain indexing
+bytes32 constant UNCHANGED_PACKED_COUNTER = bytes32(uint256(1));
+
```
alvrs, line 19: do we still need this? doesn't seem like it's used

alvrs (dismissed):
lgtm, amazing work @dk1a @holic. Let's get this over the finish line. Only thing missing is a changeset.

packages/store/src/PackedCounter.sol
holic: we're not using it yet so feel free to trash, can add later

holic:
added follow up issue for splice tests: https://github.com/latticexyz/mud/issues/1515

PR #1360, opened 2023-08-24
docs: how to use MUD from onchain code

qbzzt:


docs/pages/tutorials/minimal/use-onchain.mdx
```diff
@@ -0,0 +1,255 @@
+# Use MUD from onchain code
+
+In this tutorial you learn how to use a `World` and the systems in it from code that runs on the blockchain.
+You can use this knowledge to add sub-systems to a `World` or to use MUD as part of a bigger dapp.
+
+## Setup
+
+[Create a minimal application](../minimal) and run it.
+
+## Use the `World` from Foundry
```
alvrs, line 10: I feel like a much easier way to interact with a MUD contract from the CLI would be to write a simple forge script (https://book.getfoundry.sh/reference/forge/forge-script) and execute it via the CLI. The actual code would be similar to the functions you're executing via `cast` here, but it would be much less cluttered. 

The thing I want to avoid is that people read this and think it's very complicated to interact with MUD contracts, while there would be a much easier to understand recommended method.

Another thing we can do would be to add a simple CLI command to the MUD cli to read table values that automatically requests the table's schema and handles the encoding/decoding.

docs/pages/tutorials/minimal/use-onchain.mdx
```diff
@@ -0,0 +1,255 @@
...
+The constructor receives the value for it.
+
+```solidity
+function readCounter() public view returns (uint32) {
+  return Counter.get(world);
+}
+```
+
+To get the value of the counter we use the `Counter` library we imported from `codegen/Tables.sol`.
+There are two versions of `get` in that library.
```
alvrs, line 199: How was this table created? I think it would be helpful if the example project included a `mud.config.ts` to generate this or other table libraries

alvrs (commented):
On a higher level: I think this tutorial could be simplified by focussing on forge scripts (https://book.getfoundry.sh/reference/forge/forge-script). A forge script is already very similar to a contract but can be run from the command line, so we can combine the two sections. The first part could be a plain script that fetches the table schema and then fetches the table value, and in the second part you could explain how this process can be simplified by generating a table library and using it in the forge script.

examples/onchain-client/.env
```diff
@@ -0,0 +1,12 @@
...
+# This should usually be excluded via .gitignore and the env vars attached to
+# your deployment enviroment, but we're including this here for ease of local
+# development. Please do not commit changes to this file!
+#
+# A different private key from the one that is used to deploy the world
+PRIVATE_KEY=0x2a871d0798f97d79848a013d4936a73bf4cc922c825d33c1cf7073dff6d409c6
+
+# Assuming this is the first World deployed by the `pnpm dev`.
+# Otherwise you need to change this.
+WORLD_ADDR=0x5FbDB2315678afecb367f032d93F642f64180aa3
```
alvrs, line 12: as a general rule, let's always try to avoid abbreviations
```suggestion
WORLD_ADDRESS=0x5FbDB2315678afecb367f032d93F642f64180aa3
```

docs/pages/tutorials/minimal/use-onchain.mdx
```diff
@@ -0,0 +1,328 @@
+# Use MUD from onchain code
+
+In this tutorial you learn how to use a `World` and the systems in it from code that runs on the blockchain.
```
alvrs, line 3:
```suggestion
In this tutorial you will learn how to use a `World` and the systems in it from Solidity code (via forge scripts in your command line or via deployed contracts).
```

docs/pages/tutorials/minimal/use-onchain.mdx
```diff
@@ -0,0 +1,328 @@
...
+You can use this knowledge to add sub-systems to a `World` or to use MUD as part of a bigger dapp.
```
alvrs, line 4:
```suggestion
You can use this knowledge to add sub-systems to a `World` or to access MUD state in contracts outside the `World`.
```

docs/pages/tutorials/minimal/use-onchain.mdx
```diff
@@ -0,0 +1,328 @@
...
+
+This is the actual script.
+Note that it has access to `Counter.get(world)` to read the data, but to modify it you have to go through a `System`.
+Remember, _there are no secrets on the blockchain_ and therefore we can assume that anybody can read anything at any time after it is posted - so we might as well provide read access directly.
+
+<details>
+
+<summary>Try to bypass security</summary>
+
+There is another script `BypassSecurity`, which attempts to bypass the `System` logic and set `Counter` directly.
```
alvrs, line 147: Should we call this script `AttemptBypassSecurity`? 😄

docs/pages/tutorials/minimal/use-onchain.mdx
```diff
@@ -0,0 +1,328 @@
...
+    Counter.set(world, 0);
+    console.log("Just after trying to reset the counter");
+```
+
+We know the the failure happens in the `Counter.set(world, 0);` line because we see the log entry for the line before it, but not the one for the line immediately after it.
+
+</details>
+
+```solidity
+    // Stop broadcasting transactions from te account
```
alvrs, line 178:
```suggestion
    // Stop broadcasting transactions from the account
```

docs/pages/tutorials/minimal/use-onchain.mdx
```diff
@@ -0,0 +1,328 @@
...
+   . .env
+   ```
+
+1. Deploy the sample contact.
+
+   ```sh copy
+   forge create OnchainClient --private-key $PRIVATE_KEY --constructor-args $WORLD_ADDR
+   ```
+
+1. Set `CLIENT_ADDR` to the address to which the contract is deployed.
```
alvrs, line 204: Where should they set this? In the `.env` file? (Also let's use `CLIENT_ADDRESS`)

docs/pages/tutorials/minimal/use-onchain.mdx
```diff
@@ -0,0 +1,328 @@
...
+It is the version you use in code which is not part of the `World` with the relevant table.
+
+```
+  function add1() public {
+    world.increment();
+  }
+```
+
+When a `System` is in the root namespace of a `World`, its functions are available from the `World` itself.
+This is the way to activate the logic in a `System`.
```
alvrs, line 283:
```suggestion
When a `System` is in the root namespace of a `World`, its functions are available from the `World` itself, so we can call the function on the `World` to execute the `System` logic.
```

docs/pages/tutorials/minimal/use-onchain.mdx
```diff
@@ -0,0 +1,328 @@
...
+1. Increment `Counter` and read the value again.
+   See it is one more.
+
+   ```sh copy
+   cast send --private-key $PRIVATE_KEY $CLIENT_ADDR "add1()"
+   cast call $CLIENT_ADDR "readCounter()"
+   ```
+
+1. Attempt to reset the counter, which bypasses the `System` logic and modifies `Counter` directly.
+   See that this attempt fails.
```
alvrs, line 221:
```suggestion
   See that this attempt fails, since our contract doesn't have the necessary access rights to write to the `Counter` table.
```

docs/pages/tutorials/minimal/use-onchain.mdx
```diff
@@ -0,0 +1,328 @@
...
+```
+
+It fails with this error:
+
+```
+== Logs ==
+  Connecting to world at  0x5FbDB2315678afecb367f032d93F642f64180aa3
+  Just before trying to reset the counter
+Error:
+e64c0348:ROOT_NAMESPACE/Counter
```
alvrs, line 161: The error here is `AccessDenied`. Sad that forge is not able to display that properly, but maybe we can mention it here?

docs/pages/tutorials/minimal/use-onchain.mdx
```diff
@@ -0,0 +1,328 @@
...
+<summary>Try to bypass security</summary>
+
+There is another script `BypassSecurity`, which attempts to bypass the `System` logic and set `Counter` directly.
+You can run it with this command.
+
+```sh copy
+forge script --rpc-url http://localhost:8545 --broadcast script/ScriptClient.s.sol:BypassSecurity
+```
+
+It fails with this error:
```
alvrs, line 154:
```suggestion
It fails with this error, because we don't have the necessary access rights to write to the `Counter` table:
```

docs/pages/tutorials/minimal/use-onchain.mdx
```diff
@@ -0,0 +1,328 @@
...
+
+```
+  function reset() public {
+    Counter.set(world, 0);
+  }
+}
+```
+
+This function attempts to bypass the `System` logic and set the counter value directly.
+As you have already seen, it fails.
```
alvrs, line 293:
```suggestion
As you have already seen, it fails because the contract doesn't have the necessary access rights to write to the `Counter` table.
```

docs/pages/tutorials/minimal/use-onchain.mdx
```diff
@@ -0,0 +1,328 @@
...
+   ```sh copy
+   forge init
+   ```
+
+1. At the root of the Foundry project, create a Node project for the libraries.
+   Solidity libraries are usually published as [npm packages](https://www.npmjs.com/) and ours are no exception.
+
+   ```sh copy
+   pnpm init
+   pnpm install @latticexyz/world@next @latticexyz/store@next @latticexyz/schema-type@next @latticexyz/cli
```
alvrs, line 312:
```suggestion
   pnpm install @latticexyz/world@next @latticexyz/store@next @latticexyz/schema-type@next @latticexyz/cli@next
```

qbzzt:
> for consistency, can we use the same project setup as in eg. the minimal example here 

I would rather use the template, for these reasons:

- It is more likely to be kept up to date.
- It is simpler, which means less time spent on understanding it before we can get the point of the tutorial. For example, if people want to learn how to use code on-chain, they don't need to think about table keys.
- Consistency with the other tutorials, which all use the template.

qbzzt:
> > I would rather use the template
> 
> I agree - can we update the `onchain-client` example to use this setup? (e.g. no `lib` directory but installing dependencies via pnpm. This should be automatically be the case if the project is initialized via `pnpm create mud`)

Yes. Is there a standard way to create a forge project with `pnpm`, or do I just copy and modify [the template `packages.json`](https://github.com/latticexyz/mud/blob/main/templates/react/packages/contracts/package.json)?

qbzzt:
> I agree - can we update the `onchain-client` example to use this setup? (e.g. no `lib` directory but installing dependencies via pnpm. This should be automatically be the case if the project is initialized via `pnpm create mud`)

Consider it done.

qbzzt:
> On a higher level: I think this tutorial could be simplified by focussing on forge scripts (https://book.getfoundry.sh/reference/forge/forge-script). A forge script is already very similar to a contract but can be run from the command line, so we can combine the two sections. The first part could be a plain script that fetches the table schema and then fetches the table value, and in the second part you could explain how this process can be simplified by generating a table library and using it in the forge script.

I changed `onchain-client` to look like a MUD project instead of a Foundry project, but that caused the script to stop working. [Here is the script](https://github.com/latticexyz/mud/blob/230822-use-onchain/examples/onchain-client/script/Script.s.sol).

When I try to run it, I get this error:

```
qbzzt@Oris-MacBook-Pro onchain-client % forge script --rpc-url http://localhost:8545 --broadcast script/ScriptClient.s.sol:ScriptClient
Error: 
Could not canonicalize the target path

Context:
- No such file or directory (os error 2)
```

We need to either:

1. Spend time to figure a solution. I don't know where to look and what to do, so I think it'll require time from a developer.
2. Give up on the script by reverting to [this commit](https://github.com/latticexyz/mud/tree/34f5c72b04a8e77d7a3fa3d395e99f7362463958/examples/onchain-client).

I vote for #2, but it's not my call. @alvrs , what do you think?

qbzzt:
Too many things changed, no longer relevant.

PR #1364, opened 2023-08-25
feat(world): add callFrom entry point

### LINKED ISSUE #1358: Proposal: account delegation via `callFrom`
## Abstract

This proposal introduces a modular mechanism allowing accounts to delegate other accounts to execute functions on their behalf. By modularizing the delegation logic, we can cater to various use cases, from simple unlimited delegations to complex, conditional ones.

## Design Goals

- Enable accounts to authorize other accounts for delegated function calls.
- Provide flexibility for various delegation types and restrictions.
- Aim for minimal complexity in the core protocol to maintain simplicity and modularity.

## Sample use cases

1. **Session accounts**
    - Allow main accounts to delegate a burner wallet with restricted permissions.
    - Set conditions such as the number of calls, specific function access, or duration of the delegation.
    - More use case specific restrictions could be “allow delegation for the duration of this match”
2. **Trusted Forwarde**r**: [[EIP-2771](https://eips.ethereum.org/EIPS/eip-2771)](https://eips.ethereum.org/EIPS/eip-2771)**, where users can authorize their preferred forwarder.
3. **Atomic Operations**: Like ERC20's `approve` + `transferFrom` mechanism.
    - Ex. 1: Let a user delegate a module to call `registerHook` without transferring namespace ownership.
    - Ex. 2: Flash lend an army in an onchain strategy game.
4. **Modular Systems:** Build higher-level abstractions using pre-existing systems. Refer to https://github.com/latticexyz/mud/issues/319

### Overview

- An account can register a delegation in the world.
- The delegation can be specific to a particular caller, or general to all callers who wish to call functions on behalf of the account.
- The delegation can be unlimited or limited. Limited delegations require an associated contract which validate’s a caller’s attempt to perform an actions on behalf of the account.
- Delegated actions are executed via the `callFrom` entry point.

### Implementation

- Any contract implementing the `DelegationCheck` interface can be associated with a limited delegation.
    
    ```solidity
    interface DelegationCheck {
    	function check(address grantor, bytes32 systemId, bytes funcSelectorAndArgs) returns (bool);
    }
    
    contract World {
    	function registerLimitedDelegation(address grantee, address delegationCheck) {
    		Delegations.set({ grantor: msg.sender, grantee: grantee, delegation: delegationCheck });
    	}
    	
    	function revokeDelegation(address grantee) {
    		Delegations.deleteRecord({ grantor: msg.sender, grantee: grantee });
    	}
    }
    ```
    
- For efficiency, a constant (`address(1)`) can represent unlimited delegations, circumventing the need for contract calls in this case.
    
    ```solidity
    address constant UNLIMITED_DELEGATION = address(1);
    
    contract World {
    	function registerUnlimitedDelegation(address grantee) {
    		Delegations.set({ grantor: msg.sender, grantee: grantee, delegation: UNLIMITED_DELEGATION });
    	}
    }
    ```
    
- Once a delegation is registered, the “grantee” can call functions on behalf of the “grantor” using `callFrom`. The world checks the validity of a delegation by calling the delegation check system registered for this grantee/grantor. There may be a different delegation check contract for each grantee, and/or a default delegation check for each grantor.
    
    ```solidity
    contract World {
    	function callFrom(address from, bytes32 systemId, bytes funcSelectorAndArgs) {
    		// Check if there is an explicit delegation check set for the msg.sender
    		Delegation explicitDelegation = Delegations.wrap(Delegations.get({ grantor: from, grantee: msg.sender }));
    		if(explicitDelegation.check(from, msg.sender, systemId, funcSelectorAndArgs)) {
    			// forward the call with `from` as `msgSender`
    		}
    	
    		// Check if `from` has a fallback delegation check set
    		Delegation fallbackDelegation = Delegation.wrap(Delegations.get({ grantor: from, grantee: address(0) }));
    		if(fallbackDelegation.check(from, msg.sender, systemId, funcSelectorAndArgs)) {
    			// forward the call with `from` as `msgSender`
    		}
    		
    		revert NoDelegationFound();
    	}
    }
    
    type Delegation is address;
    using DelegationInstance for Delegation;
    
    library DelegationInstance {
    	function exists(address self) returns (bool) {
    		return self != address(0);
    	}
    
    	function isUnlimited(address self) returns (bool) {
    		return self == UNLIMITED_DELEGATION;
    	}
    
    	function isLimited(address self) returns (bool) {
    		return exists(self) && !isUnlimited(self);
    	}
    
    	function check(address self, address grantor, address grantee, bytes32 systemId, bytes funcSelectorAndArgs) returns (bool) {
    		// Early return if there is no valid delegation
    		if(!exists(self)) return false;
    
    		// Early return if there is an unlimited delegation
    		if(isUnlimited()) return true;
    
    		// Check the delegation if it's a limited delegation
    		// Note: this is pseudo-code; in reality we'd use a library that appends `msg.sender` to the calldata so `msgSender()` works in the system
    		try DelegationCheck(self).check(grantor, systemId, funcSelectorAndArgs) returns (bool success) {
    			return success;
    		} catch {
    			return false;
    		}
    	}
    }
    ```
    
- With this setup in place delegation types such as limiting the number of calls can be added via a module.
    
    ```solidity
    // The DecayingDelegationModule installs a DecayingDelegationSystem and DecayingDelegationTable in the world. It allows users to create delegations that are limited in the number of calls.
    
    // Like other systems, this system is only ever called by the World contract it reads/writes to/from.
    // (It can be called by anyone, but always reads from and writes to the caller)
    contract DecayingDelegationSystem is System, DelegationCheck {
    	function check(address grantee, bytes32 systemId, bytes funcSelectorAndArgs) returns (bool) {
    		uint256 numCallsLeft = DecayingDelegationTable.getNumCalls(_msgSender(), grantee, systemId, keccak256(funcSelectorAndArgs));
    		if(numCallsLeft == 0) {
    			revert NoCallsLeft();
    		}
    		DecayingDelegationTable.set(_msgSender(), grantee, systemId, funcSelectorAndArgs, numCallsLeft - 1);
    		return true;
    	}
    
    	function initializeDelegation(address grantee, bytes32 systemId, bytes funcSelectorAndArgs, uint256 numCalls) {
    		DecayingDelegationTable.set({ 
    			grantor: _msgSender(),
    			grantee: grantee,
    			funcSelectorAndArgsHash: keccak256(funcSelectorAndArgs),
    			numCalls: numCalls
    		});
    	}
    }
    ```
    
    - We can add similar systems for all primitive delegation times (e.g. approve delegation until block number), and more advanced ones like “approve until the current match is finished”
    - A user can create a new delegation using this system by setting this system as the delegation check and calling the system to set the number of calls. The UX of this can be improved using libraries (for the client and contracts).
    
    ```solidity
    world.registerLimitedDelegation(grantee, decayingDelegationSystem);
    world.DecayingDelegationSystem_initializeDelegation(grantee, systemId, funcSelectorAndArgs, numCalls);
    ```
    
    - We could also extend the `registerLimitedDelegation` function to include a parameter for arbitrary bytes, which are passed to the system’s init function. This would simplify the UX of creating and initializing a limited delegation
    
    ```solidity
    function registerLimitedDelegation(address grantee, address delegationCheck, bytes args) {
    	// Store the delegation check contract as before
    	Delegations.set({ grantor: msg.sender, grantee: grantee, delegation: delegationCheck });
    
    	// Initialize the delegation
    	delegationCheck.call(abi.encodePacked(DelegationCheck.init.selector, args, msg.sender));
    }
    ```

## Related issues

- https://github.com/latticexyz/mud/issues/327
- https://github.com/latticexyz/mud/issues/393
- https://github.com/latticexyz/mud/issues/319

alvrs:
fixes #1358
see changeset for details on this change


packages/world/src/interfaces/IWorldErrors.sol
```diff
@@ -10,4 +10,5 @@ interface IWorldErrors {
   error FunctionSelectorExists(bytes4 functionSelector);
   error FunctionSelectorNotFound(bytes4 functionSelector);
   error ModuleAlreadyInstalled(string module);
+  error NoDelegationFound(address grantor, address grantee);
```
holic, line 13:
```suggestion
  error DelegationNotFound(address grantor, address grantee);
```
(for consistency with our other errors)

alvrs:
* main
  * **PR #1372** <a href="https://app.graphite.dev/github/pr/latticexyz/mud/1372" target="_blank">Graphite</a> 
    * **PR #1370** <a href="https://app.graphite.dev/github/pr/latticexyz/mud/1370" target="_blank">Graphite</a> 
      * **PR #1364** <a href="https://app.graphite.dev/github/pr/latticexyz/mud/1364" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/latticexyz/mud/1364?utm_source=stack-comment).

packages/world/src/modules/core/implementations/WorldRegistrationSystem.sol
```diff
@@ -152,4 +153,26 @@ contract WorldRegistrationSystem is System, IWorldErrors {
 
     return worldFunctionSelector;
   }
+
+  /**
+   * Register a delegation from the caller to the given delegatee.
+   */
+  function registerDelegation(
+    address delegatee,
+    bytes32 delegationControl,
```
holic, line 162: took me a bit to realize we switched from address to resource selector/ID, and wondering if we should clarify the naming here to `delegationControlId` or something?

packages/world/src/modules/std-delegations/DisposableDelegationControl.sol
```diff
@@ -0,0 +1,55 @@
...
+
+    if (availableCalls > 0) {
+      // Decrement the number of available calls
+      DisposableDelegations.set({
+        delegator: delegator,
+        delegatee: _msgSender(),
+        resourceSelector: resourceSelector,
+        funcSelectorAndArgsHash: funcSelectorAndArgsHash,
+        availableCalls: availableCalls - 1
+      });
```
holic, line 30: should we delete once we get to zero?

packages/world/src/modules/std-delegations/DisposableDelegationControl.sol
```diff
@@ -0,0 +1,55 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
+
+import { DelegationControl } from "../../DelegationControl.sol";
+import { DisposableDelegations } from "./tables/DisposableDelegations.sol";
+
+contract DisposableDelegationControl is DelegationControl {
```
holic, line 7: I like this better than "decaying" 👍 

Another potential name idea is "callbound" (to go with "timebound")

holic (dismissed):

packages/world/src/modules/std-delegations/DisposableDelegationControl.sol
```diff
@@ -0,0 +1,55 @@
+// SPDX-License-Identifier: MIT
```
holic, line 1: thoughts on just `delegations` instead of `std-delegations` for the dir name?

packages/world/src/modules/std-delegations/constants.sol
```diff
@@ -0,0 +1,12 @@
...
+import { ResourceSelector } from "../../ResourceSelector.sol";
+import { ROOT_NAMESPACE } from "../../constants.sol";
+
+bytes16 constant MODULE_NAME = bytes16("stddelegations.m");
+
+// Disposable delegation
+bytes32 constant DISPOSABLE_DELEGATION = bytes32(abi.encodePacked(ROOT_NAMESPACE, bytes16("disposable.d")));
+
+// Timebound delegation
+bytes32 constant TIMEBOUND_DELEGATION = bytes32(abi.encodePacked(ROOT_NAMESPACE, bytes16("timebound.d")));
```
dk1a, line 12: should it perhaps have a `MODULE_NAMESPACE`?

dk1a (dismissed):

packages/world/src/modules/std-delegations/constants.sol
alvrs: My thinking was that this module should ideally be installed as a root module (so it can install the systems as root systems), because it reduces the gas cost of all systems that read or change state (around 8k for the calls to `DisposableDelegationSystem`)

packages/world/src/modules/std-delegations/CallboundDelegationControl.sol
```diff
@@ -0,0 +1,67 @@
...
+        delegatee: _msgSender(),
+        resourceSelector: resourceSelector,
+        funcSelectorAndArgsHash: funcSelectorAndArgsHash
+      });
+      return true;
+    }
+
+    if (availableCalls > 0) {
+      // Decrement the number of available calls
+      unchecked {
```
holic, line 35: I don't know the consequences of `unchecked` but is it safe for us to wrap this whole statement vs just the decrement?

packages/world/src/modules/std-delegations/constants.sol
holic: I don't think I have my head wrapped around the end-to-end, but are these delegation/callfrom checks _always_ routed through the world? Ultimately wondering if it would be useful to inherit the namespace or allow the namespace to be configured, or if that only has negative consequences (only used by world, only more gas)

alvrs: Yes, they are always routed through the World. The main challenge is that the delegation check system needs to know which table to access. We could make this dynamic and store the namespace in which the module is registered in storage and then load it from storage in the system, but that would cost more gas. A cheaper alternative would be to just use two different modules for root and non-root installations (it only needs to be installed once per world, so the namespace for non-root could also be fixed). For now, I'd just go with providing the root version, because I'd expect this standard module to be installed by default - if we find there's a need for a non-root version (to be installed on worlds that don't have the root version installed by default), we can just add another version of the module later

holic: root only for now as the default/expected/most common path is totally fine by me

packages/world/src/modules/std-delegations/CallboundDelegationControl.sol
dk1a: I don't think it would

alvrs: moved it outside of the block, gas didn't change so seems like unchecked doesn't impact the nested calls, but fells still cleaner to have it explicit

holic (approved):

PR #1366, opened 2023-08-27
feat(store-sync): improve waitForTransaction handling

holic:
Before this would wait for transaction receipt via `waitForTransactionReceipt` (which can timeout/stall) and then check if the mined block has been processed.

Now this caches the last 10 blocks of storage operations and checks if the tx exists, otherwise for each new block it requests the tx receipt (effectively polling) and waits for the first `true` value from the stream (tx was found).


packages/store-sync/src/common.ts
```diff
@@ -119,5 +119,5 @@ export type SyncResult<TConfig extends StoreConfig = StoreConfig> = {
   latestBlockNumber$: Observable<bigint>;
   blockLogs$: Observable<BlockLogs>;
   blockStorageOperations$: Observable<BlockStorageOperations<TConfig>>;
-  waitForTransaction: (tx: Hex) => Promise<{ receipt: TransactionReceipt }>;
+  waitForTransaction: (tx: Hex) => Promise<void>;
```
holic, line 122: Sometimes the tx is found via the processed storage operations, thus I don't have a transaction receipt, so we can't return one here. If downstream usage needs one, they can always do something like:

```ts
await waitForTransaction(tx);
const receipt = publicClient.getTransactionReceipt({ hash: tx });
```

packages/store-sync/src/createStoreSync.ts
```diff
@@ -229,42 +230,45 @@ export async function createStoreSync<TConfig extends StoreConfig = StoreConfig>
...
-    if (lastBlockNumberProcessed == null || lastBlockNumberProcessed < receipt.blockNumber) {
-      debug("waiting for tx block to be processed", tx, receipt.blockNumber);
-      await firstValueFrom(
-        blockStorageOperations$.pipe(filter(({ blockNumber }) => blockNumber >= receipt.blockNumber))
-      );
-    }
-    debug("tx block was processed", tx, receipt.blockNumber);
 
-    return { receipt };
+    await firstValueFrom(hasTransaction$.pipe(filter(identity)));
```
alvrs, line 271: why `filter(identity)`?

alvrs (approved):
one minor question, not blocking

packages/store-sync/src/createStoreSync.ts
alvrs: wouldn't `await firstValueFrom(hasTransactions$)` work?

alvrs: nvm, somehow i didn't realize there is a `filter` and was thinking it's just mapping to the same value again

PR #1370, opened 2023-08-30
refactor(world): separate call utils into `WorldContextProvider` and `SystemCall`

alvrs:
- We previously had a `Call.withSender` util that managed appending the `msg.sender` to the calldata, and took a flag to use `call` or `delegatecall`. Since appending `msg.sender` to the calldata is tightly coupled with `WorldContext`, the utils for calling/delegatecalling with `msg.sender` appended to the calldata were moved to `WorldContextProvider.callWithContext/delegatecallWithContext`.
- `WorldContext` is renamed to `WorldContextConsumer` (to clarify the interaction between `WorldContextProvider` (appending context) and `WorldContextConsumer` (extracting context))
- The `World` previously had an internal `_call` method which handled access control checks and converting a `resourceSelector` into the system's address. Since this logic is useful for other libraries (most notably the delegation checks coming in #1364) it is moved to library functions `SystemCall.call/callWithHooks`
- System hooks' interface changed to take the `resourceSelector` instead of the system address. This seems more useful and simplifies the code. If needed a hook can still load the system's address from storage via `Systems.get(resourceSelector)` (which would be cheap since the storage slot is warm)


packages/world/src/SystemCall.sol
```diff
@@ -0,0 +1,103 @@
...
+    }
+  }
+
+  /**
+   * Call the system at the given namespace and call any hooks registered for it.
+   * If the system is not public, the caller must have access to the namespace or name.
+   * Reverts with the error if the call was not successful.
+   * Else returns any return data.
+   */
+  function callWithHooksOrRevert(
```
alvrs, line 93: this is the previous behavior of `_call`. In addition we now have library functions that return a `success` flag instead of reverting, which is useful if we want to execute more logic if a call failed (eg. will be needed in #1364)

packages/world/src/SystemCall.sol
```diff
@@ -0,0 +1,103 @@
...
+
+library SystemCall {
+  using ResourceSelector for bytes32;
+
+  /**
+   * Call the system at the given namespace.
+   * If the system is not public, the caller must have access to the namespace or name.
+   * Returns the success status and any return data.
+   */
+  function call(
```
holic, line 27: Ohhh this is now a nice entry point for adding a system call event

packages/world/src/utils.sol
```diff
@@ -5,7 +5,7 @@ import { StoreSwitch } from "@latticexyz/store/src/StoreSwitch.sol";
 import { ResourceSelector } from "./ResourceSelector.sol";
 import { SystemRegistry } from "./Tables.sol";
 
-library Utils {
+library SystemUtils {
```
alvrs, line 8: renamed this to make the purpose a bit clearer, and to add more utility functions to the same file. (Unfortunately it's not possible to turn the `SystemUtils.systemNamespace()` function into a free function because it uses `address(this)`)

alvrs: Undid this change - felt unclean to have a mixed utils file, so i moved `revertWithBytes` to its own file. Still unhappy with this file being called just "Utils", but changing that should be its own PR.

alvrs:
* main
  * **PR #1372** <a href="https://app.graphite.dev/github/pr/latticexyz/mud/1372" target="_blank">Graphite</a> 
    * **PR #1370** <a href="https://app.graphite.dev/github/pr/latticexyz/mud/1370" target="_blank">Graphite</a>  👈
      * **PR #1364** <a href="https://app.graphite.dev/github/pr/latticexyz/mud/1364" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/latticexyz/mud/1370?utm_source=stack-comment).

dk1a (dismissed):

alvrs:
Merging since no change since last approval

holic:
👏 this is great

.changeset/quiet-squids-share.md
```diff
@@ -0,0 +1,116 @@
...
+
+- The `World` contract previously had a `_call` method to handle calling systems via their resource selector, performing accesss control checks and call hooks registered for the system.
+
+  ```solidity
+  library SystemCall {
+    /**
+     * Calls a system via its resource selector and perform access control checks.
+     * Does not revert if the call fails, but returns a `success` flag along with the returndata.
+     */
+    function call(
```
holic, line 67: should this have a corresponding `callOrRevert`?

.changeset/quiet-squids-share.md
```diff
@@ -0,0 +1,116 @@
...
+      address caller,
+      bytes32 resourceSelector,
+      bytes memory funcSelectorAndArgs,
+      uint256 value
+    ) internal returns (bool success, bytes memory data);
+
+    /**
+     * Calls a system via its resource selector, perform access control checks and trigger hooks registered for the system.
+     * Does not revert if the call fails, but returns a `success` flag along with the returndata.
```
holic, line 76: I assume hooks are only called on success?

PR #1378, opened 2023-08-31
feat(world): add support for upgrading systems

### LINKED ISSUE #1140: Add support for upgrading systems
- It should be possible for the owner of a namespace to upgrade systems in the namespace
- One important thing to consider is how to handle permissions that have been granted to the previous system address. We probably have to store those permissions in a list to then be able to revoke all of them before granting them to the new system address.

alvrs:
fixes #1140. Context in the changeset.

TODO:
- [x] Add tests for upgrading systems
- [x] ~Decide on whether we should store the list of addresses with access to a certain resource somewhere to be able to remove access for the previous system.~ -> I'm tending towards no since granting a system access is a manual action, so it feels wrong to automatically revoke access and grant it to a new system.

TODO in followup: [Add CLI support for upgrading systems](https://github.com/latticexyz/mud/issues/1382)

### DIFF
--- a/packages/world/src/modules/core/implementations/WorldRegistrationSystem.sol
+++ b/packages/world/src/modules/core/implementations/WorldRegistrationSystem.sol
@@ -62,13 +62,13 @@ contract WorldRegistrationSystem is System, IWorldErrors {
    * If the namespace doesn't exist yet, it is registered.
    * The system is granted access to its namespace, so it can write to any table in the same namespace.
    * If publicAccess is true, no access control check is performed for calling the system.
+   *
+   * Note: this function doesn't check whether a system already exists at the given selector,
+   * making it possible to upgrade systems.
    */
   function registerSystem(bytes32 resourceSelector, WorldContextConsumer system, bool publicAccess) public virtual {
     // Require the name to not be the namespace's root name
     if (resourceSelector.getName() == ROOT_NAME) revert InvalidSelector(resourceSelector.toString());
-
-    // Require the system to not exist yet
-    if (SystemRegistry.get(address(system)) != 0) revert SystemExists(address(system));
 
     // If the namespace doesn't exist yet, register it
     // otherwise require caller to own the namespace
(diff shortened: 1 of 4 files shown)


packages/world/src/modules/core/implementations/WorldRegistrationSystem.sol
```diff
@@ -62,27 +62,46 @@ contract WorldRegistrationSystem is System, IWorldErrors {
...
+   * Note: this function doesn't check whether a system already exists at the given selector,
+   * making it possible to upgrade systems.
    */
   function registerSystem(bytes32 resourceSelector, WorldContextConsumer system, bool publicAccess) public virtual {
     // Require the name to not be the namespace's root name
     if (resourceSelector.getName() == ROOT_NAME) revert InvalidSelector(resourceSelector.toString());
 
-    // Require the system to not exist yet
-    if (SystemRegistry.get(address(system)) != 0) revert SystemExists(address(system));
+    // Require this system to not be registered at a different resource selector yet
```
holic, line 73: ooc why? what if I want the same system in multiple namespaces?

packages/world/src/modules/core/implementations/WorldRegistrationSystem.sol
```diff
@@ -62,27 +62,46 @@ contract WorldRegistrationSystem is System, IWorldErrors {
...
+    if (existingSystem != address(0)) {
+      // Remove the existing system from the system registry
+      SystemRegistry.deleteRecord(existingSystem);
+
+      // Remove the existing system's access to its namespace
+      ResourceAccess.deleteRecord(namespace, existingSystem);
+    } else {
+      // Otherwise, this is a new system, so register its resource type
+      ResourceType.set(resourceSelector, Resource.SYSTEM);
+    }
```
holic, line 104: idle thought: I wonder if it's worth encoding the resource type into the selector, so selectors become something like
```
type (bytes2) + namespace (bytes14) + name (bytes16)
```
this would save a little on storage (but maybe not in any meaningful way if it's only used during registration) and could help during introspection (it's part of the selector/identifier itself rather than a separate lookup)

holic (approved):
implementation looks good, one question around registering the same system at multiple locations (specifically namespaces)

packages/world/src/modules/core/implementations/WorldRegistrationSystem.sol
alvrs: Since access control is based on addresses that would be a security risk (namespace A might have given system X access, now you register system X in namespace B without realising it might mean whoever has access to the system based on namespace A can now also call the system to access namespace B)

packages/world/src/modules/core/implementations/WorldRegistrationSystem.sol
alvrs: agree that might make sense!

dk1a (approved):

packages/world/src/modules/core/implementations/WorldRegistrationSystem.sol
alvrs: Since access control is also used for EOAs it couldn't be in instead of address (since EOAs don't have a registered selector). It would probably be possible to have an additional concept of access control based on the selector, but there are more complexities: if a system is registered at two selectors, it's not possible to infer the selector based on the system address (since there is no 1-1 mapping anymore), so a system would have to provide its own selector for every call that requires access control. That would require the system to be aware of where its registered (currently it doesn't care) and might have more security implications. Overall I think it's not worth the additional complexity to allow a system to be registered multiple times.

PR #1384, opened 2023-09-01
refactor(cli): split up deploy into modules

johngrantuk:
Initial refactor of cli deploy script. Should make it easier to do follow up work - using Viem and create2 deployments
- Added TxHelper class that manages nonce state during deploy/transactions
- Create separate classes for World/CoreModule/Systems/Modules with independent deploy and install functions
- Simplify promises/handling


alvrs:
Thank you for picking this up! One high level feedback i have is we try to use utility functions + simple objects instead of classes, because it tends to produce small chunks of code that are easier to reuse and test. (We should probably note this decision with more context on the reasoning down somewhere). Would you be down to refactor the refactor to use a more functional approach?

packages/cli/src/utils/modules.ts
```diff
@@ -0,0 +1,70 @@
...
+    modules: ModuleConfig;
+    moduleContracts: Record<string, Promise<string>>;
+    tableIds: TableIds;
+  }
+): Promise<number> {
+  console.log(chalk.blue("Installing Modules"));
+  const { confirmations, worldContract, modules, moduleContracts, tableIds } = input;
+  // Install modules - non-blocking for tx, await all at end
+  const modulePromisesNew: Promise<TransactionResponse | TransactionReceipt>[] = [];
+  for (const module of modules) {
```
alvrs, line 40: if we have a `installModule` util we can do something like

```ts
const modulePromises = modules.map(installModule);
await Promise.all(modulePromises);
```

packages/cli/src/utils/systems.ts
```diff
@@ -0,0 +1,192 @@
...
+    accessListSystems: string[];
+  }
+>;
+
+interface FunctionSignature {
+  functionName: string;
+  functionArgs: string;
+}
+
+export async function registerSystems(
```
alvrs, line 24: same here, feels like it would be nice to have a small `registerSystem` and `registerFunctionSelector` util and then use it to create a small `registerSystems` util

packages/cli/src/utils/deploy.ts
```diff
@@ -54,552 +79,151 @@ export async function deploy(
...
+  if (!worldAddress)
+    txConfig.nonce = await installCoreModule({
+      ...txConfig,
+      worldContract,
+      coreModuleAddress: await coreModulePromise,
+    });
 
-  // Await all promises before executing PostDeploy script
-  await Promise.all(promises); // ----------------------------------------------------------------------------------------------
+  txConfig.nonce = await registerNamesSpace({
```
alvrs, line 156: nit
```suggestion
  txConfig.nonce = await registerNamespace({
```

packages/cli/src/utils/world.ts
```diff
@@ -0,0 +1,129 @@
...
+    ? Promise.resolve(ip.worldAddress)
+    : ip.worldContractName
+    ? deployContractByName({ ...ip, contractName: ip.worldContractName })
+    : deployContract({
+        ...ip,
+        contract: { abi: IBaseWorldData.abi, bytecode: WorldData.bytecode, name: "World" },
+      });
+}
+
+export async function registerNamesSpace(
```
alvrs, line 36:
```suggestion
export async function registerNamepace(
```

packages/cli/src/utils/systems/registerFunctionCalls.ts
```diff
@@ -0,0 +1,61 @@
+import { tableIdToHex } from "@latticexyz/common";
+import { System } from "./types";
+import { loadFunctionSignatures, toFunctionSelector } from "./utils";
+import { CallData } from "../utils/types";
+
+export function registerFunctionCalls(input: {
```
alvrs, line 6: should we call this `getRegisterFunctionSelector` to make it consistent with the other function names? Or could even rename all to `get<MethodName>CallData` to make it extra clear (i was surprised for a second that this function doesn't actually `register` yet but just prepares the call by returning the calldata)

packages/cli/src/utils/deploy.ts
```diff
@@ -41,580 +44,202 @@ export async function deploy(
...
-      console.log(chalk.blue(`Registering table ${tableName} at ${namespace}/${name}`));
-
-      // Store the tableId for later use
-      tableIds[tableName] = toResourceSelector(namespace, name);
-
-      // Register table
-      const schemaTypes = Object.values(schema).map((abiOrUserType) => {
-        const { schemaType } = resolveAbiOrUserType(abiOrUserType, mudConfig);
-        return schemaType;
+  // Deploy all contracts - World, Core, Systems, Module. Non-blocking.
```
alvrs, line 76:
```suggestion
  // Deploy the World contract
```

packages/cli/src/utils/deploy.ts
```diff
@@ -41,580 +44,202 @@ export async function deploy(
...
-
-  await setInternalFeePerGas(priorityFeeMultiplier);
+  const txParams = await setInternalFeePerGas(signer, priorityFeeMultiplier);
 
-  // Catch all to await any promises before exiting the script
-  let promises: Promise<unknown>[] = [];
+  const txConfig = {
+    ...txParams,
+    signer,
+    debug: !!debug,
```
alvrs, line 67:
```suggestion
    debug: Boolean(debug),
```

packages/cli/src/utils/modules/constants.ts
```diff
@@ -0,0 +1,23 @@
+import KeysWithValueModuleData from "@latticexyz/world/out/KeysWithValueModule.sol/KeysWithValueModule.json" assert { type: "json" };
+import KeysInTableModuleData from "@latticexyz/world/out/KeysInTableModule.sol/KeysInTableModule.json" assert { type: "json" };
+import UniqueEntityModuleData from "@latticexyz/world/out/UniqueEntityModule.sol/UniqueEntityModule.json" assert { type: "json" };
+import { ContractCode } from "../utils/types";
+
+// These modules are always deployed
+export const defaultModules: ContractCode[] = [
```
alvrs, line 7: for consistency with the other variable names
```suggestion
export const defaultModuleContracts: ContractCode[] = [
```

alvrs (commented):
So much cleaner! And can't wait for the viem refactor. Added a couple nits and will push a commit to resolve them shortly.

alvrs (approved):

PR #1385, opened 2023-09-01
feat(world): add world factory

johngrantuk:
WorldFactory allows new worlds to be deployed (with a predefined CoreModule installed) using create2
  - Used an internal count as Salt. Makes it easier for a user. (And allows all previous deployed World addresses to be calculated.)
  - For CoreModule - passed in to Factory constructor so we associate one CoreModule with all pools created from factory.
Create2Factory is a stand alone helper contract that can be used to deploy any contract using create2. Eventually this could be used to deploy WorldFactory, CoreModule, Systems, etc.

Deploy script will be updated to use factory in a future PR after this and #1384 have been finalised.


packages/world/src/factories/WorldFactory.sol
```diff
@@ -0,0 +1,29 @@
...
+    @dev Deploy a new World and install coreModule.
+  */
+  function deployWorld() public {
+    bytes memory bytecode = type(World).creationCode;
+    address worldAddress = Create2.create2Deploy(bytecode, worldCount);
+    IBaseWorld world = IBaseWorld(worldAddress);
+    world.installRootModule(coreModule, new bytes(0));
+    emit WorldDeployed(worldAddress);
+    worldCount++;
+  }
```
alvrs, line 28: Should we transfer the ownership of the root module to the caller after installing the root module? (Or maybe we could make it optional and emit the selected option as part of the event? The interesting thing with a deploy function that does not transfer ownership of the root module is that users can be sure the World behaves exactly as the audited vanilla World)

packages/world/test/Factories.t.sol
```diff
@@ -0,0 +1,59 @@
...
+    // Address we expect for World
+    address calculatedAddress = calculateAddress(worldFactoryAddress, bytes32(0), type(World).creationCode);
+    // Check for HelloWorld event from World
+    vm.expectEmit(true, false, false, false);
+    emit HelloWorld();
+    // Check for WorldDeployed event from Factory
+    vm.expectEmit(true, false, false, false);
+    emit WorldDeployed(calculatedAddress);
+    worldFactory.deployWorld();
+    // TODO - Can we test if core module has been installed correctly? (Confirmed visually via forge trace)
```
alvrs, line 55: We can use the `InstalledModules` table! (See https://github.com/latticexyz/mud/blob/main/packages/world/src/World.sol#L62 for the expected value)

packages/world/test/Factories.t.sol
```diff
@@ -0,0 +1,59 @@
...
+    // Deploy WorldFactory with current CoreModule
+    address worldFactoryAddress = address(new WorldFactory(new CoreModule()));
+    IWorldFactory worldFactory = IWorldFactory(worldFactoryAddress);
+    // Address we expect for World
+    address calculatedAddress = calculateAddress(worldFactoryAddress, bytes32(0), type(World).creationCode);
+    // Check for HelloWorld event from World
+    vm.expectEmit(true, false, false, false);
+    emit HelloWorld();
+    // Check for WorldDeployed event from Factory
+    vm.expectEmit(true, false, false, false);
```
alvrs, line 52: high level style nit - let's add an empty line before every comment to improve readability

packages/world/src/factories/WorldFactory.sol
johngrantuk: As it currently stands it would be the Factory address that would be the owner? I imagine the caller would probably assume they will be the owner so maybe that is best although I'm happy to take direction as I don't have a full understanding of the knock on effects.

packages/world/test/Factories.t.sol
```diff
@@ -0,0 +1,59 @@
...
+  function testCreate2Factory() public {
+    Create2Factory create2Factory = new Create2Factory();
+    // Encode constructor arguments for WorldFactory
+    bytes memory encodedArguments = abi.encode(new CoreModule());
+    bytes memory combinedBytes = abi.encodePacked(type(WorldFactory).creationCode, encodedArguments);
+    // Address we expect for deployed WorldFactory
+    address calculatedAddress = calculateAddress(address(create2Factory), bytes32(0), combinedBytes);
+    // Confirm event for deployment
+    vm.expectEmit(true, false, false, false);
+    emit ContractDeployed(calculatedAddress, uint256(0));
```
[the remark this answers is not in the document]
johngrantuk: I can't seem to make this work. I noticed in other tests the events are redefined, e.g. World.t.sol. Could you point me to somewhere I can use as an example if it is possible?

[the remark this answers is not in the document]
alvrs: seems to not be possible yet, but soon: https://github.com/ethereum/solidity/pull/14274

alvrs (approved):

packages/world/src/factories/Create2Factory.sol
```diff
@@ -0,0 +1,19 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
+
+import { Create2 } from "./Create2.sol";
+
+/**
+  @dev Helper Contract to facilitate create2 deployment of Contracts.
+*/
+contract Create2Factory {
+  event ContractDeployed(address addr, uint256 salt);
```
holic, line 10: is this a standard somewhere? should we have an index on addr?

packages/world/test/Factories.t.sol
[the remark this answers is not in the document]
holic: looks like it's part of the 0.8.21 release: https://github.com/ethereum/solidity/releases/tag/v0.8.21
> Allow qualified access to events from other contracts.

PR #1386, opened 2023-09-01
docs: fix table references in store hooks example

nishuzumi:
Chang 
```
if (table != table) revert("invalid table");
```
to
```
if (_table != table) revert("invalid table");
```


nishuzumi:
@alvrs @holic

holic (approved):

PR #1390, opened 2023-09-03
Advance world contract nonce

curtischong:
- when we call viem functions with our burner wallet (e.g. deploy a codegened contract, the nonce in the worldContract is not updated. So we expose the nonceManager to allow us to manually increment the nonce (so subsequent system calls will have the proper nonce number)


curtischong:
wrong repo :'(

alvrs:
have you tried `createContract` from `@latticexyz/common`? (see the changeset in https://github.com/latticexyz/mud/commit/939916bcd5c9f3caf0399e9ab7689e77e6bef7ad for context)

PR #1399, opened 2023-09-05
feat(store,world): more granularity for onchain hooks

### LINKED ISSUE #1159: Consider registering individual hook methods (instead of all-or-nothing)
frolic — Today at 1:40 PM
thinking about store hooks and we should prob move to either 1) separate registration array per hook method (eg set record) or 2) the registration includes a bitmap of all the hook methods used
registering a hook right now means every operation goes through that hook, even if it’s a no op?

alvarius — Today at 1:42 PM
we could say we can expect a store hook to implement all functions because data can be modified via each of them, it's similar to how an off-chain indexer has to support all store events and can't just ignore one

frolic — Today at 1:43 PM
I’m just thinking this is the worst approach for gas in the hot code path

alvarius — Today at 1:44 PM
we could do a bitmap if it doesn't add much gas (which might be reasonable since we can store it in the same storage slot)
rn i can't think of a use case of only implementing one store hook function but not another though

frolic — Today at 1:45 PM
yeah, there’s a trade off of my two proposals: iterate through all hooks but skip some (per store op) vs more registration gas cost (one time)

alvarius — Today at 1:45 PM
do you have something in mind?

frolic — Today at 1:45 PM
maybe I’m thinking more about system hooks where you may only want before or after
could see potentially doing a before and after for store events too
but can’t think of a specific use case right now
one thought, prob not a great example: a hook that makes an array field append only 

alvarius — Today at 1:48 PM
we currently do that for field but not for record, because for handling a field update you might want to just read the entire record from storage before and after the update instead of manually reconstructing the new record with the field update applied in the hook
could allow this for setRecord too but the usecase is less clear

frolic — Today at 1:50 PM
this feels like one of those “I can’t think of a good example but maybe someone will think of one” and we get some potential gas benefits 

curious if there’s a strong case for store hooks implementing all methods

alvarius — Today at 1:51 PM
curious if there’s a strong case for store hooks implementing all methods
just that if you want to react to data changes you probably want to react to all possible ways the data can change
i see the point for system hooks though
a use case for afterSetRecord might be if you want to actually change the storage value of what has been set

frolic — Today at 2:01 PM
I think my main point is: first order effect is better gas (in the hot code path), second order effect is more granularity 

alvarius — Today at 2:05 PM
yeah i understand
the argument for one array in terms of gas would be: if we expect that every hook implements all functions, one array would be cheaper if a table is modified twice in a transaction with different methods (setRecord, setField, deleteRecord, ...) because the storage is hot for every access after the first one
but i see advantages for system hooks, and the potential disadvantages for store hooks are small enough to justify making it consistent for both imo

frolic — Today at 2:17 PM
this makes sense, at the very least we could couple a “which hooks” bitmap with the address into the same slot

alvrs:
Fixes #1159 (that issue also includes more context on this approach vs different approaches)

TODOs
- [x] Split `onSetRecord` into `onBeforeSetRecord` and `onAfterSetRecord`
- [x] Split `onDeleteRecord` into `onBeforeDeleteRecord` and `onAfterDeleteRecord`
- [x] Add a bitmap to the store hooks and system hooks: One bit for each operation, so we can skip unnecessary calls.
- [x] Rename `EnableHooks` to `EnableStoreHooks` and each element from `beforeX` to `onBeforeX` etc

TODOs in follow ups
- [ ] Add ability to remove hooks (by filtering the array). Since it's possible to upgrade systems now it seems useful to be able to remove hooks and register new ones. 
- [ ] Consider renaming all mentions of "Hook" to "Hooks" (since now that individual functions can be enabled and disabled that makes more sense)

TODOs depending on #1354
- [ ] Remove `onBeforeSetField` and `onAfterSetField`
- [ ] Add `onBeforeSpliceRecord` and `onAfterSpliceRecord`


packages/store/src/Hook.sol
```diff
@@ -0,0 +1,43 @@
...
+
+using HookInstance for Hook global;
+
+library HookLib {
+  /**
+   * Encode enabled hooks into a bitmap with 1 bit per hook, and pack the bitmap with the store hook address into a bytes21 value
+   */
+  function encode(address hookAddress, uint8 encodedHooks) internal pure returns (Hook) {
+    // Move the address to the leftmost 20 bytes and the bitmap to the rightmost byte
+    return Hook.wrap(bytes21(bytes20(hookAddress)) | bytes21(uint168(encodedHooks)));
```
holic, line 15: ooc is this more gas efficient than the reverse?

packages/store/src/Hook.sol
```diff
@@ -0,0 +1,43 @@
...
+    return address(bytes20(Hook.unwrap(self)));
+  }
+
+  /**
+   * Get the store hook's bitmap
+   */
+  function getBitmap(Hook self) internal pure returns (uint8) {
+    // Extract the bitmap from the rightmost bytes
+    return uint8(uint168(Hook.unwrap(self)));
+  }
```
holic, line 42: this is all clever, can we have some tests for these to keep confidence?

packages/store/test/StoreCore.t.sol
```diff
@@ -838,7 +839,18 @@ contract StoreCoreTest is Test, StoreMock {
...
+      table,
+      subscriber,
+      EnabledStoreHooks({
+        onBeforeSetRecord: true,
+        onAfterSetRecord: false,
+        onBeforeSetField: true,
+        onAfterSetField: false,
+        onBeforeDeleteRecord: true,
+        onAfterDeleteRecord: false
+      })
```
holic, line 852: I wonder if this would also work
```solidity
EnabledStoreHooks({
  onBeforeSetRecord: true,
  onBeforeSetField: true,
  onBeforeDeleteRecord: true
})
```
(only toggling on the ones we want and letting the others default, but not sure if solc will complain)

holic: I also wonder if we should just use a `unit8` here instead of the struct so that the ABI (and presumably abi-encoded calldata) is a bit more compact vs `(bool,bool,bool,bool,...)`

packages/store/src/Hook.sol
alvrs: tests for this are in `StoreHook` (i ripped out `Hook` from `StoreHook` after realizing that `SystemHook` has a lot in common) - can definitely move the general tests to `Hook` though!

packages/store/test/StoreCore.t.sol
alvrs: unfortunately `solc` does complain :(

alvrs: > I also wonder if we should just use a unit8 here

yeah i think you're right here - we just have to document how to use the helper libraries to encode the bitmap, but it would be much cheaper to encode it in eg. the module the hook is registered and only pass the `uint8` between contracts.

holic (approved):

holic (approved):

PR #1403, opened 2023-09-06
feat(cli,create-mud,store,world): add abi-ts to replace typechain

holic:
big diff because I removed checked-in types from example and e2e dirs (don't think we need em in here for now and they were a pain to update if core stuff changed)

TODO:
- [x] migrate typechain usage in dev-contracts (need to build ABIs and pass into abi-ts)
- [x] migrate store/world

related to https://github.com/microsoft/TypeScript/issues/32063


alvrs (commented):
this is very exciting! just minor nits

packages/cli/src/commands/trace.ts
```diff
@@ -66,7 +65,7 @@ const commandModule: CommandModule<Options, Options> = {
 
     // Create World contract instance from deployed address
     const provider = new ethers.providers.StaticJsonRpcProvider(rpc);
-    const WorldContract = new ethers.Contract(worldAddress, IBaseWorldData.abi, provider) as IBaseWorld;
```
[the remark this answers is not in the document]
holic: correct! this would have been the last remaining spot to use typechain, so I opted to ditch it all for the short time that we're between ethers and viem for deploys

packages/cli/src/commands/dev-contracts.ts
```diff
@@ -142,8 +143,10 @@ const commandModule: CommandModule<Options, Options> = {
       // Build the contracts
       await forge(["build"]);
 
-      // Run typechain to rebuild typescript types for the contracts
-      await worldtypes();
+      // Generate TS-friendly ABI files
+      // We rebuild into a separate dir to have a clean set of ABIs without test/script contracts
```
alvrs, line 147: is this comment still up to date?

alvrs (approved):

packages/cli/src/commands/dev-contracts.ts
holic: yep, still relevant

holic: more context:

We need forge to build everything to run things like tests, scripts, etc. But for a clean ABI dir, we want to ignore these. It's tricky to use the same target dir (e.g. `out`) because then tests won't find the build files they need to run. So we use a separate dir just for ABIs.

Since abi-ts can now target/build a specific ABI file or glob, we could simplify this to just a single output dir and only build IWorld ABI, but has trade-offs in that it's a bit harder for users to expand that if they want to include more ABIs in their TS code.

Issue #1344, opened 2023-08-21
System call event


holic:
Something to identify and group system calls and their side effects (events/store operations). This isn't as important for one-off system calls, since each store operation has the associated transaction (and thus can be decoded to a system call), but is important once you have systems calling systems.

alvrs:
Could this be implemented as a module? Assuming https://github.com/latticexyz/mud/issues/1392 would expose a client stream of all registered events (that keeps the order of the events in the transaction/block), we could create a module that adds a hook on every new system (via a hook on `registerSystem`), and emits one event in `onBeforeCallSystem` and one in `onAfterCallSystem`. Then there could be a small client library that takes the event stream from #1392 and infer which store events belong to which system call (which is possible because the events are ordered).

Alternatively we could emit a built-in event before and after every system call and also build in the logic for grouping store updates by system call. We should measure how much gas overhead it would be to always emit these events - if it's not significant it might be worth doing it by default, since the module approach likely has much more overhead (due to all the hooks and indirections).

holic:
My impression is gas is quite cheap for events, but not sure if we're also appending all the original arguments in the system call. If it's cheap, including it by default would be great. But it might make sense to keep this opt-in, via custom event + manual emission where needed. (I don't think implementing this entirely as a module is great, due to overhead you mentioned.)

alvrs:
Context from #1408 - the event introduces a significant gas overhead and we've decided to approach the issue of a "system call stream" / grouping storage operations by systems from another angle and do it via traces.

PR #1408, opened 2023-09-06
feat(world): add system call event

alvrs:
Draft for #1344

### DIFF
--- a/packages/world/gas-report.json
+++ b/packages/world/gas-report.json
@@ -39,13 +39,13 @@
     "file": "test/KeysInTableModule.t.sol",
     "test": "testInstallComposite",
     "name": "install keys in table module",
-    "gasUsed": 1411774
+    "gasUsed": 1438729
   },
   {
     "file": "test/KeysInTableModule.t.sol",
     "test": "testInstallGas",
     "name": "install keys in table module",
-    "gasUsed": 1411774
+    "gasUsed": 1438729
   },
   {
     "file": "test/KeysInTableModule.t.sol",
@@ -57,13 +57,13 @@
     "file": "test/KeysInTableModule.t.sol",
     "test": "testInstallSingleton",
     "name": "install keys in table module",
-    "gasUsed": 1411774
+    "gasUsed": 1438729
   },
   {
     "file": "test/KeysInTableModule.t.sol",
     "test": "testSetAndDeleteRecordHookCompositeGas",
     "name": "install keys in table module",
-    "gasUsed": 1411774
+    "gasUsed": 1438729
   },
   {
     "file": "test/KeysInTableModule.t.sol",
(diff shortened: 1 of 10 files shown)


packages/world/gas-report.json
```diff
@@ -231,67 +231,73 @@
...
     "test": "testCallFromTimeboundDelegation",
     "name": "call a system via a timebound delegation",
-    "gasUsed": 34791
+    "gasUsed": 36952
+  },
+  {
+    "file": "test/SystemCall.t.sol",
+    "test": "testGasSystemCallEvent",
+    "name": "emit a system call event",
+    "gasUsed": 3158
```
alvrs, line 258: Unfortunately it seems like emitting this event is pretty significant. With the current design we'd have to emit it twice per system call.

ludns:
one thing we could do is not using events at all (which cost gas) and instead use traces as a way to get these "events" into the sync stack.
traces allow you to see EVM CALL and DELEGATECALL with their calldata, which is enough to reconstruct these events.

packages/world/gas-report.json
ludns: why is it so expensive? it seems like the data is not that big

alvrs:
It seems like emitting an event with sufficient data to connect it to a specific system function call (`bytes32 systemId, bytes4 functionSelector`) would still cost 1440 gas per event, and we'd need two events (start and end) to be able to assign every store event to a system call (for situations where a system updates a table, then calls another system, and then updates a table again). 

2880 gas overhead for every system call for this feature is definitely too much. One alternative approach would be to not emit the events for every system call, but only for systems where it is needed. The `Systems` table still has space in the first storage slot of static data, so we could add a `emitSystemCallEvents` flag and allow it to be set when registering systems.

alvrs:
> one thing we could do is not using events at all (which cost gas) and instead use traces as a way to get these "events" into the sync stack.
traces allow you to see EVM CALL and DELEGATECALL with their calldata, which is enough to reconstruct these events.

Yeah thought about this as well. Unfortunately I don't think traces are supported by all RPC nodes and I assume they're pretty expensive operations, but we could this feature to the indexer (so at least it's one trace per system call, not one trace per system call per client). After seeing the gas cost of these events I do prefer a solution via traces.

holic:
Given that folks can manually backfill for previous usages of system call stream via offchain tables or (soon) custom events, I think it's fine punting on this and being more of a feature you can enable if your RPC supports tracing.

ludns:
it's interesting that custom events will have the same "significant increased gas usage" issue; i wonder if that is going to tip the balance back to system calls via traces again.

alvrs:
Seems like we're aligned on going with the trace route, closing this for now

PR #1413, opened 2023-09-07
docs: rework abi-ts changesets

holic:
I wasn't happy with the changelog in https://github.com/latticexyz/mud/pull/1412 so I separated the changesets for each package.


holic:
yolomerging

PR #1414, opened 2023-09-07
build: wait for snapshot/prerelease before pushing docker container

holic:
I think we're not seeing Docker images tagged with `next` because the tag doesn't exist at the time the docker job is run (because the snapshot/prerelease jobs haven't completed, thus haven't created tags).

This PR attempts to wait for these jobs to complete before running the docker job.


holic:
this might be fixed by https://github.com/latticexyz/mud/pull/1417

holic:
gonna close in favor of https://github.com/latticexyz/mud/pull/1417 🤞

PR #1416, opened 2023-09-07
build: use git context when creating docker metadata

holic:
https://github.com/docker/metadata-action#context-input

Hoping that this will help the workflow find the next tags, since they may not be available in e.g. `workflow_dispatch` event (manual trigger).


holic:
since this change would apply after the latest next commit, we won't really be able to test it via manual dispatch until there's another next release

hoping that https://github.com/latticexyz/mud/pull/1414 fixes this instead

holic:
gonna close in favor of https://github.com/latticexyz/mud/pull/1417 🤞

PR #1417, opened 2023-09-07
build: trigger docker for tag push

holic:
missed that we needed to configure the workflow for tags being pushed

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - main
+    tags:
+      - "*"
   workflow_dispatch:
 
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yml
```diff
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - main
+    tags:
+      - "*"
```
holic, line 8: If this does get triggered per tag (i.e. once per package per version release), we could restrict this to just one of the packages we expect to always exist (e.g. `common`).
```suggestion
      - "@latticexyz/common@*"
```

ludns (approved):

PR #1422, opened 2023-09-08
feat(store,world): add ability to unregister hooks

### LINKED ISSUE #1405: Add ability to remove hooks
followup to #1399

alvrs:
Fixes #1405 
Followup to #1399 
Context in the changeset


packages/world/src/modules/core/implementations/WorldRegistrationSystem.sol
```diff
@@ -64,6 +64,40 @@ contract WorldRegistrationSystem is System, IWorldErrors {
...
+          newHooksIndex++;
+        }
+      }
+    }
+
+    // Set the new hooks table length in place
+    // (Note: this does not update the free memory pointer)
+    assembly {
+      mstore(newHooks, newHooksIndex)
+    }
```
holic, line 95: Since the "remove a hook from the list of hooks" is duplicated, should we move this into some sort of hooks lib?

holic (dismissed):
Just one comment about code duplication but not blocking!

packages/store/src/Hook.sol
```diff
@@ -14,6 +16,36 @@ library HookLib {
     // Move the address to the leftmost 20 bytes and the bitmap to the rightmost byte
     return Hook.wrap(bytes21(bytes20(hookAddress)) | bytes21(uint168(encodedHooks)));
   }
+
+  /**
+   * Filter the given hook from the hook list at the given key in the given hook table
+   */
+  function filterListByAddress(bytes32 hookTableId, bytes32 key, address hookAddressToRemove) internal {
+    bytes21[] memory currentHooks = Hooks.get(hookTableId, key);
```
holic, line 24: Is it better/more gas efficient to do this than to pass in the array and get an array back out?

holic (approved):

packages/store/src/Hook.sol
alvrs: yeah, unfortunately much more gas efficient to pass as little memory around as possible - passing arrays is pretty expensive. (Otherwise would have preferred a more functional approach with arrays as input/output too)

PR #1423, opened 2023-09-08
WIP playing around with ffi test deploys

holic:
just wanted to see if it is possible to run our deploy command within the context of a test so we can avoid the need to have `.mudtest` file

### DIFF
--- a/packages/store/src/MudTest.sol
+++ b/packages/store/src/MudTest.sol
@@ -1,13 +1,37 @@
 // SPDX-License-Identifier: MIT
 pragma solidity >=0.8.0;
 
-import { Test } from "forge-std/Test.sol";
+import { Test, console } from "forge-std/Test.sol";
+import { Vm, VmSafe } from "forge-std/Vm.sol";
 import { StoreSwitch } from "../src/StoreSwitch.sol";
 
 contract MudTest is Test {
-  address worldAddress;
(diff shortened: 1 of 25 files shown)


packages/store/src/MudTest.sol
```diff
@@ -1,13 +1,37 @@
...
 
-import { Test } from "forge-std/Test.sol";
+import { Test, console } from "forge-std/Test.sol";
+import { Vm, VmSafe } from "forge-std/Vm.sol";
 import { StoreSwitch } from "../src/StoreSwitch.sol";
 
 contract MudTest is Test {
-  address worldAddress;
+  address public worldAddress = vm.envOr("WORLD_ADDRESS", address(0));
+  // TODO: figure out how to uniquely identify this among other tests, or run tests in series rather than parallel, to avoid anvil nonce issues
```
holic, line 10: currently stuck here

then the next step is being able to run the deploy and have it only return the world address or somehow parse the world address from the deploy logs

holic: https://github.com/foundry-rs/foundry/issues/5826

holic: got past these things, but now stuck again on generic evm reverts

holic:
closing as outdated

PR #1425, opened 2023-09-08
feat(world): add Balance table and BalanceTransferSystem

### LINKED ISSUE #1283: store value in world instead of system contracts
- currently any `msg.value` sent to non-root systems is forwarded to the system contract
- this works as long as the system contract is aware of this and makes sure the system is written in a way that prevents anyone from just calling the system contract directly or via a "fake work" to trigger a value transfer outside of the system
- it seems like this is an easy way for devs to shoot themselves in the foot. Systems are supposed to not have any state and read/write all state on the world contract, but the eth value is the only state that is in fact stored in the system itself.
- I think instead all eth balance should be managed in the world and separated by namespace, and there should be a way for systems to access eth balance in their namespace

alvrs:
Fixes #1283 
TODO
- [x] fix `WorldContextConsumer`
- [x] add test for `WorldContextConsumer` / `WorldContextProvider` with msgSender and msgValue
- [x] add logic for updating root namespace balance when sending to world without calling a system
- [x] add test for calling root systems with balance
- [x] add test for calling non-root systems with balance
- [x] add test for sending balance to world without calling a system
- [x] add test for transferring balance between namespaces
- [x] add test for transferring balance to address


packages/world/src/modules/core/implementations/BalanceTransferSystem.sol
```diff
@@ -0,0 +1,49 @@
...
+
+    // Get current namespace balance
+    uint256 balance = Balances.get(fromNamespace);
+
+    // Require the balance balance to be greater or equal to the amount to transfer
+    if (amount > balance) revert InsufficientBalance(balance, amount);
+
+    // Update the balances
+    Balances.set(fromNamespace, balance - amount);
+    payable(toAddress).transfer(amount);
```
holic, line 47: I don't know the "why" but I've seen folks prefer this approach over the above:
```solidity
(bool success, ) = toAddress.call{value: amount}("");
```

packages/world/src/SystemCall.sol
```diff
@@ -41,18 +42,26 @@ library SystemCall {
     // Allow access if the system is public or the caller has access to the namespace or name
     if (!publicAccess) AccessControl.requireAccess(resourceSelector, caller);
 
+    // If the msg.value is non-zero, update the namespace's balance
+    if (value > 0) {
+      bytes16 namespace = resourceSelector.getNamespace();
+      uint256 currentBalance = Balances.get(namespace);
+      Balances.set(namespace, currentBalance + value);
+    }
```
holic, line 50: anything to be worried about here wrt reentry?

ludns: yes this smells like a re-entry attack

holic:
- do we need to protect against systems receiving value if not called through the world?
  - similarly but not related to this PR, should we prevent systems being called without the world entry point?
- does the world/root namespace have a way to withdraw all? if so, how should it update its accounting of balances? or will you have to specify the namespace to withdraw from?
- what if I just send value to the world? do we need to account for that or prevent it?

ludns:
you can never prevent someone from sending eth to a contract, a contract can always self destruct and send value and not trigger your fallback

packages/world/src/modules/core/implementations/BalanceTransferSystem.sol
alvrs: good catch! probably because of the 2300 gas limit of `transfer` (see https://twitter.com/obatirou/status/1697167795077698038 and https://medium.com/coinmonks/solidity-transfer-vs-send-vs-call-function-64c92cfc878a)

packages/world/src/SystemCall.sol
alvrs: Can you elaborate how a reentry attack would work here?

This is a utility that is used by `call`, `callFrom` and the World's fallback function, in each case just passing the `msg.value` to the `SystemCall` util. If the contract that's called in here would call one of these entry points again, it would still only pass that new call's `msg.value` to the function.

I think we need to update the balance table value before calling the system, because the system might want to transfer the balance somewhere else in this call.

alvrs:
> * do we need to protect against systems receiving value if not called through the world?
>   * similarly but not related to this PR, should we prevent systems being called without the world entry point?

Calling systems directly is not intended, but I don't think we should protect against it. As long as systems don't hold state it doesn't matter by who they're called since they will read/write on their caller by default. And there might be some advanced custom use cases where systems are limited to one World or do hold some internal state and the developers have some method for updating the internal state by calling them directly.

> * does the world/root namespace have a way to withdraw all? if so, how should it update its accounting of balances? or will you have to specify the namespace to withdraw from?

Whoever has root access can theoretically withdraw everything by deploying a new root system to transfer all the world contract's balance out, but it's basically a rug pull and not intended (we can't prevent it though). In the ideal world namespaces can only manage their own balances, which is the case once root namespace access is burned.

> * what if I just send value to the world? do we need to account for that or prevent it?

Right now the world can receive ETH without being attributed to any namespace via its `receive` function. We could update the root namespace's balance in that case?

packages/world/src/SystemCall.sol
alvrs: Yes, `requireAccess` only checks an internal table.

Agree this is definitely a critical part of the system that we should be extra careful with and should also point this out as a focus area in the audit.

.changeset/pink-tips-give.md
```diff
@@ -0,0 +1,23 @@
+---
+"@latticexyz/world": major
+---
+
+The World now maintains a balance table per namespace. When a system is called with value, the value stored in the World contract and credited to the system's namespace.
```
alvrs, line 5:
```suggestion
The World now maintains a balance per namespace.
When a system is called with value, the value stored in the World contract and credited to the system's namespace.
```

.changeset/pink-tips-give.md
```diff
@@ -0,0 +1,23 @@
...
+There are two new World methods to transfer balance between namespaces (`transferBalanceToNamespace`) or to an address (`transferBalanceToAddress`).
+
+```solidity
+interface IBaseWorld {
+  function transferBalanceToNamespace(bytes16 fromNamespace, bytes16 toNamespace, uint256 amount) external;
+
+  function transferBalanceToAddress(bytes16 fromNamespace, address toAddress, uint256 amount) external;
+
+  // ...
+}
```
alvrs, line 22:
```suggestion
interface IBaseWorld {
  function transferBalanceToNamespace(bytes16 fromNamespace, bytes16 toNamespace, uint256 amount) external;

  function transferBalanceToAddress(bytes16 fromNamespace, address toAddress, uint256 amount) external;
}
```

packages/world/src/World.sol
```diff
@@ -221,9 +221,12 @@ contract World is StoreRead, IStoreData, IWorldKernel {
    ************************************************************************/
 
   /**
-   * Allow the World to receive ETH
+   * ETH sent to the world without calldata is added to the root namespace's balance
```
alvrs, line 224:
```suggestion
   * ETH sent to the World without calldata is added to the root namespace's balance
```

packages/world/src/modules/core/CoreModule.sol
```diff
@@ -57,7 +60,9 @@ contract CoreModule is IModule, WorldContextConsumer {
    * Register core tables in the World
    */
   function _registerCoreTables() internal {
+    Balances.register();
     InstalledModules.register();
+    Delegations.register();
```
alvrs, line 65: this is unrelated, should rip it out

holic (approved):
Are there ways for funds to get into the world without triggering `receive`?

If so, do we want some way for the root namespace owner to withdraw `address(world).balance - totalOfCurrentBalances` as a safety/backstop?

ludns:
Yes it is possible to not trigger receive (see my comment on selfdestruct)

alvrs:
> If so, do we want some way for the root namespace owner to withdraw address(world).balance - totalOfCurrentBalances as a safety/backstop?

I'd say because it requires real effort to send balance to the world without triggering `receive` we don't need to solve for this case (maintaining a `totalOfCurrentBalances` would increase the gas cost of every system call with value).

For anyone with root access to the World it's still possible to withdraw all balance from the World (with or without adjusting the `Balances` table by registering a new root system.

holic:
> For anyone with root access to the World it's still possible to withdraw all balance from the World (with or without adjusting the Balances table by registering a new root system.

How so?

alvrs:
> How so?

Root systems are delegatecalled and therefore can access the world state directly (including the World contract's balance)

PR #1429, opened 2023-09-08
feat(dev-tools): hide tabs and custom tabs

0xhank:
I want to continue using my modified ecs-browser but also want access to the new mud dev-tools. This PR allows me to continue using my custom tooling within the mud dev tools window. It also adds options for hiding unused tools.

### DIFF
--- a/packages/dev-tools/src/App.tsx
+++ b/packages/dev-tools/src/App.tsx
@@ -1,10 +1,11 @@
-import "./preflight.css";
-import "tailwindcss/tailwind.css";
-import { useEffect, useState } from "react";
-import { twMerge } from "tailwind-merge";
-import { router } from "./router";
+import { useEffect, useMemo } from "react";
 import { RouterProvider } from "react-router-dom";
+import { twMerge } from "tailwind-merge";
+import "tailwindcss/tailwind.css";
 import useLocalStorageState from "use-local-storage-state";
+import { useDevToolsContext } from "./DevToolsContext";
+import "./preflight.css";
+import { createRouter } from "./router";
 
 // TODO: fix tab index so that it's not possible to tab around in the UI when it's hidden
 
@@ -23,6 +24,8 @@ export function App() {
     return () => window.removeEventListener("keypress", listener);
   });
 
+  const settings = useDevToolsContext();
+  const router = useMemo(() => createRouter(settings), []);
   return (
     <div className="fixed inset-0 pointer-events-none">
       <div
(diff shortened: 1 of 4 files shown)


holic:
Ohhh cool idea! Some thoughts/questions:

- Because we compile dev tools as a separate bundle (including bundling React), I wonder what the implications are of allowing you to add any component to this bundle. I assume you'd need to keep React versions aligned with dev tools? I haven't really given extensibility much thought.
- I am a little worried about the horizontal real estate we have for the dev tools and allowing any arbitrary tabs to be added. We have plans to add a few more but haven't built them yet, and that might mean if you added any then the tabs would overflow or we'd need to design around this. Definitely solvable, just something we didn't think we'd have to solve for a while.

In terms of API, making each tab toggleable will lead to configuration bloat and lots of conditionals everywhere. I think we'll get more mileage if we 1) make all the routes/tabs configurable 2) export each tab component and 3) have some sane defaults for routes/tabs that are easy to extend or copy/paste. I think this should help with the second point above, where you'd have an override of the default tabs so any upstream tabs added wouldn't get added until you extend your override.

Thinking something like:

```ts
import { AgnosticDataRouteObject } from "react-router-dom";

type Tab = {
 label: string,
 route: AgnosticDataRouteObject, // not sure if this is exactly the right type
};

type DevToolsOptions = {
  ...
  tabs: Tab[];
};
```

```ts
export defaultTabs: Tab[] = [
  {
    label: "Summary",
    route: { index: true, element: <SummaryPage /> },
  },
  {
    label: "Actions",
    route: { path: "actions", element: <ActionsPage /> },
  },
  ...
];
```

then rework the `createMemoryRouter` to use the route/tab config instead of the React elements approach.

Then you can opt to pass in your own tabs config to override like

```ts
import { mount as mountDevTools } from "@latticexyz/dev-tools";

mountDevTools({
  ...
  tabs: [
    // your own tabs here
  ],
});
```

or extend with

```ts
import { mount as mountDevTools, defaultTabs } from "@latticexyz/dev-tools";


mountDevTools({
  ...
  tabs: [
    ...defaultTabs,
    // your own tabs here
  ],
});
```

0xhank:
To your first point (react version), this is a requirement across mud packages -- for example, if you want to use rxjs locally and with mud, it must be 7.5.5. This extends to the rest of the peer dependencies and using `@latticexyz/react`. So this requirement doesn't seem out of scope with general mud usage.

To your second point, for tab overflow, the two solutions are either a topbar overflow (multiple rows of tabs) or convert the topbar to a sidebar. seems like the general best practice for a pane with many tabs is a sidebar instead of a topbar.

Finally, I agree that the approach Id prefer is to have a module for each tab and a separate module for the containing browser component. This is the method I went for in the mud-game-tools package (just with worse styling). It would  mean the Browser component I wrote could be deprecated for the mud dev-tools browser component.

holic:
Closing for now to clear up our PR backlog. Feel free to reopen if you plan to address comments above.

PR #1430, opened 2023-09-09
feat(create-mud): add worlds.json types

### LINKED ISSUE #1428: add `worlds.json.d.ts` file
so we can just import `worlds` rather than `worldsJson` and overwriting the type 
https://github.com/latticexyz/mud/blob/aa96d4ccaf7dad7569d57a41fef0a981fdce6f7f/e2e/packages/client-vanilla/src/mud/getNetworkConfig.ts#L2-L5

holic:
fixes #1428

### DIFF
--- a/e2e/packages/client-vanilla/src/mud/getNetworkConfig.ts
+++ b/e2e/packages/client-vanilla/src/mud/getNetworkConfig.ts
@@ -1,8 +1,6 @@
 import { getBurnerPrivateKey } from "@latticexyz/common";
-import worldsJson from "contracts/worlds.json";
+import worlds from "contracts/worlds.json";
 import { supportedChains } from "./supportedChains";
-
-const worlds = worldsJson as Partial<Record<string, { address: string; blockNumber?: number }>>;
 
 export async function getNetworkConfig() {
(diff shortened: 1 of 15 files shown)


alvrs (approved):
is the `world.d.ts` file generated or manually created? Should we automatically create it when `world.json` is written (i believe in the `deploy` script)? Could even make it strongly typed like the abi types (might be useful to have type safety for which chains there is a deployment?)

holic:
> is the `world.d.ts` file generated or manually created? Should we automatically create it when `world.json` is written (i believe in the `deploy` script)? Could even make it strongly typed like the abi types (might be useful to have type safety for which chains there is a deployment?)

We intentionally check in an initial `worlds.json` to avoid import/type issues when building the client, otherwise Vite throws up unless you deploy contracts first.

And we intentionally don't use stronger types/downcast it because we do a look up based on a dynamic chain ID, which fails type checks if it's a strict object with only specific keys included.

holic:
All that said, I still don't love the approach for tracking deploys and using the contract addresses and would like to rethink it in the future but this is an 80%+ solution for now.

PR #1431, opened 2023-09-09
docs: Delete expired links

Mirror-Tang:
Removed two expired Github links.


qbzzt:
I think the module was removed from the repository, so it's no longer supported.

PR #1443, opened 2023-09-11
feat(protocol-parser): add keySchema/valueSchema helpers

holic:
related to https://github.com/latticexyz/mud/issues/1296

pulled out of https://github.com/latticexyz/mud/pull/1354


packages/protocol-parser/src/decodeKey.ts
```diff
@@ -0,0 +1,15 @@
+import { Hex } from "viem";
+import { SchemaToPrimitives, KeySchema } from "./common";
+import { decodeKeyTuple } from "./decodeKeyTuple";
+
+export function decodeKey<TSchema extends KeySchema>(
+  keySchema: TSchema,
+  data: readonly Hex[]
+): SchemaToPrimitives<TSchema> {
+  // TODO: refactor and move all decodeKeyTuple logic into this method so we can delete decodeKeyTuple
```
alvrs, line 9: is `decodeKeyTuple` still used somewhere else? If not, should we just move the logic in there in this PR?

packages/protocol-parser/src/decodeRecord.test.ts
```diff
@@ -21,4 +21,16 @@ describe("decodeRecord", () => {
       ]
     `);
   });
+
+  it("can decode an out of bounds array", () => {
+    const schema = { staticFields: [], dynamicFields: ["uint32[]"] } as const;
+    const values = decodeRecord(schema, "0x0000000000000000000000000000000000000000000000000400000000000004");
```
alvrs, line 27: The hex represents an encoded dynamic length field with total length 4 and length of the first array of 4, correct? Can you add more context for why the decoded array has one element with value 0? Somehow I would have expected either an empty because there is no data after the encoded data length, or an array with 4 elements that are all `0`?

packages/protocol-parser/src/encodeField.ts
```diff
@@ -8,10 +8,13 @@ export function encodeField<TSchemaAbiType extends SchemaAbiType>(
 ): Hex {
   if (isArrayAbiType(fieldType) && Array.isArray(value)) {
     const staticFieldType = arrayAbiTypeToStaticAbiType(fieldType);
-    return encodePacked(
-      value.map(() => staticFieldType),
-      value
-    );
+    // TODO: we can remove conditional once this is fixed: https://github.com/wagmi-dev/viem/pull/1147
```
alvrs, line 11: seems like it got merged already!

packages/protocol-parser/src/encodeKey.ts
```diff
@@ -0,0 +1,10 @@
+import { isStaticAbiType } from "@latticexyz/schema-type";
+import { Hex } from "viem";
+import { SchemaToPrimitives, KeySchema } from "./common";
+import { encodeKeyTuple } from "./encodeKeyTuple";
+
+export function encodeKey<TSchema extends KeySchema>(keySchema: TSchema, key: SchemaToPrimitives<TSchema>): Hex[] {
+  const staticFields = Object.values(keySchema).filter(isStaticAbiType);
+  // TODO: refactor and move all encodeKey logic into this method so we can delete encodeKey
```
alvrs, line 8:
```suggestion
  // TODO: refactor and move all encodeKeyTuple logic into this method so we can delete encodeKeyTuple
```

packages/protocol-parser/src/padSliceHex.ts
```diff
@@ -0,0 +1,8 @@
+import { Hex } from "viem";
+
+export function padSliceHex(data: Hex, start: number, end?: number): Hex {
```
alvrs, line 3: should we add a small typedoc comment with something like "slices the hex string while padding data out of bound data with `0`; including start index, excluding end index"?

packages/protocol-parser/src/encodeField.ts
holic: yep! we'll need to bump viem to get this change so gonna leave this here for now

packages/protocol-parser/src/decodeKey.ts
holic: it's used by `blockLogsToStorage` which is getting refactored in the other PR

didn't wanna refactor everything quite yet to avoid a big rebase of the other PR, just wanted to add new methods and deprecate old ones for an easy first pass

packages/protocol-parser/src/decodeRecord.test.ts
holic: I think I had added this test for proof that `padSliceHex` was doing the right thing internally when used within `decodeRecord`. I can't recall where exactly I was getting a the hex value but I copied it from some tests or logs that were failing without `padSliceHex`.

To explain what's going on:

The total length and first array length is the byte length (4 bytes in the case of `uint32`). Meaning that there's one item/value for the `uint32[]` field.

The record hex is "trimmed" to the right-most byte (end of the counter). On chain, any bytes after the right-most byte would be treated as zeros, so we want to have the same behavior in the client.

So if we were to read the same value on chain, we should get a value of `[0]` for this record (one `uint32` value with unset bytes, i.e. `0`). This test ensures that the same behavior happens in the client.

alvrs: Ahh makes sense, had forgotten that the length corresponds to bytes and assumed elements. Thanks for the explanation!

packages/protocol-parser/src/padSliceHex.ts
holic: done! also renamed to `readHex` because I didn't like `padSliceHex`

packages/protocol-parser/src/decodeKey.ts
holic: updated the original issue to remind me to refactor this: https://github.com/latticexyz/mud/issues/1296

alvrs (approved):

Issue #1125, opened 2023-07-10
Use `0x00` as key for singleton tables instead of empty array, require at least one element in a keyTuple


alvrs:
the current empty array leads to a couple subtle issues: in “KeysWithValue” it’s not possible to distinguish the “singleton key” from “no keys”, and in MODE and the client we currently need custom handling of singleton tables because of the empty array. It might be easier to just use a known key, maybe just “0x00”, to use for singleton tables, so that all other tools and integrations don’t need a special case for it

holic:
Hmm, I can't think of anything. Maybe keys aren't an issue with zero/uninitialized values, only values (where it's hard to tell if something is set or not set)

holic:
>  in “KeysWithValue” it’s not possible to distinguish the “singleton key” from “no keys”

This seems fine, since a singleton key/empty schema is for the entire table (only one value stored rather than one per key), so you wouldn't even use/need `KeysWithValue`, right?

> in MODE and the client we currently need custom handling of singleton tables because of the empty array

I think I've got a decent fix for this in the TS sync stack/indexer, where a singleton key is effectively `0x` and when you decode it it becomes `[]`.

alvrs:
In my opinion it would be cleaner to use `0x00` as key for the singleton table to avoid the need for special cases for handling this (ie checking the table's key length). One example use case is a module like the previous `SnapSync` module that installs `KeysInTable` on all tables and then uses it to fetch the table's values. It doesn't feel right to handle singleton tables differently here - the `SnapSync` module would have to be aware of which tables are singleton tables and which are not and then fetch the singleton table's value via a hardcoded `[]` key. And even with that it wouldn't be possible to distinguish onchain whether the table has a value or no value (it would return the default value even if no value is set).

alvrs:
Another approach for how the `KeysInTable` / `KeysWithValue` modules could handle this might be storing `abiEncodedKeyTuple` instead of storing each key in a separate array. That way we would be able to infer whether there is a value set in the table and also would be able to support an arbitrary number of keys (currently the limit is 5). And we'd be able to continue to use the empty array for singleton tables, which is a better "conceptual" match, and slightly cheaper in gas.

holic:
Started fiddling with this but quickly realized we can't take that approach without support for `bytes[]`. KeysWithValue and KeysInTable both return an array of keys, and if we switch to ABI-encoded keys (arbitrary length `bytes`), then we won't be able to easily store these.

[...]

PR #1446, opened 2023-09-11
feat(world): don't index keys of singleton tables

holic:
see https://github.com/latticexyz/mud/issues/1125#issuecomment-1672892617


holic:
closing as outdated

PR #1452, opened 2023-09-12
fix(world): register Delegations table in CoreModule

alvrs:
small patch to register the Delegations table so it can be synced by the sync stack


.changeset/quiet-dancers-prove.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/world": patch
+---
+
+- Register `Delegations` table in the `CoreModule`
```
holic, line 5: do we need the bullet here?

packages/world/src/modules/core/CoreModule.sol
```diff
@@ -58,6 +59,7 @@ contract CoreModule is IModule, WorldContextConsumer {
    */
   function _registerCoreTables() internal {
     InstalledModules.register();
+    Delegations.register();
```
holic, line 62: any easy way to add some tests to ensure we're registering all of our expected core tables?

holic (dismissed):

.changeset/quiet-dancers-prove.md
alvrs, line 5:
```suggestion
Register `Delegations` table in the `CoreModule`
```

packages/world/src/modules/core/CoreModule.sol
alvrs: no automated way, maintaining the test would have the same issue as maintaining the actual registration

alvrs: well we could add some sort of e2e test that checks that all tables in the world config are registered, but for that we first need to separate the test tables

holic: opened an issue to track here: https://github.com/latticexyz/mud/issues/1456

PR #1457, opened 2023-09-12
feat(world): change requireOwnerOrSelf to requireOwner

alvrs:


alvrs:
* main
  * **PR #1457** <a href="https://app.graphite.dev/github/pr/latticexyz/mud/1457" target="_blank">Graphite</a>  👈
    * **PR #1458** <a href="https://app.graphite.dev/github/pr/latticexyz/mud/1458" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/latticexyz/mud/1457?utm_source=stack-comment).

packages/world/src/modules/uniqueentity/UniqueEntityModule.sol
```diff
@@ -19,12 +19,16 @@ import { NAMESPACE, MODULE_NAME, SYSTEM_NAME, TABLE_NAME } from "./constants.sol
 contract UniqueEntityModule is IModule, WorldContextConsumer {
```
holic, line 19: should this and others use `is Module` instead of `is IModule`?

packages/world/src/modules/core/CoreModule.sol
```diff
@@ -81,11 +85,9 @@ contract CoreModule is IModule, WorldContextConsumer {
       msgSender: _msgSender(),
       msgValue: 0,
       target: coreSystem,
-      funcSelectorAndArgs: abi.encodeWithSelector(
-        WorldRegistrationSystem.registerSystem.selector,
-        ResourceSelector.from(ROOT_NAMESPACE, CORE_SYSTEM_NAME),
-        coreSystem,
-        true
+      funcSelectorAndArgs: abi.encodeCall(
```
holic, line 88: not blocking this PR but would it make more sense to call this `calldata`?

packages/world/src/modules/keysintable/KeysInTableHook.sol
```diff
@@ -58,68 +59,68 @@ contract KeysInTableHook is IStoreHook {
 
   function onBeforeDeleteRecord(bytes32 tableId, bytes32[] memory key, Schema) public {
     bytes32 keysHash = keccak256(abi.encode(key));
-    (bool has, uint40 index) = UsedKeysIndex.get(tableId, keysHash);
+    (bool has, uint40 index) = UsedKeysIndex.get(UsedKeysIndexTableId, tableId, keysHash);
```
holic, line 62: does StoreSwitch not work here anymore?

dk1a: I assume it's cause they got namespaced?

packages/world/src/modules/keyswithvalue/KeysWithValueModule.sol
```diff
@@ -23,43 +24,55 @@ import { getTargetTableSelector } from "../utils/getTargetTableSelector.sol";
...
+            onAfterSetRecord: false,
+            onBeforeSetField: true,
+            onAfterSetField: true,
+            onBeforeDeleteRecord: true,
+            onAfterDeleteRecord: false
+          })
+        )
+      )
     );
+    if (!success) revertWithBytes(returnData);
```
holic, line 72: would it make sense to move this pattern into a lib for the commonly used methods for ergonomics?

```solidity
WorldRoot.registerStoreHook(...)
```

packages/world/src/modules/uniqueentity/UniqueEntityModule.sol
```diff
@@ -19,12 +19,16 @@ import { NAMESPACE, MODULE_NAME, SYSTEM_NAME, TABLE_NAME } from "./constants.sol
...
   // Since the UniqueEntitySystem only exists once per World and writes to
   // known tables, we can deploy it once and register it in multiple Worlds.
-  UniqueEntitySystem immutable uniqueEntitySystem = new UniqueEntitySystem();
+  UniqueEntitySystem private immutable uniqueEntitySystem = new UniqueEntitySystem();
 
   function getName() public pure returns (bytes16) {
     return MODULE_NAME;
   }
 
+  function installRoot(bytes memory args) public {
```
dk1a, line 28: is it ok that `installRoot` won't delegatecall anything here?

dk1a (approved):

packages/world/src/modules/uniqueentity/UniqueEntityModule.sol
alvrs: yeah this module doesn't require any root access, it's just registering the table and system in its own namespace

packages/world/src/modules/keyswithvalue/KeysWithValueModule.sol
alvrs: agree it would be more ergonomic to use, but unfortunately more gas expensive because of all the parameters that need to be passed around in memory, and that library would be pretty painful to maintain (because it would have to include all the function signatures of all systems of `IBaseWorld` which includes all the functions the `CoreModule` registers)

packages/world/src/modules/core/CoreModule.sol
alvrs: we use `funcSelectorAndArgs` in a couple places, could consider changing that to `callData` (`calldata` wouldn't work because it's a reserved word). Let's track in an issue!

alvrs: https://github.com/latticexyz/mud/issues/1460

packages/world/src/modules/keysintable/KeysInTableHook.sol
alvrs: yeah exactly, it's not using the root namespace anymore but we don't support separate namespaces in the config yet, so need to do it manually here for now by creating a separate table id

alvrs: i think technically it doesn't have to be namespaced though but could continue using the root namespace since we only support installing it as root module for now anyway

alvrs: Reverted this change. Had originally started to refactor the module to support non-root installations but then realized it would be a bigger change because we need to register a store hook on an arbitrary table.

alvrs:
no major changes since last approvals, going ahead with mege

PR #1464, opened 2023-09-13
test: add script to generate test data

### LINKED ISSUE #1402: create a script to generate block logs fixture for tests
We have a few hardcoded test fixtures in client code that uses block logs that I generated on the fly. As we've made protocol changes, I keep having to do one-off generating of these test fixtures/block logs. We should have a script that generates the subset of block logs we need to write tests and then put those logs in a common place to import across tests.

holic:
closes #1402

Adds an e2e package to generate viem logs file that we can use as test fixtures in tests in protocol-parser, store-sync, etc. This gives us an easy way to generate new logs when we do protocol-level changes.


packages/block-logs-stream/package.json
```diff
@@ -25,8 +25,6 @@
   },
   "dependencies": {
     "@latticexyz/common": "workspace:*",
-    "@latticexyz/config": "workspace:*",
-    "@latticexyz/schema-type": "workspace:*",
```
holic, line 29: these weren't being used

holic:
I opted for a full World deployment for now, but we can add more deployment types/files for specific use cases as needed.

alvrs (approved):
non-blocking thought: would it be useful to have a matching json of expected decoded values? Could achieve this by defining some input data json and then only fetching logs after the world deployment. It seems like right now most of the logs in `world-logs.json` come from the deployment.

holic:
> non-blocking thought: would it be useful to have a matching json of expected decoded values? Could achieve this by defining some input data json and then only fetching logs after the world deployment. It seems like right now most of the logs in `world-logs.json` come from the deployment.

these will soon be useful! for example, we're about to wire up `callFrom` support in `createContract` and in dev tools, which will depend on a `getFunctionSelector` function in the `store-sync/recs` tooling, which will need to look up values in `FunctionSelectors` table (registered + populated in the deploy)

alvrs:
> these will soon be useful! for example, we're about to wire up callFrom support in createContract and in dev tools, which will depend on a getFunctionSelector function in the store-sync/recs tooling, which will need to look up values in FunctionSelectors table (registered + populated in the deploy)

My question is more about how do we know "what these logs are", like what their expected decoded value is. Do we need to manually define that somewhere? At least for the logs for setting values after the deployment is complete it might be useful to have the "input data" defined as json somewhere and a way to match the input value to a log.

PR #1466, opened 2023-09-13
feat(common,create-mud): route contract calls to use callFrom

y77cao:


packages/store-sync/src/recs/syncToRecs.ts
```diff
@@ -59,9 +66,49 @@ export async function syncToRecs<TConfig extends StoreConfig = StoreConfig>({
...
+
+    // If we can't find selectors due to not being synced yet, we can try to read them from the world contract
+    if (!selectors) {
+      const encodedValueSchema = valueSchemaToHex(components.FunctionSelectors.metadata.valueSchema);
+
+      const selectorRecord = (await publicClient.readContract({
+        address: address as Hex,
+        abi,
+        functionName: "getRecord",
+        args: [components.FunctionSelectors.id, [entity], encodedValueSchema],
```
holic, line 97: FYI if you rebase or merge main, the `valueSchema` arg here has changed to `fieldLayout`, which we'll need to compute from the schema (or store it in the component metadata when the table is registered)

packages/store-sync/src/recs/syncToRecs.ts
```diff
@@ -59,9 +68,61 @@ export async function syncToRecs<TConfig extends StoreConfig = StoreConfig>({
...
+
+    let selectors = getComponentValue(components.FunctionSelectors, entity);
+
+    // If we can't find selectors due to not being synced yet, we can try to read them from the world contract
+    if (!selectors) {
+      // TODO move to computation after table registration?
+      const schemaTypes = Object.values(components.FunctionSelectors.metadata.valueSchema).map((abiOrUserType) => {
+        const { schemaType } = resolveAbiOrUserType(abiOrUserType, config);
+        return schemaType;
+      });
```
holic, line 97: the values of `valueSchema` should already be abi types (not user types)

packages/store-sync/src/recs/syncToRecs.ts
```diff
@@ -59,9 +68,61 @@ export async function syncToRecs<TConfig extends StoreConfig = StoreConfig>({
 
   world.registerDisposer(stopSync);
 
+  const functionSelectorToResourceSelector = new Map<Hex, Hex>();
+
+  const getResourceSelector = async (functionName: string): Promise<Hex> => {
+    const functionSignature = getAbiItem({
```
holic, line 74: can this throw?

packages/store-sync/src/recs/syncToRecs.ts
```diff
@@ -59,9 +68,61 @@ export async function syncToRecs<TConfig extends StoreConfig = StoreConfig>({
...
+      abi,
+      name: functionName,
+    }) as AbiFunction;
```
holic, line 77: instead of passing the ABI around, I wonder if `getResourceSelector` should take in `functionSignature` or `functionSelector` as an argument (rather than function name) and we do the `functionSignature`/`functionSelector` calculation inside `createContract` (where we already have the ABI)

packages/store-sync/src/recs/syncToRecs.ts
y77cao: No - will be undefined if function doesn't exist. I can add a custom error for that

packages/store-sync/src/recs/syncToRecs.ts
y77cao: Agree - but abi is also needed to make the getRecord rpc call. I can move the abi to be a param of getResourceSelector instead of syncToRecs

holic: since we're doing a generic Store call (`getRecord` comes from `Store`), we could use the store ABI for that (available as an import)

e2e/packages/client-vanilla/src/mud/setupNetwork.ts
```diff
@@ -26,20 +26,21 @@ export async function setupNetwork() {
...
     walletClient: burnerWalletClient,
-  });
-
-  const { components, latestBlock$, blockStorageOperations$, waitForTransaction } = await syncToRecs({
-    world,
-    config: mudConfig,
-    address: networkConfig.worldAddress as Hex,
-    publicClient,
-    startBlock: BigInt(networkConfig.initialBlockNumber),
-    indexerUrl: networkConfig.indexerUrl ?? undefined,
```
holic, line 42: this got removed, that's why e2e tests are failing

packages/common/src/createContract.ts
```diff
@@ -47,6 +53,7 @@ export type CreateContractOptions<
   TPublicClient extends PublicClient<TTransport, TChain>,
   TWalletClient extends WalletClient<TTransport, TChain, TAccount>
 > = Required<GetContractParameters<TTransport, TChain, TAccount, TAbi, TPublicClient, TWalletClient, TAddress>> & {
+  getResourceSelector: (functionSelector: Hex) => Promise<Hex>;
```
holic, line 56: since we already have a conditional below to figure out if we should use callFrom, we could make this arg optional to avoid this becoming a breaking change

packages/common/src/createContract.ts
```diff
@@ -150,13 +158,43 @@ export function createContract<
...
+              request = {
+                address,
+                abi,
+                functionName: "callFrom",
+                args: argsForCallFrom,
+                ...options,
+              };
+            }
 
             const result = write(request);
```
holic, line 199: since we made this function `async`, I'd like to add a note here about intentionally not using `await` here?

```suggestion
            // We intentionally don't use `await` here so we can pass the `result` promise immediately to the `onWrite` call and consumers (i.e. dev tools) can use the write request before it hits the RPC.
            const result = write(request);
```

packages/common/src/createContract.ts
```diff
@@ -150,13 +158,43 @@ export function createContract<
               }
             >getFunctionParameters(parameters as any);
 
-            const request = {
+            const functionSignature = getAbiItem<Abi, string>({
+              abi,
+              name: functionName,
+            }) as AbiFunction;
+            const functionSelector = getFunctionSelector(functionSignature);
+            if (!functionSignature) throw new Error(`Unable to get function signature for ${functionName}`);
```
holic, line 166:
```suggestion
            });
            if (!functionSignature) throw new Error(`Unable to get function signature for ${functionName}`);
            const functionSelector = getFunctionSelector(functionSignature);
```
casting `as AbiFunction` and hiding the possible `undefined` seems unwise if we're already checking for the `functionSignature` later, can just move up the conditional

holic: furthermore, do you think throwing is the right outcome here? an alternative is to `console.warn` and let this pass through as a regular write request (i.e. don't route it through `callFrom`)

packages/common/src/createContract.ts
```diff
@@ -150,13 +158,43 @@ export function createContract<
...
+
+            let request: WriteContractParameters = {
               address,
               abi,
               functionName,
               args,
               ...options,
             };
+            const resourceSelector = await getResourceSelector(functionSelector);
+            const shouldUseCallFrom = resourceSelector && resourceSelector !== staticAbiTypeToDefaultValue.bytes32;
```
holic, line 176: I see you're doing a falsy check here. Should we update our `getResourceSelector` type to allow for `Promise<Hex | null>`?

packages/store-sync/src/recs/syncToRecs.ts
```diff
@@ -15,6 +19,7 @@ type SyncToRecsOptions<TConfig extends StoreConfig = StoreConfig> = SyncOptions<
 type SyncToRecsResult<TConfig extends StoreConfig = StoreConfig> = SyncResult & {
   components: RecsStorageAdapter<TConfig>["components"];
   stopSync: () => void;
+  getResourceSelector: (functionSelector: Hex) => Promise<Hex>;
```
holic, line 22: might be worth adding this to `SyncResult` type and filling this in for sqlite/postgres too

packages/store-sync/src/recs/syncToRecs.ts
```diff
@@ -59,9 +64,48 @@ export async function syncToRecs<TConfig extends StoreConfig = StoreConfig>({
...
+  const getResourceSelector = async (functionSelector: Hex): Promise<Hex> => {
+    if (functionSelectorToResourceSelector.has(functionSelector)) {
+      return functionSelectorToResourceSelector.get(functionSelector) as Hex;
+    }
+
+    const entity = encodeEntity<KeySchema>(components.FunctionSelectors.metadata.keySchema, {
+      functionSelector,
+    });
+
+    const selectors = getComponentValue(components.FunctionSelectors, entity);
```
holic, line 78: this only returns
```suggestion
    const selector = getComponentValue(components.FunctionSelectors, entity);
```
this only returns one selector

packages/store-sync/src/recs/syncToRecs.ts
```diff
@@ -59,9 +64,48 @@ export async function syncToRecs<TConfig extends StoreConfig = StoreConfig>({
...
+
+    // If we can't find selectors due to not being synced yet, we can try to read them from the world contract
+    if (!selectors) {
+      // TODO make fieldLayout a table metadata field
+      const encodedFieldLayout = valueSchemaToFieldLayoutHex(components.FunctionSelectors.metadata.valueSchema);
+
+      const selectorRecord = (await publicClient.readContract({
+        address: address as Hex,
+        abi: IStoreAbi,
```
holic, line 87: this is a small thing that won't impact functionality, but given that function selectors exist as a world concept, might be clearer to use the `IBaseWorld` ABI here (it inherits `IStore` so it'll still have `getRecord`)

packages/store-sync/src/recs/syncToRecs.ts
```diff
@@ -59,9 +64,48 @@ export async function syncToRecs<TConfig extends StoreConfig = StoreConfig>({
 
   world.registerDisposer(stopSync);
 
+  const functionSelectorToResourceSelector = new Map<Hex, Hex>();
```
holic, line 67: it might make sense to keep this function not-cached and move the caching of it into `createContract`

packages/store-sync/src/recs/syncToRecs.ts
```diff
@@ -59,9 +64,48 @@ export async function syncToRecs<TConfig extends StoreConfig = StoreConfig>({
...
+    if (!selectors) {
+      // TODO make fieldLayout a table metadata field
+      const encodedFieldLayout = valueSchemaToFieldLayoutHex(components.FunctionSelectors.metadata.valueSchema);
+
+      const selectorRecord = (await publicClient.readContract({
+        address: address as Hex,
+        abi: IStoreAbi,
+        functionName: "getRecord",
+        args: [components.FunctionSelectors.id as Hex, [entity as Hex], encodedFieldLayout],
+      })) as Hex;
```
holic, line 90: why do we have to cast this?

packages/store-sync/src/recs/syncToRecs.ts
```diff
@@ -59,9 +64,48 @@ export async function syncToRecs<TConfig extends StoreConfig = StoreConfig>({
...
+
+      const decodedSelectors = decodeValue<ValueSchema>(
+        components.FunctionSelectors.metadata.valueSchema,
+        selectorRecord
+      );
+
+      functionSelectorToResourceSelector.set(functionSelector, decodedSelectors.resourceSelector as Hex);
+      return decodedSelectors.resourceSelector as Hex;
```
holic, line 98: why do we have to cast this?

packages/common/src/createContract.ts
y77cao: `getFunctionSelector` will throw slicing error if `functionSignature` is an unexpected type, so might be better to throw a clearer message earlier. This case is probably unlikely to happen since the abi should contain the signature for the given functionName.

packages/store-sync/src/recs/syncToRecs.ts
y77cao: Looks like it returns both `resourceSelector` and `systemFunctionSelector`...? We are only grabbing `resourceSelector` here for sure

packages/store-sync/src/recs/syncToRecs.ts
y77cao: for args - these are required to have `0x${string}` type as inferred from the abi, but the id and entity are both `string` so I casted it to a stricter type here

for selectorRecord - don't really need to cast that. Will update

packages/store-sync/src/recs/syncToRecs.ts
y77cao: Actually don't need to - will update

packages/common/src/createContract.ts
y77cao: don't need a falsy check here actually since that's handled in the method and should always return Hex. Will update

packages/store-sync/src/recs/syncToRecs.ts
holic: you're right, sorry! didn't realize we had two properties on this table

packages/common/src/createContract.ts
holic: would we want to try/catch in that case?

packages/common/src/createContract.ts
y77cao: actually nvm needed falsy check still after making `getResourceSelector ` optional. But don't need to update its return type

packages/common/src/createContract.ts
y77cao: that seems equivalent to throwing early here since the only case that `getFunctionSelector ` will throw is the input being either null or undefined. Do we have a preference to use try catch instead?

packages/store-sync/src/recs/syncToRecs.ts
y77cao: the current implementation depends on Components which is unique to RECS it seems? Not very familar with the code there yet but looks like for both we can only use the RPC call. Do you mind me handling it as a follow-up?

e2e/packages/test-data/generate-test-data.ts
```diff
@@ -68,6 +68,7 @@ const worldContract = createContract({
   abi: IWorldAbi,
   publicClient,
   walletClient,
+  getResourceSelector: async (selector) => selector,
```
holic, line 71:
```suggestion
```
this is optional now, so we can delete

packages/store-sync/src/recs/syncToRecs.ts
holic: yes, we'd need to add implementations for each sqlite and postgres, but it should be fairly straightforward: we use drizzle for queries and there should be strongly typed schemas for built-in tables like FunctionSelectors

packages/store-sync/src/recs/syncToRecs.ts
```diff
@@ -59,9 +64,46 @@ export async function syncToRecs<TConfig extends StoreConfig = StoreConfig>({
 
   world.registerDisposer(stopSync);
 
+  const getResourceSelector = async (functionSelector: Hex): Promise<Hex> => {
+    const entity = encodeEntity<
+      KeySchema & {
+        functionSelector: "bytes4";
+      }
```
holic, line 71: why do we need this? I would expect it to be inferred from `components.FunctionSelectors.metadata.keySchema`

packages/store-sync/src/recs/syncToRecs.ts
```diff
@@ -59,9 +64,46 @@ export async function syncToRecs<TConfig extends StoreConfig = StoreConfig>({
...
+        abi: IBaseWorldAbi,
+        functionName: "getRecord",
+        args: [components.FunctionSelectors.id as Hex, [entity as Hex], encodedFieldLayout],
+      });
+
+      const decodedSelectors = decodeValue<
+        ValueSchema & {
+          resourceSelector: "bytes32";
+          systemFunctionSelector: "bytes4";
+        }
```
holic, line 94: same as above, this should be inferred from `components.FunctionSelectors.metadata.valueSchema`

packages/store-sync/src/recs/syncToRecs.ts
y77cao: Sg!

packages/store-sync/src/recs/syncToRecs.ts
y77cao: `SchemaToPrimitives` didn't seem to pick up the inferred type but it works now. Will remove

PR #1467, opened 2023-09-13
test(store-sync): add recs storage tests

holic:
much easier to write full tests now that we have logs from #1464 

kept these tests simple for now since part of it (`blockLogsToStorage`) will be going away in #1354 and didn't wanna cause a big refactor there


e2e/packages/test-data/generate-test-data.ts
```diff
@@ -86,10 +87,7 @@ const logs = (
 ).flatMap((range) => range.logs);
 
 console.log("writing logs to", logsFilename);
-await fs.writeFile(
-  logsFilename,
-  JSON.stringify(logs, (k, v) => (typeof v === "bigint" ? v.toString() : v), 2)
-);
```
holic, line 92: manually transforming `bigint`s didn't work well because we want to pass this into functions that expect `Log` (which expect `bigint` not strings)

packages/store-sync/src/recs/recsStorage.test.ts
```diff
@@ -0,0 +1,47 @@
...
+import superjson from "superjson";
+// TODO: figure out if this needs to be imported from a package rather than relative path
+// TODO: if relative imports are fine, probably should put the e2e test data with its mud config?
+import mudConfig from "../../../../e2e/packages/contracts/mud.config";
+import testLogsJson from "../../../../test-data/world-logs.json";
+import { groupLogsByBlockNumber } from "@latticexyz/block-logs-stream";
+import { StoreEventsLog } from "../common";
+import { singletonEntity } from "./singletonEntity";
+
+const testLogs = superjson.deserialize<readonly StoreEventsLog[]>(testLogsJson as any);
```
holic, line 14: superjson wants a more generic type while the import has a specific type, so we cast to `any`

image

I wanted to cast as `SuperJSONResult` but superjson doesn't export this, and it wouldn't let me cast to `ReturnType<typeof superjson.serialize>` (effectively `SuperJSONResult`)

packages/store-sync/src/recs/recsStorage.test.ts
```diff
@@ -0,0 +1,47 @@
+import { describe, expect, it } from "vitest";
+import { blockLogsToStorage } from "../blockLogsToStorage";
+import { recsStorage } from "./recsStorage";
+import { createWorld, getComponentEntities, getComponentValue } from "@latticexyz/recs";
+import superjson from "superjson";
+// TODO: figure out if this needs to be imported from a package rather than relative path
+// TODO: if relative imports are fine, probably should put the e2e test data with its mud config?
+import mudConfig from "../../../../e2e/packages/contracts/mud.config";
+import testLogsJson from "../../../../test-data/world-logs.json";
```
holic, line 9: @alvrs any opinions here?

e2e/packages/test-data/generate-test-data.ts
holic: ended up moving to rpc logs instead, which are json serializable

packages/store-sync/src/recs/recsStorage.test.ts
holic: ended up moving to rpc logs instead, which are json serializable

packages/store-sync/src/recs/recsStorage.ts
```diff
@@ -13,25 +12,44 @@ import { isDefined } from "@latticexyz/common/utils";
...
-  const componentsByTableId = Object.fromEntries(
-    Object.entries(components).map(([id, component]) => [component.id, component])
-  );
+  const components = {
+    ...configToRecsComponents(world, config),
+    ...configToRecsComponents(world, storeConfig),
+    ...configToRecsComponents(world, worldConfig),
+    ...defineInternalComponents(world),
+  };
+  // TODO: do we need to store block number?
```
alvrs, line 50: I see this TODO was already there before so not directly related to this PR, but could we add some more context / track this in an issue? This is a more general thing, I feel like for TODO comments to be useful it needs to be clear how to resolve them when eg. when we find them in a couple weeks/months and we don't have the implicit context anymore that we had when writing the TODO

alvrs (dismissed):
nothing blocking

packages/store-sync/src/recs/recsStorage.test.ts
```diff
@@ -0,0 +1,56 @@
...
+      ]
+    `);
+
+    expect(getComponentValue(components.NumberList, singletonEntity)).toMatchInlineSnapshot(`
+      {
+        "value": [
+          420,
+          69,
+        ],
+      }
```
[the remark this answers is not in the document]
holic: there's also some nuance to figure out around inputs (e.g. method calls or txs to get us here) and the intended/expected final state (what's shown in these snapshots)

packages/store-sync/src/recs/recsStorage.ts
holic: agreed on the meta point! been trying to be more diligent about that

I think this specific comment refers to how we track block numbers per row in sqlite/postgres adapters to keep track of when a record was last updated (for the future purposes of streaming changed data to client), but not sure how necessary it is here

holic:
FYI I only removed comments + merged main since last approval

PR #1470, opened 2023-09-13
feat(cli,world): batch register function selectors 

CrazyNorman:
The old pull request seems to lag behind for a long time due to conflicts with later commits. Therefore, I have created a new updated pull request here for easier merging.

Please refer to the rationale from the old PR: https://github.com/latticexyz/mud/pull/1287.

fix https://github.com/latticexyz/mud/issues/1280


packages/world/src/modules/core/implementations/WorldRegistrationSystem.sol
```diff
@@ -174,6 +174,19 @@ contract WorldRegistrationSystem is System, IWorldErrors {
...
+   */
+  function registerFunctionSelectors(
+    bytes32 resourceSelector,
+    string[] calldata systemFunctionNames,
+    string[] calldata systemFunctionArguments
+  ) public {
+    for (uint256 i = 0; i < systemFunctionNames.length; i++) {
+      registerFunctionSelector(resourceSelector, systemFunctionNames[i], systemFunctionArguments[i]);
+    }
+  }
```
holic, line 188: one small optimization we could do here is move the access control check outside this loop so we're not checking it on each iteration (via `registerFunctionSelector`)

packages/world/src/modules/core/implementations/WorldRegistrationSystem.sol
```diff
@@ -200,6 +213,21 @@ contract WorldRegistrationSystem is System, IWorldErrors {
...
+  /**
+   * Register root World function selectors in batch (without namespace / name prefix).
+   * Requires the caller to own the root namespace.
+   */
+  function registerRootFunctionSelectors(
+    bytes32 resourceSelector,
+    bytes4[] calldata worldFunctionSelectors,
+    bytes4[] calldata systemFunctionSelectors
+  ) public {
+    // Register the function selectors
```
holic, line 225:
```suggestion
```

packages/world/src/modules/core/implementations/WorldRegistrationSystem.sol
```diff
@@ -200,6 +213,21 @@ contract WorldRegistrationSystem is System, IWorldErrors {
...
+    for (uint256 i = 0; i < worldFunctionSelectors.length; i++) {
+      registerRootFunctionSelector(resourceSelector, worldFunctionSelectors[i], systemFunctionSelectors[i]);
```
holic, line 227: do we need to validate that `worldFunctionSelectors.length == systemFunctionSelectors.length`?

packages/world/src/modules/core/implementations/WorldRegistrationSystem.sol
```diff
@@ -174,6 +174,19 @@ contract WorldRegistrationSystem is System, IWorldErrors {
     FunctionSelectors.set(worldFunctionSelector, resourceSelector, systemFunctionSelector);
   }
 
+  /**
+   * Register World function selectors in batch for the given namespace, name and system function.
+   */
+  function registerFunctionSelectors(
+    bytes32 resourceSelector,
+    string[] calldata systemFunctionNames,
+    string[] calldata systemFunctionArguments
```
holic, line 183: would `calldata` here allow for calling this on-chain? I thought that the reason for `memory` was to allow for both on-chain calls and off-chain calls

packages/world/src/modules/core/implementations/WorldRegistrationSystem.sol
CrazyNorman: Reasonable. I have moved the access control check from the inner function to the batch function now.

packages/world/src/modules/core/implementations/WorldRegistrationSystem.sol
CrazyNorman: OK Idk we need to call it on-chain. I have updated it to `memory` now

packages/world/src/modules/core/implementations/WorldRegistrationSystem.sol
```diff
@@ -151,9 +151,6 @@ contract WorldRegistrationSystem is System, IWorldErrors {
     string memory systemFunctionName,
     string memory systemFunctionArguments
   ) public returns (bytes4 worldFunctionSelector) {
-    // Require the caller to own the namespace
-    AccessControl.requireOwner(resourceSelector, _msgSender());
```
holic, line 155: you've removed a very important check from a public function 🙈

CrazyNorman: OK I see🙈 I should add an internal function to refactor it, updated.

CrazyNorman:
I have checked the test error and it seems that I need to update the gas report. I have generated the report locally and committed it.

CrazyNorman:
I found the gasUsed in my local machine is exactly 2 gas less than the one in test. So should I setup a container with the same env as test script to test it, and then commit?

holic:
> I found the gasUsed in my local machine is exactly 2 gas less than the one in test. So should I setup a container with the same env as test script to test it, and then commit?

Haha I don't know what's going on here, but maybe try `foundryup` to make sure you're on a recent version of foundry?

We're also thinking of trying a more generic "multicall" approach so we can batch calls of ~anything, so going to put this PR on pause for a bit.

CrazyNorman:
> > I found the gasUsed in my local machine is exactly 2 gas less than the one in test. So should I setup a container with the same env as test script to test it, and then commit?
> 
> Haha I don't know what's going on here, but maybe try `foundryup` to make sure you're on a recent version of foundry?
> 
> We're also thinking of trying a more generic "multicall" approach so we can batch calls of ~anything, so going to put this PR on pause for a bit.

Agree. It would be better if we could batch all the registration tasks. Multicall could improve this process in a long way. 

In addition, it also would be great if we could have a simple mitigation method for function registration, which currently dominates most of the registration tasks XD

holic:
we're now able to call any system method in batch: https://github.com/latticexyz/mud/pull/1500

wanna make a new PR for the CLI changes to register function selectors in batch using this new approach?

PR #1472, opened 2023-09-13
feat(world,store): add initialize method, initialize core tables in core module

alvrs:
purpose is to save a lot of bytecode in the `World` and `WorldFacory` contracts

| Before (contract,size,margin) | After (contract,size,margin) |
| --- | --- |
| CleanShot 2023-09-13 at 21 53 52@2x | CleanShot 2023-09-13 at 21 54 44@2x |


packages/world/src/modules/core/CoreModule.sol
```diff
@@ -50,6 +51,7 @@ contract CoreModule is Module {
   }
 
   function installRoot(bytes memory) public override {
+    StoreCore.registerCoreTables();
```
holic, line 54: wondering if this should be in `_registerCoreTables`

.changeset/tricky-beds-kiss.md
```diff
@@ -0,0 +1,19 @@
+---
+"@latticexyz/cli": patch
+"@latticexyz/store": patch
+"@latticexyz/world": patch
```
holic, line 4: should we add another changeset for https://github.com/latticexyz/mud/pull/1385?

holic (dismissed):
wooo 4kb is a good chunk of savings

.changeset/tricky-beds-kiss.md
alvrs: ohh true forgot it there. Could do that in a separate followup though, this PR doesn't actually use the `WorldFactory` yet (because calling `initialize` in a separate transaction is supported it's not strictly necessary.)

.changeset/tricky-beds-kiss.md
```diff
@@ -0,0 +1,19 @@
...
+  This change allows the registration of all core tables to happen in the `CoreModule`, so no table metadata has to be included in the `World`'s bytecode.
+
+  ```solidity
+  interface IBaseWorld {
+    function initialize(IModule coreModule) public;
+  }
+  ```
+  
+- The `StoreRead` contract no longer calls `StoreCore.initialize` in its constructor.
+  `StoreCore` consumers are expected to call `StoreCore.initialize` and `StoreCore.registerCoreTable` in their own setup logic.
```
alvrs, line 17:
```suggestion
```
moving this to its own changeset bc it's the only `major` change

holic (approved):

PR #1473, opened 2023-09-13
feat: bump solidity to 0.8.21

### LINKED ISSUE #833: move `foundry.toml` from `solc_version` to `solc`
foundry updated this at some point but it still works for us thanks to a backwards compat layer
https://github.com/foundry-rs/foundry/blob/58a272997516046fd745f4b3c37f91d0eb113358/config/src/lib.rs#L179-L181

### LINKED ISSUE #1448: bump to latest solc version

holic:
closes #1448
closes #833


holic:
tried moving some events around as noted in #1448 but got weird natspec errors

```
Error: Internal exception in StandardCompiler::compile: /solidity/libsolidity/interface/Natspec.cpp(89): Throw in function static Json::Value solidity::frontend::Natspec::userDocumentation(const solidity::frontend::ContractDefinition &)
Dynamic exception type: boost::wrapexcept<solidity::langutil::InternalCompilerError>
std::exception::what: Solidity assertion failed
[solidity::util::tag_comment*] = Solidity assertion failed

Dynamic exception type: boost::wrapexcept<solidity::langutil::InternalCompilerError>
std::exception::what: Solidity assertion failed
[solidity::util::tag_comment*] = Solidity assertion failed
```

alvrs (dismissed):
looks good! let's add a changeset, i think this might be a breaking change as it requires consumers to update their solc version too

holic:
> looks good! let's add a changeset, i think this might be a breaking change as it requires consumers to update their solc version too

I am now kinda wondering if we want to leave the pragma alone to support a wider range of Solidity versions and just bump the foundry solc?

alvrs (approved):

PR #1475, opened 2023-09-13
draft: gas optimization

Boffee:
Pasted from discord:

Feature request: bring real world mud2 read/write gas cost to parity with sload and sstore.

Almost all of the overhead that mud2 adds on top of sload and sstore comes from computing static constants onchain. 

For the current version of mud2, warm Table.get[StaticField] cost in the ballpark of 2000 + 500 * (# of fields in key schema) + 500 * (# of fields in value schema). This can probably be reduced by 90%+ to < 400 gas:
- Hardcoding the values for getValueSchema and getKeySchema reduces the gas down to a flat 1500.
- Hardcoding everything inside StoreCore.getField except for storage location reduces the gas down to 500.
- simplifying the storage location hash function can reduces the gas down to Storage.load which is 400.
- a custom Storage.loadField might reduce gas further, though not sure by how much.

The same logic is applicable to writes.

### DIFF
--- a/packages/cli/contracts/src/codegen/tables/Dynamics1.sol
+++ b/packages/cli/contracts/src/codegen/tables/Dynamics1.sol
@@ -9,7 +9,7 @@ import { SchemaType } from "@latticexyz/schema-type/src/solidity/SchemaType.sol"
 // Import store internals
 import { IStore } from "@latticexyz/store/src/IStore.sol";
 import { StoreSwitch } from "@latticexyz/store/src/StoreSwitch.sol";
-import { StoreCore } from "@latticexyz/store/src/StoreCore.sol";
+import { StoreCore, StoreCoreInternal } from "@latticexyz/store/src/StoreCore.sol";
 import { Bytes } from "@latticexyz/store/src/Bytes.sol";
 import { Memory } from "@latticexyz/store/src/Memory.sol";
 import { SliceLib } from "@latticexyz/store/src/Slice.sol";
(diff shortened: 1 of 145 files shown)


dk1a (commented):
so far looks really good, I wanted to make basically the same thing and hadn't gotten to it yet. Do tell if you want me to take it off your hands or if u wanna finish it urself

packages/cli/contracts/src/codegen/tables/Dynamics1.sol
```diff
@@ -32,9 +36,7 @@ struct Dynamics1Data {
 library Dynamics1 {
   /** Get the table values' field layout */
   function getFieldLayout() internal pure returns (FieldLayout) {
-    uint256[] memory _fieldLayout = new uint256[](0);
-
-    return FieldLayoutLib.encode(_fieldLayout, 5);
+    return _fieldLayout;
   }
```
holic, line 40: do we even need `getFieldLayout` if we have it as a constant above? wonder if it's worth just exporting that as a named constant like we do for table ID

Boffee:
> so far looks really good, I wanted to make basically the same thing and hadn't gotten to it yet. Do tell if you want me to take it off your hands or if u wanna finish it urself

Looks like I snatched up all the low hanging fruits lol. Take over if your hands are free. I can help if you're busy with other stuff.

holic:
I think ~all of these were merged as separate smaller PRs, so closing this one!

PR #1476, opened 2023-09-14
feat(protoocl-parser): add valueSchemaToFieldLayoutHex

holic:
I lost track of where the comment was but I said I'd swap out `fieldLayoutToHex` with a thing that uses `ValueSchema`. This should help with https://github.com/latticexyz/mud/pull/1466 cc @y77cao 

Unfortunately we don't have Solidity tests that map a raw `bytes32` value to `FieldLayout` so I can copy that over to make sure the Solidity and TS line up in terms of behavior, padding, etc. So I skipped adding tests for now.

### DIFF
--- a/.changeset/silent-rice-argue.md
+++ b/.changeset/silent-rice-argue.md
@@ -1,0 +1,5 @@
+---
+"@latticexyz/protocol-parser": minor
+---
+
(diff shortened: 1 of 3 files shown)


alvrs (dismissed):

alvrs:
let's add a changeset!

PR #1483, opened 2023-09-14
feat: move forge build + abi + abi-ts to out

holic:
`forge` has a default `out` directory for build artifacts. When you `forge build` or `forge test`, everything gets built and written to this directory, including tests and scripts.

For the purposes of a clean ABI directory with matching TS types, I thought I would try to side step this by separately writing `forge build` output to an `abi` dir (via `--out abi`) that would also skip script (`.s.sol`) and test (`.t.sol`) files (via `--skip script test`).

This ended up working against us in our current CLI approach, where deploys need the systems ABI artifacts to generate an `IWorld` interface, and uses the configured forge output dir (via `forge config`) to get it. So we either had to build to both places (`out` and `abi`) or build to just `abi` but then be missing systems ABIs because they may not have been built to `out`, where forge is configured. And we can't update the `foundry.toml` to `out = abi` because that would cause `forge test` to put its artifacts there too.

We could work around this by updating our deploy pipeline to be configurable with a forge output dir (and fall back to `forge config` -> `out`), but then we'd still have `forge build --out abi --extra-files abi ...` commands littered around and that felt gross too.

So I've gone the simpler route for now and just default everything to `forge build` (with skips for tests/scripts when we're just deploying or building for ABI artifacts) and make sure to include the proper ABI files we need in our npm package artifacts.

This also moves `src/MudTest.sol` to `test/MudTest.t.sol` so it's included in `--skip test` and we don't have to manually skip it. Since this isn't used in store package and has a reference to a world, should we move this to the world package instead?


alvrs:
> This also moves src/MudTest.sol to test/MudTest.t.sol so it's included in --skip test and we don't have to manually skip it. Since this isn't used in store package and has a reference to a world, should we move this to the world package instead?

good point, should be in world!

.gitattributes
```diff
@@ -1,4 +1,2 @@
 # suppress diffs for codegen in PRs
-**/abi/**/*.json linguist-generated=true
-**/abi/**/*.json.d.ts linguist-generated=true
```
alvrs, line 3: should we suppress `out` diffs now then?

e2e/packages/contracts/package.json
```diff
@@ -5,11 +5,11 @@
   "license": "MIT",
   "scripts": {
     "build": "pnpm run build:mud && pnpm run build:abi && pnpm run build:abi-ts",
-    "build:abi": "forge build --extra-output-files abi --out abi --skip test script MudTest.sol",
-    "build:abi-ts": "mud abi-ts --input 'abi/IWorld.sol/IWorld.abi.json' && prettier --write '**/*.abi.json.d.ts'",
+    "build:abi": "forge build --skip test script",
```
alvrs, line 8: do we not need the `--extra-output-files abi` anymore?

alvrs: ahh i see it's in the forge config now

.gitattributes
alvrs: ahh i see abis are no longer checked in to git

alvrs (approved):

.changeset/many-pumpkins-cry.md
```diff
@@ -0,0 +1,18 @@
+---
+"@latticexyz/store": minor
+"@latticexyz/world": minor
+---
+
+Store and World contract ABIs are now exported from `out` directory. You'll need to update your imports like:
```
alvrs, line 6:
```suggestion
Store and World contract ABIs are now exported from the `out` directory. You'll need to update your imports like:
```

.changeset/many-pumpkins-cry.md
```diff
@@ -0,0 +1,18 @@
+---
+"@latticexyz/store": minor
+"@latticexyz/world": minor
```
alvrs, line 3:
```suggestion
"@latticexyz/store": major
"@latticexyz/world": major
```
i'd say everything that requires consumers to update something in their code is a `major` change

packages/world/test/MudTest.t.sol
```diff
@@ -2,10 +2,10 @@
 pragma solidity >=0.8.0;
```
alvrs, line 2: just realized this and other tests here are still using `0.8.0`, should we update to `0.8.21`?

.changeset/many-pumpkins-cry.md
holic: yeah originally had major but then thought just a rename felt less like a major, but happy to go with that

alvrs (approved):

PR #1485, opened 2023-09-14
Revert "fix(cli): don't build twice on contract changes"

holic:
Reverts latticexyz/mud#1480

I believe this is the cause of
```
MUDError: Error reading file at out/System.sol/System.json
```


alvrs (approved):
could we use `out` as input dir for `abi-ts` to avoid building twice here?

holic:
> could we use `out` as input dir for `abi-ts` to avoid building twice here?

working on that in a separate RP

PR #1490, opened 2023-09-14
fix: release bytecode on npm and import abi in cli deploy

alvrs:
followup to #1483 
- release bytecode of world and store on npm


holic (approved):
thanks!

PR #1495, opened 2023-09-14
docs: using the indexer with PostgreSQL

qbzzt:


docs/pages/indexer.mdx
```diff
@@ -290,3 +290,82 @@ pnpm start:sqlite
...
+   ```sh copy
+   pnpm start:postgres:local
+   ```
+
+   Otherwise, set these environment variables:
+
+   | Variable     | Value                                                            |
+   | ------------ | ---------------------------------------------------------------- |
+   | DATABASE_URL | URL for the database, of the form `postgres://<host>/<database>` |
+   | CHAIN_ID     | ChainID for the chain you index                                  |
```
ludns, line 311: we prob want to list the chain_id we currently support (and a link to that file with constants), along with instruction on how to bring custom rpc and chain_id in case it is not listed in the MUD constant file

docs/pages/indexer.mdx
```diff
@@ -290,3 +290,82 @@ pnpm start:sqlite
...
+1. Get the list of tables for a specific world (if your world address is different, use that instead of `0x5FbDB2315678afecb367f032d93F642f64180aa3`).
+
+   ```sql copy
+   \dt "0x5FbDB2315678afecb367f032d93F642f64180aa3__".*
+   ```
+
+1. Get the value of the counter.
+
+   ```sql copy
+   select * from "0x5FbDB2315678afecb367f032d93F642f64180aa3__"."Counter";
```
ludns, line 370: to make this easier, there is a way to select the schema (ie: a mud world + namespace) in SQL; and then be able to do `select * from Counter;`, which is much nicer

docs/pages/indexer.mdx
qbzzt: Done. I assume we use the same mechanism we do with sqlite, environment variables.

docs/pages/indexer.mdx
qbzzt: I'd rather wait on namespaces until MUD is up to using them in a stable manner. Are we ready for "permissionlessly extending MUD" already?

qbzzt: Other than that, done.

docs/pages/indexer.mdx
```diff
@@ -364,8 +372,14 @@ You can now use [`psql`](https://www.postgresql.org/docs/current/app-psql.html)
    \dt "0x5FbDB2315678afecb367f032d93F642f64180aa3__".*
    ```
 
+1. Specify that the default schema is the world being indexed.
```
ludns, line 375: Should we say that it is <world_address>__<namespace>?
And here our tables are in the default (root) namespace hence we use the empty string. Otherwise this lgtm!

docs/pages/indexer.mdx
```diff
@@ -388,11 +388,22 @@ If you are indexing a different `World`, change the world address and use the na
...
+
+### Deleting the PostgreSQL data
+
+PostgreSQL is a persistent data storage.
+This means that if you restart the blockchain without manually deleting the data, your indexer will use old data and provide incorrect results.
+
+To avoid this, enter `psql` and run this command:
+
+```sql copy
+DROP OWNED BY <your user name>;
```
ludns, line 408: what happens when people execute within docker? docker has different usernames I think.
should we drop the DB instead?

qbzzt:
The docker directions cover this

docs/pages/indexer.mdx
qbzzt: The docker directions cover this.

ludns (approved):

ludns (approved):

PR #1496, opened 2023-09-14
feat: replace ephemeral with offchain tables

dk1a:
fixes #1496


packages/store/src/StoreCore.sol
```diff
@@ -443,24 +444,40 @@ library StoreCore {
...
    */
-  function emitEphemeralRecord(
+  function emitSetRecord(
     bytes32 tableId,
-    bytes32[] memory key,
+    bytes32[] memory keyTuple,
     bytes memory data,
     FieldLayout fieldLayout
   ) internal {
+    if (!Tables.getOffchainOnly(tableId)) {
```
holic, line 460: do we want/need these checks?

holic: this might be another good use case for a "resource type" encoded in the resource selector, i.e. table ID

dk1a: it's pretty important to avoid divergence of offchain and onchain state. Yeah a resource type would be a nice alternative

holic:
Design question: since we're using regular events and soon we'll encode the notion of "offchain table" into the resource selector/table ID (which we'll ~always have access to in these methods), would it make sense to not have separate methods for emitting and just use e.g. `setRecord` and `deleteRecord`?

dk1a:
> would it make sense to not have separate methods for emitting and just use e.g. setRecord and deleteRecord?

I think it'd make sense, but it'd also add a little gas overhead to onchain tables, since we'll always need to extract resource type and check it. The overhead is possibly worth having 2 fewer methods

holic:
from discord:
- we're gonna pause on this until #1498 is done so we can use it here (avoids one extra refactor PR and API churn)
- we're gonna remove offchain-specific methods (`emitSet`, etc.) in favor of the emission-only check (via table type encoded into table ID) being inside existing methods (`setRecord`, etc.)
- we're gonna keep the `offchainOnly` flag in `registerTable` for clarity of intent and so we can check that the table ID has the proper table type prefix
-

holic:
closing in favor of https://github.com/latticexyz/mud/pull/1558

alvrs:
For context: decided to start over because we changed the approach away from a separate method for ephemeral tables and towards conditionally early returning in the main StoreCore methods based on the table's resource type, so it seemed like a smaller lift to start fresh than to fix the merge conflicts of this PR with the recent changes

PR #1497, opened 2023-09-15
feat(dev-tools): update actions to display function name instead of callFrom

y77cao:
Follow up for https://github.com/latticexyz/mud/pull/1466

Before:
Screenshot 2023-09-14 at 9 45 53 PM
After
Screenshot 2023-09-14 at 9 42 26 PM

### DIFF
--- a/packages/dev-tools/src/actions/WriteSummary.tsx
+++ b/packages/dev-tools/src/actions/WriteSummary.tsx
@@ -56,10 +56,16 @@ export function WriteSummary({ write }: Props) {
           .filter(isDefined)
       : null;
 
-  const functionSelectorAndArgs: Hex = write.request?.args?.length
-    ? (write.request.args[write.request.args.length - 1] as Hex)
-    : `0x`;
-  const functionData = decodeFunctionData({ abi: worldAbi, data: functionSelectorAndArgs });
+  let functionName = write.request.functionName;
+  let functionArgs = write.request.args;
+  if (functionName === "call" || functionName === "callFrom") {
+    const functionSelectorAndArgs: Hex = write.request?.args?.length
+      ? (write.request.args[write.request.args.length - 1] as Hex)
+      : `0x`;
+    const functionData = decodeFunctionData({ abi: worldAbi, data: functionSelectorAndArgs });
+    functionName = functionData.functionName;
+    functionArgs = functionData.args;
+  }
 
   return (
     <details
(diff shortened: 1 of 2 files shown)


packages/dev-tools/src/actions/WriteSummary.tsx
```diff
@@ -56,6 +56,11 @@ export function WriteSummary({ write }: Props) {
           .filter(isDefined)
       : null;
 
+  const functionSelectorAndArgs: Hex = write.request?.args?.length
```
holic, line 59: I think this won't always be the case if a system is called with arguments. Unfortunately the template doesn't have a good example of this but I think the example/minimal does (look for `InventorySystem.sol`).

Instead we might have to check if the request method matches `call` or `callFrom` and then decode accordingly.

packages/dev-tools/src/actions/WriteSummary.tsx
```diff
@@ -56,6 +56,17 @@ export function WriteSummary({ write }: Props) {
           .filter(isDefined)
       : null;
 
+  let functionName = write.request.functionName;
+  let functionArgs = write.request.args;
+  if (functionName === "call" || functionName === "callFrom") {
```
holic, line 61: note that `call` and `callFrom` have different signatures:
https://github.com/latticexyz/mud/blob/4c7fd3eb29e3d3954f2f1f36ace474a436082651/packages/world/src/interfaces/IWorldKernel.sol#L16-L32

y77cao: I see. Looks like funcSelectorAndArgs is the last param for both so should be ok here?

holic: oh yep, you're right, I missed that you were grabbing the last element, not slicing off the first element 🙈 my bad

packages/dev-tools/src/actions/WriteSummary.tsx
```diff
@@ -73,7 +84,7 @@ export function WriteSummary({ write }: Props) {
         )}
       >
         <div className="flex-1 font-mono text-white whitespace-nowrap overflow-hidden text-ellipsis">
-          {write.request.functionName}({write.request.args?.map((value) => serialize(value)).join(", ")})
+          {functionName}({functionArgs?.map((value) => serialize(value)).join(", ")})
```
holic, line 87:
```suggestion
          {functionName}({functionArgs?.map((value) => serialize(value)).join(", ")})
          {" "}
          {write.request.functionName !== functionName ? <span className="text-xs text-white/40">via {write.request.functionName}</span> : null}
```

holic (approved):
Looks great!

My comment above isn't blocking. If you do decide to commit that, it might need a pass through prettier because I just typed it into GitHub and no idea if I got the formatting correct.

holic:
just realized we can actually land this independent of #1466!

holic (approved):

PR #1498, opened 2023-09-15
WIP resource type

holic:


packages/world/src/resourceTypes.sol
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
+
+bytes2 constant RESOURCE_NAMESPACE = "ns";
+bytes2 constant RESOURCE_TABLE = "tb";
+bytes2 constant RESOURCE_TABLE_OFFCHAIN = "tf";
+bytes2 constant RESOURCE_SYSTEM = "sy";
+bytes2 constant RESOURCE_MODULE = "md";
```
holic, line 8: I wish Solidity had something like TypeScript's enums where you can define the value of each, e.g.
```ts
enum ResourceType {
  NAMESPACE = 'ns',
  TABLE = 'tb',
  ...
}
```

packages/world/src/ResourceSelector.sol
```diff
@@ -3,53 +3,65 @@ pragma solidity >=0.8.0;
 import { ROOT_NAMESPACE, ROOT_NAME } from "./constants.sol";
 import { Bytes } from "@latticexyz/store/src/Bytes.sol";
 
-bytes16 constant ROOT_NAMESPACE_STRING = bytes16("ROOT_NAMESPACE");
-bytes16 constant ROOT_NAME_STRING = bytes16("ROOT_NAME");
-
+/**
+ * A ResourceSelector is a 32-byte value that uniquely identifies a resource
+ * - 2 bytes for the resource type
+ * - 14 bytes for the namespace
```
holic, line 9: decided to steal from the namespace rather than name, but we could also steal 1 byte each (`bytes2`+`bytes15`+`bytes15`)

packages/world/src/resourceTypes.sol
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
+
+bytes2 constant RESOURCE_NAMESPACE = "ns";
+bytes2 constant RESOURCE_TABLE = "tb";
+bytes2 constant RESOURCE_TABLE_OFFCHAIN = "tf";
```
alvrs, line 6: why not `to` or `ot` 👀

holic: considered it but "offchain" and "onchain" are both "o"

also `tb` and `to` look too similar and wanted to visually differentiate

🤷

alvrs: fair, `f` is the only diff between `onchain` and `offchain` 🙈

packages/world/src/ResourceSelector.sol
holic: > feels right! agree with stealing from the namespace (I actually like two different types bytes14 and bytes16 to distinguish them better)

packages/world/src/resourceTypes.sol
holic: @dk1a in case you weren't following our chat, we probably want to define the table ID prefix separately and inside store, and the other resource prefixes here in world

alvrs:
closing in favor of #1544

PR #1500, opened 2023-09-15
feat(world): add CallBatchSystem to core module

### LINKED ISSUE #1160: Support batch-calling world methods
In an autonomous world, where an external system has to call back to the world entry point, I am wondering if it's more gas efficient to batch calls (i.e. multicall) when you know you're going to make several writes, etc.

I don't know if this is worth blocking the protocol finalization and audit for and seems like more of an additive improvement (if it makes sense), but just wanted to start the conversation.

yonadaaa:
closes #1160
closes #1470 

The `BatchCallSystem` enables calling many systems in a World in a single transaction. 

It leverages `callFrom` to execute each system call with the user as `msgSender()`. Users delegate to the `BatchCallSystem` system, then call `batchCall` with the resource selectors and calldata for the systems they wish to call.


packages/world/src/modules/batchcall/BatchCallSystem.sol
```diff
@@ -0,0 +1,14 @@
...
+pragma solidity >=0.8.0;
+import { System } from "../../System.sol";
+import { IBaseWorld } from "../../interfaces/IBaseWorld.sol";
+
+contract BatchCallSystem is System {
+  function batchCall(bytes32[] memory resourceSelectors, bytes[] memory funcSelectorAndArgss) public {
+    IBaseWorld world = IBaseWorld(_world());
+
+    for (uint256 i; i < resourceSelectors.length; i++) {
+      world.callFrom(_msgSender(), resourceSelectors[i], funcSelectorAndArgss[i]);
```
alvrs, line 11: should we make this a root system and then do this so the users don't need to create a delegation?

```solidity
(bool success, bytes memory returnData) = address(world).delegatecall(abi.encodeCall(
  world.call,
  (
    _msgSender(),
    resourceSelector[i],
    funcSelectorAndArgs[i]
  )
));
if(!success) revertWithBytes(returnData);
```

We can actually do both, have a `BatchCallSystem` and a `BatchCallRootSystem`, and install then in `installModule` and `installRootModule` respectively

holic: do we need this to be a module? if we intend to use this for deploys anyway, would it make sense to just have this under `core/implementations` as a plain root system?

alvrs: yeah it can also be part of the core module

packages/world/src/modules/batchcall/BatchCallSystem.sol
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
+import { System } from "../../System.sol";
+import { IBaseWorld } from "../../interfaces/IBaseWorld.sol";
+
+contract BatchCallSystem is System {
+  function batchCall(bytes32[] memory resourceSelectors, bytes[] memory funcSelectorAndArgss) public {
```
holic, line 7: would it be worse on calldata length if this was something like

```solidity
struct SystemCall {
  bytes32 systemId;
  bytes callData;
}

function batchCall(SystemCall[] memory systemCalls) public {
```
```

File: packages/world/src/modules/batchcall/BatchCallSystem.sol (hunk 1 above)
Comment (yonadaaa) [reply packages/world/src/modules/batchcall/BatchCallSystem.sol]: I like this approach without the delegation. I'm going to add the root version to the core module. 

I'll also remove the non-root version, because the root version will be installed on worlds by default.

File: packages/world/src/modules/core/implementations/CallBatchSystem.sol [hunk 3]
```diff
@@ -0,0 +1,27 @@
...
+import { IBaseWorld } from "../../../interfaces/IBaseWorld.sol";
+import { revertWithBytes } from "../../../revertWithBytes.sol";
+
+contract CallBatchSystem is System {
+  /**
+   * Call multiple systems at the given resourceSelectors, on behalf of _msgSender().
+   */
+  function callBatch(
+    bytes32[] memory resourceSelectors,
+    bytes[] memory callDatas
```
holic, line 13: same question as before: would it be better or worse gas-wise to have an array of `SystemCall` structs instead of two arrays? I assume a little worse on calldata but we wouldn't have to do a length check on the arrays to make sure they match (though I guess the whole thing would revert with an out of bounds error in that case)

packages/world/test/CallBatch.t.sol
```diff
@@ -0,0 +1,54 @@
...
+    callDatas[1] = abi.encodeWithSelector(WorldTestSystem.getStoreAddress.selector);
+
+    vm.prank(caller);
+    startGasReport("call systems with callBatch");
+    bytes[] memory datas = world.callBatch(resourceSelectors, callDatas);
+    endGasReport();
+
+    assertEq(abi.decode(datas[0], (address)), caller, "wrong address returned");
+    assertEq(abi.decode(datas[1], (address)), address(world), "wrong store returned");
+  }
```
holic, line 53: should we add some tests to make sure this goes through access control checks properly and fails on system calls we don't have access to?

packages/world/src/modules/core/implementations/CallBatchSystem.sol
```diff
@@ -0,0 +1,27 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
+import { System } from "../../../System.sol";
+import { IBaseWorld } from "../../../interfaces/IBaseWorld.sol";
+import { revertWithBytes } from "../../../revertWithBytes.sol";
+
+contract CallBatchSystem is System {
+  /**
+   * Call multiple systems at the given resourceSelectors, on behalf of _msgSender().
```
alvrs, line 9:
```suggestion
   * Call multiple systems at the given resourceSelectors and return the array of return data.
```

packages/world/src/modules/core/implementations/CallBatchSystem.sol
alvrs: my feeling is it would be a little _better_ calldata wise because the structs are encoded like tuples, so there shouldn't be overhead on the encoding of the individual elements, but we might save one word because we only have one array (and therefore only one length encoding)

yonadaaa: Going to implement the struct for the calldata improvements and to improve usability (only dealing with a single array)

packages/world/src/modules/core/implementations/SystemCall.sol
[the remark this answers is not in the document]
alvrs: i'd propose calling this one `SystemCallData` (aligned with how the tablegen structs are called, and also alllgins with "callData")

alvrs (approved):

PR #1503, opened 2023-09-15
docs: cli changeset after deploy changes

holic:
forgot a changeset in https://github.com/latticexyz/mud/pull/1384


.changeset/nasty-trains-drop.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/cli": patch
```
holic, line 2: patch because it doesn't change any external behavior

alvrs (approved):

PR #1508, opened 2023-09-15
feat(store): compute FieldLayout at compile time

### LINKED ISSUE #1327: Encode table schemas and fieldLayout at compile time
Depends on https://github.com/latticexyz/mud/issues/1326
If Schemas are just lengths, it's pretty trivial to make a typescript Schema builder and pre-render a hex string, since table schemas are immutable.
`getSchema` consumes quite a lot of gas (e.g. `2814` for 6 entries) and this would remove all of it

Boffee:
Followup to #1336
Fixes #1327 

Computing FieldLayout bytes from schema field type array is expensive (~500 + 500 * # of schema fields). This change precomputes the FieldLayout offchain inside codegen and hardcodes the byte into the auto-generated table libraries.


packages/cli/contracts/src/codegen/tables/Dynamics1.sol
```diff
@@ -32,9 +36,7 @@ struct Dynamics1Data {
 library Dynamics1 {
   /** Get the table values' field layout */
   function getFieldLayout() internal pure returns (FieldLayout) {
```
holic, line 38: Do we need this method anymore?

packages/store/ts/codegen/renderFieldLayout.ts
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  if (dynamicFields > MAX_DYNAMIC_FIELDS) throw new Error(`FieldLayout: invalid length ${dynamicFields}`);
+
+  fieldLayout |= BigInt(totalLength) << BigInt(LayoutOffsets.TOTAL_LENGTH);
+  fieldLayout |= BigInt(staticFields) << BigInt(LayoutOffsets.NUM_STATIC_FIELDS);
+  fieldLayout |= BigInt(dynamicFields) << BigInt(LayoutOffsets.NUM_DYNAMIC_FIELDS);
+
+  return `0x${fieldLayout.toString(16).padStart(64, "0")}`;
+}
```
alvrs, line 30: slightly rearanged and renamed some variables to make it as close as possible to `FieldLayout.sol` (which makes auditing easier)

```suggestion
export function encodeFieldLayout(fields: RenderType[]) {
  const staticFields = fields.filter(({ isDynamic }) => !isDynamic);
  const numDynamicFields = fields.length - staticFields.length;
  
  let fieldLayout = 0n;
  let totalLength = 0;
  const totalFields = fields.length;
  
  if (totalFields > MAX_TOTAL_FIELDS) throw new Error(`FieldLayout: invalid length ${totalFields}`);
  if (numDynamicFields > MAX_DYNAMIC_FIELDS) throw new Error(`FieldLayout: invalid length ${numDynamicFields}`);

  for (let i = 0; i < staticFields.length; i++) {
    const { isDynamic, staticByteLength } = fields[i];
    if (isDynamic) throw new Error(`FieldLayout: static type after dynamic type`);

    totalLength += staticByteLength;
    fieldLayout |= BigInt(staticByteLength) << BigInt((WORD_LAST_INDEX - 4 - i) * BYTE_TO_BITS);
  }

  fieldLayout |= BigInt(totalLength) << BigInt(LayoutOffsets.TOTAL_LENGTH);
  fieldLayout |= BigInt(staticFields.length) << BigInt(LayoutOffsets.NUM_STATIC_FIELDS);
  fieldLayout |= BigInt(numDynamicFields) << BigInt(LayoutOffsets.NUM_DYNAMIC_FIELDS);

  return `0x${fieldLayout.toString(16).padStart(64, "0")}`;
}
```

packages/store/ts/codegen/renderFieldLayout.ts
```diff
@@ -0,0 +1,30 @@
+import { RenderType } from "@latticexyz/common/codegen";
+import { BYTE_TO_BITS, LayoutOffsets, MAX_DYNAMIC_FIELDS, MAX_TOTAL_FIELDS, WORD_LAST_INDEX } from "../constants";
+
+export function renderFieldLayout(fields: RenderType[]) {
+  return `FieldLayout constant _fieldLayout = FieldLayout.wrap(${encodeFieldLayout(fields)});`;
+}
+
```
alvrs, line 7:
```suggestion
// Make sure this logic stays aligned with @latticexyz/store/src/FieldLayout.sol
```

packages/store/ts/constants.ts
```diff
@@ -0,0 +1,22 @@
+/* Shared constants */
+
```
alvrs, line 2: // Make sure these stay aligned with @latticexyz/store/src/constants.sol

alvrs, line 2:
```suggestion
// Make sure these stay aligned with @latticexyz/store/src/constants.sol

```

alvrs (commented):
amazing, thank you!!

packages/store/ts/codegen/renderFieldLayout.ts
holic: can we add tests for this too? seems important to get this right (and would be nice to have the same tests both here and in solidity)

we have a version of this in protocol-parser but it's also untested and doesn't match the solidity code as well

packages/cli/contracts/src/codegen/tables/Dynamics1.sol
alvrs: seems fine for completeness (so the methods match what you can get from the `Tables` table), if it's not used anywhere it's also not included in the bytecode

alvrs (approved):

alvrs (approved):

packages/cli/contracts/src/codegen/tables/Dynamics1.sol
holic: I don't know why anyone would choose the less efficient one (function call) over the more efficient one though (importing the `_fieldLayout` directly)

PR #1509, opened 2023-09-15
feat(store): optimize storage location hash

Boffee:
The existing storage location hash encodes all parameters using `abi.encode` and hashes the result with `keccak256`. The 2 optimizations are:

1. Move non-user dependent parameters (everything except `keyTuple` and `tableId`) outside of the hash and xor against the hash.
2. Use `abi.encodePacked` because it produces a shorter encoding which reduces the gas cost for `keccak256`.


packages/store/src/StoreCore.sol
```diff
@@ -811,14 +811,16 @@ library StoreCoreInternal {
...
+    // assumes schemaIndex < 255
+    return uint256(SLOT ^ tableId ^ bytes1(schemaIndex + 1) ^ keccak256(abi.encodePacked(keyTuple)));
   }
 
   /**
    * Compute the storage location for the length of the dynamic data
    */
   function _getDynamicDataLengthLocation(bytes32 tableId, bytes32[] memory keyTuple) internal pure returns (uint256) {
-    return uint256(keccak256(abi.encode(SLOT, tableId, keyTuple, "length")));
+    return uint256(SLOT ^ tableId ^ bytes32("length") ^ keccak256(abi.encodePacked(keyTuple)));
```
alvrs, line 823: what if i register a table with id `tableIdA` which equals `tableIdB ^ bytes32(length)`? would that create a conflict between `_getStaticDataLocation` for `tableA` and `_getDynamicDataLengthLocation` for `tableB`?

I wonder if, since `tableId` is also a user influenceable parameter we need to also hash `tableId`?

Boffee: Ah ya forgot that tableId is a string instead of a hash... that's a problem.

Boffee: Should create a hash for "length" like SLOT.

alvrs (dismissed):
🔥🔥🔥

alvrs (approved):

PR #1511, opened 2023-09-15
feat(store,world): emit Store/World versions

### LINKED ISSUE #1130: Add a protocol version to Store and World
- Eg a view function returning the current version (since version doesn't change within an existing Store/World)
- This allows for upgrades to the protocol while keeping support for previous versions

holic:
closes #1130


packages/store/src/IStore.sol
```diff
@@ -7,6 +7,10 @@ import { Schema } from "./Schema.sol";
 import { IStoreHook } from "./IStoreHook.sol";
 
 interface IStoreRead {
+  event HelloStore(bytes32 indexed storeVersion);
+
+  function storeVersion() external view returns (bytes32);
```
holic, line 12: I really wanted this to just be `version()` but this overlaps with `World.version()` since World is a Store.

alvrs (dismissed):

alvrs (approved):

PR #1512, opened 2023-09-15
feat(store): add Storage.loadField for optimized loading of 32 bytes or less from storage

Boffee:
Loading field currently uses `Storage.load`, which can load arbitrary number of slots. However, fields are always within bytes32. 

`Storage.loadField` removes the logic for handling arbitrary lengths and allocating memory. 

This results in a 160 gas saving for fields that span a single slot and 80 gas saving for fields that span multiple slots. Decoding will also be cheaper because it would simple type casting.

`Storage.loadField` should be used for `getStaticField` once `getField` is split into `getStaticField` and `getDynamicField`.


alvrs (dismissed):

packages/store/src/Storage.sol
```diff
@@ -209,4 +209,38 @@ library Storage {
       }
     }
   }
+
+  /**
+   * Load up to 32 bytes from storage at the given storagePointer and offset.
+   * The return value is left-aligned, the bytes beyond the length are not zeroed out,
+   * and the caller is expected to truncate as needed.
+   * The field can span up to two storage slots.
```
holic, line 217: why? can we expand this comment to explain the case in which this happens? (I think it's because fields are packed tightly so two `bytes20` fields together will take up a 32-byte slot + an extra 8 bytes of the next slot?)

alvrs: Yeah so since we tightly pack a field can span over two slots (eg the example you mentioned, or eg a `bytes1` field followed by a `bytes32` field), but since the max field length is 32 bytes it can maximally span two fields.

PR #1513, opened 2023-09-15
feat(common,store,world): add StoreCore table functions and use them for internal mud logic

Boffee:
All table methods currently go through `StoreSwitch` which adds ~300 gas on top of StoreCore (due `_layout().storeAddress`). This is necessary for non-root systems to resolve the store address.

However, MUD internal systems as well as almost all systems during the development phase of a new project are root systems. `StoreSwitch` is rarely needed in these cases.

This pr adds codegen to generate an internal version of every table function that touches `Store` to bypass `StoreSwitch` and call `StoreCore` directly. the internal table functions are differentiated from existing tables functions with an `_` prefix.

The MUD internal systems have been updated to use internal `_set[...]`, `_get[...]`, and `deleteRecord`.


holic:
OOC why the extra methods everywhere instead of just changing the internals of each method based on table config?

Boffee:
> OOC why the extra methods everywhere instead of just changing the internals of each codegen table method based on table config?

Because root tables can be read by non-root systems.

alvrs (dismissed):
agree with the approach of generating an additional set of prefixed functions (using `_` which matches Solidity conventions for "internal" functions) so that other non-root systems can import the same table libraries!

alvrs (dismissed):

alvrs (approved):

alvrs:
only changeset changed since last passing tests (only vercel failed due to missing permissions)

PR #1517, opened 2023-09-16
feat(faucet): add faucet service

holic:
This is the last thing left in services package before we can 🔥


alvrs (dismissed):

alvrs:
amazing!

alvrs:
(let's add a changeset with minimal usage instructions before merging)

examples/minimal/packages/client-react/src/mud/setupNetwork.ts
```diff
@@ -44,28 +45,20 @@ export async function setupNetwork() {
...
-      if (lowBalance) {
-        console.info("[Dev Faucet]: Balance is low, dripping funds to player");
-        // Double drip
-        await faucet.dripDev({ address });
-        await faucet.dripDev({ address });
-      }
+    const drip = async () => {
+      console.log("dripping");
+      const tx = await faucet.drip.mutate({ address: burnerAccount.address });
+      console.log("got drip", tx);
```
alvrs, line 55: should we still only do this if the balance is below a certain threshold?

alvrs (approved):

examples/minimal/packages/client-react/src/mud/setupNetwork.ts
holic: this is just the example app, but yes the templates should once we get this wired up to k8s

holic: or better yet, consider putting the threshold in the faucet itself? so folks don't try to load up

PR #1520, opened 2023-09-16
feat(store): indexed `tableId` in store events

### LINKED ISSUE #1127: Add `indexed` to `tableId` in all Store update events
This makes it easier to only sync selected table from RPC

holic:
closes #1127


packages/store/gas-report.json
```diff
@@ -549,217 +549,217 @@
...
     "name": "pop from field (cold, 1 slot, 1 uint32 item)",
-    "gasUsed": 21393
+    "gasUsed": 22232
   },
   {
     "file": "test/StoreCoreDynamic.t.sol",
     "test": "testPopFromSecondField",
     "name": "pop from field (warm, 1 slot, 1 uint32 item)",
-    "gasUsed": 15405
+    "gasUsed": 16244
```
holic, line 606: any idea why this would add so much gas here when other operations are like +100 gas?

alvrs: it looks like it adds around 900 gas for operations where we emit an event, while the other methods are getters (why does it even increase gas for getters at all 🤔)

holic: I think the gas for getters is just the usual sway we see between gas reports for some unexplained reason (optimizer differences when bytecode changes maybe?)

alvrs (approved):

alvrs (approved):

Issue #1510, opened 2023-09-15
Use splice events for hooks to avoid the linear storage reads


alvrs:
Followup to #1354

alvrs:
implemented in #1531

PR #1527, opened 2023-09-17
feat(store,world): move hooks to bit flags

### LINKED ISSUE #1459: use bitwise OR for hook bitmaps
We can replace instances of

```solidity
uint8 bitmap = StoreHookLib.encodeBitmap({
  onBeforeSetRecord: true,
  onAfterSetRecord: false,
  onBeforeSetField: true,
  onAfterSetField: true,
  onBeforeDeleteRecord: true,
  onAfterDeleteRecord: false
});
```

with something like:
```solidity
import { BEFORE_SET_RECORD, BEFORE_SET_FIELD, AFTER_SET_FIELD, BEFORE_DELETE_RECORD } from './StoreHookType.sol';

uint8 bitmap = BEFORE_SET_RECORD | BEFORE_SET_FIELD | AFTER_SET_FIELD | BEFORE_DELETE_RECORD;
```
where each of the above is a `uint8` constant that is the bit position.

holic:
closes #1459

TODO:
- [ ] expand bitmap to uint16 or uint32 for more bits/hooks (will need for #1510)

### DIFF
--- a/packages/store/src/Hook.sol
+++ b/packages/store/src/Hook.sol
@@ -52,9 +52,8 @@ library HookInstance {
   /**
    * Check if the given hook type is enabled in the hook
    */
-  function isEnabled(Hook self, uint8 hookType) internal pure returns (bool) {
-    // Pick the bitmap encoded in the rightmost byte from the hook and check if the bit at the given hook type is set
-    return (getBitmap(self) & (1 << uint8(hookType))) != 0;
+  function isEnabled(Hook self, uint8 hookTypes) internal pure returns (bool) {
+    return (getBitmap(self) & hookTypes) == hookTypes;
   }
 
   /**
--- a/packages/store/src/storeHookTypes.sol
+++ b/packages/store/src/storeHookTypes.sol
@@ -1,0 +1,10 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
+
+uint8 constant BEFORE_SET_RECORD = 1 << 0;
+uint8 constant AFTER_SET_RECORD = 1 << 1;
+// TODO: do we need to differentiate between static and dynamic set field?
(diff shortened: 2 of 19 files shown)


packages/store/src/Hook.sol
```diff
@@ -52,9 +52,8 @@ library HookInstance {
   /**
    * Check if the given hook type is enabled in the hook
    */
-  function isEnabled(Hook self, uint8 hookType) internal pure returns (bool) {
-    // Pick the bitmap encoded in the rightmost byte from the hook and check if the bit at the given hook type is set
-    return (getBitmap(self) & (1 << uint8(hookType))) != 0;
+  function isEnabled(Hook self, uint8 hookTypes) internal pure returns (bool) {
+    return (getBitmap(self) & hookTypes) == hookTypes;
```
alvrs, line 56: ✨✨✨ so much nicer

packages/store/src/storeHookTypes.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
+
+uint8 constant BEFORE_SET_RECORD = 1 << 0;
+uint8 constant AFTER_SET_RECORD = 1 << 1;
+// TODO: do we need to differentiate between static and dynamic set field?
```
alvrs, line 6: cross posting here from discord for context: the "set field" hook is soon replaced with a "splice" hook (see #1510)

alvrs (dismissed):
🥇

alvrs (approved):

alvrs:
> expand bitmap to uint16 or uint32 for more bits/hooks

don't think this will be necessary since we'll only have 6 hooks (before/after set, splice, delete)

Issue #1525, opened 2023-09-16
move to `abi.encodeCall` for function calls


holic:
We can move usages of `abi.encodeWithSelector` to `abi.encodeCall` to get some extra type checking by the compiler

See https://soliditylang.org/blog/2021/12/20/solidity-0.8.11-release-announcement/

PR #1528, opened 2023-09-17
refactor(world): use abi.encodeCall instead of abi.encodeWithSelector

holic:
closes https://github.com/latticexyz/mud/issues/1525

### DIFF
--- a/packages/world/src/Delegation.sol
+++ b/packages/world/src/Delegation.sol
@@ -44,7 +44,7 @@ library DelegationInstance {
     (bool success, bytes memory data) = SystemCall.call({
       caller: delegatee,
       resourceSelector: Delegation.unwrap(self),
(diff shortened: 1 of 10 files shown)


packages/world/src/Delegation.sol
```diff
@@ -44,7 +44,7 @@ library DelegationInstance {
     (bool success, bytes memory data) = SystemCall.call({
       caller: delegatee,
       resourceSelector: Delegation.unwrap(self),
-      callData: abi.encodeWithSelector(IDelegationControl.verify.selector, delegator, systemId, callData),
+      callData: abi.encodeCall(IDelegationControl.verify, (delegator, systemId, callData)),
```
holic, line 47: changing this found a bug(?) in the `IDelegationControl` interface, where we had `bytes memory` here but the interface wanted `bytes calldata`

alvrs (approved):

PR #1531, opened 2023-09-17
feat(store,world): add splice hooks, expose spliceStaticData, spliceDynamicData

alvrs:
- remove `onBeforeSetField` and `onAfterSetField`, replace with `onBeforeSpliceStaticData`, `onAfterSpliceStaticData`, `onBeforeDynamicData`, `onAfterDynamicData` so we don't have to read the entire dynamic field when calling partial field methods (push, pop, update)
- while replacing the hooks I realized that exposing `spliceStaticData` and `spliceDynamicData` as methods on `IStore` would be really useful for hooks (otherwise even simple hooks like "mirror the data to another table" are very complicated to implement to because the hook data doesn't match any store method)
  - exposing those methods is not a concern for diverging onchain and offchain indexer data, since it's already possible to modify records in arbitrary ways by providing different `FieldLayout` values
  - this change triggered a larger refactor of `StoreCore` to use the new `spliceStaticData` and `spliceDynamicData` instead of duplicating the splice logic in every single dynamic data function

will add more detailed context to the changeset once this PR is ready


packages/store/src/IStoreHook.sol
```diff
@@ -8,16 +8,18 @@ import { PackedCounter } from "./PackedCounter.sol";
...
+  IStoreHook.onAfterSpliceDynamicData.selector ^
   IStoreHook.onBeforeDeleteRecord.selector ^
   IStoreHook.onAfterDeleteRecord.selector ^
   ERC165_INTERFACE_ID;
 
 interface IStoreHook is IERC165 {
   function onBeforeSetRecord(
     bytes32 tableId,
-    bytes32[] memory keyTuple,
+    bytes32[] calldata keyTuple,
```
holic, line 22: will we have similar issues as https://github.com/latticexyz/mud/pull/1528#discussion_r1328132000 by changing this?

packages/store/src/storeHookTypes.sol
```diff
@@ -3,8 +3,9 @@ pragma solidity >=0.8.0;
...
-uint8 constant BEFORE_SET_FIELD = 1 << 2;
-uint8 constant AFTER_SET_FIELD = 1 << 3;
-uint8 constant BEFORE_DELETE_RECORD = 1 << 4;
-uint8 constant AFTER_DELETE_RECORD = 1 << 5;
+uint8 constant BEFORE_SPLICE_STATIC_DATA = 1 << 2;
+uint8 constant AFTER_SPLICE_STATIC_DATA = 1 << 3;
+uint8 constant BEFORE_SPLICE_DYNAMIC_DATA = 1 << 4;
+uint8 constant AFTER_SPLICE_DYNAMIC_DATA = 1 << 5;
+uint8 constant BEFORE_DELETE_RECORD = 1 << 6;
+uint8 constant AFTER_DELETE_RECORD = 1 << 7;
```
holic, line 11: this is what I was referring to btw when I thought we maybe needed more than 8 slots for hooks

do we wanna consider a bump to uint16 or uint32 for future proofing?

packages/store/test/EchoSubscriber.sol
```diff
@@ -16,7 +16,9 @@ contract EchoSubscriber is StoreHook {
     bytes memory dynamicData,
     FieldLayout fieldLayout
   ) public {
-    emit HookCalled(abi.encode(tableId, keyTuple, staticData, encodedLengths, dynamicData, fieldLayout));
+    emit HookCalled(
+      abi.encodeCall(this.onBeforeSetRecord, (tableId, keyTuple, staticData, encodedLengths, dynamicData, fieldLayout))
```
holic, line 20: neat

packages/store/test/MirrorSubscriber.sol
```diff
@@ -48,21 +48,53 @@ contract MirrorSubscriber is StoreHook {
...
     bytes32 tableId,
     bytes32[] memory keyTuple,
-    uint8 fieldIndex,
-    bytes memory data,
-    FieldLayout fieldLayout
+    uint48 start,
+    uint40 deleteCount,
+    bytes memory data
   ) public {
-    if (tableId != tableId) revert("invalid tableId");
```
holic, line 58: 😅

packages/store/test/setDynamicDataLengthAtIndex.sol
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
+
+import { PackedCounter } from "../src/PackedCounter.sol";
+import { StoreCoreInternal } from "../src/StoreCore.sol";
+import { Storage } from "../src/Storage.sol";
+
+/**
+ * Tes helper function to set the length of the dynamic data (in bytes) for the given value field layout and index
```
holic, line 9:
```suggestion
 * Test helper function to set the length of the dynamic data (in bytes) for the given value field layout and index
```

packages/store/test/setDynamicDataLengthAtIndex.sol
```diff
@@ -0,0 +1,26 @@
...
+ */
+function setDynamicDataLengthAtIndex(
+  bytes32 tableId,
+  bytes32[] memory keyTuple,
+  uint8 dynamicSchemaIndex, // fieldIndex - numStaticFields
```
holic, line 14: should this 
```suggestion
  uint8 dynamicFieldIndex, // fieldIndex - numStaticFields
```

packages/store/test/setDynamicDataLengthAtIndex.sol
```diff
@@ -0,0 +1,26 @@
...
+  uint256 newLengthAtIndex
+) {
+  // Load dynamic data length from storage
+  uint256 dynamicDataLengthSlot = StoreCoreInternal._getDynamicDataLengthLocation(tableId, keyTuple);
+  PackedCounter encodedLengths = PackedCounter.wrap(Storage.load({ storagePointer: dynamicDataLengthSlot }));
+
+  // Update the encoded lengths
+  encodedLengths = encodedLengths.setAtIndex(dynamicSchemaIndex, newLengthAtIndex);
```
holic, line 22:
```suggestion
  encodedLengths = encodedLengths.setAtIndex(dynamicFieldIndex, newLengthAtIndex);
```

packages/world/src/modules/keysintable/KeysInTableHook.sol
```diff
@@ -63,11 +63,41 @@ contract KeysInTableHook is StoreHook {
     // NOOP
   }
 
-  function onBeforeSetField(bytes32 tableId, bytes32[] memory keyTuple, uint8, bytes memory, FieldLayout) public {
+  function onBeforeSpliceStaticData(bytes32, bytes32[] calldata, uint48, uint40, bytes calldata) public pure {
     // NOOP
   }
```
holic, line 68: if we add these method stubs to `StoreHook` with a `NotImplemented` error, we could remove these here

packages/store/test/StoreHook.t.sol
```diff
@@ -54,30 +58,36 @@ contract StoreHookTest is Test, GasReporter {
...
           BEFORE_SET_RECORD |
             AFTER_SET_RECORD |
-            BEFORE_SET_FIELD |
-            AFTER_SET_FIELD |
+            BEFORE_SPLICE_STATIC_DATA |
+            AFTER_SPLICE_STATIC_DATA |
+            BEFORE_SPLICE_DYNAMIC_DATA |
+            AFTER_SPLICE_DYNAMIC_DATA |
             BEFORE_DELETE_RECORD |
             AFTER_DELETE_RECORD
```
holic, line 66: a nice thing we can do now that we are operating with bit flags is add shortcuts, e.g. `BEFORE_ALL` and `AFTER_ALL` like

```solidity
uint8 constant BEFORE_ALL = BEFORE_SET_RECORD | BEFORE_SPLICE_STATIC_DATA | BEFORE_SPLICE_DYNAMIC_DATA | BEFORE_DELETE_RECORD;
```

packages/store/src/StoreCore.sol
```diff
@@ -984,53 +971,13 @@ library StoreCoreInternal {
   }
 
   /**
-   * Get the length of the dynamic data for the given value field layout and index
+   * Load the encoded dynamic data length from storage for the given tableId and key tuple
```
holic, line 974: should prob be
> given table ID and key tuple

or

> given tableId and keyTuple

packages/store/src/StoreCore.sol
```diff
@@ -797,44 +823,24 @@ library StoreCoreInternal {
...
-    // Update the encoded length
-    uint256 oldFieldLength = encodedLengths.atIndex(dynamicSchemaIndex);
-    encodedLengths = encodedLengths.setAtIndex(dynamicSchemaIndex, oldFieldLength - byteLengthToPop);
+    uint8 dynamicFieldIndex = fieldIndex - uint8(fieldLayout.numStaticFields());
 
-    // Set the new length
-    Storage.store({ storagePointer: dynamicDataLengthSlot, data: encodedLengths.unwrap() });
+    // Load the previous length of the field to set from storage to compute where to start to push
+    PackedCounter oldEncodedLengths = _loadEncodedDynamicDataLength(tableId, keyTuple);
+    uint40 oldFieldLength = uint40(oldEncodedLengths.atIndex(dynamicFieldIndex));
```
holic, line 830: I tend to prefer "previous" instead of "old" for these sorts of operations, feels a bit more specific/accurate.

packages/store/test/StoreMock.sol
```diff
@@ -29,6 +29,29 @@ contract StoreMock is IStore, StoreRead {
...
+    bytes32[] calldata keyTuple,
+    uint8 dynamicFieldIndex,
+    uint40 startWithinField,
+    uint40 deleteCount,
+    bytes calldata data
+  ) public virtual {
+    StoreCore.spliceDynamicData(tableId, keyTuple, dynamicFieldIndex, startWithinField, deleteCount, data);
+  }
+
   // Set partial data at schema index
```
holic, line 55:
```suggestion
  // Set partial data at field index
```

packages/store/src/IStoreHook.sol
alvrs: Changed all to `memory`. Had `calldata` initially bc it would be cheaper if called with calldata, but we're in fact calling all these functions with memory from the `World`, so needs to be memory here.

packages/store/src/storeHookTypes.sol
alvrs: I'd prefer to only bump to `uint16` once we actually need it (don't think we'll need it for this iteration of the protocol) - although one could argue we could avoid a breaking change if we already use `uint16` now and later add additional functionality for other types of hook 🤔

holic: yeah my main motivation is reducing breaking changes later

I guess either way it'd require a redeploy of the world, so maybe it's fine to leave this and expand when we need it

packages/world/test/SystemHook.t.sol
```diff
@@ -45,4 +45,8 @@ contract SystemHookTest is Test, GasReporter {
     assertEq(systemHook.getAddress(), hookAddress);
     assertEq(systemHook.getBitmap(), enabledHooksBitmap);
   }
+
+  function testShorthand() public {
+    assertEq(ALL, BEFORE_CALL_SYSTEM | AFTER_CALL_SYSTEM);
+  }
```
holic, line 51: is this a useful test? 😆

holic (dismissed):

packages/world/test/SystemHook.t.sol
alvrs: haha just want to make sure `ALL` is updated in case we add more hooks

packages/store/src/StoreCore.sol
```diff
@@ -658,69 +678,89 @@ library StoreCoreInternal {
...
+          startWithinField: startWithinField,
+          deleteCount: deleteCount,
+          data: data,
+          encodedLengths: updatedEncodedLengths
+        });
       }
     }
 
-    // Emit event to notify indexers
+    // Emit event to offchain notify indexers
```
alvrs, line 756:
```suggestion
    // Emit event to notify offchain indexers
```

alvrs:
skipping CI, only updated a comment since last CI pass

Issue #1518, opened 2023-09-16
separate `data` into `staticData`, `encodedLengths`, `dynamicData` in Store getter functions


alvrs:
followup to #1354

PR #1532, opened 2023-09-17
refactor: separate data into staticData, encodedLengths, dynamicData in getRecord

dk1a:
fixes https://github.com/latticexyz/mud/issues/1518

### DIFF
--- a/.changeset/hip-tables-check.md
+++ b/.changeset/hip-tables-check.md
@@ -8,6 +8,6 @@ Store's `getRecord` has been updated to return `staticData`, `encodedLengths`, a
 If you use codegenerated libraries, you will only need to update `encode` calls.
 
 ```diff
-- bytes memory data = Mixed.encode(mixed.u32, mixed.u128, mixed.a32, mixed.s);
-+ (bytes memory staticData, PackedCounter encodedLengths, bytes memory dynamicData) = Mixed.encode(mixed.u32, mixed.u128, mixed.a32, mixed.s);
+- bytes memory data = Position.encode(x, y);
++ (bytes memory staticData, PackedCounter encodedLengths, bytes memory dynamicData) = Position.encode(x, y);
 ```
--- a/.changeset/hip-tables-check.md
+++ b/.changeset/hip-tables-check.md
@@ -1,6 +1,6 @@
 ---
 "@latticexyz/store": major
-"@latticexyz/world": major
+"@latticexyz/world": patch
 ---
 
 Store's `getRecord` has been updated to return `staticData`, `encodedLengths`, and `dynamicData` instead of a single `data` blob, to match the new behaviour of Store setter methods.
--- a/.changeset/hip-tables-check.md
+++ b/.changeset/hip-tables-check.md
@@ -1,6 +1,6 @@
 ---
 "@latticexyz/store": major
-"@latticexyz/world": patch
+"@latticexyz/world": major
 ---
 
 Store's `getRecord` has been updated to return `staticData`, `encodedLengths`, and `dynamicData` instead of a single `data` blob, to match the new behaviour of Store setter methods.
--- a/.changeset/hip-tables-check.md
+++ b/.changeset/hip-tables-check.md
@@ -1,0 +1,13 @@
+---
+"@latticexyz/store": major
+"@latticexyz/world": major
+---
+
+Store's `getRecord` has been updated to return `staticData`, `encodedLengths`, and `dynamicData` instead of a single `data` blob, to match the new behaviour of Store setter methods.
+
+If you use codegenerated libraries, you will only need to update `encode` calls.
+
+```diff
+- bytes memory data = Mixed.encode(mixed.u32, mixed.u128, mixed.a32, mixed.s);
++ (bytes memory staticData, PackedCounter encodedLengths, bytes memory dynamicData) = Mixed.encode(mixed.u32, mixed.u128, mixed.a32, mixed.s);
+```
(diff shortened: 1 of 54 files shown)


e2e/packages/contracts/src/codegen/tables/Vector.sol
```diff
@@ -277,13 +304,13 @@ library Vector {
   }
 
   /** Tightly pack full data using this table's field layout */
-  function encode(int32 x, int32 y) internal pure returns (bytes memory) {
+  function encode(int32 x, int32 y) internal pure returns (bytes memory, PackedCounter, bytes memory) {
```
holic, line 307: does it help at all gas wise to name these return values and use them in the function body?

```solidity
function encode(int32 x, int32 y) internal pure returns (bytes memory _staticData, PackedCounter, bytes memory) {
  _staticData = encodeStatic(x, y);

```

packages/store/src/StoreCore.sol
```diff
@@ -521,54 +521,31 @@ library StoreCore {
     bytes32 tableId,
     bytes32[] memory keyTuple,
     FieldLayout fieldLayout
-  ) internal view returns (bytes memory) {
+  ) internal view returns (bytes memory staticData, PackedCounter encodedLengths, bytes memory dynamicData) {
```
holic, line 524: since we're now passing around this tuple a lot, do you think we'd get any value from encoding this as a struct? or would that make it less convenient?

```solidity
struct ValueData {
  bytes staticData;
  PackedCounter encodedLengths;
  bytes dynamicData;
}
```

packages/store/ts/codegen/record.ts
```diff
@@ -115,69 +120,95 @@ function renderDecodeFunction({ structName, fields, staticFields, dynamicFields
...
-              } else {
-                return `
-                  _start = _end;
-                  unchecked {
-                    _end += _encodedLengths.atIndex(${index});
-                  }
-                  ${fieldNamePrefix}${field.name} = ${renderDecodeDynamicFieldPartial(field)};
-                `;
+    result += `
+    /**
```
holic, line 146: nit: can we indent this one level (and all the backticks below) for readability?

.changeset/hip-tables-check.md
```diff
@@ -0,0 +1,13 @@
+---
+"@latticexyz/store": major
+"@latticexyz/world": major
```
holic, line 3:
```suggestion
"@latticexyz/world": patch
```
As far as I can tell, this didn't break any world APIs, this is just a bump of the store dep?

packages/store/src/StoreCore.sol
holic: I see a lot of folks asking about stack too deep errors, is this anything to consider?

.changeset/hip-tables-check.md
dk1a: yeah, keyswithvalue got an internal breaking change, but that's not api

packages/store/src/StoreCore.sol
dk1a: yes, it would reduce vars from 3 to 1

.changeset/hip-tables-check.md
```diff
@@ -0,0 +1,13 @@
+---
+"@latticexyz/store": major
+"@latticexyz/world": patch
```
holic, line 3: this got dismissed with applying the change but I am wondering about moving this to its own changeset

otherwise, the world changelog will have info about store internals changes (which aren't super relevant if you're using world)

alvrs: Since `StoreRead`'s API was changed, also the `World` API changed, so i'd say this is a `major` change for world?

holic: Ohh right, very true

packages/store/src/StoreCore.sol
alvrs: might be worth it to support more variables in generated libraries? Since the functions are not used a lot outside codegen I think the slightly decreased dev ex of having to import a struct for the return data is fine

> edit: nevermind, talked myself out of this idea

curious why?

alvrs: the main advantage of using the struct is in the `set` function, where I believe we'd currently run into "stack too deep" issues for tables where num keys + num fields is greater than 7 (because the set function has 5 additional variables besides the key+fields). That might be worked around if we make the version of setter that uses the table's data struct a "first class citizen" instead of internally calling the version with arguments.

In any case, i think we can do this as a follow up since it involves refactors to code unrelated to this PR. Opened an issue to track here: #1535

alvrs (dismissed):
Looks good! Let's do the struct refactor in a follow up (tracked in #1535 and #1536). Only nits left are the indentation mentioned by @holic and changing the changeset back to major.

packages/store/ts/codegen/record.ts
dk1a: sure, I went with this style for all renderers so just sticking to it here. If we wanna change it I'd rather make a separate PR and change it everywhere

holic: we already did this: https://github.com/latticexyz/mud/pull/1474

alvrs (approved):

e2e/packages/contracts/src/codegen/tables/Vector.sol
dk1a: tested on `Callbacks` and `Vector2` but it doesn't seem to make a difference. Intuitively I'd also assume that this shouldn't affect bytecode

PR #1537, opened 2023-09-18
refactor(world): move codegen files

holic:
when I was working on #1318, some files got left around because world doesn't use a codegen dir, so it was harder to identify+clean up files

I still don't love the approach, maybe mostly because of how we do module codegen? I think that would be fixed if each module had its own mud config?

in the meantime, instead of all the codegen being lumped together, could customize each table to have its own codegen dir within its respective module?


alvrs:
closing in favor of #1592

PR #1541, opened 2023-09-18
build: pin react/react-dom types versions

holic:
something about a recent package push to `@types/react` caused CI to fall over and seemed to only resolve it by pinning these versions so they all align

### DIFF
--- a/examples/minimal/packages/client-phaser/package.json
+++ b/examples/minimal/packages/client-phaser/package.json
@@ -40,8 +40,8 @@
     "zustand": "^4.3.8"
   },
   "devDependencies": {
-    "@types/react": "^18.2.6",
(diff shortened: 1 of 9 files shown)


alvrs (approved):

holic:
> now that I think about it, I think this makes sense
> we have the versions ~pinned in our repo packages with pnpm-lock.yaml but we don't have a similar pnpm-lock.yaml for templates
> so the templates are allowed to bump their patch versions to whatever matches the semver when installing in CI (because there's no lockfile to follow) 
> looking at the diff of @types/react, I don't see any reason this would have just started failing
> except I guess that they haven't published in a month so maybe something else happened in that last month in our code that wasn't caught until just now

Issue #1536, opened 2023-09-18
Consider making tablegen's struct setters first class citizens instead of calling variable setters


alvrs:
Context in https://github.com/latticexyz/mud/pull/1532/files#r1328697853

PR #1542, opened 2023-09-18
refactor(store): inline logic in codegenned set method which uses struct

dk1a:
fixes https://github.com/latticexyz/mud/issues/1536


holic:
was this meant to be an improvement on gas or stack size?

dk1a:
> was this meant to be an improvement on gas or stack size?

stack size, tho gas might be slightly improved too

.changeset/olive-pigs-fold.md
```diff
@@ -0,0 +1,6 @@
+---
+"@latticexyz/store": major
+---
+
+- Always render field methods with a suffix in tablegen (they used to not be rendered if field methods without a suffix were rendered).
+- Add `withSuffixlessFieldMethods` to `RenderTableOptions`, which indicates that field methods without a suffix should be rendered.
```
holic, line 6: looks like these are from #1550? does this need a rebase?

dk1a: updated

.changeset/tall-buses-kick.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/store": major
```
holic, line 2: is this a major change?

dk1a: kind of, I add an argument to `renderRecordData`, which is technically an exported function that somebody could be using, albeit unlikely

holic: since we tend to default this to `""` in a few places, we could make the argument optional with `'""` as the default and make it a minor change

dk1a: updated, don't have a strong opinion here and will trust ur judgement. Initially I'd considered this, but I'm kinda biased against optional args

alvrs (approved):

alvrs (approved):

Issue #1139, opened 2023-07-10
Improve handling of custom types in tablegen


alvrs:
- [ ] Allow generating 1 struct for several tables with identical data, so the data can be seamlessly passed between tables
- [x] Allow generating user types from the config to be used in tables
- [ ] Allow structs to be imported from a manual user file

alvrs:
partially fixed by #1566

holic:
closing in favor of https://github.com/latticexyz/mud/issues/1461

Issue #1126, opened 2023-07-10
Prefix all events and errors with library/contract name


alvrs:
We currently prefix `StoreCore` events with `StoreCore`, but not `World` events/errors. We should align on one approach, and prefixing makes it easier to debug where an event came from.

TODO:
- [ ] prefix events
- [ ] try consolidating event defs of the same name (solidity 0.8.21 bump might allow us to reference events defined elsewhere)
- [ ] prefix errors
- [ ] make module error names consistent (one uses "root mode" and the other just says "non root"): https://github.com/latticexyz/mud/blob/b18d49171bf1750da9d659ac71a6152592e57193/packages/world/src/interfaces/IModule.sol#L14-L15

dk1a:
> We should align on one approach, and prefixing makes it easier to debug where an event came from

agree with prefixing

holic:
Also forge flattens ABI directory structures, which means there's higher chance of collision if we had something named `Context` but also pull in a lib that has `Context` contract already.

holic:
when we do events, we should see if we can consolidate into one place (e.g. interface or library) and reference them directly

I believe the latest solc version should enable this but I encountered errors when I tried (but didn't dig too deep): https://github.com/latticexyz/mud/pull/1473#issuecomment-1718405334

PR #1544, opened 2023-09-18
feat(store,world): replace `ResourceSelector` with `ResourceId` and `WorldResourceId`

### LINKED ISSUE #1367: Rename resourceSelector to `tableId` and `systemId` in function signatures for clarity
- `resourceSelector` is an umbrella term for `tableId` and `systemId`.
- For clarity we should use `tableId` and `systemId` in the respective functions that only interact with either of them (eg `registerTable` vs `registerSystem`)
- In remaining places we should rename `resourceSelector` to `resourceId` for consistency with `tableId` and `systemId`

### LINKED ISSUE #1394: Consider adding resource type to resource selector
Right now we generally use the pattern:
```
namespace (bytes16) + name (bytes16)
```

I am proposing we explore a pattern like:
```
resource type (bytes2) + namespace (bytes14) + name (bytes16)
```

This comes from seeing a few patterns emerge:
- `ResourceType` table/enum for registering systems: https://github.com/latticexyz/mud/blob/5150544354816949ab3d47950df7fc67ded797a7/packages/world/mud.config.ts#L95-L102
- module namespace suffix: https://github.com/latticexyz/mud/blob/5150544354816949ab3d47950df7fc67ded797a7/packages/world/src/modules/uniqueentity/constants.sol#L5
- onchain table vs offchain tables: https://github.com/latticexyz/mud/pull/1496#discussion_r1326925880

alvrs:
fixes #1394 
fixes #1367 

TODO
- [x] Turn `ResourceId` into a user type - had a bunch of runtime errors where `bytes14` was automatically casted into `bytes32`, which would have been caught at compile time if resource type was a user type

TODOs in followups
- [x] Move away from using `bytes14 namespace` as argument, and use `ResourceId namespaceId` instead (https://github.com/latticexyz/mud/pull/1555)
- [x] Consider removing `ResourceType` table. Need to find a way to check whether a given resource already exists though (https://github.com/latticexyz/mud/pull/1557)
- [x] Add requirement checks to all registration functions to revert if the resource id doesn't match the expected type (https://github.com/latticexyz/mud/pull/1557)


packages/world/src/ResourceId.sol
```diff
@@ -0,0 +1,92 @@
...
+// (I ran into a bunch of runtime errors while refactoring this where bytes14 was automatically casted to bytes32)
+
+library ResourceId {
+  /**
+   * Create a 32-byte resource ID from a namespace, name and type.
+   *
+   * A resource ID is a 32-byte value that uniquely identifies a resource.
+   * The first 14 bytes represent the namespace,
+   * the next 16 bytes represent the name,
+   * the last 2 bytes represent the type.
```
holic, line 24: ooc why last two bytes? in terms of information hierarchy, type feels like it should be first

alvrs: we check the namespace more often and in hotter paths (in access control) and it's cheaper to check the most significant bits because we don't need to shift

holic: does that mean type is least often used compared to name?

e2e/packages/contracts/src/codegen/tables/Multi.sol
```diff
@@ -17,9 +17,13 @@ import { EncodeArray } from "@latticexyz/store/src/tightcoder/EncodeArray.sol";
 import { FieldLayout, FieldLayoutLib } from "@latticexyz/store/src/FieldLayout.sol";
 import { Schema, SchemaLib } from "@latticexyz/store/src/Schema.sol";
 import { PackedCounter, PackedCounterLib } from "@latticexyz/store/src/PackedCounter.sol";
+import { ResourceId } from "@latticexyz/store/src/ResourceId.sol";
+import { RESOURCE_TABLE } from "@latticexyz/store/src/storeResourceTypes.sol";
 
-bytes32 constant _tableId = bytes32(abi.encodePacked(bytes16(""), bytes16("Multi")));
-bytes32 constant MultiTableId = _tableId;
+ResourceId constant _tableId = ResourceId.wrap(
+  bytes32(abi.encodePacked(bytes14(""), bytes16("Multi"), RESOURCE_TABLE))
```
holic, line 24: could we just hardcode the result (hex) of this like we do for field layout?

alvrs: there wouldn't be a gas advantage (since this is computed statically at compile time) and think it would be more complex / less obvious what's going on if we only render hex here

packages/config/src/library/validation.ts
```diff
@@ -2,6 +2,7 @@ import { utils } from "ethers";
 import { ZodIssueCode, RefinementCtx } from "zod";
 
 export const STORE_SELECTOR_MAX_LENGTH = 16;
```
holic, line 4: since this is no longer a selector, should we rename this to `STORE_RESOURCE_NAME_MAX_LENGTH` or something?

packages/store/src/IStore.sol
```diff
@@ -6,23 +6,24 @@ import { PackedCounter } from "./PackedCounter.sol";
...
 import { IStoreHook } from "./IStoreHook.sol";
+import { ResourceId } from "./ResourceId.sol";
 
 interface IStoreRead {
   event HelloStore(bytes32 indexed storeVersion);
 
   function storeVersion() external view returns (bytes32);
 
-  function getFieldLayout(bytes32 tableId) external view returns (FieldLayout fieldLayout);
+  function getFieldLayout(ResourceId tableId) external view returns (FieldLayout fieldLayout);
```
holic, line 16: Did you decide against calling this type TableId and having specific types for each resource type?

packages/store/src/StoreCore.sol
```diff
@@ -87,8 +88,8 @@ library StoreCore {
   /**
    * Get the field layout for the given tableId
    */
-  function getFieldLayout(bytes32 tableId) internal view returns (FieldLayout fieldLayout) {
-    fieldLayout = FieldLayout.wrap(Tables._getFieldLayout(tableId));
+  function getFieldLayout(ResourceId tableId) internal view returns (FieldLayout fieldLayout) {
+    fieldLayout = FieldLayout.wrap(Tables._getFieldLayout(ResourceId.unwrap(tableId)));
```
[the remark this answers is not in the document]
alvrs: It would be very nice, but i'm not sure if we have capacity before the freeze. It's tracked here: https://github.com/latticexyz/mud/issues/1139

packages/store/src/IStore.sol
alvrs: Unfortunately that would make it much harder to share the `Instance` libraries between the different types (and since all resource ids are structured the same way i think it makes sense to call the type the more general `ResourceId`

packages/config/src/library/validation.ts
alvrs: could do!

packages/world/src/ResourceId.sol
alvrs: i'd say `type` and `name` are used similarly often (in access control checks, and in the future in "type" checks if logic depends on it). I did some quick benchmarks and it seems like it's not significant how much you shift, just whether or not you shift, so it doesn't really matter if `name` or `type` would come first

packages/store/src/IStore.sol
holic: ahhh I see now that `ResourceId` is generic name+type and `WorldResourceId` is generic namespace+name+type

alvrs: Yeah! They share the same user type `ResourceId`, but by doing `using WorldResourceIdInstance for ResourceId` we get `getNamespace`, `getName`, etc (while `getType` is on `ResourceIdInstance`)

e2e/packages/contracts/src/codegen/tables/Multi.sol
holic: ahh if it's computed at compile time then great!

packages/store/src/IStore.sol
holic: woah TIL you can "extend" like that

packages/store/src/StoreCore.sol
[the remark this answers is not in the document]
holic: not sure I follow! why can't `_getFieldLayout` take a `ResourceId` arg instead of `bytes32`?

packages/config/src/library/validation.ts
alvrs: done

.changeset/stale-seahorses-pay.md
```diff
@@ -0,0 +1,70 @@
...
+---
+
+- `ResourceSelector` is replaced with `ResourceId`, `ResourceIdLib`, `ResourceIdInstance`, `WorldResourceIdLib` and `WorldResourceIdInstance`.
+
+  Previously a "resource selector" was a `bytes32` value with the first 16 bytes reserved for the resource's namespace, and the last 16 bytes reserved for the resource's name.
+  Now a "resource ID" is a `bytes32` value with the first 14 bytes reserved for the resource's namespace, the next 16 bytes reserved for the resource's name, and the last 2 bytes reserved for the resource type.
+
+  Previously `ResouceSelector` was a library and the resource selector type was a plain `bytes32`.
+  Now `ResourceId` is a user type, and the functionality is implemented in the `ResourceIdInstance` (for type) and `WorldResourceIdInstance` (for namespace and name) libraries.
+  We split the logic into two libraries, because `Store` also uses `ResourceId` now and needs awareness of resource types, but not of namespaces/names.
```
alvrs, line 15:
```suggestion
  We split the logic into two libraries, because `Store` now also uses `ResourceId` and needs to be aware of resource types, but not of namespaces/names.
```

packages/common/src/hexToResourceId.test.ts
```diff
@@ -0,0 +1,11 @@
+import { describe, it, expect } from "vitest";
+import { hexToResourceId } from "./hexToResourceId";
+
+describe("hexToResourceId", () => {
+  it("can convert from hex string", () => {
+    const resourceId = hexToResourceId("0x6e616d65737061636500000000006e616d650000000000000000000000007462");
+    expect(resourceId.namespace).toMatchInlineSnapshot('"namespace"');
+    expect(resourceId.name).toMatchInlineSnapshot('"name"');
+    expect(resourceId.type).toMatchInlineSnapshot('"table"');
```
alvrs, line 9: ooc what's the difference between `toMatchInlineSnapshot` and `toEqual`?

packages/common/src/hexToResourceId.ts
```diff
@@ -0,0 +1,30 @@
...
+  const namespace = hexToString(sliceHex(hex, 0, 14)).replace(/\0+$/, "");
+  const name = hexToString(sliceHex(hex, 14, 30)).replace(/\0+$/, "");
+  const resourceTypeId = hexToString(sliceHex(hex, 30, 32)).replace(/\0+$/, "");
+  const type = getResourceType(resourceTypeId);
+
+  if (!type) {
+    throw new Error(`Unknown resource type: ${resourceTypeId}`);
+  }
+
+  return { namespace, name, type };
```
alvrs, line 29: i would have expected `type` to be a reserved keyword in typescript, interesting that it's not

packages/common/src/resourceIdToHex.ts
```diff
@@ -0,0 +1,23 @@
+import { Hex, stringToHex, concatHex } from "viem";
+import { ResourceId } from "./common";
+import { ResourceType } from "./resourceTypes";
+
+/** @internal */
```
alvrs, line 5: ooc what does `@internal` do?

packages/common/src/hexToResourceId.test.ts
holic: inline snapshots are serialized and will give you when running tests to update the snapshot

a bit less of a hassle than having a hardcoded value in `toEqual` that you have to manually update

packages/common/src/resourceIdToHex.ts
holic: ideally keeps it from being exported outside of the package: https://typedoc.org/tags/internal/

packages/world/src/modules/keyswithvalue/getTargetTableId.sol
```diff
@@ -0,0 +1,31 @@
...
+uint256 constant MODULE_NAMESPACE_BYTES = 7;
+uint256 constant TABLE_NAMESPACE_BYTES = 7;
+uint256 constant TABLE_NAME_BYTES = 16;
+uint256 constant BYTES_TO_BITS = 8;
+
+/**
+ * Get a deterministic selector for the reverse mapping table for the given source table.
+ * The selector is constructed as follows:
+ *  - The first 7 bytes are the module namespace
+ *  - The next 7 bytes are the first 8 bytes of the source table namespace
```
holic, line 16:
```suggestion
 *  - The next 7 bytes are the first 7 bytes of the source table namespace
```

packages/world/src/modules/keyswithvalue/getTargetTableId.sol
```diff
@@ -0,0 +1,31 @@
...
+ *    -- This is to avoid collisions between tables with the same name in different namespaces
+ *       (Note that collisions are still possible if the first 8 bytes of the namespace are the same, in which case installing the module fails)
```
holic, line 18:
```suggestion
 *       (Note that collisions are still possible if the first 7 bytes of the namespace are the same, in which case installing the module fails)
```

holic (commented):
Since we have codegen and helpers for interacting with resource IDs, would you consider adjusting the API so that the type is first in all args? Can still keep the encoded version as `{namespace}{name}{type}` but intuitively I want to read the type first so I know what kind of resource I am working with.

(I took a similar approach in the frontend where the underlying encoded hex is all abstracted away, users can just interact with an object in any order)

packages/world/test/WorldBalance.t.sol
```diff
@@ -6,13 +6,14 @@ import { GasReporter } from "@latticexyz/gas-report/src/GasReporter.sol";
...
 import { ROOT_NAMESPACE } from "../src/constants.sol";
 import { CoreModule } from "../src/modules/core/CoreModule.sol";
 import { Balances } from "../src/modules/core/tables/Balances.sol";
 import { IWorldErrors } from "../src/interfaces/IWorldErrors.sol";
+import { RESOURCE_SYSTEM } from "../src/worldResourceTypes.sol";
 
-using ResourceSelector for bytes32;
+using WorldResourceIdInstance for ResourceId;
 
 contract WorldBalanceTestSystem is System {
```
holic, line 18: this keeps the world in balance 😇

packages/world/ts/library/config/worldConfig.ts
```diff
@@ -9,7 +9,7 @@ const zSystemAccessList = z.array(zSystemName.or(zEthereumAddress)).default(SYST
 // The system config is a combination of a name config and access config
 const zSystemConfig = z.intersection(
   z.object({
-    name: zSelector.optional(),
+    name: zName.optional(),
```
holic, line 12: ahh I guess if we adjusted `zName` to validate length, it'd fail for system contract names?

packages/common/src/codegen/render-solidity/common.ts
```diff
@@ -166,11 +166,11 @@ export function renderTableId(staticResourceData: StaticResourceData): {
   hardcodedTableId: string;
   tableIdDefinition: string;
 } {
-  const hardcodedTableId = `bytes32(abi.encodePacked(bytes16("${staticResourceData.namespace}"), bytes16("${staticResourceData.name}")))`;
+  const hardcodedTableId = `ResourceId.wrap(bytes32(abi.encodePacked(bytes14("${staticResourceData.namespace}"), bytes16("${staticResourceData.name}"), RESOURCE_TABLE)))`;
```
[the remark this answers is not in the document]
alvrs: yes! i planned to that in the #1496 PR

.changeset/stale-seahorses-pay.md
```diff
@@ -0,0 +1,70 @@
...
+  IBaseWorld {
+  function callFrom(
+    address delegator,
+  -   bytes32 resourceSelector,
+  +   ResourceId systemId,
+    bytes memory callData
+  ) external payable returns (bytes memory);
+
+  // Same for all other methods
+  }
```
alvrs, line 70:
```suggestion
  IBaseWorld {
    function callFrom(
      address delegator,
  -   bytes32 resourceSelector,
  +   ResourceId systemId,
      bytes memory callData
    ) external payable returns (bytes memory);

    // Same for all other methods
  }
```

packages/cli/contracts/src/codegen/tables/Ephemeral.sol
```diff
@@ -17,9 +17,13 @@ import { EncodeArray } from "@latticexyz/store/src/tightcoder/EncodeArray.sol";
 import { FieldLayout, FieldLayoutLib } from "@latticexyz/store/src/FieldLayout.sol";
 import { Schema, SchemaLib } from "@latticexyz/store/src/Schema.sol";
 import { PackedCounter, PackedCounterLib } from "@latticexyz/store/src/PackedCounter.sol";
+import { ResourceId } from "@latticexyz/store/src/ResourceId.sol";
+import { RESOURCE_TABLE } from "@latticexyz/store/src/storeResourceTypes.sol";
 
-bytes32 constant _tableId = bytes32(abi.encodePacked(bytes16(""), bytes16("Ephemeral")));
-bytes32 constant EphemeralTableId = _tableId;
+ResourceId constant _tableId = ResourceId.wrap(
+  bytes32(abi.encodePacked(bytes14(""), bytes16("Ephemeral"), RESOURCE_TABLE))
```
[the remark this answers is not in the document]
alvrs: I didn't touch any of the ephemeral/offchain table logic yet, planned to do that in #1496

packages/config/src/library/validation.ts
```diff
@@ -1,7 +1,8 @@
 import { utils } from "ethers";
 import { ZodIssueCode, RefinementCtx } from "zod";
 
-export const STORE_SELECTOR_MAX_LENGTH = 16;
+export const STORE_NAME_MAX_LENGTH = 16;
```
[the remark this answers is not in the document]
alvrs: Names are inferred from the key in the `tables` object and truncated automatically

packages/store/ts/config/storeConfig.ts
```diff
@@ -170,7 +171,7 @@ export const zTablesConfig = z.record(zTableName, zTableConfig).transform((table
   // default name depends on tableName
   for (const tableName of Object.keys(tables)) {
     const table = tables[tableName];
-    table.name = tableName.slice(0, STORE_SELECTOR_MAX_LENGTH);
+    table.name = tableName.slice(0, STORE_NAME_MAX_LENGTH);
```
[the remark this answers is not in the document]
alvrs: `table.name` is not configurable but inferred from the key in the tables object, see https://github.com/latticexyz/mud/pull/818. (I see you already had a comment in there about warning users if the name is truncated: https://github.com/latticexyz/mud/pull/818#discussion_r1193163480, we should open a followup issue to track that)

System `name` is configurable, but still has to fit in the 16 bytes of the `systemId`'s name (the name we register the function selector should match the system id). We could add a `zod` error for when the system name is longer than 16 char though.

packages/world/src/WorldResourceId.sol
```diff
@@ -0,0 +1,102 @@
...
+    bytes2 resourceType = ResourceIdInstance.getType(resourceId);
+    return
+      string(
+        abi.encodePacked(
+          resourceNamespace == ROOT_NAMESPACE ? ROOT_NAMESPACE_STRING : resourceNamespace,
+          "_",
+          resourceName == ROOT_NAME ? ROOT_NAME_STRING : resourceName,
+          ".",
+          resourceType
+        )
```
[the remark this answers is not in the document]
alvrs: yeah that was a different PR but can also move it here

packages/world/src/interfaces/IModule.sol
```diff
@@ -10,7 +10,7 @@ bytes4 constant MODULE_INTERFACE_ID = IModule.getName.selector ^
   ERC165_INTERFACE_ID;
 
 interface IModule is IERC165 {
-  error RequiredModuleNotFound(string resourceSelector);
+  error RequiredModuleNotFound(string moduleId);
```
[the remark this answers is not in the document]
alvrs: > Is this not bytes32 or ResourceId because it's a debug value? like a human readable one?

exactly

alvrs, line 13:
```suggestion
  error RequiredModuleNotFound(string moduleIdString);
```

packages/world/mud.config.ts
```diff
@@ -63,7 +63,7 @@ export default mudConfig({
     Systems: {
```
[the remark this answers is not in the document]
alvrs: coming in #1555

packages/world/src/modules/core/implementations/BalanceTransferSystem.sol
```diff
@@ -3,21 +3,21 @@ pragma solidity >=0.8.0;
...
 
 contract BalanceTransferSystem is System, IWorldErrors {
-  using ResourceSelector for bytes32;
+  using WorldResourceIdInstance for ResourceId;
 
   /**
    * Transfer balance to another namespace in the World
    */
-  function transferBalanceToNamespace(bytes16 fromNamespace, bytes16 toNamespace, uint256 amount) public virtual {
+  function transferBalanceToNamespace(bytes14 fromNamespace, bytes14 toNamespace, uint256 amount) public virtual {
```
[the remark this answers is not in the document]
alvrs: coming in #1555

packages/world/src/modules/core/implementations/StoreRegistrationSystem.sol
```diff
@@ -31,56 +32,53 @@ import { WorldRegistrationSystem } from "./WorldRegistrationSystem.sol";
...
-      // an external call, so msg.sender would be the address of the World contract
-      (address systemAddress, ) = Systems._get(ResourceSelector.from(ROOT_NAMESPACE, CORE_SYSTEM_NAME));
-      WorldContextProvider.delegatecallWithContextOrRevert({
-        msgSender: _msgSender(),
-        msgValue: 0,
-        target: systemAddress,
-        callData: abi.encodeCall(WorldRegistrationSystem.registerNamespace, (namespace))
-      });
+    ResourceId namespaceId = tableId.getNamespaceId();
+    if (ResourceType._get(ResourceId.unwrap(namespaceId)) == Resource.NONE) {
```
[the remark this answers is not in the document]
alvrs: yep, #1553

packages/world/src/modules/core/implementations/StoreRegistrationSystem.sol
```diff
@@ -31,56 +32,53 @@ import { WorldRegistrationSystem } from "./WorldRegistrationSystem.sol";
...
+      // Since this is a root system, we're in the context of the World contract already,
+      // so we can use delegatecall to register the namespace
```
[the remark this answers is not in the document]
alvrs: `StoreRegistrationSystem` is part of the `CoreModule`, which registers the `CoreSystem` as root system

packages/world/src/modules/core/implementations/WorldRegistrationSystem.sol
```diff
@@ -88,54 +88,55 @@ contract WorldRegistrationSystem is System, IWorldErrors {
...
     // If the namespace doesn't exist yet, register it
     // otherwise require caller to own the namespace
-    bytes16 namespace = resourceSelector.getNamespace();
-    if (ResourceType._get(namespace) == Resource.NONE) registerNamespace(namespace);
-    else AccessControl.requireOwner(namespace, _msgSender());
+    ResourceId namespaceId = systemId.getNamespaceId();
+    if (ResourceType._get(ResourceId.unwrap(namespaceId)) == Resource.NONE) registerNamespace(systemId.getNamespace());
+    else AccessControl.requireOwner(namespaceId, _msgSender());
 
+    // TODO: this check is unnecessary with resource types, need to replace with a requirement that the system type is encoded correctly
```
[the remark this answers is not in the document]
alvrs: already queued in #1557

packages/world/src/modules/core/implementations/WorldRegistrationSystem.sol
```diff
@@ -181,44 +182,43 @@ contract WorldRegistrationSystem is System, IWorldErrors {
...
-    Delegations.set({ delegator: _msgSender(), delegatee: delegatee, delegationControlId: delegationControlId });
+    Delegations.set({
+      delegator: _msgSender(),
+      delegatee: delegatee,
+      delegationControlId: ResourceId.unwrap(delegationControlId)
+    });
 
     // If the delegation is not unlimited...
-    if (delegationControlId != UNLIMITED_DELEGATION && initCallData.length > 0) {
+    if (ResourceId.unwrap(delegationControlId) != ResourceId.unwrap(UNLIMITED_DELEGATION) && initCallData.length > 0) {
```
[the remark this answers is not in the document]
alvrs: unfortunately not 😭

alvrs:
> Since we have codegen and helpers for interacting with resource IDs, would you consider adjusting the API so that the type is first in all args? Can still keep the encoded version as {namespace}{name}{type} but intuitively I want to read the type first so I know what kind of resource I am working with.

> (I took a similar approach in the frontend where the underlying encoded hex is all abstracted away, users can just interact with an object in any order)

I feel like it would be most intuitive if the argument order would match the order in which they are encoded. (Sometimes it's necessary to manually encode via `abi.encodePacked` because the library call is not considered a compile-time static function, and i think it would be a tripwire if the order was different in that case compared to the libraries.)

But it's already possible to change the order by using named arguments:

```solidity
WorldResourceIdLib.encode({ 
  resourceType: RESOURCE_SYSTEM,
  resourceNamespace: "namespace",
  resourceName: "name"
});
```

I also personally think it would be most intuitive to match the encoded order in the stringified version, but would be open to putting the type first there if you feel strongly about it, since we now also log the hex value next to the stringified version.

packages/store/ts/config/storeConfig.ts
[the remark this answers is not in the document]
holic: yeah for sure! this comment isn't blocking, just noticed it as I was passing through

(though validation is a thing I'm kinda punting on right now haha)

holic:
from discord:
> I’m fine sticking with encoding order for resources
> not my favorite but I’m optimizing for consistency and gas

.changeset/stale-seahorses-pay.md
```diff
@@ -0,0 +1,70 @@
...
+      bytes32[] calldata keyTuple,
+      bytes calldata staticData,
+      PackedCounter encodedLengths,
+      bytes calldata dynamicData,
+      FieldLayout fieldLayout
+    ) external;
+
+    // Same for all other methods
+  }
+  ```
```
holic, line 55:
```suggestion
  ```diff
    import { ResourceId } from "@latticexyz/store/src/ResourceId.sol";
  
    IStore {
      function setRecord(
  -     bytes32 tableId,
  +     ResourceId tableId,
        bytes32[] calldata keyTuple,
        bytes calldata staticData,
        PackedCounter encodedLengths,
        bytes calldata dynamicData,
        FieldLayout fieldLayout
      ) external;
  
      // Same for all other methods
    }
  ```
```

.changeset/stale-seahorses-pay.md
```diff
@@ -0,0 +1,70 @@
...
+  function callFrom(
+    address delegator,
+  -   bytes32 resourceSelector,
+  +   ResourceId systemId,
+    bytes memory callData
+  ) external payable returns (bytes memory);
+
+  // Same for all other methods
+  }
+  ```
```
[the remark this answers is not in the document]
holic: ugh gihub messed up my attempt at suggesting a chance, but I think for diffs you want a "gutter" where the + and - live 

so basically everything in the code block should indent one level then dedent the +- to be at the left edge
(see your diff above)

holic (approved):

.changeset/stale-seahorses-pay.md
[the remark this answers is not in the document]
alvrs: oof yes that's what i did, i think prettier keeps messing this up again

.changeset/stale-seahorses-pay.md
```diff
@@ -0,0 +1,70 @@
...
+"@latticexyz/cli": major
+"@latticexyz/common": major
+"@latticexyz/config": major
+"@latticexyz/store": major
+---
+
+- `ResourceSelector` is replaced with `ResourceId`, `ResourceIdLib`, `ResourceIdInstance`, `WorldResourceIdLib` and `WorldResourceIdInstance`.
+
+  Previously a "resource selector" was a `bytes32` value with the first 16 bytes reserved for the resource's namespace, and the last 16 bytes reserved for the resource's name.
+  Now a "resource ID" is a `bytes32` value with the first 14 bytes reserved for the resource's namespace, the next 16 bytes reserved for the resource's name, and the last 2 bytes reserved for the resource type.
```
alvrs, line 11:
```suggestion
  Now a "resource ID" is a `bytes32` value with the first 2 bytes reserved for the resource type, the next 14 bytes reserved for the resource's namespace, and the last  16 bytes reserved for the resource's name.
```

.changeset/stale-seahorses-pay.md
```diff
@@ -0,0 +1,70 @@
...
+  We split the logic into two libraries, because `Store` now also uses `ResourceId` and needs to be aware of resource types, but not of namespaces/names.
+
+  ```diff
+  - import { ResourceSelector } from "@latticexyz/world/src/ResourceSelector.sol";
+  + import { ResourceId, ResourceIdInstance } from "@latticexyz/store/src/ResourceId.sol";
+  + import { WorldResourceIdLib, WorldResourceIdInstance } from "@latticexyz/world/src/WorldResourceId.sol";
+  + import { RESOURCE_SYSTEM } from "@latticexyz/world/src/worldResourceTypes.sol";
+
+  - bytes32 systemId = ResourceSelector.from("namespace", "name");
+  + ResourceId systemId = WorldResourceIdLib.encode("namespace", "name", RESOURCE_SYSTEM);
```
alvrs, line 24:
```suggestion
  + ResourceId systemId = WorldResourceIdLib.encode(RESOURCE_SYSTEM, "namespace", "name");
```

packages/store/test/ResourceId.t.sol
```diff
@@ -0,0 +1,37 @@
...
+    startGasReport("get type from a table ID");
+    bytes2 resourceType = tableId.getType();
+    endGasReport();
+
+    assertEq(resourceType, RESOURCE_TABLE);
+  }
+
+  function testFuzz(bytes30 name, bytes2 resourceType) public {
+    bytes2 NOT_TYPE = bytes2("xx");
+    vm.assume(resourceType != NOT_TYPE);
```
alvrs, line 33:
```suggestion
```

e2e/packages/contracts/src/codegen/tables/Multi.sol
```diff
@@ -17,9 +17,13 @@ import { EncodeArray } from "@latticexyz/store/src/tightcoder/EncodeArray.sol";
 import { FieldLayout, FieldLayoutLib } from "@latticexyz/store/src/FieldLayout.sol";
 import { Schema, SchemaLib } from "@latticexyz/store/src/Schema.sol";
 import { PackedCounter, PackedCounterLib } from "@latticexyz/store/src/PackedCounter.sol";
+import { ResourceId } from "@latticexyz/store/src/ResourceId.sol";
+import { RESOURCE_TABLE } from "@latticexyz/store/src/storeResourceTypes.sol";
 
-bytes32 constant _tableId = bytes32(abi.encodePacked(bytes16(""), bytes16("Multi")));
-bytes32 constant MultiTableId = _tableId;
+ResourceId constant _tableId = ResourceId.wrap(
+  bytes32(abi.encodePacked(RESOURCE_TABLE, bytes14(""), bytes16("Multi")))
```
holic, line 24: ooc is there any gas benefit to `bytes.concat` instead of `abi.encodePacked` here?

packages/common/src/hexToResourceId.test.ts
```diff
@@ -0,0 +1,11 @@
+import { describe, it, expect } from "vitest";
+import { hexToResourceId } from "./hexToResourceId";
+
+describe("hexToResourceId", () => {
+  it("can convert from hex string", () => {
+    const resourceId = hexToResourceId("0x74626e616d65737061636500000000006e616d65000000000000000000000000");
+    expect(resourceId.namespace).toMatchInlineSnapshot('"namespace"');
+    expect(resourceId.name).toMatchInlineSnapshot('"name"');
+    expect(resourceId.type).toMatchInlineSnapshot('"table"');
```
holic, line 9:
```suggestion
    expect(resourceId.type).toMatchInlineSnapshot('"table"');
    expect(resourceId.namespace).toMatchInlineSnapshot('"namespace"');
    expect(resourceId.name).toMatchInlineSnapshot('"name"');
```

packages/common/src/hexToResourceId.ts
```diff
@@ -0,0 +1,30 @@
...
+  if (resourceTypes.includes(type as ResourceType)) {
+    return type;
+  }
+}
+
+export function hexToResourceId(hex: Hex): ResourceId {
+  const resourceTypeId = hexToString(sliceHex(hex, 0, 2)).replace(/\0+$/, "");
+  const namespace = hexToString(sliceHex(hex, 2, 16)).replace(/\0+$/, "");
+  const name = hexToString(sliceHex(hex, 16, 32)).replace(/\0+$/, "");
+  const type = getResourceType(resourceTypeId);
```
holic, line 23:
```suggestion
  const type = getResourceType(resourceTypeId);
  const namespace = hexToString(sliceHex(hex, 2, 16)).replace(/\0+$/, "");
  const name = hexToString(sliceHex(hex, 16, 32)).replace(/\0+$/, "");
```

packages/common/src/hexToResourceId.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  if (!type) {
+    throw new Error(`Unknown resource type: ${resourceTypeId}`);
+  }
+
+  return { namespace, name, type };
```
holic, line 29:
```suggestion
  return { type, namespace, name };
```

packages/world/src/interfaces/IModule.sol
```diff
@@ -10,7 +10,7 @@ bytes4 constant MODULE_INTERFACE_ID = IModule.getName.selector ^
   ERC165_INTERFACE_ID;
 
 interface IModule is IERC165 {
-  error RequiredModuleNotFound(string resourceSelector);
+  error RequiredModuleNotFound(string moduleIdString);
   error RootInstallModeNotSupported();
   error NonRootInstallNotSupported();
```
holic, line 15: maybe for followup: should we make these more consistent by removing "mode" from one or adding it to the other?

holic (dismissed):

packages/world/src/interfaces/IModule.sol
```diff
@@ -10,7 +10,7 @@ bytes4 constant MODULE_INTERFACE_ID = IModule.getName.selector ^
   ERC165_INTERFACE_ID;
 
 interface IModule is IERC165 {
-  error RequiredModuleNotFound(string resourceSelector);
+  error RequiredModuleNotFound(string moduleIdString);
```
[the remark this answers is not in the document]
alvrs: we could actually just remove this, it's not used anywhere and we don't have a module resource type either

packages/world/src/interfaces/IModule.sol
holic: Oh I mean "root install _mode_ not supported" vs "non root install not supported", we call it a "mode" in one instance but not the other

but yeah, can resolve as part of naming pass in #1126

alvrs: > Oh I mean "root install mode not supported" vs "non root install not supported", we call it a "mode" in one instance but not the other

I understood, I meant we can do this renaming as part of the naming pass in 1126

holic (dismissed):

holic (approved):

Issue #1319, opened 2023-08-17
Proposal: Keep get{FieldName} function in the single value table


CrazyNorman:
Now if the table has only one filed in the schemas, the get{FileName} and set{FiledName} function will be removed from the table code generating. However this can cause code refactoring when I add one filed in the shema. Maybe we can reserve the  get{FileName} and set{FiledName} to avoid this troublesome. The extra gas cost for reserve these two functions should be acceptable.

dk1a:
I think more functions should be free gas-wise, they aren't kept because that'd double the functions in table lib making it harder to read/use (there're more methods than set/get, and with 1-field tables just using get/read/push/pop/... is really convenient imo).
But I agree that at least an option to force the postfixes in 1-field tables, so they're like multi-fields ones (for ppl planning to potentially add more fields) would be useful

CrazyNorman:
Yes, having a config option to control the function convention will be very helpful. It is very common you define a table config with single value at first then add a field later during the development process.

holic:
Just ran into this when prototyping a thing, and I agree that it would be clearer if we kept `getField` methods.

PR #1550, opened 2023-09-19
refactor(store): always render field methods with suffix and conditionally without

dk1a:
fixes https://github.com/latticexyz/mud/issues/1319


alvrs (approved):
🚢

PR #1555, opened 2023-09-20
feat(world): use `ResourceId namespaceId` for all methods acting on the namespace as a resource

### LINKED ISSUE #1552: Consider using `ResourceId namespaceId` instead of `bytes14 namespace` for registering namespaces and in BalanceTransferSystem
Followup to #1544

alvrs:
Fixes #1552 
Followup to #1544 

All methods acting on namespaces as resources (ownership and balances of namespaces) now use `ResourceId namespaceId` instead of `bytes14 namespace` for clarity and improved compile time type safety (no more implicit casting)
- `WorldRegistrationSystem` now uses `ResourceId namespaceId` instead of `bytes14 namespace` for `registerNamespace`
- `BalanceTransferSystem` now uses `ResourceId namespaceId` instead of `bytes14 namespace` for all methods
- `AccessManagementSystem` now uses `ResourceId namespaceId` instead of `bytes14 namespace` for `transferOwnership` and internal ownership checks


packages/world/src/WorldResourceId.sol
```diff
@@ -92,7 +92,7 @@ library WorldResourceIdInstance {
       string(
         abi.encodePacked(
           resourceNamespace == ROOT_NAMESPACE ? ROOT_NAMESPACE_STRING : resourceNamespace,
-          "_",
+          "/",
           resourceName == ROOT_NAME ? ROOT_NAME_STRING : resourceName,
           ".",
           resourceType
```
holic, line 98: I think this is leftover from our "file" analogy, which I think we've moved on from?

We should agree on a stringified version of these resource IDs, and my vote would be something like
```
type_namespace_name
type:namespace:name
```
(Stripe uses e.g. `acct_1234` and `ch_1234`)

packages/world/src/SystemCall.sol
```diff
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: MIT
 pragma solidity >=0.8.0;
 
+import { console } from "forge-std/console.sol";
+
```
holic, line 5:
```suggestion
```

packages/world/src/interfaces/IAccessManagementSystem.sol
```diff
@@ -10,5 +10,5 @@ interface IAccessManagementSystem {
 
   function revokeAccess(ResourceId resourceId, address grantee) external;
 
-  function transferOwnership(bytes14 namespace, address newOwner) external;
+  function transferOwnership(ResourceId namespaceId, address newOwner) external;
```
holic, line 13: We sorta talked about this in the other PR but I wonder if there's any value to having a `NamespaceId` wrapper rather than full resource ID where we mask/shift/slice the namespace from it.

packages/world/src/WorldResourceId.sol
alvrs: don't have a strong opinion on the separator string, `_` was just a bit confusing in `ROOT_NAMESPACE_ROOT_NAME`. `:` works for me!

packages/world/src/interfaces/IAccessManagementSystem.sol
alvrs: would prefer one user type for all of these because i want to add checks like `namespaceId.isType(RESOURCE_NAMESPACE)` in the next PR - if `namespaceId` uses a different user type, we'd have to duplicate the `ResourceId` library for it. (We can't trust the id provided as argument to actually be a valid namespace id, even if it would be of type `NamespaceId`, because anyone can just wrap whatever with `NamespaceId`)

holic: oh right, I forgot we have a namespace type, that makes sense

packages/world/src/WorldResourceId.sol
alvrs: how about `namespace:name:type`? (to make the order match the encoding)

holic: I would still argue for the more human readable form here, and asked in the other PR about moving the order for our helpers

alvrs: from discord for future reference

> I’m fine sticking with encoding order for resources
> not my favorite but I’m optimizing for consistency and gas

holic (approved):

PR #1557, opened 2023-09-20
feat(store,world): replace `ResourceType` table with `ResourceId` table and corresponding checks

### LINKED ISSUE #1553: Remove `ResourceType` table and replace with checks based on `ResourceId`

alvrs:
Fixes #1553 

TODO:
- [ ] update deploy script to encode ids correctly


packages/store/src/StoreCore.sol
```diff
@@ -134,6 +131,11 @@ library StoreCore {
     string[] memory keyNames,
     string[] memory fieldNames
   ) internal {
+    // Verify the table ID is of type RESOURCE_TABLE
+    if (!tableId.isType(RESOURCE_TABLE)) {
```
holic, line 135: should this also include offchain tables?

holic: is it worth including a free function in `storeResourceTypes.sol` something like
```
function isTable(ResourceId resourceId) { ... }
```
that we can use here? or is the indirection too gassy?

alvrs: increases gas by 40, see https://github.com/latticexyz/mud/pull/1557/commits/ef89c43d3a9316bcfb523c752ef0c39a130f5cdd - not super significant, but still don't think it's worth the small increase in ergonomics (agree it would read nicer, sad that the Solidity compiler doesn't just optimize this away)

packages/store/src/IStoreErrors.sol
```diff
@@ -7,6 +7,7 @@ interface IStoreErrors {
   // Errors include a stringified version of the tableId for easier debugging if cleartext tableIds are used
   error StoreCore_TableAlreadyExists(ResourceId tableId, string tableIdString);
   error StoreCore_TableNotFound(ResourceId tableId, string tableIdString);
```
holic, line 9: nice, we already had this pattern for table errors

PR #1558, opened 2023-09-20
feat(store,world): replace `ephemeral` tables with `offchain` tables

### LINKED ISSUE #1123: Replace `ephemeral` tables with `offchain` tables
* Ephemeral tables are essentially just events that take advantage of the same network stack as tables. We should rename them to events to make their use case more clear.

Todos:
- separate from tables in config
- `StoreEphemeralRecord` -> `StoreEvent`
- `MyEvent.emit` instead of `MyEvent.emitEphemeral`
- Still use existing encoding/packing

alvrs:
Fixes #1123 
Replacement for #1496 (since we changed the approach)


packages/store/src/StoreCore.sol
```diff
@@ -302,6 +305,20 @@ library StoreCore {
...
+      keyTuple: keyTuple,
+      start: start,
+      deleteCount: deleteCount,
+      data: data
+    });
+
+    // Early return if the table is an offchain table
+    if (tableId.getType() == RESOURCE_OFFCHAIN_TABLE) {
+      return;
+    }
```
holic, line 320: do we support splicing for offchain tables?

holic (approved):
so glad to get rid of all the ephemeral specific stuff, this feels a lot cleaner!

but I am surprised gas went up across the board

alvrs:
> but I am surprised gas went up across the board

it's because we're doing an additional check in all store methods to early return if it's an offchain table, and that check costs 200 gas

packages/store/src/StoreCore.sol
```diff
@@ -688,6 +684,12 @@ library StoreCoreInternal {
     bytes memory data,
     PackedCounter previousEncodedLengths
   ) internal {
+    // Splicing dynamic data is not supported for offchain tables, because it
+    // requires reading the previous encoded lengths from storage
+    if (tableId.getType() == RESOURCE_OFFCHAIN_TABLE) {
+      revert IStoreErrors.StoreCore_InvalidResourceType(string(bytes.concat(RESOURCE_OFFCHAIN_TABLE)));
+    }
```
alvrs, line 691: re: no support for dynamic splice

packages/store/src/StoreCore.sol
alvrs: we do for static fields! Since splicing for static fields doesn't require reading anything from storage, can just support it for static fields without any further changes. For dynamic fields it's not supported because it would require reading the dynamic field's length from storage (but it's not stored in storage), see https://github.com/latticexyz/mud/pull/1558#discussion_r1332249965

Codegen doesn't render the dynamic field methods for offchain tables, but does render all other ones, so much more "real" tables than before with ephemeral

holic: ohh I see, I was thinking about push/pop but that's a dynamic field thing 👌

holic (approved):

.changeset/small-chicken-repair.md
```diff
@@ -0,0 +1,19 @@
...
+
+What used to be known as `ephemeral` table is now called `offchain` table.
+The previous `ephemeral` tables only supported an `emitEphemeral` method, which emitted a `StoreSetEphemeralRecord` event.
+
+Now `offchain` tables support all regular table methods, except partial operations on dynamic fields (`push`, `pop`, `update`).
+Unlike regular tables they don't store data on-chain but emit the same events as regular tables (`StoreSetRecord`, `StoreSpliceStaticData`, `StoreDeleteRecord`), so their data can be indexed by offchain indexers/clients.
+
+```diff
+- EphemeralTable.emitEphemeral(value);
++ OffchainTable.setRecord(value);
```
holic, line 18: should this be `OffchainTable.set(value)`?

.changeset/small-chicken-repair.md
```diff
@@ -0,0 +1,19 @@
+---
+"@latticexyz/block-logs-stream": patch
+"@latticexyz/cli": patch
+"@latticexyz/common": major
+"@latticexyz/dev-tools": patch
+"@latticexyz/store-sync": patch
+"@latticexyz/store": major
```
holic, line 7:
```suggestion
"@latticexyz/store": major
"create-mud": minor
```

.changeset/small-chicken-repair.md
alvrs, line 18:
```suggestion
+ OffchainTable.set(value);
```

holic (approved):

PR #1563, opened 2023-09-21
feat(world): prevent the `World` from calling itself

### LINKED ISSUE #1551: Remove access control case for World calling itself

alvrs:
fixes #1551 

The world should never call itself. All operations to internal tables should happen as internal library calls, and all calls to root system should happen as a delegatecall to the system.

If it was possible to make the `World` call itself, it would be possible to access internal tables that only the `World` should have access to. It should already not be possible to make the `World` call itself, but since this is a very important invariant, we decided to make it explicit and revert if `msg.sender` is `address(this)` in all `World` methods.


holic:
chatted IRL about the world calling itself and @alvrs is gonna play with adding some reverts to keep this from happening (if it's behavior we expect not to happen and ideally want to prevent)

alvrs:
> chatted IRL about the world calling itself and @alvrs is gonna play with adding some reverts to keep this from happening (if it's behavior we expect not to happen and ideally want to prevent)

The gas increase is insignificant, so I think it's worth it as an additional line of defence and to make this invariant more explicit

holic (dismissed):
Should fallback have this check too?

alvrs:
> Should fallback have this check too?

good catch, it should!

.changeset/brown-garlics-lie.md
```diff
@@ -0,0 +1,26 @@
+---
+"@latticexyz/world": major
+---
+
+All `World` methods now revert if the `World` calls itself.
+The `World` should never need to externally call itself, since all internal table operations happen via library calls, and all root system operations happen via delegate call.
+
+It should not be possible to make the `World` call itself as an external actor.
+If it were possible to make the `World` call itself, it would be possible to write to internal tables that only the `World` should have access to.
+As this is a very important invariance, we decided to codify it in an explicit requirement check in every `World` method, rather than of just relying on making it impossible to trigger the `World` to call itself.
```
alvrs, line 10:
```suggestion
As this is a very important invariance, we made it explicit in a requirement check in every `World` method, rather than just relying on making it impossible to trigger the `World` to call itself.
```

.changeset/brown-garlics-lie.md
```diff
@@ -0,0 +1,26 @@
...
+
+This is a breaking change for modules that previously used external calls to the `World` in the `installRoot` method.
+In the `installRoot` method, the `World` can only be called via `delegatecall`, and table operations should be performed via the internal table methods (e.g. `_set` instead of `set`).
+
+Example for how to replacing external calls to `world` in `installRoot` with `delegatecall`:
```
alvrs, line 15:
```suggestion
Example for how to replace external calls to `world` in `installRoot` with `delegatecall`:
```

packages/world/src/World.sol
```diff
@@ -72,12 +82,12 @@ contract World is StoreRead, IStoreData, IWorldKernel {
...
    * The module is delegatecalled and installed in the root namespace.
    */
-  function installRootModule(IModule module, bytes memory args) public {
+  function installRootModule(IModule module, bytes memory args) public requireNoCallback {
     AccessControl.requireOwner(ROOT_NAMESPACE_ID, msg.sender);
     _installRootModule(module, args);
   }
 
-  function _installRootModule(IModule module, bytes memory args) internal {
+  function _installRootModule(IModule module, bytes memory args) internal requireNoCallback {
```
holic, line 90: given that this is internal, do we need this check?

packages/world/src/modules/core/constants.sol
```diff
@@ -1,5 +1,14 @@
...
+
+import { ROOT_NAMESPACE } from "../../constants.sol";
+import { RESOURCE_SYSTEM } from "../../worldResourceTypes.sol";
+
 bytes16 constant CORE_MODULE_NAME = bytes16("core.m");
 bytes16 constant CORE_SYSTEM_NAME = bytes16("core.s");
+
+ResourceId constant CORE_SYSTEM_ID = ResourceId.wrap(
+  bytes32(abi.encodePacked(RESOURCE_SYSTEM, ROOT_NAMESPACE, CORE_SYSTEM_NAME))
+);
```
alvrs, line 14:
```suggestion
bytes16 constant CORE_MODULE_NAME = bytes16("core");

ResourceId constant CORE_SYSTEM_ID = ResourceId.wrap(
  bytes32(abi.encodePacked(RESOURCE_SYSTEM, ROOT_NAMESPACE, CORE_MODULE_NAME))
);
```

packages/world/src/modules/keyswithvalue/KeysWithValueModule.sol
```diff
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: MIT
 pragma solidity >=0.8.21;
 
+import { console } from "forge-std/console.sol";
+
```
alvrs, line 5:
```suggestion
```

alvrs, line 5:
```suggestion

```

packages/world/test/World.t.sol
```diff
@@ -689,27 +696,30 @@ contract WorldTest is Test, GasReporter {
...
-      Bool.getFieldLayout(),
-      defaultKeySchema,
-      Bool.getValueSchema(),
-      new string[](1),
-      new string[](1)
+      TwoFields.getFieldLayout(),
+      TwoFields.getKeySchema(),
+      TwoFields.getValueSchema(),
+      new string[](0),
+      new string[](2)
```
alvrs, line 703: the `Bool` table used `setField` internally because it only has a single field, but we're trying to test `setRecord` here

packages/world/src/World.sol
alvrs: nope, already checked by `installRootModule` and `installModule`, good catch

holic (approved):
A thing we might consider doing in a follow up is a TS test that parses the World contract, looks at all the public/external non-view/pure functions, and make sure they also use the modifier.

PR #1564, opened 2023-09-21
feat(world): prevent calls to self

alvrs:
Followup to #1563


alvrs:
Gonna combine this with #1563

PR #1566, opened 2023-09-21
feat(common,store): add support for user-defined types

dk1a:


packages/common/src/codegen/utils/extractUserTypes.ts
```diff
@@ -0,0 +1,57 @@
...
+ * @param userTypeNames names of the user types to extract
+ */
+export function extractUserTypes(
+  data: string,
+  userTypeNames: string[],
+  fromPath: string
+): Record<string, SolidityUserDefinedType> {
+  const ast = parse(data);
+
+  const isRelativePath = fromPath.indexOf(".") === 0;
```
alvrs, line 24: nit, don't need to check the whole string for index of `"."` if it's not at the first position
```suggestion
  const isRelativePath = fromPath.at(0) === ".";
```

alvrs (approved):
amazing!

PR #1567, opened 2023-09-21
fix(store,world): avoid `.unwrap()` for improved gas

alvrs:


alvrs:
gas report says it doesn't impact gas, so

PR #1568, opened 2023-09-21
refactor(store,world): prefix errors with library/contract name

### LINKED ISSUE #1126: Prefix all events and errors with library/contract name

alvrs:
partially fixes #1126

### DIFF
--- a/packages/world/src/interfaces/IModule.sol
+++ b/packages/world/src/interfaces/IModule.sol
@@ -10,7 +10,7 @@ bytes4 constant MODULE_INTERFACE_ID = IModule.getName.selector ^
   ERC165_INTERFACE_ID;
 
 interface IModule is IERC165 {
-  error RootInstallModeNotSupported();
+  error RootInstallNotSupported();
   error NonRootInstallNotSupported();
 
   /**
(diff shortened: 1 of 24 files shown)


alvrs:
should we rename store events to `Store_SetRecord` etc for consistency?

packages/world/src/interfaces/IModule.sol
```diff
@@ -10,7 +10,7 @@ bytes4 constant MODULE_INTERFACE_ID = IModule.getName.selector ^
   ERC165_INTERFACE_ID;
 
 interface IModule is IERC165 {
-  error RootInstallModeNotSupported();
+  error RootInstallNotSupported();
   error NonRootInstallNotSupported();
```
holic, line 14: should these be `World_` or `Module_` prefixed?

holic (approved):

alvrs (commented):
✅

PR #1571, opened 2023-09-22
WIP sky strife playtest

holic:
This is branched off of the commit Sky Strife deployed with and has some Sky Strife-specific adjustments to the indexer and sync stack to help the playtest run smoothly.

PR #1572, opened 2023-09-22
feat(store-indexer,store-sync): filter findAll by tableIds

holic:
pulled out of #1571


holic:
we should probably default-include all store/world tables since weird things happen if these are ignored

holic:
can also filter events by table IDs

holic:
can't filter logs by topics/table IDs yet: https://github.com/latticexyz/mud/issues/1658

packages/store-indexer/src/postgres/createQueryAdapter.ts
```diff
@@ -13,11 +14,11 @@ import { getAddress } from "viem";
...
-    async findAll(chainId, address) {
-      const internalTables = buildInternalTables();
-      const tables = (await getTables(database)).filter(
-        (table) => address != null && getAddress(address) === getAddress(table.address)
-      );
+    async findAll({ chainId, address, tableIds = [] }) {
+      const includedTableIds = new Set(tableIds.length ? [...internalTableIds, ...tableIds] : []);
+      const tables = (await getTables(database))
+        .filter((table) => address == null || getAddress(address) === getAddress(table.address))
+        .filter((table) => !includedTableIds.size || includedTableIds.has(table.tableId));
```
alvrs, line 21: general q, not related to this PR: should we combine this adapter with the one defined in `sqlite`? 

On the same note, a couple lines below this we have `const sqliteTable = builtTable(table)`, should that be `postgresTable`? Is it the same object type? (Can't comment directly on it bc it didn't change in this PR)

packages/store-sync/src/createStoreSync.ts
```diff
@@ -49,13 +50,15 @@ type CreateStoreSyncOptions<TConfig extends StoreConfig = StoreConfig> = SyncOpt
...
-  address,
   publicClient,
+  address,
+  tableIds = [],
   startBlock: initialStartBlock = 0n,
   maxBlockRange,
   initialState,
   indexerUrl,
 }: CreateStoreSyncOptions<TConfig>): Promise<SyncResult> {
+  const includedTableIds = new Set(tableIds.length ? [...internalTableIds, ...tableIds] : []);
```
alvrs, line 61: it looks like we actively pass `internalTableIds` here, but also always include it internally in the query adapter - doesn't really hurt bc it's a set, but feels like it could be confusing (e.g. someone removing `internalTableIds` here and then wondering why they still show up in the results). What are your thoughts on only putting it in one of the two places and making it an explicit assumption? (e.g. "`findAll` always returns internal tables", or "`findAll` returns only the tables passed in the query, it is recommended internal tables are always included in the query")

packages/store-sync/src/internalTableIds.ts
```diff
@@ -0,0 +1,23 @@
+import { resourceIdToHex } from "@latticexyz/common";
+import storeConfig from "@latticexyz/store/mud.config";
+import worldConfig from "@latticexyz/world/mud.config";
+
+// TODO: refactor config to include table IDs (https://github.com/latticexyz/mud/pull/1561)
+
+export const storeTableIds = Object.keys(storeConfig.tables).map((name) =>
+  resourceIdToHex({
+    type: storeConfig.tables[name as keyof typeof storeConfig.tables].offchainOnly ? "offchainTable" : "table",
```
alvrs, line 9: `as keyof typeof storeConfig.table` sad that typescript can't infer that from `Object.keys(storeConfig.tables)` 😢

alvrs (approved):
nothing blocking!

packages/store-indexer/src/postgres/createQueryAdapter.ts
holic: I would like to, but there's some API differences between drizzle's postgres and sqlite adapters that makes it hard to combine these.

Also the sqlite patterns here are a bit behind - I cleaned things up in the postgres, but didn't go back to refactor sqlite to match.

packages/store-sync/src/createStoreSync.ts
holic: I think making the indexer dumber (uses whatever you pass in for table IDs) makes sense, and only the store-sync includes internal table IDs.

packages/store-sync/src/internalTableIds.ts
holic: agreed! I think lodash has a thing for this, just waiting for them to finish their TS rewrite

PR #1574, opened 2023-09-22
feat(world): change registerFunctionSelector signature to accept system signature as a single string

alvrs:
- registerFunctionSelector now accepts the system function signature as a single string. There was no point in separating name and arguments.

PR #1575, opened 2023-09-22
feat(world): add `FunctionSignatures` offchain table

alvrs:
The FunctionSelectors offchain table now stores a mapping from function selector to signature for all methods registered in the World. This will enable the automatic generation of interfaces for a given world address in the future.


alvrs:
* main
  * **PR #1574** <a href="https://app.graphite.dev/github/pr/latticexyz/mud/1574" target="_blank">Graphite</a> 
    * **PR #1575** <a href="https://app.graphite.dev/github/pr/latticexyz/mud/1575" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/latticexyz/mud/1575?utm_source=stack-comment).

packages/world/src/interfaces/IWorldRegistrationSystem.sol
```diff
@@ -23,9 +23,9 @@ interface IWorldRegistrationSystem {
 
   function registerRootFunctionSelector(
```
holic, line 24: Should this be just “register root function”?

holic (commented):
Beaut!

holic (approved):

packages/world/src/interfaces/IWorldRegistrationSystem.sol
alvrs: i personally find it more intuitive to keep the selector, since what happens is that a new function selector is added to the world contract (it's debatable, but there isn't really a new function added to the world, the function is part of the system and it was already possible to call it before a function selector was registered for it)

holic: but we are effectively adding a function to the world, regardless of how it’s implemented

I kinda like that “register function” communicates “world is mutable/extendable”

but happy to keep as is if you feel strongly

alvrs: In my mine we're just adding a "link" to an existing function by adding a new "function selector" to the World here. The World was extended when the system was registered (and it's even cheaper to call a function via the World's `call` function than going the indirection over the "function selector link" in the World)

packages/world/src/modules/core/implementations/WorldRegistrationSystem.sol
```diff
@@ -179,12 +188,15 @@ contract WorldRegistrationSystem is System, IWorldErrors {
...
     ResourceId systemId,
-    bytes4 worldFunctionSelector,
+    string memory worldFunctionSignature,
     bytes4 systemFunctionSelector
-  ) public returns (bytes4) {
+  ) public returns (bytes4 worldFunctionSelector) {
     // Require the caller to own the root namespace
     AccessControl.requireOwner(ROOT_NAMESPACE_ID, _msgSender());
 
+    // Compute the function signature from the provided signature
```
alvrs, line 197:
```suggestion
    // Compute the function selector from the provided signature
```

alvrs (commented):
✅ (only gas report changed since last approval)

packages/world/src/interfaces/IWorldRegistrationSystem.sol
holic: ohh I see, we're aliasing a function's name? I thought this is what system registration used to enable you to call the system methods via the world

PR #1581, opened 2023-09-22
feat(store): prevent onchain state and indexer state from diverging

### LINKED ISSUE #1137: Revisit naming of core functions
Function names to revisit:
- `updateInField` (see https://github.com/latticexyz/mud/pull/525#discussion_r1150872608)
- `startByteIndex` (see https://github.com/latticexyz/mud/pull/525#discussion_r1150987039)
- `TightCoder`

anything else?

Please leave comments here for alternative names, and let's set ourselves a deadline after which we commit to whatever name we have (I propose Aug 15th 2023)

### LINKED ISSUE #1449: make sure there is no way to make onchain state and indexer state diverge via exposed store methods
and remove onchain validation of fieldLayout in methods where it doesn't matter

### LINKED ISSUE #1523: Revisit argument order on core functions
- eg should `FieldLayout` be the last argument or not? https://github.com/latticexyz/mud/pull/1521#discussion_r1328005455

alvrs:
Fixes #1449 
Fixes #1137 
Fixes #1523 

Details on the change in the changeset.

<details>
<summary>More context</summary>

- `setRecord` static data, dynamic data lengths, dynamic datas
    - The event includes the entire data that is being written to storage (static, packed lengths, dynamic data)
    - For static data the onchain way of handling it is the same as offchain
    - The provided field layout is used for separating the provided data blob.
        - If the `numDynamicFields` value in `fieldLayout` is invalid (e.g. 0), the indexer would override the dynamic field data, while it would stay its previous value onchain.
            - we could solve this by either loading `fieldLayout` from storage instead of accepting it as a passed in parameter
            - or by emitting `fieldLayout` as part of the event, so indexers can imitate the onchain behavior
    - We “validate the input” onchain, but can actually remove that check, because all we’re testing is
        - provided static data matches the field layout static data length (doesn’t matter, if data is too long or too short it’s handled the same onchain and offchain)
        - check that the encoded total length matches the dynamic data length
            - i’m not sure if we currently do this, but it would be possible to append 0 to the dynamic data blob to mirror the onchain behavior. In this case we don’t need the onchain check
- `spliceStaticData` static data
    - onchain deleteCount is always `data.length`, so we should imitate that behavior offchain
    - no need to include `deleteCount` in the event or function signature
- `deleteRecord` static data, dynamic data lengths
    - the user provided field layout determines how much of the static data is actually deleted
        - to solve this, either emit the field layout as part of the event and handle the same way as onchain, or load the field layout from storage
    - the user provided field layout determines whether or not we set the dynamic lengths to 0 (= deleting the dynamic data)
        - also, since we don’t actually set the dynamic data to 0, we get invalid data onchain in `getFieldSlice` after deleting the data
            - this can be solved by adding a bound check for reading slices of fields
            - alternatively we could make it more explicit that users have to check the length before reading from an index, but that would be more expensive (if they decide to do it)
- `spliceDynamicData` dynamic data length, dynamic data ✅
    - start depends on user provided field, previous encoded lengths, and startWithin field
        - we load `previousEncodedLengths` from storage, so can’t be tampered with
        - `startWithinField` is what we use onchain, and we use the provided field index to find the right storage slot
    - we make sure onchain that `deleteCount` is at the end of the field to guarantee the indexer doesn’t cut into another field or shifts data at the end of the field
    - the updated encoded length is emitted with the event and stored onchain
</details>


packages/store/gas-report.json
```diff
@@ -363,19 +363,19 @@
...
     "name": "register Mixed table",
-    "gasUsed": 582081
+    "gasUsed": 582351
   },
   {
     "file": "test/Mixed.t.sol",
     "test": "testSetAndGet",
     "name": "set record in Mixed",
-    "gasUsed": 103978
+    "gasUsed": 104908
```
alvrs, line 372: gas only increased around 1k for set record here (i expected 2k for the additional sload, but seems like some of it was offset by less memory being passed around between functions and the optimized `getFieldLayout`)

.changeset/wicked-squids-do.md
```diff
@@ -0,0 +1,131 @@
...
+      uint48 start,
+  -   uint40 deleteCount,
+      bytes calldata data
+    ) external;
+  }
+
+  ```
+
+- All methods that are only valid for dynamic fields (`pushToField`, `popFromField`, `updateInField`, `getFieldSlice`)
+  have been renamed to make this more explicit (`pushToDynamicField`, `popFromDynamicField`, `updateInDynamicField`, `getDynamicFieldSlice`).
```
holic, line 62: should we make `updateInDynamicField` more like a `spliceDynamicField`?

.changeset/wicked-squids-do.md
```diff
@@ -0,0 +1,131 @@
...
+  +   uint8 dynamicFieldIndex
+  + ) external view returns (uint256);
+  }
+
+- `IStore` now has additional overloads for `getRecord`, `getField`, `getFieldLength` and `setField` that don't require a `FieldLength` to be passed, but instead load it from storage.
+
+- `IStore` now exposes `setStaticField` and `setDynamicField` to save gas by avoiding the dynamic inference of whether the field is static or dynamic.
+
+- The `getDynamicFieldSlice` method no longer accepts reading outside the bounds of the dynamic field.
+  This is to avoid returning invalid data, as the data of a dynamic field is not deleted when the record is deleted, but only its length is set to zero.
```
holic, line 129: I'm curious about the gas difference between actually deleting and setting length to zero - isn't there a gas return on storage slot deletion?

docs/pages/store/spec.mdx
```diff
@@ -8,7 +8,7 @@ When a record or a single field is edited, or when a record is deleted, Store em
 
 ```solidity
 event Store_SetRecord(bytes32 indexed tableId, bytes32[] keyTuple, bytes staticData, bytes32 encodedLengths, bytes dynamicData),
-event Store_SpliceStaticData(bytes32 indexed tableId, bytes32[] keyTuple, uint48 start, uint40 deleteCount, bytes data),
+event Store_SpliceStaticData(bytes32 indexed tableId, bytes32[] keyTuple, uint48 start, bytes data),
```
holic, line 11: sorry lermchair

.changeset/wicked-squids-do.md
alvrs: there is a `spliceDynamicField` function too! `updateInDynamicField` was there before but now basically just calls `spliceDynamicField` with the correct `deleteCount` value and does a range check. In fact the range check is not even necessary, since `_spliceDynamicField` does a range check already. Could argue `updateInDynamicField` not really necessary.

packages/common/src/getByteLength.ts
```diff
@@ -0,0 +1,6 @@
+import { Hex } from "viem";
+
+export function getByteLength(data: Hex): number {
```
holic, line 3: viem has `size` for this purpose: https://github.com/wagmi-dev/viem/blob/main/src/utils/data/size.ts#L12

packages/world/gas-report.json
```diff
@@ -285,91 +285,91 @@
...
-    "test": "testPopFromField",
+    "test": "testPopFromDyanmicField",
     "name": "pop 1 address (cold)",
-    "gasUsed": 24499
+    "gasUsed": 23714
   },
   {
     "file": "test/WorldDynamicUpdate.t.sol",
-    "test": "testPopFromField",
+    "test": "testPopFromDyanmicField",
```
holic, line 358:
```suggestion
    "test": "testPopFromDynamicField",
```

packages/world/gas-report.json
```diff
@@ -285,91 +285,91 @@
...
     "file": "test/World.t.sol",
     "test": "testSetRecord",
     "name": "Write data to the table",
-    "gasUsed": 36340
+    "gasUsed": 36970
   },
   {
     "file": "test/WorldDynamicUpdate.t.sol",
-    "test": "testPopFromField",
+    "test": "testPopFromDyanmicField",
```
holic, line 352:
```suggestion
    "test": "testPopFromDynamicField",
```

.changeset/wicked-squids-do.md
holic: yeah I am leaning towards us removing, the name was always confusing to me

.changeset/wicked-squids-do.md
alvrs: I thought so too and actually implemented the actual removal today. Surprisingly setting it to 0 was _a lot_ more expensive, and the cost grew linearly. Eventually I realized that you still have to provide the entire gas for the transaction to complete, and _at the end_ of the tx receive up to 20% refund.

packages/common/src/getByteLength.ts
alvrs: oh nice, i figured there must exist something for a super general util like this one

.changeset/wicked-squids-do.md
alvrs: See https://github.com/latticexyz/mud/pull/1581/commits/004fe9ce9ec3012429a498de0a121a32b9639841 and the following 3 commits

holic: sorry ethereum for the state bloat we're about to create

alvrs: that's what ethereum gets for being so stingy with the refund

holic (approved):

PR #1583, opened 2023-09-23
WIP skystrife playtest

holic:


Kooshaba:
it seems like this breaks initial sync from RPC, noticed this in dev

if you run `pnpm dev` at sky strife root and then visit `localhost:1337/match` you'll get to the admin match page. you should be able to hit Create Match and the `debug` map should be there, but it isn't

turning on debugger in console shows a bunch of `Skipping update because we don't know about this table` messages

i looked into it a little bit and i believe table registration is only happening given the list of tables from the indexer snapshot, but i wasn't able to find where it was happening previously

holic:
> it seems like this breaks initial sync from RPC, noticed this in dev
> 
> if you run `pnpm dev` at sky strife root and then visit `localhost:1337/match` you'll get to the admin match page. you should be able to hit Create Match and the `debug` map should be there, but it isn't
> 
> turning on debugger in console shows a bunch of `Skipping update because we don't know about this table` messages
> 
> i looked into it a little bit and i believe table registration is only happening given the list of tables from the indexer snapshot, but i wasn't able to find where it was happening previously

this was an issue with table ID formatting in sky strife repo, added some commits to fix it

PR #1585, opened 2023-09-24
fix(common): always import relative sol files from ./

alvrs:
- solc doesn't like relative imports without `./`
- solc is fine with `./../`

### DIFF
--- a/packages/common/src/codegen/utils/loadUserTypesFile.ts
+++ b/packages/common/src/codegen/utils/loadUserTypesFile.ts
@@ -31,8 +31,9 @@ function loadUserTypesFile(
   data: string;
 } {
   if (unresolvedFilePath.at(0) === ".") {
+    const relativePath = path.relative(outputBaseDirectory, unresolvedFilePath);
     return {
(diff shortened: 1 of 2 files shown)


packages/common/src/codegen/utils/loadUserTypesFile.ts
```diff
@@ -31,8 +31,9 @@ function loadUserTypesFile(
   data: string;
 } {
   if (unresolvedFilePath.at(0) === ".") {
+    const relativePath = path.relative(outputBaseDirectory, unresolvedFilePath);
     return {
-      filePath: path.relative(outputBaseDirectory, unresolvedFilePath),
+      filePath: "./" + relativePath, // solc doesn't like relative paths without "./"
```
holic, line 36: is this gonna muck with windows at all?

holic (approved):

packages/common/src/codegen/utils/loadUserTypesFile.ts
alvrs: no, it's just relevant for imports in solidity files, which are processed by `solc`. We do the same for other relative imports in table libraries already.

PR #1561, opened 2023-09-21
WIP config parser

holic:
I started playing around with #994 and at the time was trying to refactor some of our config parsing, but realized I was fighting Zod most of the time. I didn't realize it until well into refactoring that Zod doesn't carry through strong types (i.e. string literals from const types), so you have to both define the Zod type/validation/parsing and the TS transformation.

This seemed like a lot more than we need right now, especially if we feel good about leaning on TS for most of the validation, so I started a new config parser that is pure TS and carries strong types. It's made up of small composable pieces that should hopefully make it easier to maintain, extend, etc.

I'm sure we can extract some common patterns into a Zod-like interface, but we can figure that out later.

TODO:
- [ ] figure out why the union of config root-level tables and namespace-level tables don't intersect properly in some cases
- [ ] figure out what to do about tables with the same name but diff namespaces (this doesn't work well when organizing config output by table key)
- [ ] prototype plugin system with input/output interfaces

PR #1586, opened 2023-09-24
feat(store,world): use user-types for `ResourceId`, `FieldLayout` and `Schema` in table libraries

alvrs:
Thanks to #1566 we can now use the real user-types in table libraries and don't need to manually `wrap`/`unwrap` everywhere


packages/store/gas-report.json
```diff
@@ -603,73 +603,73 @@
     "file": "test/StoreCoreDynamic.t.sol",
     "test": "testGetDynamicFieldSlice",
     "name": "get field slice (cold, 1 slot)",
-    "gasUsed": 10240
+    "gasUsed": 5609
```
alvrs, line 606: these gas changes come from not loading the `_tableId` and `_keyTuple` from storage in the gas test

holic: same for all tests or just this one? I noticed some other gas improvements below

holic (dismissed):
So nice!

packages/store-sync/src/common.ts
```diff
@@ -76,7 +76,12 @@ export type StorageAdapterBlock = { blockNumber: BlockLogs["blockNumber"]; logs:
 export type StorageAdapter = (block: StorageAdapterBlock) => Promise<void>;
 
 // TODO: adjust when we get namespace support (https://github.com/latticexyz/mud/issues/994) and when table has namespace key (https://github.com/latticexyz/mud/issues/1201)
-export const schemasTable = storeConfig.tables.Tables;
+// TODO: adjust when schemas are automatically resolved
+export const schemasTable = {
+  ...storeConfig.tables.Tables,
+  valueSchema: resolveUserTypes(storeConfig.tables.Tables.valueSchema, storeConfig.userTypes),
+};
+
```
alvrs, line 84: kind of annoying, but we need this workaround for now to land the contract changes (until we have a better way to resolve the user type config with strong types)

packages/store/test/StoreCoreDynamic.t.sol
```diff
@@ -96,41 +98,43 @@ contract StoreCoreDynamicTest is Test, GasReporter, StoreMock {
...
-    assertEq(StoreCore.getField(_tableId, _keyTuple, 2, fieldLayout), thirdDataBytes);
+    assertEq(bytes32(StoreCore.getField(tableId, keyTuple, 0, fieldLayout)), firstDataBytes);
+    assertEq(StoreCore.getField(tableId, keyTuple, 2, fieldLayout), thirdDataBytes);
   }
 
   function testPopFromThirdField() public {
     FieldLayout fieldLayout = StoreCore.getFieldLayout(_tableId);
     bytes memory dataBytes = thirdDataBytes;
+    ResourceId tableId = _tableId;
+    bytes32[] memory keyTuple = _keyTuple;
```
dk1a, line 137: we could use immutable too, it's about as convenient as storage, but without the gas overhead. It's only useful for constant-like values tho

dk1a (dismissed):

packages/store-sync/src/common.ts
holic: hopefully coming in #1561

holic (dismissed):

packages/store/gas-report.json
alvrs: `wrapping`/`unwrapping` in itself should not significantly impact gas, so i think most of the gas variation in here comes from the changed gas tests

PR #1591, opened 2023-09-24
refactor(world,world-modules): move optional modules from `world` to `world-modules` package

alvrs:
Moving all modules except the core module from `world` to `world-modules` to slim down the `world` package


.changeset/unlucky-cups-fetch.md
```diff
@@ -0,0 +1,8 @@
+---
+"@latticexyz/cli": path
```
holic, line 2:
```suggestion
"@latticexyz/cli": patch
```

.changeset/pre.json
```diff
@@ -29,7 +29,8 @@
     "@latticexyz/store-sync": "1.42.0",
     "@latticexyz/utils": "1.42.0",
     "@latticexyz/world": "1.42.0",
-    "@latticexyz/abi-ts": "2.0.0-next.5"
+    "@latticexyz/abi-ts": "2.0.0-next.5",
+    "@latticexyz/world-modules": "1.42.0"
```
holic, line 33: iirc I didn't have to add this manually for abi-ts, and the reason abi-ts has the version that it has is because the next prerelease set it for us

packages/world-modules/package.json
```diff
@@ -0,0 +1,57 @@
...
+    "ds-test": "https://github.com/dapphub/ds-test.git#e282159d5170298eb2455a6c05280ab5a73a4ef0",
+    "ejs": "^3.1.8",
+    "forge-std": "https://github.com/foundry-rs/forge-std.git#74cfb77e308dd188d2f58864aaf44963ae6b88b1",
+    "glob": "^8.0.3",
+    "solhint": "^3.3.7",
+    "tsup": "^6.7.0",
+    "tsx": "^3.12.6",
+    "vitest": "0.31.4"
+  },
+  "gitHead": "914a1e0ae4a573d685841ca2ea921435057deb8f"
```
holic, line 56:
```suggestion
  "publishConfig": {
    "access": "public"
  },
  "gitHead": "914a1e0ae4a573d685841ca2ea921435057deb8f"
```

packages/world/package.json
```diff
@@ -63,7 +63,6 @@
     "ejs": "^3.1.8",
     "forge-std": "https://github.com/foundry-rs/forge-std.git#74cfb77e308dd188d2f58864aaf44963ae6b88b1",
     "glob": "^8.0.3",
-    "hardhat": "^2.10.2",
```
holic, line 66: iirc this was needed for solhint?

if we don't actually need this then we should prob get rid of a lot more stuff (tasks dir etc)

holic (dismissed):
lgtm aside from a few small things

packages/world/package.json
alvrs: it seems like `solhint` does the same with and without hardhat, interestingly it doesn't find actual errors (like missing semicolon) in either case

PR #1593, opened 2023-09-24
docs(indexer): better way to ensure compatibility with app

qbzzt:


qbzzt:
Superceded by the newdocs

PR #1594, opened 2023-09-24
feat(world): add `batchCallFrom`

### LINKED ISSUE #1582: add `batchCallFrom` to `BatchCallSystem`
an extension of #1500

alvrs:
Fixes #1582


packages/world/src/modules/core/implementations/BatchCallSystem.sol
```diff
@@ -0,0 +1,44 @@
...
+      if (!success) revertWithBytes(returnData);
+
+      returnDatas[i] = abi.decode(returnData, (bytes));
+    }
+  }
+
+  /**
+   * Batch calls to multiple systems into a single transaction, return the array of return data.
+   */
+  function batchCallFrom(SystemCallFromData[] calldata systemCalls) public returns (bytes[] memory returnDatas) {
```
holic, line 31: probably best to go with this approach for extensibility/flexibility, but I wonder how often we'll want to batch call from different addresses vs the signature being

```solidity
function batchCallFrom(address from, SystemCallData[] calldata systemCalls)
```

holic (dismissed):

packages/world/src/modules/core/implementations/BatchCallSystem.sol
alvrs: to get that you can do this:

```solidity
world.callFrom(from, coreSystemId, abi.encodeCall(BatchCallSystem.batchCall, (systemCalls)));
```

PR #1598, opened 2023-09-24
fix(common): fix memory corruption for dynamic to static array conversion

### LINKED ISSUE #1005: Possible array to static array memory corruption
Reminder to self to investigate a potential memory corruption issue during the cast of an **uninitialized** dynamic array to a static array within autogenerated tables

dk1a:
fixes #1005

### DIFF
--- a/packages/cli/contracts/src/codegen/tables/Dynamics1.sol
+++ b/packages/cli/contracts/src/codegen/tables/Dynamics1.sol
@@ -1495,37 +1495,67 @@ library Dynamics1 {
 }
 
 function toStaticArray_bytes32_1(bytes32[] memory _value) pure returns (bytes32[1] memory _result) {
-  // in memory static arrays are just dynamic arrays without the length byte
-  assembly {
-    _result := add(_value, 0x20)
+  if (_value.length < 1) {
+    // ignore invalid dynamic arrays that are too small
+    return _result;
+  } else {
+    // in memory static arrays are just dynamic arrays without the length byte
+    // (without the length check this could lead to memory corruption)
+    assembly {
+      _result := add(_value, 0x20)
+    }
   }
 }
 
 function toStaticArray_int32_2(int32[] memory _value) pure returns (int32[2] memory _result) {
-  // in memory static arrays are just dynamic arrays without the length byte
-  assembly {
-    _result := add(_value, 0x20)
+  if (_value.length < 2) {
+    // ignore invalid dynamic arrays that are too small
+    return _result;
+  } else {
+    // in memory static arrays are just dynamic arrays without the length byte
+    // (without the length check this could lead to memory corruption)
+    assembly {
+      _result := add(_value, 0x20)
+    }
   }
 }
 
 function toStaticArray_uint128_3(uint128[] memory _value) pure returns (uint128[3] memory _result) {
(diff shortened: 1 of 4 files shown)


packages/common/src/codegen/render-solidity/renderTypeHelpers.ts
```diff
@@ -63,9 +63,15 @@ function renderWrapperStaticArray(
...
-        _result := add(_value, 0x20)
+      if (_value.length < ${staticLength}) {
+        // ignore invalid dynamic arrays that are too small
+        return _result;
+      } else {
+        // in memory static arrays are just dynamic arrays without the length byte
+        // (without the length check this could lead to memory corruption)
+        assembly {
+          _result := add(_value, 0x20)
+        }
```
holic, line 74: any way we can add a regression test for this to make sure we don't run into this in the future?

packages/cli/contracts/src/codegen/tables/Dynamics1.sol
```diff
@@ -1495,37 +1495,67 @@ library Dynamics1 {
 }
 
 function toStaticArray_bytes32_1(bytes32[] memory _value) pure returns (bytes32[1] memory _result) {
-  // in memory static arrays are just dynamic arrays without the length byte
-  assembly {
-    _result := add(_value, 0x20)
+  if (_value.length < 1) {
+    // ignore invalid dynamic arrays that are too small
+    return _result;
```
holic, line 1500: if this is invalid, should this revert instead of returning?

dk1a: a 0-length array is kinda valid (it's uninitialized), so maybe I should change the comment?
and other cases are weird and shouldn't happen, but it's cheaper to return empty than add more checks

packages/common/src/codegen/render-solidity/renderTypeHelpers.ts
dk1a: memory corruption tests are tricky, I think I'm too sleepy to write 1 rn, we can add an issue for later, tests aren't production code so does freeze affect em?

alvrs: we can always add more tests later! planning to do extend the test suite in general next week

packages/common/src/codegen/render-solidity/renderTypeHelpers.ts
```diff
@@ -63,9 +63,15 @@ function renderWrapperStaticArray(
     ) pure returns (
       ${elementType}[${staticLength}] memory _result
     ) {
-      // in memory static arrays are just dynamic arrays without the length byte
-      assembly {
-        _result := add(_value, 0x20)
+      if (_value.length < ${staticLength}) {
+        // ignore invalid dynamic arrays that are too small
```
alvrs, line 67:
```suggestion
        // return an uninitialized array if the length is smaller than the fixed length to avoid memory corruption
```

packages/common/src/codegen/render-solidity/renderTypeHelpers.ts
holic: added here: https://github.com/latticexyz/mud/issues/1603

PR #1600, opened 2023-09-25
refactor(store,world): move test table config out of main table config

### LINKED ISSUE #1309: move test tables out of store/world mud config
Apps shouldn't be deploying our test tables like `Mixed` and `Vector2` and `AddressArray` such. These should get moved into some test dir.

https://github.com/latticexyz/mud/blob/5294a7d5983c52cb336373566afd6a8ec7fc4bfb/packages/store/mud.config.ts#L22-L46
https://github.com/latticexyz/mud/blob/5294a7d5983c52cb336373566afd6a8ec7fc4bfb/packages/store/mud.config.ts#L22-L46

alvrs:
Fixes #1309 
- Cleaned up store's and world's `mud.config.ts` by moving out the test tables into their own config


alvrs:
self-approving because only low risk changes to test setup / codegen location

PR #1602, opened 2023-09-25
refactor(store,world): move around interfaces and base contracts

holic:
torn on whether to put interface files into an `interfaces` dir (I tend to prefer not doing this but we do this in world already and I'd like to be consistent)


alvrs (approved):

alvrs:
> torn on whether to put interface files into an interfaces dir (I tend to prefer not doing this but we do this in world already and I'd like to be consistent)

Agree with the point on consistency, moved tables to the same folder structure in both for the same reason

PR #1612, opened 2023-09-26
docs: update onchain walkthrough to next.9

qbzzt:


qbzzt:
Not needed anymore, updated to `next.11` as part of doing Docs 2.0.

PR #1618, opened 2023-09-26
feat(world): more nuanced system hooks per function selector

alvrs:
- System hooks are now registered per function on the system instead of on the system as a whole


alvrs:
cross posting from discord:
unfortunately adding this change increases gas surprisingly significantly, so i wonder if it's still worth it. It's already possible to get "system hooks per function" by only having a single function per system (e.g. the fallback function for extra gas efficiency), or by a layer on top of system hooks (i.e. SystemFunctionHook is SystemHook) and i'm not sure if a gas increase of ~500 per system call (in any case, even with no hooks on the system installed) is worth it for this feature

holic:
closing this for now since 1) significantly increases gas and 2) code is frozen for audit

PR #1622, opened 2023-09-27
ci: replace snapshot action with changeset cli commands

alvrs:
In an attempt to fix the broken snapshot action


alvrs:
Seems to work as expected: https://github.com/latticexyz/mud/actions/runs/6324842597

holic (approved):

PR #1624, opened 2023-09-27
refactor(world): expose library for WorldContextConsumer

alvrs:
Minor non-breaking change to move the logic of `WorldContextConsumer` that was previously inlined in the contract to a library, so things like `_msgSender()` can be used in internal libraries in the `System` contracts.

This is not a breaking change and doesn't modify any logic.


holic (approved):

.changeset/olive-foxes-shop.md
```diff
@@ -0,0 +1,7 @@
+---
+"@latticexyz/world-modules": patch
+"@latticexyz/world": minor
```
alvrs, line 3:
```suggestion
"@latticexyz/world": major
```

Issue #1006, opened 2023-06-08
Refactor deploy script to use viem


alvrs:
Currently the [deploy script](https://github.com/latticexyz/mud/blob/main/packages/cli/src/commands/deploy.ts) is using ethers to send transactions for deploying contracts and registering tables / systems / modules etc.

We're moving towards viem (see https://github.com/latticexyz/mud/issues/582) and should also use viem instead of ethers here too (eg. for more expressive errors in case a transaction fails).

We should also modularize this script, it's currently pretty spaghetti.

holic:
done in https://github.com/latticexyz/mud/pull/1702

Issue #1611, opened 2023-09-26
declarative deployments


holic:
Given a mud config, systems, etc. and an optional world address, we should be able to figure out what needs to be deployed (missing tables, updated systems, etc.) and turn our deploy process from imperative to declarative.

ludns:
i think it would be good to have an hour long jam on this (along with maybe a larger tooling discussion)

holic:
we should probably do #1006 and #1636 first

ludns:
I think #1636 <https://github.com/latticexyz/mud/issues/1636> is easier
when we have the primitives for declarative deployment.

holic:
done in https://github.com/latticexyz/mud/pull/1702

PR #1625, opened 2023-09-27
refactor(cli): replace ethers with viem

johngrantuk:
closes https://github.com/latticexyz/mud/issues/1006


packages/cli/src/utils/deploy.ts
```diff
@@ -17,12 +16,27 @@ import { getRegisterTableCallData } from "./tables/getRegisterTableCallData";
 import { getTableIds } from "./tables/getTableIds";
 import { confirmNonce } from "./utils/confirmNonce";
 import { deployContract } from "./utils/deployContract";
-import { fastTxExecute } from "./utils/fastTxExecute";
+import { fastTx } from "./utils/fastTxExecute";
```
holic, line 19: Might see if we can use our internal createContract helper, since it has its own nonce tracking

johngrantuk: I did originally try that but came across issues when it was sending a bunch of txs. I could have been doing something wrong though so would definitely be interested to see how to do it correctly.

holic: oh interesting! thanks for trying, happy to go with this and revisit in the future

packages/cli/package.json
```diff
@@ -57,6 +57,7 @@
     "path": "^0.12.7",
     "throttle-debounce": "^5.0.0",
     "typescript": "5.1.6",
+    "viem": "1.12.2",
```
holic, line 60: can we align this with the version we use everywhere else? `1.6.0`

packages/cli/src/utils/systems/utils.ts
```diff
@@ -1,14 +1,26 @@
...
+type Param = AbiParameter & { components: any[] };
 
 export function loadFunctionSignatures(contractName: string, forgeOutDirectory: string): string[] {
   const { abi } = getContractData(contractName, forgeOutDirectory);
 
   return abi
     .filter((item) => ["fallback", "function"].includes(item.type))
     .map((item) => {
-      return `${item.name}${parseComponents(item.inputs)}`;
+      let name: string;
```
johngrantuk, line 13: I'm not sure if this is the best approach?

holic (dismissed):

johngrantuk:
> I just spent a bit trying to remove the `chain` requirement, because folks will often deploy with just an RPC URL, and they'll have no way to edit the `supportedChains` of the built CLI package.
> 
> While messing around with that, I turned on the `--debug` flag and discovered that we're getting _tons_ of nonce errors + retries with the approach in this PR. It does eventually seem to deploy properly (see passing tests), but I feel like we should handle nonces better, especially for non-local chains.
> 
> I think I'm gonna move on to working on #1611 and overhaul our deployment process. So I'll leave it to you to decide if you want to figure out nonce handling here or if we wanna just wait for handling this better in the declarative deployment approach.

No worries. TBH I'm not sure how its possible to use the manual nonce setting with Viem without this approach as I couldn't find a way to avoid the nonce errors (as far as I could see createContract was essentially doing the same but was just resubmitting with latest nonce instead of the original, which I think ends up causing issues when lots of transactions are being handled). I also tried with a que type setup where it would submit one after the other but this was way slower and AFAIK speed is one of the main priorities here.

PR #1635, opened 2023-09-28
feat(services): remove all but faucet client

holic:
to ease MUD repo install

we still need the faucet client until we have a new faucet tRPC service spun up on our infra


holic:
tested that the faucet client still works via testnet deploy of the react template

alvrs:
🔥🔥🔥 
CleanShot 2023-09-28 at 15 40 18@2x

alvrs (approved):

alvrs:
can we remove installing `protoc` from the CI setup action?

PR #1665, opened 2023-09-29
feat(world-modules): add SystemSwitch util

alvrs:
- util for better ergonomics of calling systems from other systems


packages/world-modules/src/modules/utils/SystemSwitch.sol
```diff
@@ -0,0 +1,80 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.21;
```
holic, line 2: talked IRL but can we move this to `src/utils` instead of `stc/modules/utils` since it's not a module?

.changeset/modern-stingrays-kneel.md
```diff
@@ -0,0 +1,17 @@
...
+The recommended approach is to use `delegatecall` to the system if in the context of a root system, and an external call via the World if in the context of a non-root system.
+To bring back the developer experience of calling systems from other sysyems without caring about the context in which the call is executed, we added the `SystemSwitch` util.
+
+```diff
+- // Instead of calling the system via an external call to world...
+- IBaseWorld(_world()).callMySystem();
+
++ // ...you can now use the `SystemSwitch` util.
+* // This works independent of whether used in a root system or non-root system.
++ SystemSwitch.call(abi.encodeCall(IBaseWorld.callMySystem, ());
```
holic, line 16: does this need `abi.decode` too?

holic (approved):

holic:
since we've had a few folks ask about this, we should prob get @qbzzt to add a docs page about systems-calling-systems

.changeset/modern-stingrays-kneel.md
alvrs: yes, if the call returns a value. Will change in the changeset example.

alvrs, line 16:
```suggestion
- // Instead of calling the system via an external call to world...
- uint256 value = IBaseWorld(_world()).callMySystem();

+ // ...you can now use the `SystemSwitch` util.
* // This works independent of whether used in a root system or non-root system.
+ uint256 value = abi.decode(SystemSwitch.call(abi.encodeCall(IBaseWorld.callMySystem, ()), (uint256)); 
```

.changeset/modern-stingrays-kneel.md
```diff
@@ -0,0 +1,19 @@
...
+This made the developer experience of calling other systems via root systems worse, since calls from root systems are executed from the context of the World.
+The recommended approach is to use `delegatecall` to the system if in the context of a root system, and an external call via the World if in the context of a non-root system.
+To bring back the developer experience of calling systems from other sysyems without caring about the context in which the call is executed, we added the `SystemSwitch` util.
+
+```diff
+- // Instead of calling the system via an external call to world...
+- uint256 value = IBaseWorld(_world()).callMySystem();
+
++ // ...you can now use the `SystemSwitch` util.
+* // This works independent of whether used in a root system or non-root system.
```
holic, line 15: what's the * here for?

alvrs, line 15:
```suggestion
+ // This works independent of whether used in a root system or non-root system.
```

alvrs: typo! should be `+`

PR #1673, opened 2023-09-30
WIP indexer SSE prototype

holic:

PR #1674, opened 2023-09-30
fix(protocol-parser): allow arbitrary key order when encoding values

holic:


packages/protocol-parser/src/encodeValueArgs.ts
```diff
@@ -8,15 +15,17 @@ export function encodeValueArgs<TSchema extends ValueSchema>(
   valueSchema: TSchema,
   value: SchemaToPrimitives<TSchema>
 ): ValueArgs {
-  const staticFields = Object.values(valueSchema).filter(isStaticAbiType);
-  const dynamicFields = Object.values(valueSchema).filter(isDynamicAbiType);
+  const valueSchemaEntries = Object.entries(valueSchema);
+  const staticFields = valueSchemaEntries.filter(([, type]) => isStaticAbiType(type)) as [string, StaticAbiType][];
+  const dynamicFields = valueSchemaEntries.filter(([, type]) => isDynamicAbiType(type)) as [string, DynamicAbiType][];
+  // TODO: validate <=5 dynamic fields
+  // TODO: validate <=28 total fields
```
holic, line 22: we might want to do this via a `strict` parameter and only use it when writing to the chain

encoding without these limits is useful for #1673

alvrs (approved):

PR #1683, opened 2023-10-02
feat(cli): add batch call for function registering

### LINKED ISSUE #1645: use batch call for registration methods in deploy
like #1470 but using features added in #1582 

cc @CrazyNorman

CrazyNorman:
I plan to first implement batch calling for the function registration part in the deploy. If everything works well, I can then proceed to finish the remaining parts.

One tricky parameter is the BATCH_SIZE. It may vary based on the length of function names and the gas limit. I've chosen a conservative value for now, but I believe this can be made configurable in a certain place eventually.

fix #1645


holic:
Going to close this for now to clean up some of our open PR backlog since this is a diff against our old deploy code (overhauled in https://github.com/latticexyz/mud/pull/1702)

Feel free to re-open or start a new PR when you get to this!

Issue #1334, opened 2023-08-21
gas-report: make it work within the context of `MudTest`


holic:
if you run `mud-gas-report` on a thing that extends `MudTest`, it'll fail because it's missing a `.mudtest` file (created on MUD deploy)

since our deploy pipeline is separate (a TS script) from the rest of the foundry lifecycle, I think we should:
- make gas-report (or sub command of) more of a pure "test output parser" rather than a "test runner"
- allow overriding the test command used in the gas-report test runner

another wild idea: get `MudTest` to call out to TS deploy script during the test run, but not sure if there's an exposed RPC within the test runner we can thread through to the deploy process via CLI arg

PR #1688, opened 2023-10-03
feat(gas-reporter): allow gas-reporter to parse stdin

alvrs:
fixes https://github.com/latticexyz/mud/issues/1334

### DIFF
--- a/packages/gas-report/ts/index.ts
+++ b/packages/gas-report/ts/index.ts
@@ -31,6 +31,7 @@ type Options = {
   path: string[];
   save?: string;
   compare?: string;
+  stdin?: boolean;
 };
 
 type GasReportEntry = {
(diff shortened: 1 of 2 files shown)


packages/gas-report/ts/index.ts
```diff
@@ -52,20 +53,22 @@ const commandModule: CommandModule<Options, Options> = {
     return yargs.options({
       save: { type: "string", desc: "Save the gas report to a file" },
       compare: { type: "string", desc: "Compare to an existing gas report" },
+      stdin: { type: "boolean", desc: "Parse the gas report from stdin instead of running an internal test command" },
```
alvrs, line 56:
```suggestion
      stdin: { type: "boolean", desc: "Parse the gas report logs from stdin instead of running an internal test command" },
```

holic (approved):

Issue #1307, opened 2023-08-15
consider moving to generic viem client, passing into individual methods


holic:
Instead of carrying around `publicClient` and `walletClient` and calling specific ones depending on the context, we can support a generic client and use only the viem methods we need, to make it easier to tree-shake: https://viem.sh/docs/clients/custom.html#tree-shaking

PR #1693, opened 2023-10-03
refactor(common): move `createContract`'s internal write logic to `writeContract`

### LINKED ISSUE #1554: refactor createContract, move out writeContract
we should move out the `write` behavior into a `writeContract` replacement (like viem) and then our `createContract` can call into that instead of all the logic living in there

holic:
closes #1554 

This matches how viem abstracts this behavior and lets us use `writeContract` without having to create a contract instance via `createContract` (planning to use this in #1625)


packages/common/src/createContract.ts
```diff
@@ -1,173 +1,4 @@
...
-          };
-        },
-      }
-    );
-  }
-
-  return contract as unknown as GetContractReturnType<TAbi, TPublicClient, TWalletClient, TAddress>;
-}
+/** @deprecated use `getContract` instead */
+export const createContract = getContract;
```
holic, line 4: Slightly different approach to what we've been doing in the past to avoid a breaking change. We can just remove this in a future version as a breaking change.

alvrs (approved):
i like the new more modular approach. For context, was there something in the previous `createContract` approach that was blocking using it in #1625?

holic:
> i like the new more modular approach. For context, was there something in the previous `createContract` approach that was blocking using it in #1625?

not necessarily blocking, just feels a bit cleaner, esp if we eventually have to write to multiple contracts

and ultimately trying to move our code to the more modular/tree-shakeable approach that viem suggests (#1307)

PR #1694, opened 2023-10-03
feat(gas-report): add more logs to stdin piping

alvrs:
- minor addition to pass through logs received via stdin (since the test command can be long running it's useful to know if the tests are still running or the gas reporter is already parsing)

### DIFF
--- a/packages/gas-report/ts/index.ts
+++ b/packages/gas-report/ts/index.ts
@@ -205,13 +205,16 @@ function saveGasReport(gasReport: GasReport, path: string) {
 
 function readStdIn(): Promise<string> {
   return new Promise((resolve) => {
+    console.log("Waiting for stdin...");
     let data = "";
 
     process.stdin.on("data", (chunk) => {
+      console.log(chunk.toString());
       data += chunk;
     });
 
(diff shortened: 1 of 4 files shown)


holic (approved):

packages/gas-report/ts/index.ts
```diff
@@ -205,13 +205,16 @@ function saveGasReport(gasReport: GasReport, path: string) {
...
+    console.log("Waiting for stdin...");
     let data = "";
 
     process.stdin.on("data", (chunk) => {
+      console.log(chunk.toString());
       data += chunk;
     });
 
     process.stdin.on("end", () => {
+      console.log("Done waiting for stdin");
```
holic, line 217: `process.stdin` is a stream so you could also

```
process.stdin.pipe(process.stdout)
```

holic: and could prob lean on a generic stream package like https://github.com/stream-utils/stream-to-array

```
const stdin = (await toArray(process.stdin)).map((chunk) => Buffer.from(chunk).toString('utf8'))
```

PR #1703, opened 2023-10-05
fix(cli): deploy failing to find user modules

codynhat:
When adding a custom user module to `mud.config.ts`, I am receiving the following error when deploying:

```
MUD dev-contracts watcher failed to deploy config or contracts changes

MUDError: Error reading file at out/0.sol/0.json
```

It appears the deploy script cannot find the compiled contract of my module, because it is using the array index (`0`) as the module name instead of the `name` field.


holic (approved):

holic:
good find, thank you!

PR #1706, opened 2023-10-05
feat(common): clarify resourceId (hex) from resource (object)

frolic:
in #1702, I found myself getting a bit confused between resource ID (hex on chain) and resource ID (object in client)

I'm not sure if this is better but pulling this out in case we think this is a good direction

should `hexToResource` be `resourceIdToResource` or `decodeResourceId`?

### DIFF
--- a/.changeset/thirty-cups-provide.md
+++ b/.changeset/thirty-cups-provide.md
@@ -4,6 +4,16 @@
 
 Renames `resourceIdToHex` to `resourceToHex` and `hexToResourceId` to `hexToResource`, to better distinguish between a resource ID (hex value) and a resource reference (type, namespace, name).
 
+```diff
+- resourceIdToHex({ type: 'table', namespace: '', name: 'Position' });
++ resourceToHex({ type: 'table', namespace: '', name: 'Position' });
+```
+
+```diff
+- hexToResourceId('0x...');
++ hexToResource('0x...');
+```
+
 Previous methods still exist but are now deprecated to ease migration and reduce breaking changes. These will be removed in a future version.
 
-Also removes the previously deprecated and now-unused table ID utils (replaced by these resource ID utils).
+Also removes the previously deprecated and unused table ID utils (replaced by these resource ID utils).


packages/common/src/common.ts
```diff
@@ -1,7 +1,9 @@
+import { Hex } from "viem";
 import { ResourceType } from "./resourceTypes";
 
-export type ResourceId = {
+export type Resource = {
   namespace: string;
   name: string;
   type: ResourceType;
+  hex: Hex;
```
holic, line 8: should arguably call this `resourceId` if we're calling the object "resource"

alvrs: agree this could be `resourceId`

packages/common/package.json
```diff
@@ -30,9 +29,6 @@
       "codegen": [
         "./src/codegen/index.ts"
       ],
-      "deprecated": [
-        "./src/deprecated/index.ts"
-      ],
```
holic, line 35: since there's `@deprecated` typedoc operator that does a better job of signaling deprecated stuff, I am preferring that now over marking this as deprecated in the export path

packages/common/src/deprecated/TableId.ts
```diff
@@ -1,41 +0,0 @@
-import { Hex, stringToHex, hexToString, sliceHex, concatHex } from "viem";
-
-/** @deprecated Use `tableIdToHex` and `hexToTableId` instead. */
-export class TableId {
```
holic, line 4: this was deprecated/replaced a while ago but never removed

packages/common/src/deprecated/createContract.ts
alvrs, line 1: it seems like `/deprecated` is not exported anymore, should we keep `createContract` in the `src` directory then? Or fully remove it

alvrs: ah i see now `/deprecated` is still exported from the root `index`, so nevermind

alvrs (approved):

packages/common/src/deprecated/createContract.ts
holic: yep! we're just keeping things in `/deprecated` to make it easier to figure out what we need to remove in the future

PR #1712, opened 2023-10-06
fix(world): register store namespace during initialization

alvrs:


packages/world/src/constants.sol
```diff
@@ -8,6 +8,10 @@ import { RESOURCE_SYSTEM, RESOURCE_NAMESPACE } from "./worldResourceTypes.sol";
 bytes14 constant ROOT_NAMESPACE = "";
 bytes16 constant ROOT_NAME = "";
 
+ResourceId constant STORE_NAMESPACE_ID = ResourceId.wrap(
+  bytes32(abi.encodePacked(RESOURCE_NAMESPACE, bytes14("store"), ROOT_NAME))
+);
+
 ResourceId constant WORLD_NAMESPACE_ID = ResourceId.wrap(
   bytes32(abi.encodePacked(RESOURCE_NAMESPACE, bytes14("world"), ROOT_NAME))
 );
```
holic, line 17: we should prob have some tests eventually that check that these stay aligned with our mud config namespaces

holic (approved):

PR #1721, opened 2023-10-10
docs: v2 [WIP]

alvrs:


alvrs:
Context from discord: we decided to work on the next version of docs from the main branch (in a separate directory to not disrupt the current docs while we're aggressively moving things around for the new docs).

Self-approving this to get the ball rolling, we can iterate on the exact page structure when adding content

PR #1724, opened 2023-10-10
feat(store): add HelloStore event to TS event definitions

holic:
pulled out of https://github.com/latticexyz/mud/pull/1702


alvrs:
seems like there are a couple places that assume the type of these to be just store operation events (https://github.com/latticexyz/mud/actions/runs/6470447113/job/17566678260?pr=1724#step:6:778)

holic:
> seems like there are a couple places that assume the type of these to be just store operation events (https://github.com/latticexyz/mud/actions/runs/6470447113/job/17566678260?pr=1724#step:6:778)

Ah whoops. I moved out the `HelloStore` event for now.

alvrs (approved):

PR #1726, opened 2023-10-10
docs: add store/introduction

alvrs:


qbzzt (approved):

ludns:
Can you remove all the forced new lines? we should let the CSS wrap sentences for us and not introduce line breaks within paragraphs.

PR #1734, opened 2023-10-11
ci: add eslint max-len

techemmy:
fixes https://github.com/latticexyz/mud/issues/1729

setting a max length that is not too restrictive but avoids horizontal scroll bars if auto-wrap is not enabled (assuming the maximum line length that fits on a MBP 16" in vscode with default zoom)
unfortunately i didn't find a auto-fixing setup (see https://github.com/prettier/prettier/issues/265), but the limit seems high enough to not be too disruptive

PR #1735, opened 2023-10-11
ci: add eslint max-len

### LINKED ISSUE #1729: add linter/autofixer for comment line lengths
to avoid nitpicking on these

alvrs:
fixes #1729 

- setting a max length that is not too restrictive but avoids horizontal scroll bars if auto-wrap is not enabled (assuming the maximum line length that fits on a MBP 16" in vscode with default zoom)
- unfortunately i didn't find a auto-fixing setup (see https://github.com/prettier/prettier/issues/265), but the limit seems high enough to not be too disruptive


holic (approved):

alvrs:
closing in favor of #1734

PR #1736, opened 2023-10-11
feat(abi-ts): move logs to debug

holic:


packages/abi-ts/src/debug.ts
```diff
@@ -0,0 +1,3 @@
+import createDebug from "debug";
+
+export const debug = createDebug("abi-ts");
```
holic, line 3: intentionally didn't prefix with `mud:` because this package doesn't have any MUD dependencies and kind of lives outside the framework

alvrs (approved):

PR #1740, opened 2023-10-11
ci: allow revert PRs

### LINKED ISSUE #1486: update PR linter to allow reverts
The "Revert" button in GitHub opens a PR that looks like

> Revert "fix(cli): don't build twice on contract changes"

But our linter doesn't like that format because it doesn't match conventional commit formats. We should update that to support reverts PRs of this format.

buneeIsSlo:
Fixes #1486

Update PR linter to ignore commit messages beginning with `Revert`


.commitlintrc
```diff
@@ -1,3 +1,6 @@
 {
-  "extends": ["@commitlint/config-conventional"]
+  "extends": ["@commitlint/config-conventional"],
+  "ignores": [
+    (commit) => commit.startsWith("Revert ")
```
holic, line 4: this looks like JS - should we rename the file to clarify and get better IDE integration?

buneeIsSlo: Yes, I can do that. Any other changes I need to make?

buneeIsSlo: Hi, I'm unable to import the types needed for this action. Do you have any idea why?

commitlint

buneeIsSlo: @holic I would truly appreciate some help here.

alvrs: have you installed `@commitlint/types` as a new dependency?

buneeIsSlo: Hi @alvrs, when I tried to install the dependency before, I encountered an error message that went something like  `pnpm.ps1 cannot be loaded because running scripts is disabled on this system. For more information, see...`. I switched to a different IDE, and now I get an entirely different message. Am I not supposed to install the dependency in the root folder?

buneeIsSlo: I got it to work. Please review the changes.

pnpm-lock.yaml
```diff
@@ -1683,6 +1687,13 @@ packages:
       prettier: 2.8.4
     dev: true
 
+  /@commitlint/types@17.4.4:
```
holic, line 1690: looks like this got added to the lockfile but not to `package.json`

did you use `pnpm add` to add this? try `pnpm add -D -w @commitlint/types`

buneeIsSlo: @holic I did use `pnpm add`. I ran the command, and it updated the lock file, but there were no changes in the `package.json` :( Am I missing something?

holic:
weird, I updated the PR title but wasn't able to get this to pass the new CI

I dug in a bit more and looks like the default rules should already ignore revert messages of this shape:
https://github.com/conventional-changelog/commitlint/blob/786ecb4037d343251f43f25957af563907db2634/%40commitlint/is-ignored/src/defaults.ts#L22

holic:
Looks like this is specifically related to the CI action we use to validate PR titles, not commitlint config:
https://github.com/latticexyz/mud/blob/main/.github/workflows/pr.yml#L15
https://github.com/amannn/action-semantic-pull-request/issues/192#issuecomment-1763412115

holic:
Going to close this for now as I think this requires some upstream fixes or replacing our CI step with another approach. Thanks for the attempt!

Feel free to open another PR if you want to take another stab at this using a different approach.

buneeIsSlo:
@holic I understand, no problem.

PR #1741, opened 2023-10-11
feat(store): default off storeArgument

### LINKED ISSUE #1723: don't include store argument in tablegen by default
We should turn this off by default because it's got significant trade-offs and you'll want to be careful when/how you use it. (case study: Sky Strife just lowered gas by ~half for one of their primary actions by removing the store param/arg)

holic:
closes #1723

### DIFF
--- a/packages/world/gas-report.json
+++ b/packages/world/gas-report.json
@@ -117,7 +117,7 @@
     "file": "test/World.t.sol",
     "test": "testRegisterTable",
     "name": "Register a new table in the namespace",
-    "gasUsed": 637399
+    "gasUsed": 637367
   },
   {
     "file": "test/World.t.sol",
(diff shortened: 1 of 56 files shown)


packages/world/gas-report.json
```diff
@@ -129,13 +129,13 @@
     "file": "test/World.t.sol",
     "test": "testSetRecord",
     "name": "Write data to the table",
-    "gasUsed": 36896
+    "gasUsed": 39039
   },
```
holic, line 133: do we care that this test jumped ~2k gas due to changing this to use StoreSwitch's stored address?

If so, I could warm the store address or do two writes (one cold, one warm), but not sure if it's a thing we're concerned with benchmarking in this context

alvrs (dismissed):

PR #1742, opened 2023-10-11
docs(template): add the vanilla template

qbzzt:


next-docs/pages/templates/typescript/vanilla.mdx
```diff
@@ -1 +1,91 @@
 # Vanilla
+
+The vanilla template uses standard HTML and JavaScript to communicate with the user.
```
alvrs, line 3:
```suggestion
The vanilla template uses plain HTML and JavaScript for the web client.
```

next-docs/pages/templates/typescript/vanilla.mdx
```diff
@@ -1 +1,91 @@
...
+// Components expose a stream that triggers when the component is updated.
+components.Counter.update$.subscribe((update) => {
+  const [nextValue, prevValue] = update.value;
+  console.log("Counter updated", update, { nextValue, prevValue });
+  document.getElementById("counter")!.innerHTML = String(nextValue?.value ?? "unset");
+});
+```
+
+Specify a function to be called when the value of `Counter` changes.
+This function emits [a log message](https://www.w3schools.com/jsref/met_console_log.asp) and then updates the `counter` span.
```
alvrs, line 57: would you mind if we use the MDN docs for links like this everywhere? I personally just prefer that site over `w3schools`
```suggestion
This function emits [a log message](https://developer.mozilla.org/en-US/docs/Web/API/console/log) and then updates the `counter` span.
```

next-docs/pages/templates/typescript/vanilla.mdx
```diff
@@ -1 +1,91 @@
...
+<script type="module" src="/src/index.ts"></script>
+```
+
+Import the script that runs most of the application.
+
+```html
+<div>Counter: <span id="counter">0</span></div>
+```
+
+Create [an HTML span](https://www.w3schools.com/tags/tag_span.asp) that JavaScript code can update.
```
alvrs, line 33:
```suggestion
Create [an HTML span](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/span) that JavaScript code can update.
```

alvrs (approved):

PR #1744, opened 2023-10-11
docs: add store/tables

alvrs:


next-docs/pages/store/table-intro.mdx
alvrs, line 1: it seems useful to have this "tdlr" on tables both on the store overview page, as well as on the "tables" page, so pulled it out to a snippet that can be imported in both places

next-docs/pages/store/tables.mdx
```diff
@@ -1 +1,149 @@
...
+```
+
+### Key schema
+
+The key schema represents the data types of the (composite) key of the table.
+Internally the composite key is always handled as `bytes32[]`,
+but the key schema allows [offchain indexers](../services/indexer) to create properly typed composite key columns,
+and allows MUD to generate [typed table libraries](./table-libraries).
+
+The key schema can contain up to 28 static data types.
```
qbzzt, line 66: We use "static" and "dynamic" in an unusual sense, so the first time we use either on a page I think it's best to explain what we mean.

next-docs/pages/store/tables.mdx
```diff
@@ -1 +1,149 @@
...
+
+// NOTE: static types must come before dynamic types
+SchemaType[] memory valueSchemaTypes = new SchemaType[](1);
+valueSchemaTypes[0] = SchemaType.UINT256;
+Schema valueSchema = SchemaLib.encode(valueSchemaTypes);
+```
+
+### Key and field names
+
+Key and field names are simple string arrays,
```
qbzzt, line 102:
```suggestion
Key and (value) field names are simple string arrays,
```

next-docs/pages/store/tables.mdx
```diff
@@ -1 +1,137 @@
...
 # Tables
+
+<TableIntro />
+
+Tables are registered in the `Store` contract at runtime.
+When a table is registered, its ID, key and value types, as well as key and field names are stored in the internal `Tables` table.
+This emits an event that can be used by [offchain indexers](../services/indexer) to start replicating the state of the new table.
+
+The recommended way of reading from tables and writing to tables is via the [typed table libraries](./table-libraries).
+However, it is also possible to use the low-level `IStore` (external) or `StoreCore` (internal) [API](./reference) directly.
```
alvrs, line 13:
```suggestion
However, it is also possible to use the low-level [`IStore` (external)](./reference/store) or [`StoreCore` (internal)](./reference/store-core) API directly.
```

PR #1745, opened 2023-10-11
feat(world): add TS helpers to encode system calls

holic:


packages/world/ts/encodeSystemCalls.ts
```diff
@@ -0,0 +1,17 @@
...
+export function encodeSystemCalls<abi extends Abi, functionName extends string = string>(
+  abi: abi,
+  systemCalls: readonly Omit<SystemCall<abi, functionName>, "abi">[]
+): GetFunctionArgs<typeof IWorldCallAbi, "call">["args"][] {
+  return systemCalls.map(
+    ({ systemId, functionName, args }) =>
+      [
+        systemId,
+        encodeFunctionData({ abi, functionName, args } as unknown as EncodeFunctionDataParameters<abi, functionName>),
+      ] as const
```
alvrs, line 15: would using `encodeSystemCall` here work?

alvrs (dismissed):
wonder about the possibility of using the singular methods in the plural methods, but not blocking

packages/world/ts/encodeSystemCalls.ts
holic: now you're thinkinnnnn

alvrs (approved):

PR #1749, opened 2023-10-11
fix(cli): support enums in deploy, only deploy modules/systems once

holic:
follow up of missed TODO in #1702

### DIFF
--- a/packages/cli/src/deploy/configToTables.ts
+++ b/packages/cli/src/deploy/configToTables.ts
@@ -6,8 +6,34 @@ import { Hex } from "viem";
 
 // TODO: we shouldn't need this file once our config parsing returns nicely formed tables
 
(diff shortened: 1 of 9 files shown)


packages/cli/src/deploy/configToTables.ts
```diff
@@ -41,6 +43,10 @@ export type Tables<config extends StoreConfig = StoreConfig> = {
 };
 
 export function configToTables<config extends StoreConfig>(config: config): Tables<config> {
+  const userTypes = {
+    ...config.userTypes,
+    ...Object.fromEntries(Object.entries(config.enums).map(([key]) => [key, { internalType: "uint8" }] as const)),
+  };
```
alvrs, line 49: how did this even work before 👀

alvrs (dismissed):

PR #1756, opened 2023-10-12
feat(world-modules): only install modules once

holic:
lazy/naive fix so we can catch these errors during deploys and not panic


package.json
```diff
@@ -13,7 +13,7 @@
     "dev": "turbo run dev --concurrency 100",
     "dist-tag-rm": "pnpm recursive exec -- sh -c 'npm dist-tag rm $(cat package.json | jq -r \".name\") $TAG || true'",
     "foundryup": "curl -L https://foundry.paradigm.xyz | bash && bash ~/.foundry/bin/foundryup",
-    "gas-report": "pnpm recursive run gas-report",
+    "gas-report": "pnpm run --recursive --parallel gas-report",
```
holic, line 16: snuck this in because it makes this wildly faster

alvrs (approved):

PR #1762, opened 2023-10-12
docs(store): add API reference

alvrs:
This approach lets us generate API docs from NatSpec and render them on the docs page.
Limitations:
- No sidebar quickjumps for headlines (nextra limitation with imported markdown)
- Links to github source and links to eg other interfaces in the docs don't work (bc forge is generating invalid links for github, and the nextra sitemap is different from the one expected by forge)

We could work around both limitations by creating a script that parses the raw forge markdown output, processes it (eg to replace links and change headlines from H1 to H2 etc), and render it as plain mdx file in the docs.

Curious about your thoughts on this approach @holic @qbzzt


holic:
> We could work around both limitations by creating a script that parses the raw forge markdown output, processes it (eg to replace links and change headlines from H1 to H2 etc), and render it as plain mdx file in the docs.

another approach could be a component that wraps each markdown file and does ~the same

holic:
Looks decent, despite shortcomings!

https://mud-next-docs-fv7gsl0r0-latticexyz.vercel.app/store/reference/store-core

Seems like the imported markdown is also missing from search, which is a big downside IMO. Bigger than the ones mentioned above.

alvrs:
> Seems like the imported markdown is also missing from search, which is a big downside IMO. Bigger than the ones mentioned above.

Good catch! I think to solve this we have to print the markdown instead of using the nextra component

alvrs:
Added a script to post-process the `forge doc` output and render it as raw mdx in the docs

qbzzt (approved):
I love it. There is some formatting weirdness (compare http://localhost:3000/store/reference/store-core#getfieldlayout and http://localhost:3000/store/reference/store-core#getvalueschema, for example). Also, I'd expect the italicized test to be the one in the purple box, not what's in the green one.

Screenshot 2023-10-13 at 12 53 51 PM

However, these are minor and we have lots of docs to go through. I saw merge it now and deal with formatting later. It's clear enough.

alvrs:
Re italic font vs regular font: I think that depends on what has the
@notice tag (regular) and what has the @dev tag (italic) in the NatSpec. We
need to do another pass for NatSpec in the IStore interfaces, we had
skipped those bc they were not relevant for the audit.

PR #1763, opened 2023-10-12
docs(hello-world): add-system and add-table

qbzzt:


next-docs/pages/guides/hello-world/add-system.mdx
```diff
@@ -0,0 +1,296 @@
...
+          * The parameter list informs TypeScript that:
+          *
+          * - The first parameter is expected to be a
+          *   SetupNetworkResult, as defined in setupNetwork.ts
+          *
+          * - Out of this parameter, we only care about two fields:
+          *   - worldContract (which comes from createContract, see
+          *     https://github.com/latticexyz/mud/blob/26dabb34321eedff7a43f3fcb46da4f3f5ba3708/templates/vanilla/packages/client/src/mu$
+          *   - waitForTransaction (which comes from syncToRecs, see
+          *     https://github.com/latticexyz/mud/blob/26dabb34321eedff7a43f3fcb46da4f3f5ba3708/templates/vanilla/packages/client/src/mu$
```
alvrs, line 94: these two links seem to be cut off. Also, should we link to the most recent version on `main` instead of the permalink to a previous state of the repo?

next-docs/pages/guides/hello-world/add-system.mdx
```diff
@@ -0,0 +1,296 @@
...
+          *
+          * - From the second parameter, which is a ClientComponent,
+          *   we only care about Counter. This parameter comes to use
+          *   through createClientComponents.ts, but it originates in
+          *   syncToRecs
+        (https://github.com/latticexyz/mud/blob/26dabb34321eedff7a43f3fcb46da4f3f5ba3708/templates/vanilla/packages/client/src/mud/setupN$
```
alvrs, line 100: same here, link seems to be cut off

next-docs/pages/guides/hello-world/add-system.mdx
```diff
@@ -0,0 +1,296 @@
...
+          */
+          { worldContract, waitForTransaction }: SetupNetworkResult,
+          { Counter }: ClientComponents
+        ) {
+          const increment = async () => {
+            /*
+            * Because IncrementSystem
+            * (https://mud.dev/tutorials/walkthrough/minimal-onchain#incrementsystemsol)
```
alvrs, line 108: we should update this to the new link

next-docs/pages/guides/hello-world/add-system.mdx
```diff
@@ -0,0 +1,296 @@
...
+        ```
+
+    </CollapseCode>
+
+           <details>
+
+    <summary>Explanation</summary>
+
+        {" "}
+        <p></p>
```
alvrs, line 137: do we need this?

next-docs/pages/guides/hello-world/add-system.mdx
```diff
@@ -0,0 +1,296 @@
...
+
+        The new function is `decrement`.
+
+        ```typescript
+        const decrement = async () => {
+        ```
+
+        This function involves sending a transaction, which is a slow process, so it needs to be [asynchronous](https://www.w3schools.com/js/js_async.asp).
```
alvrs, line 145: let's use `MDN` instead of `w3schools` everywhere for consistency
```suggestion
        This function involves sending a transaction, which is a slow process, so it needs to be [asynchronous](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function).
```

next-docs/pages/guides/hello-world/add-system.mdx
```diff
@@ -0,0 +1,296 @@
...
+
+        ```typescript
+          const tx = await worldContract.write.decrement();
+        ```
+
+        This is the way we call functions in top-level systems in a world.
```
alvrs, line 151:
```suggestion
        This is the way we call functions in systems in a world.
```

I feel like "top-level" is a bit confusing here. We could explain that it's a "root" system and therefore the function selector in the World is the same as the function selector in the system, otherwise it would be `namespace_name_decrement()`, but not sure if this is the right place to explain it. Maybe we should link to another place where it is explained instead.

next-docs/pages/guides/hello-world/add-system.mdx
```diff
@@ -0,0 +1,296 @@
...
+
+        ```typescript
+          await waitForTransaction(tx);
+        ```
+
+        Await until we receive confirmation that the transaction has been added to a block.
```
alvrs, line 157:
```suggestion
        Await until the transaction has been included in a block and the corresponding state has been synchronized with the client.
```

next-docs/pages/guides/hello-world/add-system.mdx
```diff
@@ -0,0 +1,296 @@
...
+    <summary>Explanation</summary>
+
+        {" "}
+        <p></p>
+
+        ```html
+        <button onclick="window.decrement()">Decrement</button>
+        ```
+
+        Create a [`button`](https://www.w3schools.com/tags/tag_button.asp) with an [`onClick`](https://www.w3schools.com/tags/ev_onclick.asp) property.
```
alvrs, line 292:
```suggestion
        Create a [`button`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button) with an [`onClick`](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Events) property.
```

alvrs: btw, after looking at https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Events i am realizing we should probably not use the `window` and inline `onclick` approach but rather give the button an ID and then do this in the javascript code:

```ts
const incrementButton = document.querySelector("#increment-button");
incrementButton.addEventListener("click", increment);
```

One could argue that it doesn't matter much because it's unlikely folks would _actually_ use plain html/js (besides learning purposes, or for integrating in a separate framework), but I think it could give a bad first impression if we're using bad practices in the vanilla template.

next-docs/pages/guides/hello-world/add-table.mdx
```diff
@@ -0,0 +1,340 @@
...
+   }
+   ```
+
+   </CollapseCode>
+
+<details>
+<summary>Explanation</summary>
+
+```typescript
+let options: String = "";
```
alvrs, line 208:
```suggestion
let options: string = "";
```

nit, let's use the primitive type over the class (see https://www.geeksforgeeks.org/what-is-the-difference-between-string-and-string-in-typescript)

alvrs: in fact we can remove `: string` since we're initializing it in the same line, so typescript can infer the type

next-docs/pages/guides/hello-world/add-table.mdx
```diff
@@ -0,0 +1,340 @@
...
+This exclamation point tells the compiler that it's OK, there will be a real value there.
+[See here for additional information](https://blog.logrocket.com/understanding-exclamation-mark-typescript/).
+
+```typescript
+(window as any).readHistory = async counterValue => {
+  const History = components.History
+  const entity = encodeEntity(History.metadata.keySchema, { counterValue });
+```
+
+`encodeEntity` creates a key in the format that MUD uses, which is based on [ABI argument encoding](https://docs.soliditylang.org/en/latest/abi-spec.html#argument-encoding).
```
alvrs, line 240:
```suggestion
`encodeEntity` creates an entity key with the [encoding that MUD uses](/store/encoding).
```

we can link to the MUD encoding docs page here

next-docs/pages/guides/hello-world/add-table.mdx
```diff
@@ -0,0 +1,340 @@
...
+
+```typescript
+  document.getElementById("timeStamp")!.innerHTML =
+    new Date(parseInt(Number(time) * 1000))
+}
+```
+
+Solidity uses [Unix time](https://en.wikipedia.org/wiki/Unix_time).
+JavaScript uses a similar system, but it measures times in milliseconds.
+So to get a readable date, we take the time (which is a [`BigInt`](https://www.w3schools.com/js/js_bigint.asp)), multiply it by a thousand, and then convert it to a [`Date`](https://www.w3schools.com/jsref/jsref_obj_date.asp) object.
```
alvrs, line 262:
```suggestion
So to get a readable date, we take the time (which is a [`BigInt`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt)), multiply it by a thousand, and then convert it to a [`Date`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date) object.
```

next-docs/pages/guides/hello-world/add-table.mdx
```diff
@@ -0,0 +1,340 @@
...
+
+```html
+<hr />
+<h2>
+  History for value:
+  <select id="historyValue" onInput="window.readHistory(value)"></select>
+</h2>
+```
+
+This is the [input field](https://www.w3schools.com/tags/tag_select.asp) that lets the user select which counter value they'd like to get information about.
```
alvrs, line 314:
```suggestion
This is the [input field](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select) that lets the user select which counter value they'd like to get information about.
```

next-docs/pages/guides/hello-world/add-table.mdx
```diff
@@ -0,0 +1,340 @@
...
+The `id` attribute is used by `packages/client/src/index.ts` to set the options.
+The `onInput` attribute is the JavaScript code to execute when the value changes.
+
+```html
+<table border></table>
+```
+
+A standard [HTML table](https://www.w3schools.com/html/html_tables.asp).
```
alvrs, line 322:
```suggestion
A standard [HTML table](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/table).
```

next-docs/pages/guides/hello-world/add-system.mdx
qbzzt: How about "systems in the root namespace"? The place to explain the details is in Extending World, where we will actually do it.

next-docs/pages/guides/hello-world/add-system.mdx
qbzzt: Oops on the links. The problem with linking to the latest version is that line numbers could change (and those are part of the real link). But then again, the template doesn't change that often - I'll use main and people can just search if the information isn't at the exact line.

next-docs/pages/guides/hello-world/add-system.mdx
qbzzt: Weird. prettier used to break things when I committed without this, but now it does. I'll remove them.

next-docs/pages/guides/hello-world/add-system.mdx
qbzzt: Merge https://github.com/latticexyz/mud/pull/1774 and I'll fix this here too.

next-docs/pages/guides/hello-world/add-system.mdx
alvrs: yep that works

alvrs (approved):

PR #1765, opened 2023-10-13
feat(cli): add `--alwaysPostDeploy` flag to deploys

holic:
Helpful if you want to re-run a `mud deploy` with a world that may not have finished deploying. A small workaround until we can do proper failure retries.

### DIFF
--- a/packages/cli/src/commands/dev-contracts.ts
+++ b/packages/cli/src/commands/dev-contracts.ts
@@ -15,7 +15,7 @@ import { Address } from "viem";
 const devOptions = {
   rpc: deployOptions.rpc,
   configPath: deployOptions.configPath,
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/dev-contracts.ts
```diff
@@ -15,6 +15,7 @@ import { Address } from "viem";
 const devOptions = {
   rpc: deployOptions.rpc,
   configPath: deployOptions.configPath,
+  alwaysPostDeploy: deployOptions.alwaysPostDeploy,
```
holic, line 18: "always post deploy" kinda sounds like we're "posting" the deploy somewhere

maybe this should be `alwaysRunPostDeploy`?

alvrs (approved):

PR #1766, opened 2023-10-13
feat(cli): add retries to deploy

holic:


packages/cli/src/deploy/ensureFunctions.ts
```diff
@@ -40,23 +42,45 @@ export async function ensureFunctions({
...
+        return pRetry(
+          () =>
+            writeContract(client, {
+              chain: client.chain ?? null,
+              address: worldDeploy.address,
+              abi: worldAbi,
+              // TODO: replace with batchCall (https://github.com/latticexyz/mud/issues/1645)
+              functionName: "registerRootFunctionSelector",
+              args: [func.systemId, func.systemFunctionSignature, func.systemFunctionSelector],
+            }),
```
holic, line 54: we could potentially expand `writeContract`'s API to take in retry options or do finer grained retries (e.g. just retry simulation), but I want to think more about that before deviating from viem

alvrs (approved):

alvrs (approved):

PR #1767, opened 2023-10-13
fix(cli): deploy systems/modules before registering/installing them

holic:
Using the `pending` block tag to simulate/estimate gas seems to give bad results if the a contract being called does not yet exist (i.e. is itself pending in the mempool). So we need to wait for contracts to be deployed before proceeding.

### DIFF
--- a/packages/cli/src/deploy/deploy.ts
+++ b/packages/cli/src/deploy/deploy.ts
@@ -2,14 +2,7 @@ import { Account, Address, Chain, Client, Transport } from "viem";
 import { ensureDeployer } from "./ensureDeployer";
 import { deployWorld } from "./deployWorld";
 import { ensureTables } from "./ensureTables";
-import {
-  Config,
-  ConfigInput,
-  DeterministicContract,
-  WorldDeploy,
-  supportedStoreVersions,
-  supportedWorldVersions,
-} from "./common";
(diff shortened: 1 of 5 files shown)


packages/cli/src/deploy/deploy.ts
```diff
@@ -51,6 +54,27 @@ export async function deploy<configInput extends ConfigInput>({
...
+      ),
+      ...uniqueBy(config.modules, (mod) => mod.address).map((mod) =>
+        ensureContract({ client, bytecode: mod.bytecode, label: `${mod.name} module` })
+      ),
+    ])
+  ).flat();
+
+  if (contractTxs.length) {
+    debug("waiting for contracts");
+    // wait for each tx separately/serially, because parallelizing results in RPC errors
```
alvrs, line 71: > because parallelizing results in RPC errors

oh interesting

alvrs (approved):

packages/cli/src/deploy/deploy.ts
holic: I think because viem under the hood sets up a block watcher when you call `waitForTransaction`, so when we do this 100x in parallel, we get a LOT of polling of the RPC.

Since ultimately the `waitForTransactionReceipt` just does a `getTransactionReceipt` under the hood until it succeeds, it's fine to do this serially, because each transaction confirmed in the same block or blocks before will return immediately without the polling.

It'll ~resolve in the same amount of time with far less RPC calls.

PR #1769, opened 2023-10-13
fix(cli): handle module already installed

holic:


packages/cli/src/deploy/ensureModules.ts
```diff
@@ -17,49 +17,47 @@ export async function ensureModules({
...
+                : await writeContract(client, {
+                    chain: client.chain ?? null,
+                    address: worldDeploy.address,
+                    abi: worldAbi,
+                    // TODO: replace with batchCall (https://github.com/latticexyz/mud/issues/1645)
+                    functionName: "installModule",
+                    args: [mod.address, mod.installData],
+                  });
+            } catch (error) {
+              if (error instanceof BaseError && error.message.includes("Module_AlreadyInstalled")) {
```
holic, line 44: I wanted to put this in `onFailedAttempt` but there's no way to signal to `p-retry` that we should stop retrying and continue.

holic:
merging to unlock sky strife

PR #1783, opened 2023-10-15
feat(utils): remove hash utils and ethers

holic:
hash utils are no longer used internally (and can be replaced by [viem's `keccak256`](https://viem.sh/docs/utilities/keccak256.html#keccak256) anyway)


alvrs (approved):
let's add a changeset that shows how to replace with viem's keccak

alvrs (approved):

PR #1789, opened 2023-10-17
feat(world-modules): add ERC20 module

alvrs:
MUD-native ERC20 token
- anyone can register a new ERC20 token into a world. The new token will be registered into a new namespace.
- The "puppet" implements the ERC20 interface and emits the ERC20 events
- All logic lives in the ERC20System
- The ERC20System can be interacted with like a normal MUD system, including all MUD features (hooks, callFrom, access control, etc)

CleanShot 2023-10-18 at 21 31 59


packages/world-modules/src/modules/erc20/ERC20Module.sol
```diff
@@ -0,0 +1,107 @@
...
+
+  ERC20DelegationControl public immutable erc20DelegationControl = new ERC20DelegationControl();
+  ResourceId public immutable erc20DelegationControlId =
+    WorldResourceIdLib.encode({ typeId: RESOURCE_SYSTEM, namespace: MODULE_NAMESPACE, name: "ERC20Delegation" });
+
+  function getName() public pure override returns (bytes16) {
+    return MODULE_NAME;
+  }
+
+  function install(bytes memory args) public {
```
holic, line 37: do we need an "already installed" check here?

packages/world-modules/src/modules/erc20/ERC20Module.sol
```diff
@@ -0,0 +1,107 @@
...
+      name: "ERC20"
+    });
+
+    // Register the tables
+    Allowances.register(allowancesTableId);
+    Balances.register(balancesTableId);
+    Metadata.register(metadataTableId);
+
+    // Create the ERC20System
+    ERC20System erc20System = new ERC20System();
```
holic, line 71: should we deploy this via the deterministic deployer?

packages/world-modules/src/modules/erc20/ERC20Module.sol
```diff
@@ -0,0 +1,107 @@
...
+
+    // Transfer ownership of the namespace to the caller
+    world.transferOwnership(namespaceId, _msgSender());
+
+    // If the ERC20Registry is not registered yet, register it
+    if (!ResourceIds.getExists(ERC20_REGISTRY_TABLE_ID)) {
+      ERC20Registry.register(ERC20_REGISTRY_TABLE_ID);
+    }
+
+    // If the delegation control is not registered yet, register it and grant it access to the
```
holic, line 92: to the... to the...

packages/world-modules/src/modules/erc20/ERC20Module.sol
alvrs: it's one system per puppet now

packages/world-modules/src/modules/erc20/ERC20Module.sol
alvrs: it's gone

packages/world-modules/src/modules/erc20/ERC20System.sol
```diff
@@ -0,0 +1,288 @@
...
+      }
+    } else {
+      unchecked {
+        // Overflow not possible: balance + value is at most totalSupply, which we know fits into a uint256.
+        Balances.set(balanceTableId, to, Balances.get(balanceTableId, to) + value);
+      }
+    }
+
+    // Emit Transfer event on puppet
+    puppet().log(IERC20Events.Transfer.selector, _toBytes32(from), _toBytes32(to), abi.encode(value));
```
holic, line 236: wtb solidity native `encodeLog` like `encodeCall`

alvrs: we should open an EIP for that, would be nice for sure

.changeset/silent-buttons-peel.md
```diff
@@ -0,0 +1,20 @@
+---
+"@latticexyz/world-modules": minor
+---
+
+Adds the `ERC20Module` to `@latticexyz/world-modules`.
```
holic, line 5: small thing but we should figure out what tense to use in changelogs vs commits

I usually use present tense in commits (adds) and past in changelogs (added)

.changeset/silent-buttons-peel.md
```diff
@@ -0,0 +1,20 @@
...
+
+Important note: this module has not been audited yet, so any production use is discouraged for now.
+
+```solidity
+import { PuppetModule } from "@latticexyz/world-modules/src/modules/puppet/PuppetModule.sol";
+import { IERC20Mintable } from "@latticexyz/world-modules/src/modules/erc20-puppet/IERC20Mintable.sol";
+import { registerERC20 } from "@latticexyz/world-modules/src/modules/erc20-puppet/registerERC20.sol";
+
+// The ERC20 module requires the Puppet module to be installed first
+world.installModule(new PuppetModule(), new bytes(0));
```
holic, line 16: ~~do we also need to install the ERC20Module here?~~

I now see below that `registerERC20` installs the module

packages/world-modules/src/modules/erc20-puppet/ERC20Module.sol
```diff
@@ -0,0 +1,93 @@
...
+    Allowances.register(_allowancesTableId(namespace));
+    Balances.register(_balancesTableId(namespace));
+    Metadata.register(_metadataTableId(namespace));
+
+    // Register a new ERC20System
+    IBaseWorld(_world()).registerSystem(_erc20SystemId(namespace), new ERC20System(), true);
+  }
+
+  function _requireDependencies() internal view {
+    // If the PuppetModule is not installed yet, install it
```
holic, line 45: in a world where `install` becomes idempotent, can we detect+install the puppet module instead of reverting?

packages/world-modules/src/modules/erc20-puppet/ERC20System.sol
```diff
@@ -0,0 +1,286 @@
...
+      }
+    } else {
+      unchecked {
+        // Overflow not possible: balance + value is at most totalSupply, which we know fits into a uint256.
+        Balances.set(balanceTableId, to, Balances.get(balanceTableId, to) + value);
+      }
+    }
+
+    // Emit Transfer event on puppet
+    puppet().log(Transfer.selector, _toBytes32(from), _toBytes32(to), abi.encode(value));
```
holic, line 234: so cool!

ooc why do we need to convert the addresses to bytes32? is this essentially what solidity/evm does under the hood for indexed address args?

holic (approved):

packages/world-modules/src/modules/erc20-puppet/ERC20Module.sol
alvrs: Had that at some intermediate state in this PR, but it blew the module's bytecode size way over the limit (because the puppet module's bytecode had to be inlinted to the erc20 module)

packages/world-modules/src/modules/erc20-puppet/ERC20System.sol
alvrs: yeah exactly! we need a common signature on the `log` functions, so need to do this conversion outside the log function. But internally the same thing happens when emitting an event with indexed args.

holic: in that case, I wonder if it'd be easier to understand if we called this `_toTopic` or something

.changeset/silent-buttons-peel.md
alvrs: Agree with guidelines being:
- past in changelogs (added)
- [imperative present tense](https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#Commit-Guidelines) (add) in commits 😄 

we should put that down somewhere, probably the "contributing" docs page (cc @qbzzt)

packages/world-modules/src/modules/erc20-puppet/ERC20Module.sol
holic: oh interesting! makes sense to go with this approach then

I wonder if we could eventually get to a place where the install just takes in the CREATE2 address of the module and we expect it to exist on chain (deployed via our deterministic deployer)

alvrs: Agree that would be nice

.changeset/silent-buttons-peel.md
```diff
@@ -0,0 +1,20 @@
+---
+"@latticexyz/world-modules": minor
+---
+
+Adds the `ERC20Module` to `@latticexyz/world-modules`.
```
alvrs, line 5:
```suggestion
Added the `ERC20Module` to `@latticexyz/world-modules`.
```

packages/world-modules/src/modules/erc20-puppet/ERC20Module.sol
```diff
@@ -0,0 +1,93 @@
...
+    if (InstalledModules.get(PUPPET_MODULE_NAME, keccak256(new bytes(0))) == address(0)) {
```
holic, line 46: since we're starting to use this more, might be nice to lift this into `Module` as a helper function
e.g. `isInstalled()` or `isInstalled(module)`

holic (dismissed):

packages/world-modules/src/modules/erc20-puppet/ERC20Module.sol
alvrs: https://github.com/latticexyz/mud/issues/1852

PR #1793, opened 2023-10-18
feat(world-modules): add puppet module

alvrs:


packages/world-modules/src/modules/puppet/Puppet.sol
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  /**
+   * @dev Log an event with a signature and three additional topics
+   */
+  function log(
+    bytes32 eventSignature,
+    bytes32 topic1,
+    bytes32 topic2,
+    bytes32 topic3,
```
holic, line 73: are these essentially the indexed args?

alvrs: yea exactly

.changeset/happy-pants-try.md
```diff
@@ -0,0 +1,22 @@
...
+
+Adds the `PuppetModule` to `@latticexyz/world-modules`. The puppet pattern allows an external contract to be registered as an external interface for a MUD system.
+This allows standards like `ERC20` (that require a specific interface and events to be emitted by a unique contract) to be implemented inside a MUD World.
+
+The puppet serves as a proxy, forwarding all calls to the implementation system (also called the "puppet master").
+The "puppet master" system can emit events from the puppet contract.
+
+```solidity
+import { PuppetModule } from "@latticexyz/world-modules/src/modules/puppet/PuppetModule.sol";
+import { createPuppet } from "@latticexyz/world-modules/src/modules/puppet/createPuppet.sol";
```
holic, line 13: Although the puppet terminology is fun, I wonder if we should just call this a proxy 🙈

packages/world-modules/src/modules/puppet/Puppet.sol
```diff
@@ -0,0 +1,80 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.21;
+
+import { StoreSwitch } from "@latticexyz/store/src/StoreSwitch.sol";
+import { ResourceId } from "@latticexyz/store/src/ResourceId.sol";
+import { IBaseWorld } from "@latticexyz/world/src/codegen/interfaces/IBaseWorld.sol";
+import { Systems } from "@latticexyz/world/src/codegen/tables/Systems.sol";
+
+contract Puppet {
```
holic, line 9: pretty great to see how lean this is!

packages/world-modules/src/modules/puppet/PuppetMaster.sol
```diff
@@ -0,0 +1,18 @@
...
+import { ResourceId } from "@latticexyz/store/src/ResourceId.sol";
+import { SystemRegistry } from "@latticexyz/world/src/codegen/tables/SystemRegistry.sol";
+import { PuppetRegistry } from "./tables/PuppetRegistry.sol";
+import { PUPPET_TABLE_ID } from "./constants.sol";
+import { Puppet } from "./Puppet.sol";
+
+contract PuppetMaster {
+  error PuppetMaster_NoPuppet(address systemAddress, ResourceId systemId);
+
+  function puppet() internal view returns (Puppet) {
```
holic, line 12: does this mean there's only a 1:1 mapping between a system and a puppet?

if I want multiple tokens (e.g. ERC721) on my world, how would I go about that?

packages/world-modules/src/modules/puppet/createPuppet.sol
```diff
@@ -0,0 +1,23 @@
...
+
+using WorldResourceIdInstance for ResourceId;
+
+/**
+ * This free function can be used to create a puppet and register it with the puppet delegation control.
+ * Since it is inlined in the caller's context, the calls originate from the caller's address.
+ */
+function createPuppet(IBaseWorld world, ResourceId systemId) returns (address puppet) {
+  puppet = abi.decode(
+    world.call(PUPPET_FACTORY, abi.encodeCall(PuppetFactorySystem.createPuppet, (systemId))),
```
holic, line 19: ooc why don't we use SystemSwitch here?

holic (approved):
Looks good! Only real concern is what seems to be a 1:1 mapping between puppets and systems. Wondering about one to many, but also don't need to explode the scope of this right now.

.changeset/happy-pants-try.md
alvrs: Thought about that too but I'm afraid proxy has too many preconceptions around the proxy being the authority (ie proxy holds the state, proxy switches its implementation, etc), whereas here the "proxy" is just the "puppet" of the World

packages/world-modules/src/modules/puppet/PuppetMaster.sol
alvrs: > does this mean there's only a 1:1 mapping between a system and a puppet

For now yes. Agree 1:N would be nicer but makes it much more complex, so I'd think about that later (either as a change to the puppet module, or with a new module)

> if I want multiple tokens (e.g. ERC721) on my world, how would I go about that?

You'll have to register different systems (see #1844). Thought a lot about this too, but I feel like this approach is the simplest (one namespace per token, with a separate system and tables in that namespace). Otherwise everything gets more complicated (if it's just one system you'd need to pass a param to the system, which means the interface doesn't match anymore with the puppet, access management becomes more complex, etc)

packages/world-modules/src/modules/puppet/createPuppet.sol
alvrs: Agree should probably use system switch here

.changeset/happy-pants-try.md
holic: true! proxy is quite overloaded

.changeset/happy-pants-try.md
```diff
@@ -0,0 +1,22 @@
+---
+"@latticexyz/world-modules": minor
+---
+
+Adds the `PuppetModule` to `@latticexyz/world-modules`. The puppet pattern allows an external contract to be registered as an external interface for a MUD system.
```
alvrs, line 5:
```suggestion
Added the `PuppetModule` to `@latticexyz/world-modules`. The puppet pattern allows an external contract to be registered as an external interface for a MUD system.
```

.changeset/happy-pants-try.md
```diff
@@ -0,0 +1,22 @@
...
+import { PuppetModule } from "@latticexyz/world-modules/src/modules/puppet/PuppetModule.sol";
+import { createPuppet } from "@latticexyz/world-modules/src/modules/puppet/createPuppet.sol";
+
+// Install the puppet module
+world.installModule(new PuppetModule(), new bytes(0));
+
+// Register a new puppet for any system
+// The system must implement the `CustomInterface`,
+// and the caller must own the system's namespace
+CustomInterface puppet = CustomInterface(createPuppet(world, <systemId>));
```
alvrs, line 21:
```suggestion
CustomInterface puppet = CustomInterface(createPuppet(<systemId>));
```

.changeset/happy-pants-try.md
```diff
@@ -0,0 +1,22 @@
...
+---
+
+Added the `PuppetModule` to `@latticexyz/world-modules`. The puppet pattern allows an external contract to be registered as an external interface for a MUD system.
+This allows standards like `ERC20` (that require a specific interface and events to be emitted by a unique contract) to be implemented inside a MUD World.
+
+The puppet serves as a proxy, forwarding all calls to the implementation system (also called the "puppet master").
+The "puppet master" system can emit events from the puppet contract.
+
+```solidity
+import { PuppetModule } from "@latticexyz/world-modules/src/modules/puppet/PuppetModule.sol";
```
alvrs, line 12: import { StoreSwitch } from "@latticexyz/store/src/StoreSwitch.sol";

.changeset/happy-pants-try.md
```diff
@@ -0,0 +1,22 @@
...
+import { createPuppet } from "@latticexyz/world-modules/src/modules/puppet/createPuppet.sol";
+
+// Install the puppet module
+world.installModule(new PuppetModule(), new bytes(0));
+
```
alvrs, line 17:
```suggestion

// Set up the `StoreSwitch` context if not calling from within a `World` context already
StoreSwitch.setStoreAddress(address(world));

```

.changeset/happy-pants-try.md
```diff
@@ -0,0 +1,22 @@
...
+Added the `PuppetModule` to `@latticexyz/world-modules`. The puppet pattern allows an external contract to be registered as an external interface for a MUD system.
+This allows standards like `ERC20` (that require a specific interface and events to be emitted by a unique contract) to be implemented inside a MUD World.
+
+The puppet serves as a proxy, forwarding all calls to the implementation system (also called the "puppet master").
+The "puppet master" system can emit events from the puppet contract.
+
+```solidity
+import { PuppetModule } from "@latticexyz/world-modules/src/modules/puppet/PuppetModule.sol";
+import { createPuppet } from "@latticexyz/world-modules/src/modules/puppet/createPuppet.sol";
```
alvrs, line 13:
```suggestion
import { createPuppet } from "@latticexyz/world-modules/src/modules/puppet/createPuppet.sol";
import { StoreSwitch } from "@latticexyz/store/src/StoreSwitch.sol";
```

packages/world-modules/src/modules/puppet/createPuppet.sol
alvrs: update: turns out using system switch here pushes `ERC20Module` over the bytecode limit, so reverting for now (meaning you can't register a puppet from a root system for now). We should find a better implementation for `SystemSwitch` that doesn't require inlining the libraries, maybe by deploying `StoreCore` as a public library.

packages/world-modules/src/utils/AccessControlLib.sol
```diff
@@ -0,0 +1,54 @@
...
+import { ResourceAccess } from "@latticexyz/world/src/codegen/tables/ResourceAccess.sol";
+import { NamespaceOwner } from "@latticexyz/world/src/codegen/tables/NamespaceOwner.sol";
+
+/**
+ * @title AccessControlLib
+ * @dev Provides access control functions for checking permissions and ownership within a namespace.
+ * This library is functionally equivalent with the AccessControl library from world,
+ * but uses StoreSwitch instead of always reading from own storage.
+ */
+library AccessControlLib {
```
dk1a, line 15: Though I understand why it's in `world-modules`, would it maybe be better to have a single `AccessControl` lib with `hasAccess`/`_hasAccess` etc, to have all the logic in 1 place and let it mirror the table methods?

dk1a (dismissed):
Just some formatting nits

packages/world-modules/src/modules/puppet/createPuppet.sol
holic: we should prob add a follow up issue to come back to 1) this specific instance and 2) the problem overall

holic (dismissed):
agree with @dk1a's comment around centralizing `AccessControlLib` but otherwise looks great

packages/world-modules/src/modules/puppet/PuppetMaster.sol
```diff
@@ -0,0 +1,18 @@
...
+    ResourceId systemId = SystemRegistry.getSystemId(address(this));
+    address puppetAddress = PuppetRegistry.get(PUPPET_TABLE_ID, systemId);
+    if (puppetAddress == address(0)) revert PuppetMaster_NoPuppet(address(this), systemId);
```
holic, line 15: does `address(this)` still work in the context of a delegatecall?

packages/world-modules/src/modules/puppet/createPuppet.sol
alvrs: yeah i mean public as in not-internal, which means delegatecall to an external contract

packages/world-modules/src/utils/AccessControlLib.sol
alvrs: agree! unfortunately can't touch the code in `world` while it's being audited but should note this down for later

packages/world-modules/src/modules/puppet/PuppetMaster.sol
alvrs: there could only be a single puppet for root systems (because they all have the world address)

packages/world-modules/src/utils/AccessControlLib.sol
alvrs: https://github.com/latticexyz/mud/issues/1853

PR #1799, opened 2023-10-19
feat(world-modules): add `totalSupply` table, use OpenZeppelin IERC20

alvrs:


alvrs:
accidentally force pushed over #1797

Issue #1719, opened 2023-10-09
PostDeploy should set store address


holic:
should add this to PostDeploy script
```solidity
StoreSwitch.setStoreAddress(worldAddress);
```
so that we can use tables directly in PostDeploy

PR #1817, opened 2023-10-24
fix(create-mud): set store address in PostDeploy script

0xrhsmt:
fixes https://github.com/latticexyz/mud/issues/1719

### DIFF
--- a/examples/minimal/packages/contracts/script/PostDeploy.s.sol
+++ b/examples/minimal/packages/contracts/script/PostDeploy.s.sol
@@ -3,11 +3,12 @@ pragma solidity >=0.8.21;
 
 import { Script } from "forge-std/Script.sol";
 import { console } from "forge-std/console.sol";
+import { StoreSwitch } from "@latticexyz/store/src/StoreSwitch.sol";
 import { ResourceId, WorldResourceIdLib } from "@latticexyz/world/src/WorldResourceId.sol";
 import { RESOURCE_SYSTEM } from "@latticexyz/world/src/worldResourceTypes.sol";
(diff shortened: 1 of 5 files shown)


holic (approved):

holic (approved):
thank you!

alvrs:
I think we might have to import `StoreSwitch` first

```solidity
```

examples/minimal/packages/contracts/script/PostDeploy.s.sol
```diff
@@ -12,6 +12,9 @@ import { ChatNamespacedSystem } from "../src/systems/ChatNamespacedSystem.sol";
 
 contract PostDeploy is Script {
   function run(address worldAddress) external {
+    // Specify a store so that you can use tables directly in PostDeploy
+    StoreSwitch.setStoreAddress(worldAddress);
```
holic, line 16: looks like this is missing an import of `StoreSwitch`

holic, line 16: looks like this is missing an import of `StoreSwitch`

holic (changes requested):

examples/minimal/packages/contracts/script/PostDeploy.s.sol
0xrhsmt: Sorry, fixed it. 🙏

0xrhsmt: Sorry, fixed it. 🙏

PR #1823, opened 2023-10-24
chore: switch to hardhat solidity vscode extension

holic:
- updates to hardhat solidity plugin for speed
- removes juanblanco-specific settings
- updates remappings because hardhat solidity didn't like the `./` path

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,1 @@
-{
-  "solidity.monoRepoSupport": true,
-  "solidity.compileUsingRemoteVersion": "v0.8.21+commit.d9974bed"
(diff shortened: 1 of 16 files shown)


.vscode/settings.json
```diff
@@ -1,4 +1 @@
-{
-  "solidity.monoRepoSupport": true,
-  "solidity.compileUsingRemoteVersion": "v0.8.21+commit.d9974bed"
-}
+{}
```
holic, line 1: I left the files in case we wanted to add some defaults but could also remove if you prefer

alvrs (approved):

Issue #1668, opened 2023-09-29
prototype new MUD config


alvrs:
### What is the MUD config

The MUD config includes information about tables, systems, modules, and other things in the World.
It is used by different tools. Store's `tablegen` script generates Solidity table libraries from the tables config, World's `worldgen` script generates interfaces for all Systems and combines them in a World interface. The deploy script reads the tables, systems, and modules from the config and registers them in the World.

### Current approach for config parsing

We currently use `zod` to parse, validate and expand the `mud.config.ts`. The config defined by users can include shorthand definitions, which the `zod` parser turns into the full definitions. 

Example: `zod` turns a shorthand table definition into the full config: 

https://github.com/latticexyz/mud/blob/feee27d453d4e49e4c9765ee0c191fbb3d1f20ad/packages/store/ts/config/storeConfig.ts#L66-L70
```ts
// Shorthand schema
const table = "uint256"

// Full schema
const table = {
  value: "uint256"
}
```

Unfortunately `zod` does not propagate strong types through the parsing function. (For example, `{ value: "uint256" }` becomes `{ value: string }`). This is a problem for us, because we want to infer types of MUD client libraries directly from the MUD config. This is why currently we have to define all of our types (and also the expanding logic) twice - once in `zod`, and once in typescript types.

Example: typescript types corresponding to the `zod` schema:
https://github.com/latticexyz/mud/blob/feee27d453d4e49e4c9765ee0c191fbb3d1f20ad/packages/store/ts/config/storeConfig.ts#L53-L60

### Solution draft

Our config parsing approach has been pretty complex and hard to extend and we'd like to rethink it from scratch. On a related note, we're planning to add plugin support to the MUD config, which is something we should keep in mind when designing the new config parsing approach.

Conceptually, the config has an input type, and an output type. The input type allows shorthands, the output type is the fully extended version of the config, including default values etc.

A `resolve` function can turn an input type into an output type. The resolve function needs to maintain strong types! (The current [`mudConfig`](https://github.com/latticexyz/mud/blob/main/packages/store/ts/register/mudConfig.ts) parser function is a `resolve` function in this sense, but it unfortunately doesn't maintain strong types.

[...]

PR #1826, opened 2023-10-25
feat(store): add experimental config resolve helper

alvrs:
- Aims to solve one of the main pain points of our current config parser, which breaks on user types in some cases
- Note that this is just a temporary solution, we're still planning to refactor our config parsing from ground up in the medium term (see #1668 for context)

CleanShot 2023-10-25 at 19 55 22

### DIFF
--- a/.changeset/dirty-cows-lie.md
+++ b/.changeset/dirty-cows-lie.md
@@ -1,0 +1,6 @@
+---
+"@latticexyz/store": minor
+---
+
+Added an experimental `resolveConfig` helper to refine the output type of `mudConfig` and simplify downstream consumption.
+Note that it's not recommended to use this helper externally yet, since the format is expected to change soon while we're refactoring the config parsing.
--- a/.changeset/dirty-cows-lie.md
+++ b/.changeset/dirty-cows-lie.md
@@ -1,6 +1,0 @@
----
-"@latticexyz/store": minor
----
-
-Added an experimental `resolveConfig` helper to refine the output type of `mudConfig` and simplify downstream consumption.
-Note that it's not recommended to use this helper externally yet, since the format is expected to change soon while we're refactoring the config parsing.
--- a/packages/common/src/utils/mapObject.ts
+++ b/packages/common/src/utils/mapObject.ts
@@ -5,9 +5,7 @@ export function mapObject<
   Source extends Record<string | number | symbol, unknown>,
   Target extends { [key in keyof Source]: unknown }
 >(source: Source, valueMap: (key: keyof Source, value: Source[typeof key]) => Target[typeof key]): Target {
-  const target: Record<string, unknown> = {};
-  for (const [key, value] of Object.entries(source)) {
-    target[key] = valueMap(key, value as Source[keyof Source]);
-  }
-  return target as Target;
+  return Object.fromEntries(
+    Object.entries(source).map(([key, value]) => [key, valueMap(key, value as Source[keyof Source])])
+  ) as Target;
 }
--- a/packages/common/src/utils/mapObject.ts
+++ b/packages/common/src/utils/mapObject.ts
@@ -4,8 +4,8 @@
 export function mapObject<
   Source extends Record<string | number | symbol, unknown>,
   Target extends { [key in keyof Source]: unknown }
->(source: Source, valueMap: (key: keyof Source, value: Source[typeof key]) => Target[typeof key]): Target {
+>(source: Source, valueMap: (value: Source[typeof key], key: keyof Source) => Target[typeof key]): Target {
   return Object.fromEntries(
-    Object.entries(source).map(([key, value]) => [key, valueMap(key, value as Source[keyof Source])])
+    Object.entries(source).map(([key, value]) => [key, valueMap(value as Source[keyof Source], key)])
   ) as Target;
 }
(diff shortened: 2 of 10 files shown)


packages/store/ts/config/experimental/resolveConfig.test-d.ts
```diff
@@ -0,0 +1,51 @@
+import { describe, expectTypeOf } from "vitest";
+import { mudConfig } from "../../register/mudConfig";
+import { resolveConfig } from "./resolveConfig";
+
+const config = resolveConfig(
+  mudConfig({
+    // Seems like we need `as const` here to keep the strong type.
+    // Note it resolves to the strong `""` type if no namespace is provided.
+    // TODO: require the entire input config to be `const`
```
holic, line 9: I remember trying to add `as const` to our default MUD configs (world config, store config) and it broke downstream types in weird ways. Hopefully we can just add this more specific `as const` to just the namespace without those downstream effects, but yeah, should aim to `as const` the whole config.

packages/common/src/utils/mapObject.ts
```diff
@@ -0,0 +1,13 @@
...
+ */
+export function mapObject<
+  Source extends Record<string | number | symbol, unknown>,
+  Target extends { [key in keyof Source]: unknown }
+>(source: Source, valueMap: (key: keyof Source, value: Source[typeof key]) => Target[typeof key]): Target {
+  const target: Record<string, unknown> = {};
+  for (const [key, value] of Object.entries(source)) {
+    target[key] = valueMap(key, value as Source[keyof Source]);
+  }
+  return target as Target;
```
holic, line 12: totally not necessary here and no idea if there are perf consequences but my most used pattern for this is `Object.fromEntries(Object.entries(...).map(...))` to keep it functionally pure/immutable

```ts 
return Object.fromEntries(Object.entries(source).map(([key, value]) => [key, valueMap(key, value)]))l
```

packages/store/ts/config/experimental/resolveConfig.test-d.ts
alvrs: yeah the arrays for e.g. `enums` currently complain if the whole config is `as const`

packages/common/src/utils/mapObject.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Map each key of a source object via a given valueMap function
+ */
+export function mapObject<
+  Source extends Record<string | number | symbol, unknown>,
+  Target extends { [key in keyof Source]: unknown }
+>(source: Source, valueMap: (key: keyof Source, value: Source[typeof key]) => Target[typeof key]): Target {
```
holic, line 7: slight preference for reversing the key/value order here since you most often want to mutate the value and can ignore the second key param, but not sure if that makes it harder to type `value` because it depends on `key`

packages/common/src/utils/mapObject.ts
alvrs: I like it! updated

packages/store/ts/config/experimental/resolveConfig.ts
```diff
@@ -0,0 +1,168 @@
...
+      config.tables[key],
+      config.userTypes,
+      Object.keys(config.enums),
+      config.namespace,
+      key
+    ) as ReturnType<typeof resolveTable>;
+  }
+
+  return {
+    ...config,
```
holic, line 85: since this function isn't being called by default as part of `mudConfig`, we could just not include `config` here since we'd have a reference to it already being part of the input and only return the `resolved` config (and then no need to worry about naming of `_resolved`)

.changeset/dirty-cows-lie.md
```diff
@@ -0,0 +1,6 @@
+---
+"@latticexyz/store": minor
+---
+
+Added an experimental `resolveConfig` helper to refine the output type of `mudConfig` and simplify downstream consumption.
+Note that it's not recommended to use this helper externally yet, since the format is expected to change soon while we're refactoring the config parsing.
```
holic, line 6: since this is internal/not recommended for external use, we could just not document this? 🙈

packages/common/src/utils/mapObject.ts
alvrs: updated!

packages/store/ts/config/experimental/resolveConfig.test-d.ts
holic: Do we wanna add `as const` to the world/store namespaces?

packages/store/ts/config/experimental/resolveConfig.ts
alvrs: agreed, updated

packages/store/ts/config/experimental/resolveConfig.ts
```diff
@@ -0,0 +1,162 @@
...
+          TUserTypes[TSchema[key]]["internalType"]
+        : never
+      : TSchema[key] extends TEnumNames
+      ? "uint8"
+      : TSchema[key];
+    internalType: TSchema[key];
+  };
+};
+
+export function resolveConfig<TStoreConfig extends StoreConfig>(
```
holic, line 66:
```suggestion

/**
 * @internal Internal only
 * @deprecated Internal only
 */
export function resolveConfig<TStoreConfig extends StoreConfig>(
```
do we wanna mark these as internal/deprecated for now?

holic (dismissed):
🙏  thank youuuuu

PR #1831, opened 2023-10-26
docs(gas-report): add stdin example to readme

holic:


alvrs (approved):

alvrs:
@qbzzt we can use this readme also for the CLI reference for `gas-report`

PR #1840, opened 2023-10-30
feat(store-sync): extra table definitions

holic:
before

image

after

image


packages/store/ts/config/storeConfig.ts
```diff
@@ -71,7 +71,7 @@ const zShorthandSchemaConfig = zFieldData.transform((fieldData) => {
 
 export const zSchemaConfig = zFullSchemaConfig.or(zShorthandSchemaConfig);
 
-export type ResolvedSchema<
+type ResolvedSchema<
```
holic, line 74: not used anywhere and the name overlaps with our experimental config types, so I removed this export

examples/minimal/packages/client-react/src/mud/setupNetwork.ts
```diff
@@ -40,10 +40,23 @@ export async function setupNetwork() {
   const { components, latestBlock$, storedBlockLogs$, waitForTransaction } = await syncToRecs({
     world,
     config: mudConfig,
+    tables: {
```
alvrs, line 43: so `syncToRecs` syncs all tables passed `mudConfig` or in `tables` correct? What in `mudConfig` do we need in here besides the resolved tables? If it's only that, should we just pass a tables key (ie `syncToRecs({ tables: { ...mudConfig.tables, ...additionalTables } })`)? (mostly wondering about longer term when the config is resolved by default, fine to not do that here yet to avoid breaking changes)

examples/minimal/packages/client-react/src/mud/setupNetwork.ts
```diff
@@ -40,10 +40,23 @@ export async function setupNetwork() {
...
+        valueSchema: {
+          keysWithValue: { type: "bytes32[]" },
+        },
+      },
+    },
     address: networkConfig.worldAddress as Hex,
     publicClient,
     startBlock: BigInt(networkConfig.initialBlockNumber),
-  });
+  } as const);
```
alvrs, line 59: what would break without `as const` here? Does this break anything for current users who don't have `as const` yet?

examples/minimal/packages/client-react/src/mud/setupNetwork.ts
holic: long term I think we should do that, yep! I just didn't want to encourage the intermediate step of using the experimental `resolveConfig` in user land

examples/minimal/packages/client-react/src/mud/setupNetwork.ts
holic: the manually defined tables were missing strong types for namespace/name without this

packages/store-sync/src/recs/recsStorage.ts
```diff
@@ -9,39 +9,40 @@ import { Hex, size } from "viem";
...
+export type RecsStorageOptions<tables extends Record<string, Table>> = {
   world: RecsWorld;
-  // TODO: make config optional?
-  config: TConfig;
+  tables: tables;
   shouldSkipUpdateStream?: () => boolean;
 };
 
-export type RecsStorageAdapter<TConfig extends StoreConfig = StoreConfig> = {
+export type RecsStorageAdapter<tables extends Record<string, Table>> = {
```
alvrs, line 27: I'm still not used to `tables` instead of `TTables` 😭 Don't you find it confusing not to see on first sight whether you have to use `typeof tables` vs just `tables`? I'm just so used to custom types being uppercase and only primitive types being lowercase

alvrs (dismissed):

PR #1843, opened 2023-10-30
feat(store-sync): sync to zustand

holic:
see changeset


packages/store-sync/src/zustand/createStore.ts
```diff
@@ -0,0 +1,74 @@
...
+export type CreateStoreOptions<tables extends Tables> = {
+  tables: tables;
+};
+
+export function createStore<tables extends Tables>(opts: CreateStoreOptions<tables>): ZustandStore<tables> {
+  return create<ZustandState<tables>>((set, get) => ({
+    tables: {},
+    rawRecords: {},
+    records: {},
+    getRecords: <table extends Table>(table: table): TableRecords<table> => {
```
alvrs, line 49: ok maybe i'm slowly getting used to `table extends Table`

packages/store-sync/src/zustand/createStore.ts
```diff
@@ -0,0 +1,74 @@
...
+      table: table,
+      key: SchemaToPrimitives<table["keySchema"]>
+    ): TableRecord<table> | undefined => {
+      const records = Object.values(get().records);
+      const expectedKeyTuple = encodeKey(flattenSchema(table.keySchema), key);
+      const record = records.find(
+        (record) =>
+          record.table.tableId === table.tableId && concatHex([...record.keyTuple]) === concatHex(expectedKeyTuple)
+      );
+      return record as TableRecord<table> | undefined;
```
alvrs, line 65: i'm a little bit scared of performance of finding records via linear search if the table has a lot of entries (maybe not the use case of the zustand store? but still).

Couldn't we make the `id` just the concatenated tableId + key to be able to look it up immediately? What are we solving for by adding the address? (which i think is the reason why we can't just look up the record by id here correct?) If we had records from multiple worlds in here, then I think the current approach would just return the first record independent of the `address` part of its id?

alvrs (approved):
comments are not blocking but i'm curious about the [id/`getRecord` approach](https://github.com/latticexyz/mud/pull/1843/files#r1379375566)

packages/store-sync/src/zustand/createStore.ts
holic: I considered this, but all of the existing sync adapters support an optional address (i.e. either one world or all worlds), so this was here to keep compatibility/predictability. We could make `address` a required param for `syncToZustand` and then get rid of the address as part of the ID/object key and that would make look ups faster.

Now that I am writing this, I am realizing that even though we key on `address`, we don't actually return it anywhere or allow filtering on it in `getRecord`, `getRecords`, etc. So maybe it is best if we make this a required param for now, and figure out how to make it optional/multi-world later.

Another thing on my mind is potentially keeping the top-level keys (`records`, `tables`, etc.) "out of view" of downstream consumers and forcing them to use things like `getRecord` and `getRecords` so we can modify the underlying state representation.

alvrs (approved):

PR #1851, opened 2023-11-01
feat(create-mud): move react template to zustand, add react-ecs template

holic:
Moved the current `react` template to `react-ecs` and converted the `react` template to use Zustand by default.

image


alvrs (approved):
🎉

alvrs (approved):

PR #1869, opened 2023-11-03
docs(cli): add config to newdocs

qbzzt:


next-docs/pages/cli/config.mdx
```diff
@@ -1 +1,21 @@
 # Config
+
+To install the MUD command line interface (CLI) into your environment, run this command:
+
+```sh copy
+pnpm install -g @latticexyz/cli@next
+```
+
+Alternatively, you can install `@latticexyz/cli` as part of a Node project, in which case you can access it from the project's root using either `pnpm mud` or `./node_modules/.bin/mud`
```
alvrs, line 9: This is the recommended approach since the version needs to align with the project. Wonder if we should just remove the global installation to avoid confusion with misaligned versions.

alvrs (commented):
I'd move this page to a `installation` or `setup` page! For the `config` page what i had in mind was explaining the setup of the `mud.config.ts` file, which many of the cli commands use to execute their logic (ie `tablegen`, `worldgen`, `deploy`, `test`)

qbzzt:
> I'd move this page to a `installation` or `setup` page! For the `config` page what i had in mind was explaining the setup of the `mud.config.ts` file, which many of the cli commands use to execute their logic (ie `tablegen`, `worldgen`, `deploy`, `test`)

I don't think the config file documentation belongs in the CLI area. We need to think about it, but this isn't the right place.

next-docs/pages/cli/config.mdx
alvrs: Did you push the change? I still see the global installation instructions here (line 5-7)

alvrs: To be clear: the recommended way of using the MUD cli is to install it as a dev dependency in the project and then run via `pnpm mud` from the project root

qbzzt:
Use https://mud.dev/world/config

qbzzt:
Replaced by #1940

PR #1871, opened 2023-11-03
feat(cli): add the batch registration feature to systems, accesses, functions and tables

CrazyNorman:
Fixes issue #1645.

There are some parts that can be further examined.

One is the `batchCall` functiin, which is not easy to invoke correctly in Typescript. To address this, I added a helper function `getBatchCallData` to alleviate this issue.

The other part is the granularity of batching. For systems and access controls, I batched the system and its corresponding access changes together. This approach ensures atomicity of the system and its access control changes, resulting in the invocation of a number of system transactions.

For function registrations, I batched the functions per system, as it seems that a system can naturally divide the functions.

For table registrations, I used a static slicing approach. The number of packed registrations for now is 10, which is a conservative figure designed to adapt to most chain environments. This number could be added to advanced configurations in the future.

PS: I have temporarily set aside the part related to modules, as there is currently no filter available for installedModules.


holic:
closing in favor of batching RPC calls, which reduces complexity

PR #1876, opened 2023-11-07
fix(store): resolveUserTypes for static arrays

### LINKED ISSUE #1875: deploy breaks for static-length arrays

dk1a:
fixes #1875


packages/store/ts/config/storeConfig.ts
```diff
@@ -88,7 +88,18 @@ export function resolveUserTypes<
...
+
+    if (userTypes[value] !== undefined) {
+      // user-defined types
+      resolvedSchema[key] = userTypes[value].internalType as SchemaAbiType;
+    } else if (staticArray) {
+      // static arrays
+      resolvedSchema[key] = `${staticArray.elementType as StaticAbiType}[]`;
+    } else {
+      // abi types
+      resolvedSchema[key] = value as SchemaAbiType;
```
holic, line 101: we have a type-narrowing `isSchemaAbiType` that could help here

may also be slightly faster to put this as the first check and only parse the array if we don't have a schema ABI type or user type

packages/store/ts/config/storeConfig.ts
```diff
@@ -88,7 +88,18 @@ export function resolveUserTypes<
...
+    }
```
holic, line 102:
```suggestion
    const staticArray = parseStaticArray(value);

    if (isSchemaAbiType(value)) {
      resolvedSchema[key] = value;
    else if (userTypes[value] !== undefined) {
      resolvedSchema[key] = userTypes[value].internalType as SchemaAbiType;
    } else {
      const staticArray = parseStaticArray(value);
      if (!staticArray) throw new Error(`Unexpected type: ${value}`);
      resolvedSchema[key] = `${staticArray.elementType as StaticAbiType}[]`;
    }
```

e2e/packages/contracts/mud.config.ts
```diff
@@ -47,5 +47,9 @@ export default mudConfig({
         player: "address",
       },
     },
+    StaticArray: {
+      keySchema: {},
+      valueSchema: "uint256[2]",
+    },
```
holic, line 53: do we have or want tests for this? or are e2e tests just being compiled/run enough?

holic (dismissed):
just a small nit around logic and type narrowing but otherwise lgtm!

packages/store/ts/config/storeConfig.ts
holic: oh actually, I think this file deals with old schema ABI type abstractions, so that may not work

e2e/packages/contracts/mud.config.ts
dk1a: the table being deployed is enough, which happens in just e2e tests being run

packages/store/ts/config/storeConfig.ts
dk1a: added `isSchemaAbiType` and imported and used the suggestion

holic (dismissed):

holic (approved):

PR #1885, opened 2023-11-09
feat(world-modules): system bound delegation control

yonadaaa:
Creates a new delegation control called `SystemboundDelegationControl` that delegates control of a specific system for some maximum number of calls. It is almost identical to `CallboundDelegationControl` except the delegation is not for specific `calldata` - the delegatee can call the system with any function and args.

The tests are mainly copied from callbound delegations but check that you can use a single delegation to call multiple functions on a system.


packages/world-modules/mud.config.ts
```diff
@@ -1,5 +1,9 @@
 import { mudConfig } from "@latticexyz/world/register";
 
+// loading state
+// promise for buttons
+// design system
+// admin
```
holic, line 6: whatcha doin yonada

yonadaaa: committing to git so I don't forget :P

holic: so _no one_ forgets

packages/world-modules/src/modules/std-delegations/SystemboundDelegationControl.sol
```diff
@@ -0,0 +1,54 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.21;
+
+import { DelegationControl } from "@latticexyz/world/src/DelegationControl.sol";
+import { ResourceId } from "@latticexyz/world/src/WorldResourceId.sol";
+import { SystemboundDelegations } from "./tables/SystemboundDelegations.sol";
+
+contract SystemboundDelegationControl is DelegationControl {
+  /**
+   * Verify a delegation by checking if the delegator has any available calls left in the SystemboundDelegations table and decrementing the available calls if so.
```
alvrs, line 10:
```suggestion
   * Verify a delegation by checking if the caller (delegatee) has any available calls left in the SystemboundDelegations table and decrementing the available calls if so.
```

alvrs (dismissed):

PR #1897, opened 2023-11-10
fix(faucet,store-indexer): build binaries as full bundles

holic:
- adds separate `tsup.config.ts` for `bin` dirs that builds each file as a full bundle (deps included)
- adds command to build/clean bins

just a note that `dev` script won't watch these yet (since these are generally run with `tsx` locally instead of the file in `dist/bin`) and unclear how to run multiple dev runners/watchers with our current setup (and tsup doesn't let us configure each entrypoint individually)


holic:
fixed well enough in https://github.com/latticexyz/mud/pull/1900

PR #1910, opened 2023-11-11
feat(cli): link and deploy public libraries

dk1a:
fixes https://github.com/latticexyz/mud/issues/1911


packages/cli/src/deploy/getPublicLibraries.ts
```diff
@@ -0,0 +1,79 @@
...
+    contractFullyQualifiedName: string;
+    libraryFullyQualifiedName: string;
+  }[] = [];
+  const files = glob.sync(`${forgeOutDir}/**/*.json`, { ignore: "**/*.abi.json" });
+
+  for (const contractOutPath of files) {
+    const json = JSON.parse((await readFile(contractOutPath, "utf8")).trim());
+    const linkReferences = json.bytecode.linkReferences as LinkReferences;
+
+    const contractFullPath = Object.keys(json.metadata.settings.compilationTarget)[0];
```
yonadaaa, line 23: Previously this fetched `contractFullPath` [from `json.ast.absolutePath;`](https://github.com/latticexyz/mud/commit/5f5a0c156515c66f59029ce6fc9c46d7e410d66f), but the ABI's don't have that `ast` property anymore. Maybe Foundry removed it?

I don't love this change but it seems to work 🤔

dk1a: this is actually the first version I had
later I switched it to ast cause an array deep in input metadata felt less reliable, but either way works fine afaik
we could probably add `ast` back to output, but that would require overriding forge defaults, which seems worse

yonadaaa: ah interesting, I didn't know there was a Forge option for this. I will try it and see what feels best 🤔

packages/cli/src/utils/utils/getContractData.ts
```diff
@@ -21,12 +34,40 @@ export function getContractData(
...
+) {
+  let result = bytecode;
+  for (const [filename, referencedLibraries] of Object.entries(linkReferences)) {
+    for (const name of Object.keys(referencedLibraries)) {
+      const fullyQualifiedName = `${filename}:${name}`;
+      const placeholderData = libraries.find((library) => library.fullyQualifiedName === fullyQualifiedName);
+      if (placeholderData === undefined) {
+        throw new Error(`Unlinked library ${fullyQualifiedName}`);
+      }
+      const trimmedAddress = placeholderData.getAddress(deployerAddress).slice(2).toLowerCase();
```
yonadaaa, line 68: To get this address, I had to change when the deterministic deployer is ensured, and pass `deployerAddress` all the through from `runDeploy` => `resolveConfig` => `getPublicLibraries` => `getContractData` => `linkLibraries`

curious if there is a better way

packages/cli/src/deploy/getPublicLibraries.ts
dk1a: since compilationTarget is an input to solc, this might be a good question for foundry devs or someone familiar with its source
I assume it only ever has a single key, and ur approach is fine

holic: I am seeing this when running tests:
```
│ /Users/kevin/Projects/latticexyz/mud/packages/cli/src/deploy/getPublicLibraries.ts:22
│     const contractFullPath = Object.keys(json.metadata.settings.compilationTarget)[0];
│                                                        ^
│ TypeError: Cannot read properties of undefined (reading 'settings')
```

packages/cli/src/utils/modules/constants.ts
```diff
@@ -9,18 +9,21 @@ export const defaultModuleContracts = [
     name: "KeysWithValueModule",
     abi: KeysWithValueModuleData.abi as Abi,
     bytecode: KeysWithValueModuleData.bytecode.object as Hex,
+    placeholders: [], // TODO
```
yonadaaa, line 12: Why is this TODO? I can see us revisiting this and not knowing what we need "to do"

holic: will clarify! this isn't used/needed right now but would need to be filled in if these modules ever start using public libs

holic: went ahead and just moved out the placeholder logic so I could reuse it here: https://github.com/latticexyz/mud/pull/1910/commits/4fea2b3f9110f8b8c201194926658934802439ad

PR #1919, opened 2023-11-12
fix(store-sync): create table registration logs from indexer records

holic:
Not sure how hydrating from the indexer ever worked without this but some parts of the sync stack expect table registration records to fill in tables before capturing values from those tables. We spotted this issue when switching from sqlite indexer to postgres indexer and I have no idea why this would have surfaced this issue.


holic:
yolo merging to unblock sky strife

PR #1927, opened 2023-11-12
docs(guides): add emojimon to newdocs

qbzzt:


alvrs (approved):

alvrs:
prettier/vercel don't like commenting code out in mdx, so i removed those sections for now (we can always still access it via the git history)

PR #1929, opened 2023-11-13
fix(cli): always rebuild IWorld ABI

dk1a:
hack to rebuild `IWorld` abi until https://github.com/foundry-rs/foundry/issues/6241 is fixed


packages/cli/src/runDeploy.ts
```diff
@@ -63,6 +64,10 @@ export async function runDeploy(opts: DeployOptions): Promise<WorldDeploy> {
   if (!opts.skipBuild) {
     const outPath = path.join(srcDir, config.codegenDirectory);
     await Promise.all([tablegen(config, outPath, remappings), worldgen(config, getExistingContracts(srcDir), outPath)]);
+
+    // TODO remove when https://github.com/foundry-rs/foundry/issues/6241 is resolved
+    await rm(path.join(outDir, "IWorld.sol"), { recursive: true, force: true });
```
holic, line 69: I think @alvrs and I tested this and this basically kills the cache/causes it to do a fresh build. Did you happen to try and see how long this takes?

As an alternative, we played around with unsetting the forge cache's hash or timestamp for this file and it seemed to regenerate just this file afterward (~1s vs full build of ~5s).

dk1a: > Did you happen to try and see how long this takes?

a while, but my laptop is just slow, not a good benchmark

> As an alternative, we played around with unsetting the forge cache's hash or timestamp for this file and it seemed to regenerate just this file afterward (~1s vs full build of ~5s).

yeah I wanted to look into this but I dunno if I even have forge cache or where it is
`forge cache ls` returns nothing for me

alvrs: for me the cache was in the project directory

holic (approved):

PR #1933, opened 2023-11-13
docs: move next-docs to docs

alvrs:


docs/next.config.js
```diff
@@ -5,4 +5,149 @@ const withNextra = nextra({
...
+        permanent: true        
+      },
+      {
+        source: "/tutorials/emojimon",
+        destination: "/guides/emojimon",
+        permanent: true        
+      },      
+      {
+        source: "/tutorials/emojimon/preface-the-ecs-model",
+        destination: "/guides/emojimon/1-preface-the-ecs-model.mdx",
```
holic, line 148: this url looks weird - are these correct?

docs/next.config.js
```diff
@@ -5,4 +5,149 @@ const withNextra = nextra({
...
 });
 
-export default withNextra();
+export default withNextra({
+  async redirects() {
+    return [
+      {
+        source: "/what-is-mud",
+        destination: "/introduction",
+        permanent: true,
```
holic, line 14: imo while things are still in flux, would recommend using `permanent: false` 

browsers, search engines, CDNs, etc. heavily cache permanent redirects and they're hard to unwind

docs/next.config.js
qbzzt: No, the extension shouldn't be there. I'll fix it.

docs/next.config.js
qbzzt: I disagree, because I don't think we'll go back to the old docs, so the redirects will stay accurate until we redo the doc outline again, and hopefully that will be a very long time from now.

holic: I am not saying we'll go back to the old docs, but we basically kill any chance of using this URL in the future. With these redirects, we lose the ability in the future to have a page on `/store` for example.

qbzzt: You convinced me. I'll change that.

PR #1940, opened 2023-11-14
docs(cli): add the configuration file

qbzzt:


docs/pages/cli/config.mdx
```diff
@@ -1 +1,94 @@
...
+Certain CLI commands, such as [`mud tablegen`](/cli/tablegen) and [`mud worldgen`](/cli/worldgen) require the MUD configuration file.
+This file needs to be named `mud.config.ts` and be in the same folder as your `foundry.toml` file.
+
+The config is used to define:
+
+- The tables in your project in the `tables` object of your configuration.
+  See [the `Store` documentation](/store/tables) for more information on table and their format.
+- The [namespace](/world/namespaces-access-control) the [`System`s](/world/systems) and tables will be deployed in.
+- The `System`s in your project.
+  By default, the deployer will find all Solidity matching `*System*.sol` (so any file containing `System` with a `.sol` extension, in any folder) and deploy them as `System`.
```
alvrs, line 12:
```suggestion
  By default, the deployer will find all Solidity matching `*System.sol` (so any file containing `System` with a `.sol` extension, in any folder) and deploy them as `System`.
```

docs/pages/cli/config.mdx
```diff
@@ -1 +1,94 @@
...
+  If you want greater control over your systems (to change their public access or their name), you can use the `systems` object in the config
```
alvrs, line 13:
```suggestion
  If you want greater control over your systems (to change their public access or their name), you can use the `systems` object in the config.
```

docs/pages/cli/config.mdx
```diff
@@ -1 +1,94 @@
...
+- The [modules](/world/modules) that will be installed on the `World`.
```
alvrs, line 14:
```suggestion
- The [modules](/world/modules) that will be installed in the `World`.
```

docs/pages/cli/config.mdx
```diff
@@ -1 +1,94 @@
...
+      name: "increment",
+      openAccess: true,
+    },
+  },
+  tables: {
+    CounterTable: {
+      valueSchema: {
+        value: "uint32",
+      },
+    },
```
alvrs, line 37: should we add all table options here?

docs/pages/cli/config.mdx
```diff
@@ -1 +1,94 @@
...
+
+- **`modules`** an array of module definitions: each module definition has a `name`, `root` (optional), and `args` key.
+
+  - `name`: Name of the module to instantiate. The same module can be instantiated multiple times. This should be the name of the contract file without `.sol` (eg: if the contract is named `DopeModule.sol`, the name of the module is `DopeModule`)
+
+  - `root`: whether to create a `root` module or not. `root` modules have access to all tables and are not bound to namespace restrictions.
+
+  - `args`: a list of arguments to be sent to the `install` function of the module. In this array, you can use the function `resolveTableId`. This function will turn a table name from your config into its low-level ID in the World. It is useful to pass references of a table to a module.
+
+- **`systems`**: a dictionary of system definitions. The keys in the array are file names without the `.sol` extension. For example, if your system is named `SystemA.sol`, use `SystemA` as the.
```
alvrs, line 71:
```suggestion
- **`systems`**: an object of system definitions. The keys in the object are file names without the `.sol` extension. For example, if your system is named `SystemA.sol`, use `SystemA` as the key.
```

docs/pages/cli/config.mdx
```diff
@@ -1 +1,94 @@
...
+
+  The value is a dictionary of system configuration properties:
```
alvrs, line 73:
```suggestion
  The value is a record of system configuration properties:
```

docs/pages/cli/config.mdx
```diff
@@ -1 +1,94 @@
...
+
+  - `fileSelector` (optional): a `string`: the file selector for the system.
+  - `openAccess` (optional, default `true`): a `bool`: if set to `false`, only the systems in the same namespace and the addresses or systems listed from the accessList.
```
alvrs, line 76:
```suggestion
  - `openAccess` (optional, default `true`): a `bool`: if set to `false`, only the systems in the same namespace and the addresses or systems listed in the the `accessList` array have access.
```

docs/pages/cli/config.mdx
```diff
@@ -1 +1,94 @@
...
+  export default mudConfig({
+    systems: {
+      SystemA: {
+        name: "systema",
+        openAccess: true,
+      },
+    },
+    tables: {},
+  });
+  ```
```
alvrs, line 94: I think we don't need another example here, but rather add all the options the the example above. (Alternatively we could have small examples for each option, but I don't think it's necessary)

docs/pages/cli/config.mdx
qbzzt: On this page? Definitely - my bad. In the example? No. It's best to keep the code sample simple.

alvrs (approved):

PR #1941, opened 2023-11-15
fix(store-indexer): revert postgres changes

holic:


holic:
will be replaced by https://github.com/latticexyz/mud/pull/1965

PR #1953, opened 2023-11-22
docs: claify link text (remove `here`)

qbzzt:


docs/pages/guides/extending-world.mdx
```diff
@@ -6,7 +6,7 @@ If you want to learn how to modify a `World` before it is deployed, [see the hel
 ## The sample program
 
 To learn how to extend a `World`, we will extend the `Counter` example to allow users to leave a message while incrementing the counter.
-The steps to create the example `World` that we'll extend [are here](/templates/typescript/getting-started).
+[See the steps to create the example `World` that we'll extend](/templates/typescript/getting-started).
```
holic, line 9: "see" is basically the same thing as "here"

We could restructure this to avoid these terms, something like

> The [TypeScript templates] start with a `World` and example "counter" app. And now we will extend the counter app to allow users to leave a message while incrementing the counter.

qbzzt: I disagree. The problem isn't "see" or "here". The problem is non-descriptive links. "are here" is bad, and "see here" would be no better. But "See the steps to create ..." is a link that includes exactly the reason we are linking to that page.

I don't want to just tell them we're linking to the TypeScript templates. I want to tell them that these are the steps they need to use to create the example world. I'll change the verb to "follow".

alvrs (dismissed):

alvrs (approved):

PR #1954, opened 2023-11-23
docs(world/systems): explain `SwitchSystem`

qbzzt:


docs/pages/world/systems.mdx
```diff
@@ -54,6 +54,61 @@ There are two ways to call one `System` from another one.
...
+
+   ```solidity copy
+   import { SystemSwitch } from "@latticexyz/world-modules/src/utils/SystemSwitch.sol";
+   ```
+
+1. Call the function using `SystemSwitch.call`.
+   For example, here is how you can call [`IncrementSystem.increment()`](http://localhost:3000/templates/typescript/contracts#incrementsystemsol).
+
+   ```solidity copy
+    uint32 retval = abi.decode(
```
alvrs, line 71:
```suggestion
    uint32 returnValue = abi.decode(
```

docs/pages/world/systems.mdx
```diff
@@ -54,6 +54,61 @@ There are two ways to call one `System` from another one.
...
+      SystemSwitch.call(
+        abi.encodeWithSignature("increment()")
```
alvrs, line 73: I think we should recommend usage of `abi.encodeCall` over `abi.encodeWithSignature` because it gives better type safety. The "downside" is it requires importing the system/interface to call (or to define an interface locally).

alvrs (approved):

Issue #1857, opened 2023-11-01
bug(recs): `useComponentValue` doesn't trigger when record is deleted


0xhank:
the useComponentValue hook doesn't trigger a rerender when I delete a record from a table. RECS systems still work, however.

PR #1959, opened 2023-11-23
fix(react): trigger useComponentValue on deleted records

emersonhsieh:
- Fixes https://github.com/latticexyz/mud/issues/1857 when deleting records

### DIFF
--- a/packages/react/src/useComponentValue.ts
+++ b/packages/react/src/useComponentValue.ts
@@ -33,7 +33,7 @@ export function useComponentValue<S extends Schema>(
     setValue(entity != null ? getComponentValue(component, entity) : undefined);
     if (entity == null) return;
 
-    const queryResult = defineQuery([Has(component)], { runOnInit: false });
+    const queryResult = defineQuery([Has(component)], { runOnInit: true });
     const subscription = queryResult.update$.subscribe((update) => {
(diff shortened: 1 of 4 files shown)


packages/react/src/useComponentValue.ts
```diff
@@ -33,7 +33,7 @@ export function useComponentValue<S extends Schema>(
     setValue(entity != null ? getComponentValue(component, entity) : undefined);
     if (entity == null) return;
 
-    const queryResult = defineQuery([Has(component)], { runOnInit: false });
+    const queryResult = defineQuery([Has(component)], { runOnInit: true });
```
holic, line 36: do we know if this works for adding/deleting records after the initial delete? I suspect this might just help with the first one, but I could be wrong

emersonhsieh: The fix still has `useComponentValue` update if the same record is added after they're deleted. In MUDv1 where the issue had already existed, the issue happened when a record that already existed on init was then deleted, which `useComponentValue` didn't update without `{ runOnInit: true }`. If a record didn't already exist upon init, was created, and then deleted, then the hook correctly updated even without the fix.

holic: do you wanna see if you can wire up a test to show this behavior and so we don't regress on this?

holic (approved):

PR #1964, opened 2023-11-27
feat(create-mud): Svelte template 

0xMugen:
add svelte template with a few writables for abstraction:
- useSystemCalls and useComponents functions
- subscribeToComponentUpdate(component: any, store: any) that updates the selected writable.


holic:
Thanks for the PR! We're not using Svelte anywhere, so we won't have the expertise or bandwidth to maintain this long term.

Instead, you might consider making a template repo that people can fork/clone from: https://docs.github.com/en/repositories/creating-and-managing-repositories/creating-a-template-repository

Happy to link to this template repo once you've created it!

0xMugen:
Ok, no problem. Always thought that Mud would work really well with svelte writables rather than react context. It's extremely similar to native html/css with just a reative overlay.
I'll create the template as you recomend!

PR #1965, opened 2023-11-27
feat(store-sync,store-indexer): schemaless indexer

### LINKED ISSUE #1640: add schemaless indexer

holic:
closes #1640 

The client libs don't use the indexer's new `getLogs` method yet, planning to do that in a follow up PR.


packages/store-indexer/src/postgres/getLogs.ts
```diff
@@ -0,0 +1,69 @@
...
+  const logs = records
+    .filter((record) => !record.isDeleted)
+    .map(
+      (record) =>
+        ({
+          address: record.address,
+          eventName: "Store_SetRecord",
+          args: {
+            tableId: record.tableId,
+            keyTuple: decodeDynamicField("bytes32[]", record.keyBytes),
```
alvrs, line 60: at some point we should do benchmarks of this hot path, I hope the decoding here doesn't become a bottleneck if a large number of records is requested

packages/store-sync/src/postgres/createStorageAdapter.ts
```diff
@@ -0,0 +1,218 @@
...
+import { PgDatabase, QueryResultHKT } from "drizzle-orm/pg-core";
+import { and, eq } from "drizzle-orm";
+import { StoreConfig } from "@latticexyz/store";
+import { debug } from "./debug";
+import { tables } from "./tables";
+import { spliceHex } from "@latticexyz/common";
+import { setupTables } from "./setupTables";
+import { StorageAdapter, StorageAdapterBlock } from "../common";
+
+// Currently assumes one DB per chain ID
```
alvrs, line 11: is this comment still accurate?

packages/store-indexer/src/postgres/getLogs.ts
```diff
@@ -0,0 +1,69 @@
...
+    .select()
+    .from(tables.recordsTable)
+    .where(or(...conditions));
+  blockNumber = bigIntMax(
+    blockNumber,
+    records.reduce((max, record) => bigIntMax(max, record.lastUpdatedBlockNumber ?? 0n), 0n)
+  );
+
+  const logs = records
+    .filter((record) => !record.isDeleted)
```
alvrs, line 52: should we just include `!isDeleted` in the sql query?

packages/store-sync/src/postgres/tables.ts
```diff
@@ -0,0 +1,44 @@
...
+    key0: asHex("key0"),
+    key1: asHex("key1"),
+    staticData: asHex("static_data"),
+    encodedLengths: asHex("encoded_lengths"),
+    dynamicData: asHex("dynamic_data"),
+    isDeleted: boolean("is_deleted"),
+    lastUpdatedBlockNumber: asBigInt("last_updated_block_number", "numeric"),
+  },
+  (table) => ({
+    pk: primaryKey(table.address, table.tableId, table.keyBytes),
```
alvrs, line 33: if the primary key is the combination of these three columns, do we need individual indices on `address` and `tableId` for efficient querying on those (where we don't include the `keyBytes` in the query)?

packages/store-sync/src/postgres/tables.ts
```diff
@@ -0,0 +1,44 @@
...
+    key0Index: index("key0_index").on(table.address, table.tableId, table.key0),
+    key1Index: index("key1_index").on(table.address, table.tableId, table.key1),
+    // TODO: add indices for querying without table ID
+    // TODO: add indices for querying multiple keys
```
alvrs, line 37: should we also add an index for `isDeleted` to be able to efficiently use it in `getLogs` instead of having to filter the array in js?

packages/store-indexer/bin/postgres-indexer.ts
```diff
@@ -37,37 +37,26 @@ const publicClient = createPublicClient({
...
     .select()
-    .from(internalTables.chain)
-    .where(eq(internalTables.chain.chainId, chainId))
-    .execute();
-  // TODO: replace this type workaround with `noUncheckedIndexedAccess: true` when we can fix all the issues related (https://github.com/latticexyz/mud/issues/1212)
-  const currentChainState: (typeof currentChainStates)[number] | undefined = currentChainStates[0];
+    .from(tables.chainTable)
+    .where(eq(tables.chainTable.chainId, chainId))
+    .execute()
+    .then((rows) => rows.find(() => true));
```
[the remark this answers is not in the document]
holic: I'll leave some comments around but this basically sidesteps our lack of `noUncheckedIndexedAccess: true`.

``` ts
const rows = query(...);
const row = rows[0]; // type will just be `Row`, which is inaccurate for an empty array
// vs
const row = rows.find(() => true); // type will be `Row | undefined`
```

packages/store-indexer/src/postgres/createQueryAdapter.ts
```diff
@@ -14,70 +14,39 @@ import { decodeDynamicField } from "@latticexyz/protocol-parser";
...
-      const internalTables = buildInternalTables();
-      const metadata = await database
-        .select()
-        .from(internalTables.chain)
-        .where(eq(internalTables.chain.chainId, chainId))
-        .execute();
-      const { lastUpdatedBlockNumber } = metadata[0] ?? {};
+      const tablesWithRecords: TableWithRecords[] = tables.map((table) => {
+        const records = logs
+          .filter((log) => getAddress(log.address) === getAddress(table.address) && log.args.tableId === table.tableId)
```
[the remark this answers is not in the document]
holic: This does less querying than we were doing before but for the same amount of data.

Before when records were distributed across many tables, we had to query for a list of tables then query each table individually for their records.

Now we query for all logs (passing the filters to the SQL query), then we just group them by table for the purposes of `findAll`.

This is mostly a backwards compat thing for `findAll` with the new table structure. The clients will start using the new `getLogs` endpoint in a PR coming soon.

packages/store-indexer/bin/postgres-indexer.ts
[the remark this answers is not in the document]
alvrs: ohh so it's essentially "give me the first item if it exists"; i had head it as "filter"

packages/store-indexer/src/postgres/getLogs.ts
```diff
@@ -0,0 +1,69 @@
...
+  let blockNumber = chainState?.lastUpdatedBlockNumber ?? 0n;
+
+  const records = await database
+    .select()
+    .from(tables.recordsTable)
+    .where(or(...conditions));
+  blockNumber = bigIntMax(
+    blockNumber,
+    records.reduce((max, record) => bigIntMax(max, record.lastUpdatedBlockNumber ?? 0n), 0n)
+  );
```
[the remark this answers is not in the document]
holic: I only did this here because of the order of queries that are run separately, one for the chain's current block number and one for all the records. I could probably do this at query time (i.e. a single query), but I was being lazy about crafting that query since Drizzle is a bit clunky/annoying with joins.

The issue I am solving for is that the indexer may be up to date for a chain but a world may not have had any recent activity. If we query for just records and return the highest block number of those records, that would signal to the client that it needs to start fetching from the RPC from a much older block number than is necessary.

So basically: we get the chain's block number first, as the "minimum" block number value. Then we get all the records and their highest block number. If it's higher than the chain, then we know some data updated between our two queries and we should use the higher number. If it's lower than the chain, we use the chain block number, because that's where the indexer is at in terms of up-to-date data.

packages/store-indexer/src/postgres/getLogs.ts
holic: I doubt it, decoding individual fields is fairly lightweight. This one just chunks the concatenated hex into individual bytes32 hex items. There's room to optimize that method but I doubt this is where our bottleneck is.

packages/store-sync/src/postgres/createStorageAdapter.ts
holic: yes

packages/store-indexer/src/postgres/createQueryAdapter.ts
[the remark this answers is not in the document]
alvrs: Ahh gotcha, this part is just grouping the logs by table, not really filtering the data. Even though this is mostly used for backwards compatibility, I wonder should we slightly change the logic here to iterate through the logs once and group them by table (ie have an object with `address/tableId` as key and array as value, iterate over the logs to push to the right array, and then do `Object.values`), instead of iterating through all logs `number of table` times?

packages/store-indexer/src/postgres/getLogs.ts
holic: I intentionally didn't for a few reasons:
- on the whole, I expect the number of actual deleted records to be low, because it costs gas to delete vs leaving a record laying around
- doing it at the query level will change the query planning and I want to make sure we hit the indexes I've set up here (not to say that we _can't_ but I wanna make sure we can make use of the indexes)
- we'll be doing similar shaped queries for the update stream (query all records and filter or return whether a record is deleted) and seems wise to use the same SQL query path for caching

packages/store-sync/src/postgres/tables.ts
holic: the indexes below this handle that and I believe that an index on `[address, tableId, key0]` can still benefit a query using `address=... AND tableId=...`

packages/store-sync/src/postgres/tables.ts
holic: see comment above

packages/store-sync/src/postgres/createStorageAdapter.ts
alvrs: I was confused about the `chain` table but I now understand the expectation is to have only one row in that table, and we don't have a `chainId` column in the data table so couldn't support multiple chains in the same data table with the current schema

packages/store-indexer/src/postgres/getLogs.ts
alvrs: > on the whole, I expect the number of actual deleted records to be low, because it costs gas to delete vs leaving a record laying around

Independent of the number of deleted records it would save us one linear operation over the entire logs array if we don't have to filter for `isDeleted` after fetching all the logs. If we'd add another index for `isDeleted` we could avoid that linear operation, at the cost of more expensive inserts (but since inserts are throttled by the blockchain, write performance is less critical than read performance)

> we'll be doing similar shaped queries for the update stream (query all records and filter or return whether a record is deleted) and seems wise to use the same SQL query path for caching

For the update stream the query would also include the latest block number no? So wouldn't it already be a slightly different query shape?

alvrs (dismissed):
my comments are mostly about performance, which is hard to fully reason about without benchmarks and its not blocking for functionality, so I'm happy merging as is. But I think we should prioritise adding a benchmark for `getLogs` on a table with _lots_ of rows so we can more efficiently identify bottlenecks in hot paths.

packages/store-indexer/src/postgres/getLogs.ts
holic: > For the update stream the query would also include the latest block number no? So wouldn't it already be a slightly different query shape?

ah yeah, that's true

mind if we come back to this? it would be nice to set up some scaffolding for testing/measuring query performance, processing performance (SQL query -> HTTP response), etc. and I don't want to blow up this PR

packages/store-indexer/src/postgres/createQueryAdapter.ts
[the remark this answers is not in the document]
holic: will adjust here but I'd prefer to save these kinds of things for an optimization pass where we can set up some measurement tools etc. (assuming we even want to keep supporting this endpoint)

although it will prob be more performant, I find your suggested approach a bit harder to read and maintain, so it'd be good to know what perf we're trading off in terms of readability vs speed/memory

packages/store-indexer/src/postgres/getLogs.ts
alvrs: yeah definitely, this is not blocking, approved the PR https://github.com/latticexyz/mud/pull/1965#pullrequestreview-1754966405

packages/store-indexer/src/postgres/createQueryAdapter.ts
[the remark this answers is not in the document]
alvrs: agree with no premature optimization and instead prioritizing benchmarks to find the actual bottlenecks!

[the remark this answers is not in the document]
holic: ah so the other reason I did this like this is because we need the table schema to decode the log, and didn't want to create another lookup mapping

PR #1966, opened 2023-11-28
feat(world-modules): add gas reports to ERC721Module

### LINKED ISSUE #1854: add gas reports to `ERC721Module`
Followup to #1844

R-Morpheus:
closes #1854

### DIFF
--- a/packages/world-modules/test/ERC721.t.sol
+++ b/packages/world-modules/test/ERC721.t.sol
@@ -157,7 +157,9 @@ contract ERC721Test is Test, GasReporter, IERC721Events, IERC721Errors {
     vm.assume(owner != address(0));
 
     _expectMintEvent(owner, id);
+    startGasReport("mint");
     token.mint(owner, id);
+    endGasReport();
 
     assertEq(token.balanceOf(owner), 1);
     assertEq(token.ownerOf(id), owner);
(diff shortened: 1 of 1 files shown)


R-Morpheus:
> Thank you! Be sure to run ` pnpm gas-report` from the repo root to regenerate the gas report file.

yes, of course, that's what I wanted to do, but for some reason my commands are not executed. At the same time, it is clearly indicated that the process is completed and no errors have occurred.  

I executed commands both in the root directory and in the world-modules folder, and deleted them as well.a json file, but all attempts are in vain. Maybe you have any thoughts what am I doing wrong?

Снимок экрана 2023-11-28 в 12 50 14

holic:
ohh I see what's going on! these are all fuzz tests, so logs are suppressed (because each one runs hundreds if not thousands of times) and therefore gas report logs are not showing up

I assume we'll need to add a separate test suite specifically for these gas reports that runs each test once

R-Morpheus:
> ohh I see what's going on! these are all fuzz tests, so logs are suppressed (because each one runs hundreds if not thousands of times) and therefore gas report logs are not showing up
> 
> I assume we'll need to add a separate test suite specifically for these gas reports that runs each test once

Oh, I get it, is there anything I can help you with?

holic:
> Oh, I get it, is there anything I can help you with?

Yeah! If you leave the gas tests where they are and a set of separate test methods that call out to the fuzzed ones with hardcoded arguments, that should give you logs/gas reports for each.

For example:
```solidity
function testMintGas() public {
  testMint(1, vm.addr("minter"));
}
```

PR #1972, opened 2023-11-29
fix(store-sync,store-indexer): make last updated block number not null

holic:
following up #1965 

not strictly a bug but just doing this for completeness so we don't have to sort inline


packages/store-indexer/src/postgres/getLogs.ts
```diff
@@ -53,7 +53,8 @@ export async function getLogs(
   const records = await database
     .select()
     .from(tables.recordsTable)
-    .where(or(...conditions));
+    .where(or(...conditions))
+    .orderBy(asc(tables.recordsTable.lastUpdatedBlockNumber));
```
holic, line 57: originally had this as 
```ts
.orderBy(sql`${tables.recordsTable.lastUpdatedBlockNumber} ASC NULLS FIRST`)
```
but then realize there's ~no case where we'd add records to the table _without_ a block number (both RPC and another indexer will provide this info)

alvrs (approved):

PR #1973, opened 2023-11-29
feat(store-sync,store-indexer): sync from getLogs indexer endpoint

holic:


packages/store-indexer/src/postgres/getLogs.ts
```diff
@@ -54,7 +54,12 @@ export async function getLogs(
     .select()
     .from(tables.recordsTable)
     .where(or(...conditions))
-    .orderBy(asc(tables.recordsTable.lastUpdatedBlockNumber));
+    .orderBy(
+      asc(tables.recordsTable.lastUpdatedBlockNumber),
+      // TODO: add logIndex and use that to sort instead of address/tableId? (https://github.com/latticexyz/mud/issues/1979)
+      asc(tables.recordsTable.address),
+      asc(tables.recordsTable.tableId)
+    );
```
alvrs, line 62: for context, why do we have to order the result here?

alvrs (approved):

packages/store-indexer/src/postgres/getLogs.ts
holic: it was an attempt to fix the e2e tests having inconsistent results between rpc and indexer as far how data is populated, but ended up not being enough

I fixed the e2e test here and need to bring this commit over: https://github.com/latticexyz/mud/pull/1984/commits/74611df9de0c25847f626743d8b0fe135a286275

ultimately we lose some fidelity/ordering when hydrating from the indexer because everything is done at the block level, so records pulled from the DB within the same block may not be exactly the same order as records created from a list of logs from the RPC (mostly due to a lack of `logIndex`)

this isn't really an issue practically speaking, but felt like it would be nice to have some consistency between rpc and indexer in terms of data returned and its ordering

alvrs: do we still need the sorting now? unclear if it affects performance at all

alvrs (approved):
lgtm, only question i have is if we can skip the overhead of sorting if it doesn't solve a problem

packages/store-indexer/src/postgres/getLogs.ts
holic: probably not, will update

PR #1976, opened 2023-11-30
fix(world): inline debug constants [L-11]

ClaudeZsb:
## Description

The original type of `ROOT_NAMESPACE_STRING` defined in `WorldResourceId.sol` is 
```solidity
bytes16 constant ROOT_NAMESPACE_STRING = bytes16("ROOT_NAMESPACE");
```
Obviously its length is supposed to be `14`. And then problem is coming when we use it in a conditional operator `? :` and put it in the `true` condition.  See code [here](https://github.com/latticexyz/mud/blob/504e25dc83a210a1ef3b66d8487d9e292470620c/packages/world/src/WorldResourceId.sol#L96-L110):
```solidity
bytes14 resourceNamespace = getNamespace(resourceId);
...
return string(
  abi.encodePacked(
    ...,
    resourceNamespace == ROOT_NAMESPACE ? ROOT_NAMESPACE_STRING : resourceNamespace,
    ...
  )
)
```
Although `resourceNamespace` is defined as `bytes14`, the type of the value packed will be converted to `bytes16` just because the type of value at the true condition is `bytes16`. As a result, `ResourceId.toString()` will return a string of length `36=2+1+16+1+16` instead of `34=2+1+14+1+16`.

## Reproduction
Try the following test code
```solidity
console.logBytes(abi.encodePacked(bytes16("true")));
console.logBytes(abi.encodePacked(bytes14("false")));
console.logBytes(abi.encodePacked(1 == 2 ? bytes16("true") : bytes14("false")));
```
And you'll get 
```
Logs:
  0x74727565000000000000000000000000
  0x66616c7365000000000000000000
  0x66616c73650000000000000000000000
```
where `bytes14("false")` is packed as a `bytes16`.


packages/world/src/WorldResourceId.sol
```diff
@@ -10,7 +10,7 @@ import { RESOURCE_NAMESPACE, MASK_RESOURCE_NAMESPACE } from "./worldResourceType
 uint256 constant NAMESPACE_BITS = 14 * 8;
 uint256 constant NAME_BITS = 16 * 8;
 
-bytes16 constant ROOT_NAMESPACE_STRING = bytes16("ROOT_NAMESPACE");
+bytes14 constant ROOT_NAMESPACE_STRING = bytes14("ROOT_NAMESPACE");
 bytes16 constant ROOT_NAME_STRING = bytes16("ROOT_NAME");
```
holic, line 14: @alvrs A few notes/questions for us:
- Since these strings are just for debugging, they don't need to be type-aligned with the actual namespace/name types, right? They could technically be any length?
- I have been using an empty string in ~all of the client code, and wonder if we should align on that here. Alternatively, I wonder if we should use a less likely string like `<root>` or something.
- Should we inline this into `toString` (since that's the only place it's used) instead of defining up here, so we don't need to specify the type?
- Should we use `toTrimmedString` for namespace and name inside `toString`?

alvrs: > Since these strings are just for debugging, they don't need to be type-aligned with the actual namespace/name types, right? They could technically be any length?

Correct (but kinda agree using something else than `bytes14` can lead to confusion)

> I have been using an empty string for root namespace/name in ~all of the client code, and wonder if we should align on that here. Alternatively, I wonder if we should use a less likely string like <root> or something.

If we do that we don't even need the ternary operator here anymore since the actual namespace/name is already empty. The original reason why we replaced the empty string here with a more explicit one was that `ns::` might be less easy to understand than `ns:ROOT_NAMESPACE:ROOT_NAME`. (No strong opinion on which string we use, also happy to make it `ns:<root>:<root>`)

> Should we inline this into toString (since that's the only place it's used) instead of defining up here, so we don't need to specify the type?

Yeah that makes sense! (still need to specify the type tho because the ternary operator is unhappy if the result is `string | bytes14`)

> Should we use toTrimmedString for namespace and name inside toString?

We could, but if it's logged as string the `0`'s are already not displayed, and if it's logged as raw bytes it might be a tiny little bit easier to figure out what the log means if it always has the same length (34 bytes). On the other hand advantage of using `toTrimmedString` would mean it's possible to compare the result with an inline string, which also might be useful.

holic: I like it!

Given that this method is meant to be for debug/human-readability purposes and comparison/parsing should be done against the resource ID itself, I am keen for this to be more string like and use `toTrimmedString` around the whole thing. But would be curious to see gas differences before making the final call.

ClaudeZsb: With test code:
```solidity
  function testToString() public {
    ResourceId resourceId = WorldResourceIdLib.encode({
      typeId: RESOURCE_SYSTEM,
      namespace: "namespace",
      name: "name"
    });

    startGasReport("convert resource ID to string");
    string memory resourceIdString = resourceId.toString();
    endGasReport();

    // assertEq(resourceIdString, "sy:namespace:name");
  }
```
It turns out that:
```
GAS REPORT: 460 convert resource ID to string
```

After using `toTrimmedString` inside of `toString` in the following form:
```solidity
string(
  abi.encodePacked(
    resourceType,
    ":",
    resourceNamespace == bytes14("") ? "<root>" : WorldResourceIdLib.toTrimmedString(resourceNamespace),
    ":",
    resourceName == bytes16("") ? "<root>" : WorldResourceIdLib.toTrimmedString(resourceName)
  )
);
```
We got:
```
GAS REPORT: 3287 convert resource ID to string
```

ClaudeZsb: By the way, I just failed to generate `gas-report.json` by running `pnpm gas-report` no matther in directory of `mud` or `mud/packages/world/`. 

So I tested by running `GAS_REPORTER_ENABLED=true forge test --match-test testToString` in order to see the output of `gas-report`.

Do you guys know the reason? @holic @alvrs I must say that it's the first time I'm using this util. So maybe I just missed some configuration or something?

ClaudeZsb: @holic

packages/world/src/WorldResourceId.sol
```diff
@@ -102,9 +99,9 @@ library WorldResourceIdInstance {
         abi.encodePacked(
           resourceType,
           ":",
-          resourceNamespace == ROOT_NAMESPACE ? ROOT_NAMESPACE_STRING : resourceNamespace,
+          resourceNamespace == bytes14("") ? "<root>" : WorldResourceIdLib.toTrimmedString(resourceNamespace),
           ":",
-          resourceName == ROOT_NAME ? ROOT_NAME_STRING : resourceName
+          resourceName == bytes16("") ? "<root>" : WorldResourceIdLib.toTrimmedString(resourceName)
```
alvrs, line 104: small nit: let's use the `ROOT_NAMESPACE` and `ROOT_NAME` constants (imported from `constants.sol`) here instead of the inline `bytes14`

alvrs (commented):
Good to be merged and replace #2030. @yonadaaa can you take this over and finish it up (using the `ROOT_NAMESPACE` and `ROOT_NAME` constants and adding a short changeset)

holic:
Thanks @ClaudeZsb! Sorry for the delay in getting back to this: we all took a long winter break :)

Issue #1987, opened 2023-12-01
postgres indexer: consider snake-case for table/column names


holic:
postgres is default case insensitive unless you wrap all the names/identifiers in double quotes, which means our standard for table names (capitalized) and field names (camel case) translated to SQL tables and columns requires you to use quotes everywhere in your query

PR #1989, opened 2023-12-01
feat(store-sync): snake case postgres names in decoded tables

holic:
fixes https://github.com/latticexyz/mud/issues/1987

### DIFF
--- a/packages/store-sync/src/postgres-decoded/buildTable.ts
+++ b/packages/store-sync/src/postgres-decoded/buildTable.ts
@@ -1,5 +1,6 @@
 import { PgColumnBuilderBase, PgTableWithColumns, pgSchema } from "drizzle-orm/pg-core";
 import { Address, getAddress } from "viem";
+import { snakeCase } from "change-case";
 import { KeySchema, ValueSchema } from "@latticexyz/protocol-parser";
 import { asBigInt, asHex } from "../postgres/columnTypes";
 import { transformSchemaName } from "../postgres/transformSchemaName";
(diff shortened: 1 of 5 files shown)


packages/store-sync/src/postgres-decoded/buildTable.ts
```diff
@@ -46,14 +47,21 @@ export function buildTable<TKeySchema extends KeySchema, TValueSchema extends Va
   keySchema,
   valueSchema,
 }: BuildTableOptions<TKeySchema, TValueSchema>): BuildTableResult<TKeySchema, TValueSchema> {
-  const schemaName = transformSchemaName(`${getAddress(address)}__${namespace}`);
+  // We snake case the schema name and table name for ergonomics in querying
+  // (no need to double-quote), but this is destructive and may lead to
+  // overlapping namespaces (both `MyNamespace` and `my_namespace` snake case
+  // to the same value).
+  // TODO: consider not snake casing these
```
holic, line 54: @alvrs Curious if you have thoughts on this!

This snake case approach was inspired by Sky Strife analytics queries, where the SQL is quite ugly due to all the double quotes for case sensitivity:
https://github.com/latticexyz/mud/compare/historical-indexer?expand=1#diff-ec8025f1e682cd379feaad115d8f46b1220fc4a37e1859dc9a360e8b54fd9d14

holic: If we wanted to only snake case column names, the best we can do with SQL for case sensitive schemas/tables is something like

```sql
SET search_path TO "0x7203e7ADfDF38519e1ff4f8Da7DCdC969371f377__";

SELECT
    __last_updated_block_number AS created_at_block,
    created_by,
    encode(level_id, 'escape') AS map,
    key AS match_entity
FROM
    "MatchConfig" match_config;
```

- set `search_path` to the schema(s) we want to query from with their "shorthand" form (i.e. just table name)
- manually alias case-sensitive `MatchConfig` table name to `match_config`, in case we want to do joins without double quotes
- column names can be safely snake cased

For ORMs and other programmatic SQL tools, I'm not sure how easy it is to use something like `search_path`, but maybe these would also abstract over the quoting of things like schemas to ensure case sensitivity.

PR #1995, opened 2023-12-01
docs(guides/hello-world): fix `add-table` to use correct name in devtools

qbzzt:
It used the old name instead of `storedBlockLogs$`.


qbzzt:
https://github.com/latticexyz/mud/pull/1996 takes care of this.

PR #1999, opened 2023-12-03
WIP indexer SSE prototype

frolic:
- [ ] add shared input/output types
- [ ] figure out JSON encoding approach with bigints
- [ ] explore custom postgres custom types for `bytea` columns <> `Hex` values
- [ ] use row's `lastUpdatedBlockNumber` for log `blockNumber`?
- [ ] use row's `lastUpdatedBlockNumber` for SSE `id: ...` line (which is used as `Last-Event-ID` header)
  - [ ] make use of `Last-Event-ID` for filtering logs > that block number
- [ ] (not blocking) explore alternatives to `EventSource` and SSE protocol so we can control the readable stream (i.e. back-pressure)


holic:
much of this was already merged and the remainder is in #2013

Issue #1795, opened 2023-10-19
`mud set-version --link` shouldn't fetch versions


holic:
https://github.com/latticexyz/mud/blob/main/packages/cli/src/commands/set-version.ts
we can skip figuring out what the current npm version when using `--link`

PR #2000, opened 2023-12-04
(cli): mud set-version --link shouldn't fetch versions

R-Morpheus:
closing #1795


packages/cli/src/commands/set-version.ts
```diff
@@ -73,6 +76,11 @@ const commandModule: CommandModule<Options, Options> = {
 };
 
 async function resolveVersion(options: Options) {
+  // If the --link flag is used, we return the current version without requesting npm
+  if (options.link) {
+    return options.mudVersion;
+  }
```
holic, line 82: is this necessary with the above logic?

holic (approved):

PR #2001, opened 2023-12-04
chore(store-indexer): command to run decoded indexer

yonadaaa:
Adds a TS `bin` and a command to run the decoded indexer on local and testnet.


packages/store-indexer/bin/postgres-decoded-indexer.ts
yonadaaa, line 1: This file mostly mirrors [`postgres-indexer.ts`](https://github.com/latticexyz/mud/blob/main/packages/store-indexer/bin/postgres-indexer.ts)

alvrs (approved):

PR #2007, opened 2023-12-04
feat(world): require namespace to exist before registering systems/tables in it [C-01]

alvrs:


packages/world-modules/src/modules/erc20-puppet/ERC20Module.sol
```diff
@@ -68,6 +72,7 @@ contract ERC20Module is Module {
 
     // Register the ERC20 in the ERC20Registry
     if (!ResourceIds.getExists(ERC20_REGISTRY_TABLE_ID)) {
+      world.registerNamespace(MODULE_NAMESPACE_ID);
```
holic, line 75: what if the namespace exists but the table doesn't? should we put this in a separate conditional?

packages/world-modules/src/modules/erc20-puppet/ERC20Module.sol
```diff
@@ -83,6 +88,12 @@ contract ERC20ModuleRegistrationLibrary {
    * Register systems and tables for a new ERC20 token in a given namespace
    */
   function register(IBaseWorld world, bytes14 namespace) public {
+    // Register the namespace if it doesn't exist yet
+    ResourceId tokenNamespace = WorldResourceIdLib.encodeNamespace(namespace);
+    if (!ResourceIds.getExists(tokenNamespace)) {
```
holic, line 93: (not blocking)

since this is a common check, I wonder if we should add some helpers/abstractions for cases this like
```solidity
function namespaceExists(ResourceId resourceId)
function namespaceExists(bytes14 namespace)
```

mostly thinking about it feeling a little weird to sprinkle `ResourceIds` table usage everywhere, but maybe we kinda have to due to gas

packages/world-modules/src/modules/erc721-puppet/ERC721Module.sol
```diff
@@ -72,6 +76,7 @@ contract ERC721Module is Module {
 
     // Register the ERC721 in the ERC20Registry
     if (!ResourceIds.getExists(ERC721_REGISTRY_TABLE_ID)) {
+      world.registerNamespace(MODULE_NAMESPACE_ID);
```
holic, line 79: same note as above

packages/world-modules/src/modules/keyswithvalue/KeysWithValueModule.sol
```diff
@@ -45,16 +46,24 @@ contract KeysWithValueModule is Module {
...
+    ResourceId targetTableId = getTargetTableId(MODULE_NAMESPACE, sourceTableId);
 
     IBaseWorld world = IBaseWorld(_world());
 
+    // Register the target namespace if it doesn't exist yet
+    if (!ResourceIds.getExists(targetTableId.getNamespaceId())) {
+      (bool registrationSuccess, bytes memory registrationReturnData) = address(world).delegatecall(
+        abi.encodeCall(world.registerNamespace, (targetTableId.getNamespaceId()))
+      );
+      if (!registrationSuccess) revertWithBytes(registrationReturnData);
```
holic, line 58: I think we already played around with this, but is it a bad idea to wrap this logic in a free function like

```solidity
function delegateCallOrRevert(address target, bytes memory callData) returns (bytes memory) {
  (bool success, bytes memory returnData) = target.delegateCall(callData);
  if (!success) revertWithBytes(returnData);
  return returnData;
}

delegateCallOrRevert(address(world), abi.encodeCall(world.registerNamespace, (targetTableId.getNamespaceId())));
```

packages/world/src/modules/core/implementations/StoreRegistrationSystem.sol
```diff
@@ -51,20 +49,8 @@ contract StoreRegistrationSystem is System, IWorldErrors {
...
-      (bool success, bytes memory data) = coreSystemAddress.delegatecall(
-        abi.encodeCall(WorldRegistrationSystem.registerNamespace, (namespaceId))
-      );
-      if (!success) revertWithBytes(data);
-    } else {
-      // otherwise require caller to own the namespace
-      AccessControl.requireOwner(namespaceId, _msgSender());
-    }
+    // Require the caller to own the table's namespace
+    AccessControl.requireOwner(tableId, _msgSender());
```
holic, line 53: will this fail if the namespace does not exist?

packages/world-modules/src/modules/erc20-puppet/ERC20Module.sol
```diff
@@ -53,7 +54,10 @@ contract ERC20Module is Module {
 
     // Register the ERC20 tables and system
     IBaseWorld world = IBaseWorld(_world());
-    registrationLibrary.delegatecall(abi.encodeCall(ERC20ModuleRegistrationLibrary.register, (world, namespace)));
+    (bool success, bytes memory returnData) = registrationLibrary.delegatecall(
+      abi.encodeCall(ERC20ModuleRegistrationLibrary.register, (world, namespace))
+    );
+    if (!success) revertWithBytes(returnData);
```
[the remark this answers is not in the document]
alvrs: hopefully not

packages/world-modules/src/modules/erc20-puppet/ERC20Module.sol
alvrs: since both are registered in these two lines, and the table could not be registered if the namespace is not registered, i think it's fine to keep both in one conditional

packages/cli/src/deploy/ensureNamespaceOwner.ts
```diff
@@ -0,0 +1,71 @@
...
+      "found",
+      existingNamespaces.size,
+      "existing namespaces:",
+      Array.from(existingNamespaces)
+        .map((namespace) => (namespace === "" ? "<root>" : namespace))
+        .join(", ")
+    );
+  }
+
+  // Register missing namespaces
```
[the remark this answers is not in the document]
alvrs: yep can do that!

packages/world-modules/src/modules/erc20-puppet/ERC20Module.sol
alvrs: agree this might be useful, would not do it in this PR though to keep the scope smaller

packages/world-modules/src/modules/keyswithvalue/KeysWithValueModule.sol
alvrs: Sadly every time memory is passed around (even in free functions) it increases gas usage. Might still be worth for code organization purposes and to make sure handle the `success` flag, but would move this to a followup (including exploring the gas impact)

packages/world/src/modules/core/implementations/StoreRegistrationSystem.sol
alvrs: yes! if the namespace does not exist, its owner is `address(0)`

alvrs: We could make the reason for the failure more explicit with a `ResourceIds._getExists(namespaceId)` check and then reverting with `World_ResourceNotFound` though

alvrs:
There are a couple other places where we can explicitly check for resource existence, will do that in the PR for L-09

packages/world/src/AccessControl.sol
```diff
@@ -50,4 +52,15 @@ library AccessControl {
       revert IWorldErrors.World_AccessDenied(resourceId.toString(), caller);
     }
   }
+
+  /**
+   * @notice Check for existence of the given resource ID.
+   * @dev Reverts with IWorldErrors.World_ResourceNotFound if the resource does not exist.
+   * @param resourceId The resource ID to check existence for.
+   */
+  function requireExistence(ResourceId resourceId) internal view {
```
holic, line 61: 😨 so existential

holic (approved):

PR #2013, opened 2023-12-05
feat(store-indexer): add SSE logs stream endpoint

### LINKED ISSUE #1650: stream logs from indexer

holic:
closes #1650
pulled out of #1999 

TODO:
- [ ] tests
- [ ] better error handling for SSE (currently just pukes with an `Internal Server Error` if e.g. `input` query fails validation)
- [ ] move to form-encoded query string for input filters? e.g. `/sse/logs?chainId=17001&address=0x`
- [ ] bigint-safe encoding of JSON (not superjson)
- [ ] translate numeric strings from postgres to bigint (can turn on [this setting in postgres lib](https://github.com/porsager/postgres#numbers-bigint-numeric) but unclear if this will also convert e.g. chain ID, which should be a `Number`)
- [ ] add support for abort signal: https://developer.mozilla.org/en-US/docs/Web/API/Request/signal


holic:
closing as this will be solved by dozer

PR #2015, opened 2023-12-05
refactor(store,world,world-modules): use BYTE_TO_BITS constant where applicable [N-09]

yonadaaa:


alvrs (commented):
Unfortunately the changes increase gas cost in a way that makes me doubt if it's worth it. I think the gas cost increase only comes from the places where we replace an inline `8` in code, not in places where we only use it to define another constant that already exists. Maybe a good compromise would be to only change it in those constant definitions where it doesn't impact gas cost, and just add a comment in the other places?

yonadaaa:
@alvrs I agree with the direction of adding comments, but the increases actually come from constant definitions(!), not inline code. I've added comments to these files instead, and the gas reports are equivalent to `main`.  

Specifically, the increase comes from these constants in the Resource ID files:

`ResourceId.sol`
```solidity
/// @dev Number of bits reserved for the type in the ResourceId.
uint256 constant TYPE_BITS = 2 * 8;
/// @dev Number of bits reserved for the name in the ResourceId.
uint256 constant NAME_BITS = 32 * 8 - TYPE_BITS;
```

`WorldResourceId.sol`
```solidity
uint256 constant NAMESPACE_BITS = 14 * 8;
uint256 constant NAME_BITS = 16 * 8;
```

packages/store/src/ResourceId.sol
```diff
@@ -9,9 +9,9 @@ pragma solidity >=0.8.21;
 type ResourceId is bytes32;
 
 /// @dev Number of bits reserved for the type in the ResourceId.
-uint256 constant TYPE_BITS = 2 * 8;
+uint256 constant TYPE_BITS = 2 * 8; // BYTE_TO_BITS
```
alvrs, line 12:
```suggestion
uint256 constant TYPE_BITS = 2 * 8; // 2 bytes * 8 bits per byte
```

packages/store/src/ResourceId.sol
```diff
@@ -9,9 +9,9 @@ pragma solidity >=0.8.21;
...
 /// @dev Number of bits reserved for the name in the ResourceId.
-uint256 constant NAME_BITS = 32 * 8 - TYPE_BITS;
+uint256 constant NAME_BITS = 32 * 8 - TYPE_BITS; // BYTE_TO_BITS
```
alvrs, line 14:
```suggestion
uint256 constant NAME_BITS = 32 * 8 - TYPE_BITS; // 32 bytes * 8 bits per byte - type bits
```

packages/world/src/WorldResourceId.sol
```diff
@@ -7,8 +7,8 @@ import { ResourceId, ResourceIdInstance, TYPE_BITS } from "@latticexyz/store/src
 import { ROOT_NAMESPACE, ROOT_NAME } from "./constants.sol";
 import { RESOURCE_NAMESPACE, MASK_RESOURCE_NAMESPACE } from "./worldResourceTypes.sol";
 
-uint256 constant NAMESPACE_BITS = 14 * 8;
-uint256 constant NAME_BITS = 16 * 8;
+uint256 constant NAMESPACE_BITS = 14 * 8; // BYTE_TO_BITS
```
alvrs, line 10:
```suggestion
uint256 constant NAMESPACE_BITS = 14 * 8; // 14 bytes * 8 bits per byte
```

packages/world/src/WorldResourceId.sol
```diff
@@ -7,8 +7,8 @@ import { ResourceId, ResourceIdInstance, TYPE_BITS } from "@latticexyz/store/src
...
+uint256 constant NAME_BITS = 16 * 8; // BYTE_TO_BITS
```
alvrs, line 11:
```suggestion
uint256 constant NAME_BITS = 16 * 8; // 16 bytes * 8 bits per byte
```

alvrs (commented):
the named constant's purpose is mostly to make the calculations easier to parse for humans; in places where using the constant would increase gas, using a verbose comment instead is an acceptable alternative imo (no need to use the exact same `BYTE_TO_BITS` token here since this isn't a value that's ever going to change so we don't need to be able to ie search for it)

alvrs (approved):

PR #2017, opened 2023-12-05
fix(store): emit event after calling beforeSetRecord hook [L-02]

yonadaaa:
Rearranges `setRecord`, `spliceStaticData`, and `_spliceDynamicData` in `StoreCore` to emit the storage event **after** calling the `beforeSetRecord` hook, ensuring that storage events are emitted in the same order that values are stored on-chain.

~~An important side-effect is that off-chain tables also early return after the hook, which means they can now use the `beforeSetRecord` hook. I think this is okay because hooks do not necessarily need to read the current storage; they can just react to the new value.~~ EDIT: registering before hooks for offchain tables is explicitly not allowed so this is a no-op


docs/pages/store/reference/store-core.mdx
```diff
@@ -774,16 +774,6 @@ event Store_SetRecord(
...
-**Parameters**
-
-| Name             | Type            | Description                                             |
-| ---------------- | --------------- | ------------------------------------------------------- |
-| `tableId`        | `ResourceId`    | The ID of the table where the record is set.            |
-| `keyTuple`       | `bytes32[]`     | An array representing the composite key for the record. |
-| `staticData`     | `bytes`         | The static data of the record.                          |
-| `encodedLengths` | `PackedCounter` | The encoded lengths of the dynamic data of the record.  |
-| `dynamicData`    | `bytes`         | The dynamic data of the record.                         |
-
```
alvrs, line 786: these docs changes seem unintended, or am i missing something?

alvrs: possible that the docs are out of date, but let's ignore this change for this PR and move these changes to a separate PR

alvrs (approved):
lgtm, let's just remove the unrelated docs changes, add a changeset, and update the gas report

alvrs (changes requested):
Just realized we also need to apply this change to the `deleteRecord` function

packages/store/src/StoreCore.sol
```diff
@@ -583,9 +583,6 @@ library StoreCore {
    * @param fieldLayout The field layout for the record.
    */
   function deleteRecord(ResourceId tableId, bytes32[] memory keyTuple, FieldLayout fieldLayout) internal {
-    // Emit event to notify indexers
-    emit Store_DeleteRecord(tableId, keyTuple);
-
```
alvrs, line 588: we also need to move the early return for offchain tables below the event emission, otherwise there is no event emitted for offchain tables (and the onBefore hook is not triggered)

holic: can we add a test for this to make sure we don't regress?

yonadaaa: Added a couple tests

.changeset/beige-ads-melt.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/store": patch
+---
+
+Fixes the functions in `StoreCore` to emit a storage event after calling the `beforeSetRecord` hook, ensuring that storage events are emitted in the same order that values are set on-chain.
```
holic, line 5:
```suggestion
Storage events are now emitted after "before" hooks, so that the resulting logs are now correctly ordered and reflect onchain logic. This resolves issues with store writes and event emissions happening in "before" hooks.
```

holic (approved):

PR #2020, opened 2023-12-05
fix(store): onBeforeSpliceDynamicData receives the previous encoded lengths [M-01]

yonadaaa:


alvrs (approved):
let's add a small changeset

alvrs (approved):

PR #2021, opened 2023-12-05
(recs): entity values are passed by reference

R-Morpheus:
possible solution for #310


holic (commented):
I am seeing some test failures. Not sure if those were intentionally left out because you were just proposing a potential change?

PR #2022, opened 2023-12-05
refactor(world): use _getSystem when fetching system addresses [N-11]

yonadaaa:


alvrs (dismissed):

alvrs:
let's add a small changeset (we need a changeset for anything that is supposed to trigger a new version release, which is by default any code that makes it into consumer projects, even if it's a tiny refactor)

alvrs (approved):

alvrs (approved):

Issue #1936, opened 2023-11-14
indexer client should append correct path to indexer url


holic:
we currently expect/require the full URL (i.e. `https://indexer/trpc`) when we could take in just the endpoint (`https://indexer`) and append `/trpc` if the pathname is blank

PR #2026, opened 2023-12-05
feat(store-indexer, store-sync): improve query performance and enable compression, add new api

alvrs:
- most heavy lifting is pulled over from #2013 
- we should still get #2013 in to enable streaming data to the client instead of transferring in one big blob, but the improved query performance from replacing drizzle with raw SQL and the gzip compression are worth getting in as soon as possible

Query performance is increased by ~10x thanks to @holic's `queryLogs` helper to replace drizzle, and data transferred over the wire is down by ~40x thanks to @holic's `gzip` implementation

CleanShot 2023-12-05 at 22 53 04@2x


packages/store-indexer/src/compress.ts
```diff
@@ -0,0 +1,48 @@
...
+function flushEvery<stream extends Zlib & Readable>(stream: stream, bytesThreshold: number): stream {
+  let bytesSinceFlush = 0;
+  stream.on("data", (data) => {
+    bytesSinceFlush += data.length;
+    if (bytesSinceFlush > bytesThreshold) {
+      bytesSinceFlush = 0;
+      stream.flush();
+    }
+  });
+  return stream;
```
alvrs, line 27: Would it be possible that the "last" bytes are stuck in the stream if the data isn't a multiple of `bytesThreshold`? Would we need some kind of timeout after which we flush even if there is no more data event?

holic: > Would it be possible that the "last" bytes are stuck in the stream if the data isn't a multiple of `bytesThreshold`? Would we need some kind of timeout after which we flush even if there is no more data event?

This might be possible for an open and long running stream, yes. But shouldn't matter as much while we're closing the stream when we finish streaming data down.

The stream is flushed when the stream closes (which is separate from `compressed.end(ctx.body)`, which pushes+closes the entirety of the body to the compressed, for non-streaming bodies).

I think it would be best to expose a `.flush` method in this middleware that we can access downstream in the `eventStream` middleware, but it was tricky to get it right with types, so I punted on that for now.

packages/common/src/result.ts
```diff
@@ -0,0 +1,10 @@
+// Inspired by https://doc.rust-lang.org/std/result/
+export type Result<Ok, Err = unknown> = { ok: Ok } | { error: Err };
```
holic, line 2: 👀

packages/store-indexer/src/postgres/types.ts
holic, line 1: I have been preferring to put shared types in a file like `common.ts`, where you can also put things like shared constants.

I don't like naming files/dirs after their implementation details (`types`, `components`, `hooks`) because it gets hard to e.g. define and export a type based a constant or vice versa without creating circular deps. Also makes it harder to refactor/unwind certain things when you have a massive file/dir with a certain kind of "thing" instead of encapsulating features.

packages/store-indexer/src/sqlite/getLogs.ts
```diff
@@ -0,0 +1,39 @@
...
+import { debug } from "../debug";
+import { createBenchmark } from "@latticexyz/common";
+import { compress } from "../compress";
+import { getTablesWithRecords } from "./getTablesWithRecords";
+import { BaseSQLiteDatabase } from "drizzle-orm/sqlite-core";
+
+export function getLogs(database: BaseSQLiteDatabase<"sync", any>): Middleware {
+  const router = new Router();
+
+  router.get("/get/logs", compress(), async (ctx) => {
```
holic, line 15: similar to above, would prefer to use `/api/logs` and call this function/module `api` or `apiRoutes`

packages/store-sync/src/indexer-client/createIndexerClient.ts
```diff
@@ -0,0 +1,42 @@
+import { z } from "zod";
+import { input } from "./input";
+import { StorageAdapterBlock } from "../common";
+import { Result } from "@latticexyz/common";
+
+type CreateIndexerClientOptions = {
+  /**
+   * tRPC endpoint URL like `https://indexer.dev.linfra.xyz/trpc`.
```
holic, line 8: should we update this?

holic, line 8:
```suggestion
   * Indexer endpoint URL like `https://indexer.holesky.redstone.xyz/`.
```

packages/store-sync/src/indexer-client/createIndexerClient.ts
```diff
@@ -0,0 +1,42 @@
...
+export function createIndexerClient({ url }: CreateIndexerClientOptions): IndexerClient {
+  return {
+    getLogs: async (opts): Promise<Result<StorageAdapterBlock>> => {
+      const input = encodeURIComponent(JSON.stringify(opts));
+      const response = await fetch(`${url}/get/logs?input=${input}`, { method: "GET" });
+
+      // TODO: could we return a readable stream here instead of fetching the entire response right away?
+      const result = await response.json();
+      if (!isStorageAdapterBlock(result)) {
+        return { error: "Unexpected response:\n" + JSON.stringify(result) };
```
holic, line 32: why do we need to stringify when we can just return the json itself?

packages/store-sync/src/indexer-client/createIndexerClient.ts
```diff
@@ -0,0 +1,42 @@
...
+      }
+
+      return { ok: { ...result, blockNumber: BigInt(result.blockNumber) } };
```
holic, line 35: should we wrap this whole thing in a try/catch to return error type? like if the fetch fails or stringifying opts (i.e. bigint)

e2e/packages/sync-test/package.json
```diff
@@ -4,7 +4,8 @@
   "private": true,
   "license": "MIT",
   "scripts": {
-    "test": "vitest --run",
+    "create:db": "psql postgres://127.0.0.1/postgres -c \"CREATE DATABASE postgres_e2e;\"",
+    "test": "DATABASE_URL=postgres://127.0.0.1/postgres_e2e vitest --run",
     "test:ci": "pnpm run test"
```
holic, line 9: I assume the `test` script will overwrite the env var passed in by CI itself? we might have to change this to just be `vitest --run` or make a separate `test:local` script that passes in the `DATABASE_URL`

e2e/packages/sync-test/package.json
```diff
@@ -4,7 +4,8 @@
   "private": true,
   "license": "MIT",
   "scripts": {
-    "test": "vitest --run",
+    "create:db": "psql postgres://127.0.0.1/postgres -c \"CREATE DATABASE postgres_e2e;\"",
```
holic, line 7: is it feasible to do this in the test setup script?

packages/store-sync/src/indexer-client/createIndexerClient.ts
alvrs: Can do that too!

e2e/packages/sync-test/package.json
alvrs: ahh you're right, didn't realize `test:ci` just calls `test`

packages/store-indexer/bin/postgres-indexer.ts
```diff
@@ -39,7 +39,7 @@ const database = drizzle(postgres(env.DATABASE_URL));
 
 if (await shouldCleanDatabase(database, chainId)) {
   console.log("outdated database detected, clearing data to start fresh");
-  cleanDatabase(database);
+  await cleanDatabase(database);
```
holic, line 42: oooof, wonder if we should enable a linter for this

feels like we have fewer instances of calling an async function and not awaiting right away and can add a lint ignore for the line we use that approach explicitly

e2e/packages/sync-test/package.json
alvrs: I had that initially, but it felt weird to attempt to create a database that might already exist, permissions to create a new database might be different for each user, and it only has to be executed once per machine (since the db is not dropped, just cleaned, at the end), so it felt cleaner to colocate the db creation with the place where we define the db url to use

packages/store-indexer/src/postgres/apiRoutes.ts
```diff
@@ -0,0 +1,42 @@
...
+import { compress } from "../compress";
+
+export function apiRoutes(database: Sql): Middleware {
+  const router = new Router();
+
+  router.get("/api/logs", compress(), async (ctx) => {
+    const benchmark = createBenchmark("postgres:logs");
+
+    try {
+      const opts = input.parse(typeof ctx.query.input === "string" ? JSON.parse(ctx.query.input) : {});
```
holic, line 20: if this parse fails, it should prob be a 4xx error (client error) rather than a 5xx error (server error)

packages/store-sync/src/getSnapshot.ts
```diff
@@ -39,17 +41,26 @@ export async function getSnapshot({
 
   if (!indexerUrl) return;
 
-  const indexer = createIndexerClient({ url: indexerUrl });
+  const indexerOrigin = new URL(indexerUrl).origin;
+  const indexer = createIndexerClient({ url: indexerOrigin });
+  const trpcIndexer = createTrpcIndexerClient({ url: `${indexerOrigin}/trpc` });
```
holic, line 46: closes https://github.com/latticexyz/mud/issues/1936

e2e/packages/sync-test/package.json
holic: was just worried about the availability of `psql`

packages/store-indexer/bin/postgres-indexer.ts
holic: noted here:

PR #2029, opened 2023-12-06
chore(world): add explicit visibility to coreSystem [N-07]

yonadaaa:
It was also reported that [`_toBool`](https://github.com/latticexyz/mud/blob/12a9eb1afb18a7f7cb43d774cabb2d5af7c74e6b/packages/store/src/codegen/tables/ResourceIds.sol#L301) and [`requireInterface`](https://github.com/latticexyz/mud/blob/12a9eb1afb18a7f7cb43d774cabb2d5af7c74e6b/packages/world/src/requireInterface.sol#L19) are missing visibility but these are free functions.


alvrs (dismissed):
let's add a tiny changeset

alvrs (approved):

alvrs (approved):

PR #2031, opened 2023-12-06
fix(store): slice4 output should be bytes4 [M-03]

yonadaaa:
Seems like Solidity was casting the `bytes2` `output` variable to the `bytes4` return type.


alvrs (dismissed):
let's add a short changeset

PR #2033, opened 2023-12-06
refactor(store): order load function arguments [N-02]

yonadaaa:
Arguments are now in the same order as the [`store`](https://github.com/latticexyz/mud/blob/12a9eb1afb18a7f7cb43d774cabb2d5af7c74e6b/packages/store/src/Storage.sol#L40) function.


yonadaaa:
@alvrs this was easier because most consumers call the functions with named arguments. Should we align the order there for consistency? ie.

```solidity
bytes memory data = Storage.load({ storagePointer: storagePointer, length: 34, offset: 31 });
```

VS

```solidity
bytes memory data = Storage.load({ storagePointer: storagePointer, offset: 31, length: 34 });
```

alvrs:
no need to change the order in consumers with named arguments imo!

holic:
if we have fewer than a ~dozen call sites, would vote to align order for consistency, even though it's not technically necessary

yonadaaa:
Agreeing [with Frolic](https://github.com/latticexyz/mud/pull/2033#issuecomment-1882935300) I have aligned the order on the named arguments also

packages/store/gas-report.json
```diff
@@ -303,7 +303,7 @@
     "file": "test/GasStorageLoad.t.sol",
     "test": "testCompareStorageLoadMUD",
     "name": "MUD storage load field (warm, 1 word)",
-    "gasUsed": 245
+    "gasUsed": 2300
```
alvrs, line 306: do we know where this is coming from? 🤔

alvrs (approved):

packages/store/gas-report.json
holic: given that this increase isn't showing up in ~every other gas report, I am guessing something unrelated is going on

PR #2034, opened 2023-12-06
fix(store): revert if slice bound is invalid [L-10]

dk1a:


packages/store/src/IStoreErrors.sol
```diff
@@ -10,6 +10,7 @@ interface IStoreErrors {
   error Store_InvalidResourceType(bytes2 expected, ResourceId resourceId, string resourceIdString);
 
   error Store_InvalidDynamicDataLength(uint256 expected, uint256 received);
+  error Store_InvalidBounds(uint256 start, uint256 end);
```
holic, line 13: thoughts on `Store_InvalidSlice` or `Store_InvalidSliceBounds`?

.changeset/perfect-windows-reply.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/store": patch
+---
+
+Added a custom error `Store_IndexOutOfBounds` for when the `start:end` slice in `getDynamicFieldSlice` is invalid (it used to revert with the default overflow error)
```
holic, line 5:
```suggestion
Added a custom error `Store_InvalidBounds` for when the `start:end` slice in `getDynamicFieldSlice` is invalid (it used to revert with the default overflow error)
```

alvrs (dismissed):

packages/store/src/IStoreErrors.sol
holic: from @alvrs 
> no strong opinions, i'm happy with InvalidBounds, or InvalidSliceBounds (think "bounds" in there is nice for consistency with the "out of bounds" error)

PR #1974, opened 2023-11-29
fix(store,world): fix `tsc` errors in projects depending on `@latticexyz/store-sync`

tash-2s:
Projects created by `create-mud` are currently failing their `pnpm test` scripts due to TypeScript type checking errors. This pull request addresses this issue and also resolves #1958.

## Steps to Reproduce

```sh
# Create a new project
$ pnpm create mud@2.0.0-next.14 my-project --template vanilla
$ cd my-project

# Add missing dependencies to prevent other tsc errors
# I can address this issue in a separate PR
$ pnpm --filter contracts add --save-dev @types/prettier@2.7.2 @types/debug@4.1.7
$ pnpm --filter client add --save-dev @types/prettier@2.7.2 @types/debug@4.1.7 @types/react@18.2.22 @types/react-dom@18.2.7

# Test and see the errors
$ pnpm test # runs `pnpm -r run test` and fails, showing the log below
```
```
$ pnpm --filter client test
> client@0.0.0 test /Users/t/bench/my-project/packages/client
> tsc --noEmit

../../node_modules/.pnpm/@latticexyz+store-sync@2.0.0-next.14_@types+react@18.2.22_react@18.2.0_typescript
@5.1.6/node_modules/@latticexyz/store-sync/src/common.ts:6:25 - error TS2307: Cannot find module '@lattice
xyz/store/mud.config' or its corresponding type declarations.

6 import storeConfig from "@latticexyz/store/mud.config";
                          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

../../node_modules/.pnpm/@latticexyz+store-sync@2.0.0-next.14_@types+react@18.2.22_react@18.2.0_typescript
@5.1.6/node_modules/@latticexyz/store-sync/src/common.ts:7:25 - error TS2307: Cannot find module '@lattice
xyz/world/mud.config' or its corresponding type declarations.

7 import worldConfig from "@latticexyz/world/mud.config";
                          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

../../node_modules/.pnpm/@latticexyz+store-sync@2.0.0-next.14_@types+react@18.2.22_react@18.2.0_typescript
@5.1.6/node_modules/@latticexyz/store-sync/src/internalTableIds.ts:2:25 - error TS2307: Cannot find module
 '@latticexyz/store/mud.config' or its corresponding type declarations.

2 import storeConfig from "@latticexyz/store/mud.config";
[...]

PR #2036, opened 2023-12-07
fix(templates): fix templates tsc errors and add test

tash-2s:
- Follow-up to https://github.com/latticexyz/mud/pull/1974

This pull request fixes TypeScript type errors occurring in projects created from the `create-mud` templates. Currently, users have to add some `@types/*` dev dependencies to pass the tsc after creating their projects. To address this, I added the missing dependencies. Additionally, I created a GitHub Actions workflow that detects these and the errors fixed in the above PR.

The added dependencies (`@types/debug`, `@types/prettier`, `@types/react-dom`) align with the versions of these packages used elsewhere in this repository. I considered adding them as non-dev dependencies to the origin libraries, such as `common`, but they would be redundant for JavaScript projects, so I didn't take that approach.

The workflow runs after a snapshot release. It uses the release to run `create-mud` and `pnpm test` in the created projects, accurately simulating user environments. This is ensured by:

1. Not using locally linked dependencies, unlike the `/templates/` tests. This avoids missing errors related to non-included files or dependencies.
2. Running the test outside of this repository. Node.js/TypeScript might otherwise resolve modules/types using ancestor directories' `node_modules`, potentially missing certain errors. For instance, the missing `@types/prettier` error wouldn't have been detected if the test were run within `/packages/create-mud/my-project`, as the repository root's `node_modules` contains `@types/prettier`.

Here are examples of this workflow's runs: [This run](https://github.com/tash-2s/mud/actions/runs/7122062243) lacked the type fixes from this PR and thus failed, while [this run](https://github.com/tash-2s/mud/actions/runs/7122277599) had the types and passed.

Initially, I started writing this test as a locally executable script for `pnpm test`. However, I realized that implementing it as a CI workflow would be simpler. So, I shifted to creating this workflow, but I'm open to the local approach if preferred. To replicate this locally, the script should build and pack all necessary packages in this repository, add them to the projects created by `create-mud` outside of this repository, and then run the tests.


templates/vanilla/package.json
```diff
@@ -13,6 +13,8 @@
   },
   "devDependencies": {
     "@latticexyz/cli": "link:../../packages/cli",
+    "@types/debug": "4.1.7",
+    "@types/prettier": "2.7.2",
```
holic, line 17: likewise here, I don't think we're using any of these packages directly, but getting this via the cli dep?

templates/vanilla/packages/client/package.json
```diff
@@ -24,6 +24,7 @@
     "viem": "1.14.0"
   },
   "devDependencies": {
+    "@types/react-dom": "18.2.7",
```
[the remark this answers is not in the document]
tash-2s: Thank you for your review!

dev-tools is compiled into JS files, but I confirmed that TS files are referred to by template projects based on their tsconfig.json.

https://github.com/latticexyz/mud/blob/c8ffd09fb71242a3fc74e3758434ed45a1807289/packages/dev-tools/package.json#L12-L15

With moduleResolution: node/node10, the types field is prioritized over the exports field.

Other templates, except for vanilla, already include `@types/react-dom` as dependencies.

templates/vanilla/package.json
tash-2s: cli is not relevant to this diff.
prettier and debug are added as dependencies at the workspace root because both contracts and client require them. Similarly, typescript is also a dependency in the workspace root.

tash-2s: And both contracts and client use the TS files from common (via typesVersions), and common uses prettier and debug.

tash-2s:
I looked into the `@types/*` issue and will make an issue about it tomorrow. 😄

tash-2s:
I've created issue #2051 to explain why these `@types/*` are necessary.
I think this test is beneficial because it helps to find errors related to packaging and dependencies, which are easy to overlook when working within the context of this repository. What are your thoughts on the workflow?

holic:
Converted this to a draft because I think it needs a rebase after merging https://github.com/latticexyz/mud/pull/2084

tash-2s:
A rebase is not necessary because that PR doesn’t affect projects with `moduleResolution: node10` in their tsconfig, as mentioned in https://github.com/latticexyz/mud/issues/2051#issuecomment-1874375350. The templates use `node10`.

However, we can remove these added `@types/` dependencies once we complete step 2 in the issue, so we can put this PR on hold. In the meantime, template users can run the `pnpm add` commands as needed.

holic:
I wonder if instead we should just move templates to use one of the other `moduleResolution` options instead?

alvrs:
I think we can go ahead with merging this PR as a temporary fix until we get to step 2 from https://github.com/latticexyz/mud/issues/2051 (releasing with `exports`, moving the `moduleResolution` to bundler, adding `.d.ts.map` files to keep devex of jumping to source files instead of d.ts files). Right now running `pnpm test` right after creating a fresh template fails, which might lead users to believe the templates are broken, so i think this temporary fix would be worth it. (And the new CI actions to ensure this doesn't happen in the future are useful as well)

tash-2s:
Thank you for your reviews! I've now confirmed that a fresh project (`pnpm create mud@main`) set up on my local passed `pnpm test`, and the new CI has passed as well.

> I wonder if instead we should just move templates to use one of the other moduleResolution options?

I believe the `bundler` config is recommended for most users, as they are likely to use bundlers (like vite or tsup) and bundler-like runners (such as tsx) with newer versions of Node.js. Therefore, updating the configs in the templates seems like a good idea. (However, we might want to wait for `.d.ts.map`?)

https://www.typescriptlang.org/docs/handbook/modules/guides/choosing-compiler-options.html

PR #2037, opened 2023-12-07
feat(store-sync, store-indexer): order logs by logIndex

alvrs:


packages/store-sync/src/postgres/tables.ts
```diff
@@ -29,6 +29,8 @@ const recordsTable = pgSchema(schemaName).table(
     dynamicData: asHex("dynamic_data"),
     isDeleted: boolean("is_deleted"),
     lastUpdatedBlockNumber: asBigInt("last_updated_block_number", "numeric").notNull(),
+    transactionIndex: asNumber("transaction_index", "numeric").notNull(),
+    logIndex: asNumber("log_index", "numeric").notNull(),
```
holic, line 33: chatted about this in discord but I think we only need `logIndex`

from: https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_getfilterchanges

> `logIndex: QUANTITY` - integer of the log index position in the block. null when its pending log.
> `transactionIndex: QUANTITY` - integer of the transactions index position log was created from. null when its pending log.

my understanding of this is that log index is the position of the log in the block (thus already ordered by tx) and tx index is the position of the log in the tx

(I used this same approach in https://github.com/holic/chirp-city for UUIDs/pointers of specific logs [or "tweets"] in the blockchain)

holic: in terms of naming, I wonder if `logIndex` should be `lastUpdatedLogIndex` to correspond with `lastUpdatedBlockNumber`

or, which is where I am leaning, we change `lastUpdatedBlockNumber` to just `blockNumber`

holic (approved):
summary:
- I don't think we need `transactionIndex`
- should consider renaming `lastUpdatedBlockNumber` to just `blockNumber`
- need to update snapshots after incrementing version

packages/store-sync/src/postgres/tables.ts
alvrs: You're right! I thought `logIndex` is per transaction, but it is per block

PR #2038, opened 2023-12-07
feat(store-sync): numerical array column types in decoded indexer

yonadaaa:
Updates the decoded indexer to store numerical arrays (eg. `uint256[]`) as Postgres native `smallint[]`, `integer[]`, `bigint[]` or `numeric[]` column types, instead of less ergonomic JSON blobs.

Tested with Sky Strife, for example -

Previous query with JSON blob: 
```sql
SELECT
  index,
  value
FROM
  "0x7203e7adfdf38519e1ff4f8da7dcdc969371f377__".Match_Sweepstake,
  json_array_elements_text(reward_percentages :: json -> 'json') WITH ORDINALITY AS t(value, index);
```

New query with array:

```sql
SELECT
  index,
  value
FROM
  "0x7203e7adfdf38519e1ff4f8da7dcdc969371f377__".Match_Sweepstake,
  unnest(reward_percentages) WITH ORDINALITY AS t(value, index);
```


holic:
I think this needs a rebase since https://github.com/latticexyz/mud/pull/2001 landed?

yonadaaa:
TODO: add `bytesX` and `address` array types

holic (approved):

PR #2041, opened 2023-12-08
docs(world): fix the `revertWithBytes` double natspec issue

qbzzt:


qbzzt:
Double natspec is pretty common "in the wild", so it's probably not a problem.

PR #2046, opened 2023-12-08
fix(store): add missing FieldLayout and Schema validations [L-07]

dk1a:


alvrs (commented):
I'm kind of unhappy with the split into `CoreSystem` and `CoreSystem2` - would it be possible to instead split by functionality in e.g. a `CoreRegistrationSystem` (`StoreRegistration`, `WorldRegistration`, `ModuleInstallation`) and just register the other ones as separate systems? (It's only an advantage for them to be on one system if we expect them to be called in one transaction bc then the address is warmed up, but the non-registration ones seem pretty unrelated to each other). And for the `CoreModule`, would it be possible to require the systems to be deployed separately and passed as arguments to keep the module's bytecode below the limit? (Or is something else pushing the module over the limit?)

docs/pages/store/reference/store-core.mdx
```diff
@@ -774,16 +774,6 @@ event Store_SetRecord(
...
-**Parameters**
-
-| Name             | Type            | Description                                             |
-| ---------------- | --------------- | ------------------------------------------------------- |
-| `tableId`        | `ResourceId`    | The ID of the table where the record is set.            |
-| `keyTuple`       | `bytes32[]`     | An array representing the composite key for the record. |
-| `staticData`     | `bytes`         | The static data of the record.                          |
-| `encodedLengths` | `PackedCounter` | The encoded lengths of the dynamic data of the record.  |
-| `dynamicData`    | `bytes`         | The dynamic data of the record.                         |
-
```
holic, line 786: I think @yonadaaa was just seeing this, but can you try `foundryup` then regenerate docs?

dk1a:
Yep, agree with semantic system organization

> And for the CoreModule, would it be possible to require the systems to be deployed separately and passed as arguments to keep the module's bytecode below the limit? (Or is something else pushing the module over the limit?)

Yeah that's basically the alternative solution I mention on discord, will explore it, should work fine

packages/world/src/modules/core/CoreRegistrationSystem.sol
```diff
@@ -3,23 +3,18 @@ pragma solidity >=0.8.21;
...
  */
-contract CoreSystem is
+contract CoreRegistrationSystem is
   IWorldErrors,
-  AccessManagementSystem,
-  BalanceTransferSystem,
-  BatchCallSystem,
   ModuleInstallationSystem,
   StoreRegistrationSystem,
   WorldRegistrationSystem
```
holic, line 20: I kinda wonder if we should just go all the way and register all of these individually rather than grouping some of them here?

packages/world/src/modules/core/CoreModule.sol
```diff
@@ -97,34 +109,62 @@ contract CoreModule is Module {
...
    */
-  function _registerCoreSystem() internal {
+  function _registerSystem(address target, ResourceId systemId) internal {
     WorldContextProviderLib.delegatecallWithContextOrRevert({
       msgSender: _msgSender(),
       msgValue: 0,
-      target: coreSystem,
-      callData: abi.encodeCall(WorldRegistrationSystem.registerSystem, (CORE_SYSTEM_ID, CoreSystem(coreSystem), true))
+      target: coreRegistrationSystem,
+      callData: abi.encodeCall(WorldRegistrationSystem.registerSystem, (systemId, WorldContextConsumer(target), true))
```
holic, line 130: re: above, this is the bit that confused me and makes me wonder if we should just have all the systems registered individually rather than grouping some under `CoreRegistrationSystem`

it was a little unclear that `CoreRegistrationSystem` is a `WorldRegistrationSystem` and we're doing a delegatecall using the `WorldRegistrationSystem` interface rather than `CoreRegistrationSystem` interface

packages/world/src/modules/core/CoreModule.sol
```diff
@@ -97,34 +109,62 @@ contract CoreModule is Module {
...
    * @dev Iterates through known function signatures and registers them.
    */
   function _registerFunctionSelectors() internal {
-    string[19] memory functionSignatures = [
+    string[3] memory functionSignaturesAccessManagement = [
       // --- AccessManagementSystem ---
       "grantAccess(bytes32,address)",
       "revokeAccess(bytes32,address)",
-      "transferOwnership(bytes32,address)",
+      "transferOwnership(bytes32,address)"
```
holic, line 143: just thinking aloud but I wonder if it'd be worth our codegen creating a solidity file for system function selectors, so we can import from that instead of maintaining a list here

.changeset/silly-chefs-confess.md
```diff
@@ -0,0 +1,6 @@
+---
+"@latticexyz/world": major
+---
+
+- Added `CoreSystem2` and moved `BalanceTransferSystem` from `CoreSystem` to `CoreSystem2`
+- Added `CoreModule2`, which registers `CoreSystem2`
```
holic, line 6: I think this needs to be updated?

holic (commented):
thoughts on moving out the core module split into its own PR so we can keep the audit fix PRs scoped to just the things they fix?

holic:
core module split was moved to https://github.com/latticexyz/mud/pull/2128

packages/store/src/Schema.sol
yonadaaa, line 1: Heads up, there was a suggestion to simplify and split this loop into two in **N-8**:
https://github.com/latticexyz/mud/commit/7a11ba48bce1f719ff99d9f735f44c1cc4ede72c#diff-565f54743c2cafcf49b62c6c994b044ce220955a9faa2167d9c34fbe13f02ac1

Just a heads up in case that gets merged first

holic: https://github.com/latticexyz/mud/pull/2149

holic: Since this is an L-rated fix, gonna suggest we merge this first and then rebase/figure out how to apply the N-08 fix after.

packages/store/src/FieldLayout.sol
```diff
@@ -166,17 +168,30 @@ library FieldLayoutInstance {
...
 
     // Static lengths must be valid
+    uint256 _staticDataLength;
     for (uint256 i; i < _numStaticFields; ) {
       uint256 staticByteLength = fieldLayout.atIndex(i);
       if (staticByteLength == 0) {
-        revert FieldLayoutLib.FieldLayoutLib_StaticLengthIsZero();
+        revert FieldLayoutLib.FieldLayoutLib_StaticLengthIsZero(i);
       } else if (staticByteLength > WORD_SIZE) {
         revert FieldLayoutLib.FieldLayoutLib_StaticLengthDoesNotFitInAWord();
```
alvrs, line 177: should we add the index param to this error as well for consistency?

alvrs (approved):

PR #2052, opened 2023-12-10
test(store-indexer): add endpoint with static query

alvrs:


alvrs:
just for debugging purposes, closing now

PR #2053, opened 2023-12-11
feat(templates): add tests to show only the deployer can use .set

qbzzt:


templates/phaser/packages/contracts/test/CounterTest.t.sol
```diff
@@ -28,4 +28,18 @@ contract CounterTest is MudTest {
     counter = Counter.get();
     assertEq(counter, 2);
   }
+
+  function testFailCounterSet() public {
```
holic, line 32: it's now recommended practice to use `vm.expectRevert` rather than `testFail`

yonadaaa:
Closing as we can explain this in docs instead of a test templates!

Issue #1852, opened 2023-11-01
Add `isInstalled` to `Module`


alvrs:
> side note/not blocking: since we're starting to use this more, might be nice to lift this into `Module` as a helper function e.g. `isInstalled()` or `isInstalled(module)` and `requireNotInstalled()`

_Originally posted by @holic in https://github.com/latticexyz/mud/pull/1789#discussion_r1378758582_

holic:
done in https://github.com/latticexyz/mud/pull/2056

PR #2056, opened 2023-12-11
feat(world): add isInstalled to Module

R-Morpheus:
closing #1852


packages/world-modules/src/modules/erc20-puppet/ERC20Module.sol
```diff
@@ -32,16 +31,14 @@ contract ERC20Module is Module {
 
   function _requireDependencies() internal view {
     // Require PuppetModule to be installed
-    if (InstalledModules.get(PUPPET_MODULE_NAME, keccak256(new bytes(0))) == address(0)) {
+    if (!isInstalled(PUPPET_MODULE_NAME, abi.encode(keccak256(new bytes(0))))) {
```
holic, line 34: I think we're doing `keccak256` twice now (once here, once in `isInstalled`). Should this instance be removed?
```suggestion
    if (!isInstalled(PUPPET_MODULE_NAME, abi.encode(new bytes(0)))) {
```

PR #2058, opened 2023-12-11
feat(store-indexer): disable prepared statements

alvrs:


holic (dismissed):

holic:
Should we do this for the indexer backend too? I imagine it can fail there as well

PR #2071, opened 2023-12-16
docs(guides): using the token modules

qbzzt:


docs/pages/guides/adding-tokens.mdx
```diff
@@ -0,0 +1,681 @@
...
+   Normally the only entity allowed to mint ERC-20 tokens is the namespace owner.
+   The easiest way to make sure `RewardSystem` has enough tokens to give is simply to give it enough tokens, and refill as needed.
+
+   ```solidity
+       // This is a workaround for issue
+       // https://github.com/latticexyz/mud/issues/2089
+       token.mint(address(worldAddress), 10 ** 6 * RewardSize.get());
+   ```
+
+   The issue is that even though `RewardSystem` is in a non-root namespace (`DDDD`), when called from `TaskSystem` it runs as the `World` address - so we need to give that address tokens to distribute.
```
alvrs, line 528: i'd prefer to not have this in the docs, let's wait for the issue to be fixed. Alternatively we can already merge the first part of the guide and hold back on the second part (controlling from a system). Or yet another alternative would be to use a root system for now so the behavior of transferring to the world is expected.

qbzzt: Any idea how long it will be until the bug is fixed? Do you want me to try and fix it?

alvrs: it's fixed now!

alvrs:
Is this still stuck or ready for review?

qbzzt:
Should be ready, but it has been so long, maybe I should go over it again
to make sure nothing changed.

qbzzt:
Better to rewrite

PR #2079, opened 2023-12-21
fix(cli): Failed to get EIP-1559 fee

aLIEzsss4:
Fixing networks that do not support EIP-1559 fees


holic:
We unfortunately can't apply this everywhere. We'll have to find a way to detect when its appropriate to send legacy transactions and then conditionally add the flag.

aLIEzsss4:
maybe we can determine whether to add a flag [parseTransaction](https://viem.sh/docs/utilities/parseTransaction.html#parsetransaction).

PR #2080, opened 2023-12-22
fix(faucet): use MUD's sendTransaction for better nonce handling

### LINKED ISSUE #2070: faucet should use MUD's sendTransaction
for better nonce handling of lots of drips

R-Morpheus:
closes #2070

### DIFF
--- a/packages/faucet/bin/faucet-server.ts
+++ b/packages/faucet/bin/faucet-server.ts
@@ -4,7 +4,7 @@ import fastify from "fastify";
 import { fastifyTRPCPlugin } from "@trpc/server/adapters/fastify";
 import { http, createClient } from "viem";
 import { privateKeyToAccount } from "viem/accounts";
-import { AppRouter, createAppRouter } from "../src/createAppRouter";
+import { AppRouter, createAppRouter } from "../src";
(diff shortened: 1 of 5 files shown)


packages/faucet/bin/faucet-server.ts
```diff
@@ -4,7 +4,7 @@ import fastify from "fastify";
 import { fastifyTRPCPlugin } from "@trpc/server/adapters/fastify";
 import { http, createClient } from "viem";
 import { privateKeyToAccount } from "viem/accounts";
-import { AppRouter, createAppRouter } from "../src/createAppRouter";
+import { AppRouter, createAppRouter } from "../src";
```
holic, line 7: any reason for this change?

R-Morpheus: my ide was swearing and giving out an error. Remove the changes?

holic: curious to hear what the error is! but yes, my preference is not using the exports file for local imports

R-Morpheus: my new version of GoLand reacted badly to what could be shortened, had to be fixed

Issue #339, opened 2023-01-11
Remove mobx dependency


alvrs:
MUD (especially recs) started out building heavily on `mobx` as foundation for reactivity, but at some point switched to `rxjs` for performance reasons. However, `mobx` is still used in some places of the codebase, leading to "glue code" to convert `mobx` observable into `rxjs` observables and vice versa. We should remove all usages of `mobx` (eg replace them with `rxjs`'s `BehaviorSubject`) to be able to remove `mobx` as a dependency.

PR #2082, opened 2023-12-25
refactor: remove mobx

R-Morpheus:
work on #339

please see how I handled the removal of `mobx` in `utils`. if it looks good and the task is still relevant, then I will continue. 

here I rewrote the functions from `mobx` to `rxjs`, also changed the tests and deleted the file `mobx.ts`, so how can you do without it


packages/recs/src/Query.ts
```diff
@@ -508,7 +526,7 @@ export function defineQuery(
 
   return {
     matching,
```
holic, line 528: from an API standpoint, I wonder if we should return `matching.value` instead of the observable (esp. since the BehaviorSubject is mutable, which is kinda scary)

then we don't have to change all the other locations to use `matching.value`

packages/phaserx/package.json
```diff
@@ -25,7 +25,6 @@
   "dependencies": {
     "@latticexyz/utils": "workspace:*",
     "@use-gesture/vanilla": "10.2.9",
-    "mobx": "^6.7.0",
```
holic, line 28: Since this package is not really maintained (and should probably be considered deprecated), I am wondering if we can get away with not changing this here.

R-Morpheus: Okay, I'll take the changes away

R-Morpheus:
still, I can't fix one problem. I have one test that fails in `useEntityQuery.test.ts`. The problem is in the component re-rendering when calling the `setComponent()` function. 

It would be logical to assume that the problem is with `defineQuery()`, but it works fine, all tests pass. As I understand from the logs, the result is the same, even if you specify in the options `updateOnValueChange: false`

@holic what do you think? 🤔

Снимок экрана 2024-01-02 в 16 17 02

packages/recs/src/Query.ts
```diff
@@ -441,21 +452,23 @@ export function defineQuery(
...
-                matching.delete(previouslyMatchingEntity);
+                currentMatching.delete(previouslyMatchingEntity);
                 updates.push({
                   entity: previouslyMatchingEntity,
                   type: UpdateType.Exit,
                   component: update.component,
                   value: [undefined, undefined],
                 });
               }
+              matching.next(currentMatching);
```
holic, line 467: I am wondering if this line is causing the additional render?

R-Morpheus: no, deleting it did not lead to changes.

PR #2083, opened 2023-12-26
docs(faucet): `npx faucet-server` typo

emersonhsieh:
Command to start faucet should be `npx faucet-server` instead. Optional fix if moving to pnpm on docs.


holic:
I believe the current docs are correct. I think for the `npx` case, you don't need to install and want to run via `npx @latticexyz/faucet`

PR #2090, opened 2024-01-08
refactor(store,world): delete unused functions and variables [N-04]

yonadaaa:


packages/world/src/worldResourceTypes.sol
```diff
@@ -14,17 +14,9 @@ import { RESOURCE_TABLE, RESOURCE_OFFCHAIN_TABLE } from "@latticexyz/store/src/s
 // A namespace can include tables and systems.
 bytes2 constant RESOURCE_NAMESPACE = "ns";
 
-// Resource that identifies a module, an on-chain script that
-// can be executed on the World. Modules are used to install tables,
-// systems, hooks, and new entry point to a World in order to extend
-// its capabilities.
-bytes2 constant RESOURCE_MODULE = "md";
```
holic, line 21: @alvrs is it true that we don't use this anywhere?

alvrs: i don't think we do! modules are not exactly resources that are installed into the world either like other kinds of resources (namespaces, tables, systems)

packages/store/src/Bytes.sol
```diff
@@ -6,41 +6,12 @@ pragma solidity >=0.8.21;
  * @notice Utility functions for bytes.
  */
 library Bytes {
```
holic, line 8: this library feels a little incomplete with half the bytes variants removed but 🤷‍♂️

maybe in the future, we move this into the `common` package or similar so that it represents a pure lib and not dependent on anything store related

holic (approved):

PR #2091, opened 2024-01-08
refactor(store,world): rename ambiguous elements [N-03]

yonadaaa:


packages/store/src/Hook.sol
```diff
@@ -30,15 +30,15 @@ library HookLib {
    * @notice Filter a hook from the hook list by its address.
    * @dev This function writes the updated hook list to the table in place.
    * @param hookTableId The resource ID of the hook table.
-   * @param tableWithHooks The resource ID of the table with hooks to filter.
+   * @param elementWithHooks The resource ID of the table with hooks to filter.
```
holic, line 33: can we name this `resourceWithHooks`?

"element" is a term we've never used before and I think "resource" is more consistent: https://www.notion.so/latticexyz/MUD-naming-conventions-terminology-66e11356a1934465b2aa1a951c61a299

packages/world/src/World.sol
```diff
@@ -52,7 +52,7 @@ contract World is StoreData, IWorldKernel {
   /**
    * @dev Prevents the World contract from calling itself.
    */
-  modifier requireNoCallback() {
```
holic, line 55: not blocking but I feel like we could expand the docs here to explain _why_ this modifier is here and important to be used, esp since this relates to a vulnerability

holic (dismissed):

packages/world/src/World.sol
holic: added issue here: https://github.com/latticexyz/mud/issues/2106

holic:
should prob add a changeset for this one!

holic (approved):

PR #2092, opened 2024-01-08
refactor(store,world): code simplification [N-08]

yonadaaa:
I have arranged each of these tasks as a commit, which we can modify or delete seperately

TODO: 

- [x] Simplify Schema library validation loop by prioritizing static and dynamic fields.
- [ ] ~Replace redundant Memory.copy with identity precompile call.~ ([discussion]( https://github.com/latticexyz/mud/pull/2092#issuecomment-1885327493))
- [x] Optimize leftMask function by returning a right mask directly.
- [x] Resolve ResourceId.sol inconsistency and reduce duplication.
- [x] Standardize integer base in assembly blocks (hexadecimal/decimal).
- [x] Align pointer usage in Slice library (self.pointer()).
- [x] Improve EIP-165 support using interface type ID.
- [ ] ~Refactor FieldLayout.numFields using numStaticFields and numDynamicFields.~ ([discussion]( https://github.com/latticexyz/mud/pull/2092#issuecomment-1881605808))
- [x] Address inconsistency in Slice.sol and Schema.sol leading underscores.
- [x] Standardize initialization (zero/default) across the codebase.
- [x] Optimize StoreCore.getFieldLayout function using Tables._getFieldLayout.
- [x] Ensure consistency in WorldContextConsumer._world using StoreSwitch.
- [x] Make WorldFactory coreModule immutable if it's never changed.
- [x] Correct parameter type in WorldRegistrationSystem.registerSystem.
- [x] Simplify Bytes library slice functions with named return variables.
- [x] Streamline getSubslice validation condition with a concise statement.
- [x] Eliminate unnecessary extcodesize opcode in Create2.deploy call check.


yonadaaa:
The suggestion about `FieldLayout.numFields` function causes a ~10x gas increase (see [this branch]( https://github.com/latticexyz/mud/compare/yonadaaa/num-fields-refactor?expand=1)) so maybe we keep that code inlined for now.

holic:
> The suggestion about `FieldLayout.numFields` function causes a ~10x gas increase (see [this branch](https://github.com/latticexyz/mud/compare/yonadaaa/num-fields-refactor?expand=1)) so maybe we keep that code inlined for now.

OOC does the same thing happen if these are all free functions rather than being wrapped in a lib?

packages/store/src/IERC165.sol
```diff
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MIT
 pragma solidity >=0.8.21;
 
-bytes4 constant ERC165_INTERFACE_ID = IERC165.supportsInterface.selector;
+bytes4 constant ERC165_INTERFACE_ID = type(IERC165).interfaceId;
```
holic, line 4: if the `interfaceId` is now built-in, do we need the constants anymore?

yonadaaa: Nope we don't need the constants! But I wanted to avoid the bigger diff of importing contracts in each touchpoint

holic: I think each place that uses this interface ID is also using the interface itself, so we should already be importing them?

yonadaaa:
Some thoughts on the `Memory.copy` suggestion:

> The [Memory.copy function](https://github.com/latticexyz/mud/blob/12a9eb1afb18a7f7cb43d774cabb2d5af7c74e6b/packages/store/src/Memory.sol#L33) is redundant to the [identity precompile](https://www.evm.codes/precompiled#0x04?fork=shanghai). Consider replacing the function's logic with a call to the precompile for simplicity and efficiency.

It would likely decrease gas costs if we implement this, but introducing `STATICCALL` to call the precompiled contract means all the calling functions must use `view`. Plus `Memory.Copy` doesn't seemed to be used outside of tests.

dk1a:
> It would likely decrease gas costs if we implement this, but introducing STATICCALL to call the precompiled contract means all the previously pure touchpoints must use view.

we used to have the precompile, it wasn't worth it, codegen throws lots of warnings about pure/view (or requires more conditions to conditionally render some things as pure), and precompile is only really better for gas for huge blobs

packages/store/src/IERC165.sol
yonadaaa: Implemented this and amending the commit. sorry for the messy history

yonadaaa:
Closing in favour of https://github.com/latticexyz/mud/pull/2140

PR #2095, opened 2024-01-09
fix(world): check namespace exists before balance transfer [L-03]

yonadaaa:


packages/world/src/modules/core/implementations/BalanceTransferSystem.sol
```diff
@@ -37,6 +38,9 @@ contract BalanceTransferSystem is System, IWorldErrors {
       revert World_InvalidResourceType(RESOURCE_NAMESPACE, toNamespaceId, toNamespaceId.toString());
     }
 
+    // Require the namespace to exist
+    if (!ResourceIds.getExists(toNamespaceId)) revert World_ResourceNotFound(toNamespaceId, toNamespaceId.toString());
+
     // Require caller to have access to the namespace
     AccessControl.requireAccess(fromNamespaceId, _msgSender());
```
holic, line 45: I wonder how much gas we'd incur if we added this check in `requireAccess` instead?

holic (dismissed):

packages/world/src/modules/core/implementations/BalanceTransferSystem.sol
```diff
@@ -37,6 +38,9 @@ contract BalanceTransferSystem is System, IWorldErrors {
       revert World_InvalidResourceType(RESOURCE_NAMESPACE, toNamespaceId, toNamespaceId.toString());
     }
 
+    // Require the namespace to exist
+    if (!ResourceIds.getExists(toNamespaceId)) revert World_ResourceNotFound(toNamespaceId, toNamespaceId.toString());
```
alvrs, line 42: style nit (wonder if we can configure a linter to do this for consistency)
```suggestion
    if (!ResourceIds.getExists(toNamespaceId)) {
      revert World_ResourceNotFound(toNamespaceId, toNamespaceId.toString());
    }
```

alvrs:
just added a `requireExistence` helper to `AccessControl` in https://github.com/latticexyz/mud/pull/2007, we can use that for this check too!

yonadaaa:
This PR is failing integration tests because `CoreSystem` is too big to deploy. This should be fixed in https://github.com/latticexyz/mud/pull/2046.

yonadaaa:
Using the `requireExistence` helper got this under the size limit!

.changeset/seven-carpets-develop.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/world": patch
+---
+
+Require that namespaces exist before transferring ETH to them.
```
holic, line 5:
```suggestion
Namespace balances can no longer be transferred to non-existent namespaces.
```
reworded for consistent tense with other changesets and to be a bit more specific about namespace balances

holic (approved):

PR #2096, opened 2024-01-09
fix(world): prevent misconfigured delegations, allow unregistering [L-04]

yonadaaa:
Currently, we allow clearing delegations by passing empty calldata. However this could cause invalid delegations, so this PR performs full validation regardless of whether `initCallData` is empty. 

Delegations can now be explicitly unregistered by calling `unregisterDelegation`.


packages/world/src/modules/core/implementations/WorldRegistrationSystem.sol
```diff
@@ -262,6 +262,13 @@ contract WorldRegistrationSystem is System, IWorldErrors {
     }
   }
 
+  function unregisterDelegation(address delegatee) public {
+    // Delete the delegation control contract address
+    UserDelegationControl.deleteRecord({ delegator: _msgSender(), delegatee: delegatee });
+
+    // TODO: clear the specific delegation control
```
yonadaaa, line 269: Do we care about deleting the entries in the `DelegationControl`. If so should they implement some `deleteDelegation` function? 

Semantically, it's not obvious that calling `initDelegation` with empty calldata should "uninitialize" the delegation. For example, the control might increment some counter each time a delegation is registered.

yonadaaa:
This PR is failing integration tests because `CoreSystem` is too big to deploy. This should be fixed in https://github.com/latticexyz/mud/pull/2046.

holic (approved):

packages/world/src/modules/core/implementations/WorldRegistrationSystem.sol
```diff
@@ -260,6 +260,11 @@ contract WorldRegistrationSystem is System, IWorldErrors {
     }
   }
 
+  function unregisterDelegation(address delegatee) public {
```
holic, line 263: need to make sure to add this to the core module function registration step

holic:
Can we add a test while we're here, just to make sure we can unregister a delegation from the world?

yonadaaa:
Added a test that you can unregister a delegation, then the call fails

holic (approved):

PR #2097, opened 2024-01-09
refactor(store-indexer,store-sync): decoded indexer reads and writes to different DB's

yonadaaa:
Allows the decoded postgres indexer to read and write to different databases, allowing one or the other to be exposed publicly


packages/store-indexer/bin/postgres-decoded-indexer.ts
```diff
@@ -15,7 +15,8 @@ const env = parseEnv(
   z.intersection(
     indexerEnvSchema,
     z.object({
-      DATABASE_URL: z.string(),
```
holic, line 18: can we keep this for backwards compat?

packages/store-indexer/bin/postgres-decoded-indexer.ts
```diff
@@ -35,16 +36,21 @@ const publicClient = createPublicClient({
...
 
 const chainId = await publicClient.getChainId();
-const database = drizzle(postgres(env.DATABASE_URL, { prepare: false }));
+const readDatabase = drizzle(postgres(env.READ_DATABASE_URL, { prepare: false }));
+const writeDatabase = drizzle(postgres(env.WRITE_DATABASE_URL, { prepare: false }));
 
-const { storageAdapter, tables } = await createStorageAdapter({ database, publicClient });
+const { storageAdapter, tables } = await createStorageAdapter({
+  readDatabase: readDatabase,
+  writeDatabase: writeDatabase,
```
holic, line 44: can we clarify the naming here? maybe `bytesDatabase` and `decodedDatabase`? (would love to come up with some good naming for these DB types because we throw around a lot of different names)

and similar treatment above when we initiate but maybe allow for `DATABASE_URL` to be used as a fallback for both (which would give us backwards compat)

packages/store-indexer/bin/postgres-decoded-indexer.ts
yonadaaa: Yes. I was worried about readability long term

packages/store-indexer/bin/postgres-decoded-indexer.ts
yonadaaa: I like the idea of having the fallback, perhaps if WRITE_DATABASE isn't defined? we do risk trying to solve for every use case though

packages/store-indexer/bin/postgres-decoded-indexer.ts
```diff
@@ -35,16 +37,37 @@ const publicClient = createPublicClient({
...
 
 const chainId = await publicClient.getChainId();
-const database = drizzle(postgres(env.DATABASE_URL, { prepare: false }));
 
-const { storageAdapter, tables } = await createStorageAdapter({ database, publicClient });
+let readDatabaseUrl = env.READ_DATABASE_URL;
+let writeDatabaseUrl = env.WRITE_DATABASE_URL;
+
+if (env.DATABASE_URL) {
+  console.warn("DATABASE_URL is deprecated. Please use READ_DATABASE_URL and WRITE_DATABASE_URL");
```
alvrs, line 45: it doesn't have to be deprecated imo, we could just say if you want both to be the same you can use `DATABASE_URL` (which I would expect to be a common use case too)

With that we can simplify this whole logic to:

```
const readDatabaseUrl = env.READ_DATABASE_URL ?? env.DATABASE_URL;
const writeDatabaseUrl = env.WRITE_DATABASE_URL ?? env.DATABASE_URL;

if(!readDatabaseUrl || !writeDatabaseUrl) {
  throw new Error("No database URL defined. Please provide either `DATABASE_URL` or `READ_DATABASE_URL` and `WRITE_DATABASE_URL` environment variables.");
}

File: packages/store-indexer/bin/postgres-decoded-indexer.ts [hunk 4]
```diff
@@ -35,16 +37,37 @@ const publicClient = createPublicClient({
...
+
+  readDatabaseUrl = env.DATABASE_URL;
+  writeDatabaseUrl = env.DATABASE_URL;
+} else if (env.READ_DATABASE_URL && env.WRITE_DATABASE_URL) {
+  readDatabaseUrl = env.READ_DATABASE_URL;
+  writeDatabaseUrl = env.WRITE_DATABASE_URL;
+} else {
+  throw new Error("no database URL defined");
+}
```
alvrs, line 54:
```suggestion
const readDatabaseUrl = env.READ_DATABASE_URL ?? env.DATABASE_URL;
const writeDatabaseUrl = env.WRITE_DATABASE_URL ?? env.DATABASE_URL;

if(!readDatabaseUrl || !writeDatabaseUrl) {
  throw new Error("No database URL defined. Please provide either `DATABASE_URL` or `READ_DATABASE_URL` and `WRITE_DATABASE_URL` environment variables.");
}
```

yonadaaa: Applied this, thanks!

holic:
moving this to draft because the current decoded indexer implementation may make this approach not feasible while ensuring good data

PR #2099, opened 2024-01-09
fix(world): check the core module name is correct on initialisation [L-05]

yonadaaa:


packages/world/src/World.sol
```diff
@@ -75,6 +75,11 @@ contract World is StoreData, IWorldKernel {
       revert World_AlreadyInitialized();
     }
 
+    // Require the module name to be core
+    if (coreModule.getName() != CORE_MODULE_NAME) {
+      revert("Module is not the core module");
```
yonadaaa, line 80: Do we want a custom error for this?

holic: thinking more about the original issue/report, I wonder if this check is actually useful

If the implementation of `coreModule` is malicious, couldn't it easily sidestep this check by returning different values from `getName()` based on the caller?

curious if @alvrs has opinions here

Issue #2085, opened 2024-01-02
CI: disable scripts on install


holic:
from @roninjin10 
> Thinking about npm safety might be worth disabling scripts on installs https://github.com/ethereum-optimism/optimism/pull/8626

PR #2102, opened 2024-01-10
feat(CI): disable scripts on install

R-Morpheus:
closing #2085


package.json
```diff
@@ -15,6 +15,8 @@
     "docs:generate:api": "bun scripts/render-api-docs.ts",
     "foundryup": "curl -L https://foundry.paradigm.xyz | bash && bash ~/.foundry/bin/foundryup",
     "gas-report": "pnpm run --recursive --parallel gas-report",
+    "install:ci": "pnpm install --ignore-scripts",
+    "install:ci:frozen": "pnpm install --ignore-scripts --frozen-lockfile",
```
holic, line 19: my preference would be to just add `--ignore-scripts` flag to the CI command and not add new npm scripts

holic:
something weird is going on with the e2e tests for the sqlite indexer and makes me wonder if there was a script required during install of sqlite deps?

holic:
looks like yes: https://github.com/WiseLibs/better-sqlite3/blob/master/package.json#L33

we may need a workaround specific for this package

holic:
may also wanna add `--ignore-scripts` (and sqlite workaround) to the Dockerfile too: https://github.com/latticexyz/mud/blob/main/Dockerfile#L47

R-Morpheus:
I'm not sure about the solution, as I realized that `--script-install` doesn't have a specific flag to load some scripts. It is quite possible to add a `postinstall`, but after trying it, it did not work for me.

another way is to add another command to `.yml`, but I'm not sure what exactly we need to download. As I understand it, the `sqlite` package uses `store-sync`, is that what we need?

PR #2103, opened 2024-01-10
chore(common,store): remove unused codegen vars [N-06]

holic:


packages/common/src/codegen/render-solidity/common.ts
```diff
@@ -127,21 +120,33 @@ export function renderAbsoluteImports(imports: AbsoluteImportDatum[]): string {
...
-  result += callback(undefined, "StoreSwitch", "", undefined, "");
-  result += callback(undefined, "StoreCore", "", undefined, "_", true);
+  result += callback({ _typedStore: undefined, _store: "StoreSwitch", _commentSuffix: "", _methodNamePrefix: "" });
+  result += callback({
+    _typedStore: undefined,
+    _store: "StoreCore",
+    _commentSuffix: "",
+    _methodNamePrefix: "_",
+    _internal: true,
+  });
```
alvrs, line 139: i like the refactor to named arguments! should we omit `_typedStore` here if it's undefined? (similar to how we don't include `_untypedStore`)

alvrs (approved):

packages/common/src/codegen/render-solidity/common.ts
holic: whoops, merged too hastily, will do this in a follow up

holic: oh and `_untypedStore` was removed because it wasn't actually used anywhere

PR #2108, opened 2024-01-10
feat(examples): access `World` from different accounts

qbzzt:
This example would be a good starting point for a `callFrom` tutorial, and possibly other tutorials.


examples/multiple-accounts/LICENSE
```diff
@@ -0,0 +1,19 @@
+Copyright (c) 2024 Ori Pomerantz <qbzzt1@gmail.com>
```
alvrs, line 1: we don't have a licence file in the other example projects bc they inherit the monorepo root's license (MIT), don't think we need a separate one here

alvrs (approved):

alvrs:
Looks like tests are failing! Also should we add this to the CI action that tests that the example projects are working?

alvrs (approved):

qbzzt:
@alvrs , can you please hit merge? I'm not authorized to do it, even with an approval.

PR #2111, opened 2024-01-11
fix(world): limit call context of `CoreSystem` to delegatecall [C-02]

alvrs:


.changeset/lazy-foxes-applaud.md
```diff
@@ -0,0 +1,9 @@
+---
+"@latticexyz/world": patch
+---
+
+Systems are expected to be always called via the central World contract.
+Depending on whether it is a root or non-root system, the call is performed via `delegatecall` or `call`.
+Since Systems are expected to be stateless and only interact with the World state, it is not necessary to prevent direct calls to the systems.
+However, since the `CoreSystem` is known to always be registered as a root system in the World, it is always expected to be delegatecalled,
+so we made this expectation explicit by reverting if it is not delegatecalled.
```
holic, line 9: your notes here are great! can we add these notes to `LimitedCallContext` as well, so folks know the why behind using this contract/set of modifiers?

alvrs:
this one brought the core module over the size limit as well, blocked on #2046

holic (dismissed):
Not blocking but do we have/want tests for these methods to make sure they revert?

alvrs:
> Not blocking but do we have/want tests for these methods to make sure they revert?

Agree with this and started implementing it but it actually requires _a lot_ of boilerplate for testing that each method reverts as expected, and I didn't find a way to test that _no_ methods can be called directly, so i would leave this for a followup

holic (approved):

PR #2128, opened 2024-01-14
refactor(cli,world,world-modules): split and separately deploy core systems

dk1a:


.changeset/silver-ligers-grin.md
```diff
@@ -0,0 +1,9 @@
+---
+"@latticexyz/world": major
+---
+
+- Split `CoreSystem` into `AccessManagementSystem`, `BalanceTransferSystem`, `BatchCallSystem`, `CoreRegistrationSystem`
```
alvrs, line 5: Thoughts on `CoreRegistrationSystem` vs just `RegistrationSystem`?

packages/cli/src/deploy/ensureWorldFactory.ts
```diff
@@ -6,10 +10,57 @@ import { salt } from "./common";
...
+  from: deployer,
+  bytecode: coreRegistrationSystemBytecode,
+  salt,
+});
+
 export const coreModuleDeployedBytecodeSize = size(coreModuleBuild.deployedBytecode.object as Hex);
 export const coreModuleBytecode = encodeDeployData({
   bytecode: coreModuleBuild.bytecode.object as Hex,
-  abi: [],
+  abi: parseAbi(["constructor(address,address,address,address)"]),
```
alvrs, line 62: is it possible to get this form an import of the code module? so this can throw a type error when it changes instead of a runtime error if the abi doesn't match

packages/world/src/modules/core/CoreRegistrationSystem.sol
```diff
@@ -3,23 +3,18 @@ pragma solidity >=0.8.21;
...
 import { StoreRegistrationSystem } from "./implementations/StoreRegistrationSystem.sol";
 import { WorldRegistrationSystem } from "./implementations/WorldRegistrationSystem.sol";
 
 /**
- * @title Core System for World
- * @notice This system aggregates all World functionalities externalized from the World contract, aiming to keep the World contract's bytecode lean.
+ * @title Core Registration System for World
+ * @notice This system aggregates World registration and installation functionalities externalized from the World contract, aiming to keep the World contract's bytecode lean.
  * @dev Aggregates multiple system implementations for the World.
+ * Split into multiple systems because of the bytecode size limit.
```
alvrs, line 14:
```suggestion
```
This comment makes sense in the context of this PR but not when reading this by itself file afterwards

packages/world/src/modules/core/CoreModule.sol
```diff
@@ -28,16 +29,27 @@ import { WorldRegistrationSystem } from "./implementations/WorldRegistrationSyst
...
+  address internal immutable accessManagementSystem;
+  address internal immutable balanceTransferSystem;
+  address internal immutable batchCallSystem;
+  address internal immutable coreRegistrationSystem;
+
+  constructor(
+    address _accessManagementSystem,
+    address _balanceTransferSystem,
+    address _batchCallSystem,
+    address _coreRegistrationSystem
```
alvrs, line 46: If we made these the respective contract types we'd get improved compile time type checks (right now it would be possible to mess up the argument order without compile time errors). I'm not sure if i'd cause the bytecode size to increase though

packages/world/src/modules/core/constants.sol
```diff
@@ -13,9 +13,33 @@ import { RESOURCE_SYSTEM } from "../../worldResourceTypes.sol";
 bytes16 constant CORE_MODULE_NAME = bytes16("core");
 
 /**
- * @dev Resource ID for the core system.
- * @dev This ID is derived from the RESOURCE_SYSTEM type, the ROOT_NAMESPACE, and the CORE_MODULE_NAME.
+ * @dev Resource ID for access management system.
+ * @dev This ID is derived from the RESOURCE_SYSTEM type, the ROOT_NAMESPACE, and a suffixed CORE_MODULE_NAME.
```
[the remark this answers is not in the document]
dk1a: updated comments

packages/world/src/modules/core/CoreModule.sol
dk1a: They are only used as addresses so I'd rather store them as such instead of extra casts within the registration function
But the constructor should indeed receive typed addresses

packages/cli/src/deploy/ensureWorldFactory.ts
dk1a: Maybe with abitype (we don't have it in cli package yet)
I'd rather not do this as part of this PR, or right now given I wanna move on to codegen refactor

holic: I think what he means is importing the typed ABI, i.e.

```ts
import coreModuleAbi from "@latticexyz/world/out/CoreModule.sol/CoreModule.abi.json" assert { type: "json" };

export const coreModuleBytecode = encodeDeployData({
  bytecode: coreModuleBuild.bytecode.object as Hex,
  abi: coreModuleAbi,
  ...
});
```

dk1a: Oh I see, well yeah point is I remember next to no context for this

.changeset/silver-ligers-grin.md
dk1a: I guess I see `Core` as some aggregation prefix
agree that it's fine either way

holic (approved):

PR #2129, opened 2024-01-15
refactor(world): make coreModule variable immutable [n-08]

yonadaaa:
Makes the `coreModule` variable in `WorldFactory` [immutable](https://docs.soliditylang.org/en/latest/contracts.html#constant-and-immutable-state-variables) because it is only set once, in the constructor. This decreases gas cost:

> Compared to regular state variables, the gas costs of constant and immutable variables are much lower ... Immutable variables are evaluated once at construction time and their value is copied to all the places in the code where they are accessed.

### DIFF
--- a/packages/world/gas-report.json
+++ b/packages/world/gas-report.json
@@ -57,13 +57,13 @@
     "file": "test/Factories.t.sol",
     "test": "testCreate2Factory",
     "name": "deploy contract via Create2",
-    "gasUsed": 4662733
+    "gasUsed": 4647528
   },
   {
     "file": "test/Factories.t.sol",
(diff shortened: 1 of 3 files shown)


packages/world/gas-report.json
```diff
@@ -57,13 +57,13 @@
     "file": "test/Factories.t.sol",
     "test": "testCreate2Factory",
     "name": "deploy contract via Create2",
-    "gasUsed": 4662733
+    "gasUsed": 4647528
```
yonadaaa, line 60: I'm not too sure why this causes a change here

holic (approved):

PR #2130, opened 2024-01-15
refactor(store): simplify check in Slice.getSubslice [n-08]

yonadaaa:
We can simplify this by check:
```solidity
if (!(start <= end && end <= data.length)) revert Slice_OutOfBounds(data, start, end);
```
    
by negating it:
```solidity
if (start > end || end > data.length) revert Slice_OutOfBounds(data, start, end);
```

### DIFF
--- a/.changeset/thin-days-sparkle.md
+++ b/.changeset/thin-days-sparkle.md
@@ -1,0 +1,5 @@
+---
+"@latticexyz/store": patch
+---
+
+Simplified a check in `Slice.getSubslice`.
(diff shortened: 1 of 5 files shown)


.changeset/thin-days-sparkle.md
holic, line 1: I feel like this doesn't really need a changeset

holic (approved):

.changeset/thin-days-sparkle.md
yonadaaa: I would say it needs a changeset because it changes the bytecode and internal behaviour of the compiled code, even if we are confident that the output is the same.

PR #2131, opened 2024-01-15
refactor(world): extcodesize not needed in Create2 library [n-08]

yonadaaa:


holic:
Do we have gas coverage to know if this changes gas?

And if there are side effects in terms of gas, I almost prefer the current approach with `extcodesize` as it reads a bit clearer to me.

yonadaaa:
> Do we have gas coverage to know if this changes gas?

It doesn't seem like it. I'll check which codepath uses this and add some.

> I almost prefer the current approach with extcodesize as it reads a bit clearer to me.

If we remove the opcode we should add a comment because it requires somewhat contextual knowledge - that the address will be zero in this case.

holic:
Oh I see, the logic is changing from "check if the resulting contract address exists/has bytecode" into "check if the resulting contract address is address zero"?

yonadaaa:
Exactly, if the call was successful the address should be _something_ (not zero), and the bytecode will be non-empty too. I don't think you have one without the other. 

I'm not sure why the `create2` call doesn't just revert. I guess it's low-level and allows the caller to handle the error.

holic (approved):

PR #2133, opened 2024-01-15
refactor(store): slice library uses pointer consistently [n-08]

yonadaaa:
The `Slice` library uses `self.pointer()` and `pointer(self)` which are interchangeable. With this PR it only uses `pointer(self)`, which I think is more "direct" than a method attached by the global `using` statement.

### DIFF
--- a/.changeset/little-starfishes-walk.md
+++ b/.changeset/little-starfishes-walk.md
@@ -1,5 +1,0 @@
----
-"@latticexyz/store": patch
----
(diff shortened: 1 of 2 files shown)


.changeset/little-starfishes-walk.md
holic, line 1: not sure we need a changelog for stylistic changes

holic (approved):

PR #2134, opened 2024-01-15
refactor(store,world,world-modules): do not explicitly initialise to zero [n-08]

yonadaaa:
Solidity variables are initialised to zero by default so we don't need to explicitly set them.

### DIFF
--- a/.changeset/soft-panthers-develop.md
+++ b/.changeset/soft-panthers-develop.md
@@ -1,0 +1,7 @@
+---
+"@latticexyz/world-modules": patch
+"@latticexyz/store": patch
+"@latticexyz/world": patch
+---
+
+Refactored various Solidity files to not explicitly initialise variables to zero.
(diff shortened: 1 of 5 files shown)


holic (approved):
> Solidity variables are initialised to zero by default so we don't need to explicitly set them.

although technically correct, I feel like the clarity here is better than not (esp since there's no gas consequences)

but I am indifferent

yonadaaa:
> the clarity of initializing with zero is better than not 

I'm inclined to agree. The original audit suggestion was actually for consistency, so we can make a new PR that does explicit initialisation, and change [the few](https://github.com/latticexyz/mud/blob/main/packages/store/src/PackedCounter.sol#L44) places that are missing it.

alvrs:
Unfortunately I do think there are gas implications with initializing with 0. Not initializing variables with their default value is the "best practice"  in Solidity (see [here](https://detectors.auditbase.com/skip-zero-initialization-solidity) or [here](https://hacken.io/discover/solidity-gas-optimization/)) and i think we do it in many more places across the codebase so it might be hard to find and change all of them

yonadaaa:
I was trying to find a best practice guide, thanks. In that case I'll go ahead and merge this, in future I suppose we can refactor to do explicit initalisation

PR #2135, opened 2024-01-15
refactor(store,world,world-modules): global using statement for resource id [n-08]

yonadaaa:


holic (approved):
much nicer!

holic:
Could we do the same treatment for `WorldResourceId`?

yonadaaa:
> Could we do the same treatment for WorldResourceId?

I don't see why not - I'll make an issue!

PR #2136, opened 2024-01-15
refactor(store,world): use hexadecimal as integer base [n-08]

yonadaaa:
Specify integers as hex (ie. 0x20) instead of decimal (ie. 32). This is purely for consistency and does not effect the compiled code.

### DIFF
--- a/.changeset/happy-ants-lay.md
+++ b/.changeset/happy-ants-lay.md
@@ -1,0 +1,6 @@
+---
+"@latticexyz/store": patch
+"@latticexyz/world": patch
+---
+
+Refactored various files to specify integers in a hex base instead of decimals.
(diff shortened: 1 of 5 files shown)


holic (approved):
I'm a bit torn on this, as its easier for me to read/understand the decimal than the hexidecimal. But I am not an assembly maxi

dk1a:
imo hex makes more sense for pointers, cause it's indeed more consistently used for them across various places (including solidity docs)
for me personally it also helps differentiate at a glance between pointer values (which are usually in increments of 0x20) and other numbers

But I can definitely see how decimal is easier, especially with 1F

yonadaaa:
I think we should be consistent in all our assembly unless there is a particularly good counterexample, and hexadecimal is the standard now. Also most (all?) of these changes are to pointers which are definitely nice in hex.

Issue #2113, opened 2024-01-11
solidity: replace usages of `testFail` with `expectRevert`


holic:
`testFail` is considered an anti-pattern but kept for backwards compat

aditya520:
Resolved here.
https://github.com/latticexyz/mud/pull/2137

PR #2137, opened 2024-01-16
test(store): replace testFail with expectRevert

aditya520:
fixes https://github.com/latticexyz/mud/issues/2113


holic (approved):
Thanks so much! 💯

Issue #2118, opened 2024-01-12
Flakey fuzz test in ERC721 module


yonadaaa:
A fuzz run of the `testBurnRevertAccessDenined` test failed here: https://github.com/latticexyz/mud/actions/runs/7503092661/job/20427117636?pr=2117

```
packages/world-modules test:ci: Failing tests:
packages/world-modules test:ci: Encountered 1 failing test in test/ERC721.t.sol:ERC721Test
packages/world-modules test:ci: [FAIL. Reason: call did not revert as expected; counterexample: calldata=0x78eee1c80000000000000000000000000000000000000000000000000000000000003b770000000000000000000000000000000000000000000000000000000000004ac80000000000000000000000007fa9385be102ac3eac297483dd6233d62b3e1496 args=[15223 [1.522e4], 0x0000000000000000000000000000000000004Ac8, 0x7FA9385bE102ac3EAc297483Dd6233D62b3e1496]] testBurnRevertAccessDenined(uint256,address,address) (runs: 133, μ: 212238, ~: 212238)
packages/world-modules test:ci: Encountered a total of 1 failing tests, 130 tests succeeded
```

holic:
We should probably increase the verbosity of forge tests in CI so we can see where this originated from

I am guessing the `0x7FA9385bE102ac3EAc297483Dd6233D62b3e1496 ` address for the `operator` argument is probably some contract address that was created as part of the test and is failing due to some check through a series of calls.

PR #2139, opened 2024-01-16
test(world-modules): fix failing flaky erc721 test

yonadaaa:
Trying to fix https://github.com/latticexyz/mud/issues/2118

The test fails with a particular fuzzed input. So far I've added a failing test to investigate.


packages/world-modules/test/ERC721.t.sol
yonadaaa, line 1: I don't think we need a specific test for this because [`testBurn`](https://github.com/latticexyz/mud/blob/main/packages/world-modules/test/ERC721.t.sol#L176C13-L176C13) burns tokens as `address(this)`.

holic (approved):

PR #2142, opened 2024-01-16
fix(world): proper resource ID validation [L-09]

yonadaaa:


packages/world/src/modules/core/implementations/AccessManagementSystem.sol
```diff
@@ -3,22 +3,30 @@ pragma solidity >=0.8.21;
...
    * @dev Requires the caller to own the namespace.
    * @param resourceId The ID of the resource to grant access to.
    * @param grantee The address to which access should be granted.
    */
   function grantAccess(ResourceId resourceId, address grantee) public virtual {
+    // Require the resource to exist
+    AccessControl.requireExistence(resourceId);
+
     // Require the caller to own the namespace
     AccessControl.requireOwner(resourceId, _msgSender());
```
holic, line 31: would this check satisfy `requireExistence`? a non-existent resource would be "owned" by address zero, this fail this `requireOwner` check?

I guess adding this check here means the error is much more accurate/actionable

packages/world/src/modules/core/implementations/BalanceTransferSystem.sol
```diff
@@ -38,6 +39,11 @@ contract BalanceTransferSystem is System, IWorldErrors {
       revert World_InvalidResourceType(RESOURCE_NAMESPACE, toNamespaceId, toNamespaceId.toString());
     }
 
+    // Require the provided namespace name to be empty
+    if (toNamespaceId.getName() != ROOT_NAME) {
+      revert World_InvalidResourceId(toNamespaceId, toNamespaceId.toString());
+    }
```
holic, line 45: I wonder if we should add an `isNamespace()` helper that checks for type + empty name

holic (dismissed):
nothing blocking!

packages/world/src/modules/core/implementations/BalanceTransferSystem.sol
yonadaaa: Yes I'm thinking that; something like `AccessControl` but undecided exactly where it should live

holic: I think we could put this in `WorldResourceIdLib`

packages/world/src/modules/core/implementations/AccessManagementSystem.sol
yonadaaa: I had the same thought, but having the explicit check gives a better error message and is less fragile.

I guess the "theme" of this PR is methodically adding these checks everywhere, so you know the requirements without having to "figure it out" from context.

packages/store/src/ResourceId.sol
```diff
@@ -8,6 +8,8 @@ pragma solidity >=0.8.21;
  */
 type ResourceId is bytes32;
 
+using ResourceIdInstance for ResourceId global;
```
holic, line 11: this is also added in https://github.com/latticexyz/mud/pull/2140

packages/world/src/modules/core/implementations/BalanceTransferSystem.sol
```diff
@@ -10,8 +10,10 @@ import { WorldResourceIdInstance } from "../../../WorldResourceId.sol";
 import { AccessControl } from "../../../AccessControl.sol";
 import { RESOURCE_NAMESPACE } from "../../../worldResourceTypes.sol";
 import { IWorldErrors } from "../../../IWorldErrors.sol";
+import { ROOT_NAME } from "../../../constants.sol";
```
holic, line 13:
```suggestion
```

packages/world/src/modules/core/implementations/AccessManagementSystem.sol
```diff
@@ -3,22 +3,31 @@ pragma solidity >=0.8.21;
 
 import { System } from "../../../System.sol";
 import { AccessControl } from "../../../AccessControl.sol";
-import { ResourceId } from "../../../WorldResourceId.sol";
+import { ResourceId, ResourceIdInstance, WorldResourceIdInstance } from "../../../WorldResourceId.sol";
+import { RESOURCE_NAMESPACE } from "../../../worldResourceTypes.sol";
```
holic, line 7:
```suggestion
```

packages/world/src/modules/core/implementations/AccessManagementSystem.sol
```diff
@@ -3,22 +3,31 @@ pragma solidity >=0.8.21;
...
+import { IWorldErrors } from "../../../IWorldErrors.sol";
```
holic, line 8:
```suggestion
```

packages/world/src/modules/core/implementations/AccessManagementSystem.sol
```diff
@@ -3,22 +3,31 @@ pragma solidity >=0.8.21;
...
 import { ResourceAccess } from "../../../codegen/tables/ResourceAccess.sol";
 import { NamespaceOwner } from "../../../codegen/tables/NamespaceOwner.sol";
+import { requireNamespace } from "../../../requireNamespace.sol";
 
 /**
  * @title Access Management System
  * @dev This contract manages the granting and revoking of access from/to resources.
  */
-contract AccessManagementSystem is System {
+contract AccessManagementSystem is System, IWorldErrors {
```
holic, line 17:
```suggestion
contract AccessManagementSystem is System {
```

packages/world/src/modules/core/implementations/BalanceTransferSystem.sol
holic, line 13:
```suggestion
```

holic (dismissed):
just a bit of clean up but otherwise lgtm!

before we merge, let's rebase on https://github.com/latticexyz/mud/pull/2140 to pick up the `using` changes for resource IDs?

packages/world/src/requireNamespace.sol
```diff
@@ -0,0 +1,20 @@
...
+
+using WorldResourceIdInstance for ResourceId;
+
+/**
+ * @notice Checks if a given `resourceId` is a namespace.
+ * @dev Reverts with IWorldErrors.World_InvalidResourceType if the ID does not have the correct components.
+ * @param resourceId The resource ID to verify.
+ */
+function requireNamespace(ResourceId resourceId) pure {
+  if (resourceId.getType() != RESOURCE_NAMESPACE || resourceId.getName() != ROOT_NAME) {
```
holic, line 17: just realized another way we could do this is

```
if (resourceId.unwrap() == resourceId.getNamespaceId().unwrap()) {
```

yonadaaa: This suggestion is actually more gas efficient. I like it, this functionality is encapsulated anyway

holic (approved):
I was thinking `registerTable` and `registerStoreHook` could use similar resource ID type checks, but turns out these are already done inside `StoreCore` so all good 👍

PR #2146, opened 2024-01-16
refactor(world,world-modules): use eip165 interfaceId property instead of hardcoded constants [n-08]

yonadaaa:
Instead of maintaining constants like:

```solidity
bytes4 constant MODULE_INTERFACE_ID = IModule.getName.selector ^
  IModule.installRoot.selector ^
  IModule.install.selector;
```

We can use the built-in [`interfaceId`](https://docs.soliditylang.org/en/latest/units-and-global-variables.html#type-information):

```solidity
type(IModule).interfaceId
 ```


holic (approved):
cleannnn ✨ 

we should prob add a changeset for this

PR #2149, opened 2024-01-16
refactor(store): simplify schema validation loop

yonadaaa:
Instead of branching inside this loop we can split it into two


holic (approved):
If it's not too much rebasing trouble, gonna suggest we prioritize landing https://github.com/latticexyz/mud/pull/2046 first.

But I could also see about applying these changes to that PR and we can leave a note in the N-08 issue about addressing this separately.

Thoughts?

yonadaaa:
I agree we should land #2046 first and rebase the code-suggestions (https://github.com/latticexyz/mud/pull/2140) branch on. Then merge this PR.

packages/store/src/Schema.sol
```diff
@@ -169,39 +169,32 @@ library SchemaInstance {
...
       }
       unchecked {
         i++;
       }
     }
+
     // Static length sums must match
     if (_staticDataLength != schema.staticDataLength()) {
       revert SchemaLib.SchemaLib_InvalidLength(schema.staticDataLength());
     }
```
holic, line 197: should we move this check up so we can revert before the second loop starts?

yonadaaa: Yes, that's clearer IMO because it's the check is about static data. And wastes less computation if it reverts

holic (approved):

PR #2150, opened 2024-01-16
refactor(store): store core gets field layout from Tables table [n-08]

yonadaaa:


packages/store/src/StoreCore.sol
yonadaaa, line 1: I don't have much context on this and I am not 100% confident this code is equivalent

holic: given tests pass, I am inclined to trust it

I am a little concerned about the gas increase though

holic:
going to pass this to @alvrs for review as its unclear why we're seeing a bump in various gas tests (esp deletes)

packages/store/src/StoreCore.sol
yonadaaa: Some gas increase is expected in this refactor (replacing the low-level `Storage.loadField` call with the cleaner `Tables._getFieldLayout`), but it is large, like there is a few extra storage reads.

holic: I wonder if that means there's room for us to improve the table libs (use lower level calls instead of threading through various helpers)

but unclear if that'll explode the size of bytecode

alvrs: Given the significant gas increase on hot paths (~600 for setting/deleting records) I'm tending towards not doing this reactor. I think it's fine to have lower level code in the low-level StoreCore library in favor of better gas performance for consumers.

holic: agreed, esp if we have test coverage to make sure these methods stay aligned in terms of behavior

holic:
Closing since this creates more gas in the hot code path than is worth it. Let's note this in the audit issue.

PR #2151, opened 2024-01-16
refactor(store): bytes slice functions have named output [n-08]

yonadaaa:
This function:
```solidity
function sliceX(bytes memory data, uint256 start) internal pure returns (bytesX) {
  assembly {
    output := mload(add(add(data, ...), start))
  }
  return output;
}
```

can be simplified to:

```solidity
function sliceX(bytes memory data, uint256 start) internal pure returns (bytesX output) {
  assembly {
    output := mload(add(add(data, ...), start))
  }
}
```

### DIFF
--- a/packages/store/gas-report.json
+++ b/packages/store/gas-report.json
@@ -9,7 +9,7 @@
     "file": "test/Bytes.t.sol",
     "test": "testSlice32",
     "name": "slice bytes32 with offset 10",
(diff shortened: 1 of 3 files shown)


packages/store/gas-report.json
yonadaaa, line 1: interesting 👍

holic (approved):

PR #2154, opened 2024-01-17
Revert "test: increase forge test verbosity in ci"

holic:
Reverts latticexyz/mud#2119

Turns out this is too noisy and makes it hard to even see test errors due to CI log truncation


yonadaaa (approved):
LGTM, I couldn't even link to lines in the log files before

Issue #2155, opened 2024-01-17
codegen system function signatures so we can use them to register functions


frolic:
it's easy to miss that when you add a function to a system (like in the core module) you also need to make sure that function is registered on the world

we could automate some of this with codegen

holic:
worked around this for now by centralizing where the function signatures are defined so we could test them from the same source array: https://github.com/latticexyz/mud/pull/2280

we'll do the codegen step later (see #2332)

PR #2157, opened 2024-01-18
refactor(world): encapsulate functionality [N-01]

yonadaaa:
A few suggestions were fixed in previous PR's. 

https://github.com/latticexyz/mud/pull/2007 fixed `registerTable`:

> The registerTable function might [register a namespace](https://github.com/latticexyz/mud/blob/12a9eb1afb18a7f7cb43d774cabb2d5af7c74e6b/packages/world/src/modules/core/implementations/StoreRegistrationSystem.sol#L67) as well.

https://github.com/latticexyz/mud/pull/2007 fixed `registerSystem`:

> The registerSystem function might [register a namespace](https://github.com/latticexyz/mud/blob/12a9eb1afb18a7f7cb43d774cabb2d5af7c74e6b/packages/world/src/modules/core/implementations/WorldRegistrationSystem.sol#L133)

https://github.com/latticexyz/mud/pull/2096 added `unregisterDelegation`: 

> Individual delegations must be removed by overwriting them using the [regular registration function](https://github.com/latticexyz/mud/blob/12a9eb1afb18a7f7cb43d774cabb2d5af7c74e6b/packages/world/src/modules/core/implementations/WorldRegistrationSystem.sol#L241), which still leaves a stray record in the UserDelegationControl table.


packages/world/test/World.t.sol
yonadaaa, line 1: Need to add tests for `unregisterSystem` and `unregisterNamespaceDelegation`

packages/world/src/modules/core/implementations/AccessManagementSystem.sol
```diff
@@ -59,4 +59,20 @@ contract AccessManagementSystem is System {
     // Grant access to new owner
     ResourceAccess._set(namespaceId, newOwner, true);
   }
+
+  /**
+   * @notice Renounces ownership of the given namespace
+   * @dev Requires the caller to own the namespace. Revoke ResourceAccess for previous owner
+   * @param namespaceId The ID of the namespace to transfer ownership.
+   */
+  function renounceOwnership(ResourceId namespaceId) public virtual {
```
holic, line 68: need to make sure we add a `requireNamespace` check here once https://github.com/latticexyz/mud/pull/2142 lands

packages/world/src/modules/core/implementations/WorldRegistrationSystem.sol
```diff
@@ -159,6 +159,40 @@ contract WorldRegistrationSystem is System, IWorldErrors {
     ResourceAccess._set(namespaceId, address(system), true);
   }
 
+  /**
+   * @notice Unregisters a system
+   * @dev Unregisters the system at the given ID
+   * @param systemId The unique identifier for the system
+   */
+  function unregisterSystem(ResourceId systemId) public virtual {
```
holic, line 167: I think we intentionally don't allow for unregistering systems? cc @alvrs

holic: from @alvrs via discord
> yeah it's this weird situation where other contracts that interact with yours might break if you unregister function selectors or unregister systems, so I think we should only encourage upgrading systems in a non-breaking way
> if someone really wants to remove a system or function selector they can do it by editing the tables, but i don't think it should be something we encourage

packages/world/test/World.t.sol
```diff
@@ -250,7 +251,9 @@ contract WorldTest is Test, GasReporter {
       coreRegistrationSystem.registerFunctionSelector.selector,
       coreRegistrationSystem.registerRootFunctionSelector.selector,
       coreRegistrationSystem.registerDelegation.selector,
-      coreRegistrationSystem.registerNamespaceDelegation.selector
+      coreRegistrationSystem.unregisterDelegation.selector,
```
holic, line 254: oof, we didn't have this before, eh?

packages/world/src/modules/core/implementations/AccessManagementSystem.sol
```diff
@@ -71,4 +71,23 @@ contract AccessManagementSystem is System, LimitedCallContext {
...
+
+    // Require the caller to own the namespace
+    AccessControl.requireOwner(namespaceId, _msgSender());
+
+    // Delete namespace owner
+    NamespaceOwner._deleteRecord(namespaceId);
+
+    // Revoke access from old owner
+    ResourceAccess._deleteRecord(namespaceId, _msgSender());
+  }
```
alvrs, line 92: should we just call `transferOwnership` with `address(0)` here? Similar to the Ownable lib: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/e5c63635e3508a8d9d0afed091578cc4bb59a9c7/contracts/access/Ownable.sol#L77

packages/world/test/World.t.sol
holic: another reason for https://github.com/latticexyz/mud/issues/2155

packages/world/src/modules/core/implementations/AccessManagementSystem.sol
alvrs: Could transfer to address 0 and then delete the resource access record. No strong opinion though.

holic: that would basically be the same logic but a few extra writes, wouldn't it?

packages/world/test/World.t.sol
yonadaaa: Yup, this was forgotten. Although this is a test file so whatever automation we add needs to apply here too.

packages/world/src/modules/core/implementations/AccessManagementSystem.sol
```diff
@@ -71,4 +71,23 @@ contract AccessManagementSystem is System, LimitedCallContext {
...
+   * @notice Renounces ownership of the given namespace
+   * @dev Requires the caller to own the namespace. Revoke ResourceAccess for previous owner
+   * @param namespaceId The ID of the namespace to transfer ownership.
+   */
+  function renounceOwnership(ResourceId namespaceId) public virtual onlyDelegatecall {
+    // Require the namespace ID to be a valid namespace
+    requireNamespace(namespaceId);
+
+    // Require the caller to own the namespace
+    AccessControl.requireOwner(namespaceId, _msgSender());
```
holic, line 85:
```suggestion
    // Require the namespace ID to be a valid namespace
    requireNamespace(namespaceId);

    // Require the namespace to exist
    AccessControl.requireExistence(namespaceId);

    // Require the caller to own the namespace
    AccessControl.requireOwner(namespaceId, _msgSender());
```

packages/world/src/modules/core/implementations/AccessManagementSystem.sol
holic: I think this is fine how it is, but I did make a code suggestion below for completeness with other audit fixes (adding namespace existence check for better reverts)

packages/world/src/modules/core/implementations/WorldRegistrationSystem.sol
```diff
@@ -329,4 +334,20 @@ contract WorldRegistrationSystem is System, IWorldErrors, LimitedCallContext {
...
+   * @notice Unregisters a delegation for a namespace
+   * @dev Deletes the delegation control for a specific namespace
+   * @param namespaceId The ID of the namespace
+   */
+  function unregisterNamespaceDelegation(ResourceId namespaceId) public onlyDelegatecall {
+    // Require namespace ID to be a valid namespace
+    requireNamespace(namespaceId);
+
+    // Require the caller to own the namespace
+    AccessControl.requireOwner(namespaceId, _msgSender());
```
holic, line 348:
```suggestion
    // Require namespace ID to be a valid namespace
    requireNamespace(namespaceId);

    // Require the namespace to exist
    AccessControl.requireExistence(namespaceId);

    // Require the caller to own the namespace
    AccessControl.requireOwner(namespaceId, _msgSender());
```

holic (approved):

Issue #1980, opened 2023-11-30
remove system name from function signatures/selectors


holic:
Instead of `{namespace}_{systemName}_{functionName}` for world function selectors, just use `{namespace}_{functionName}`

This has a bunch of benefits:
- more closely matches root systems where the world function doesn't indicate which system the function is in
- more closely matches how namespaced tables (`{namespace}_{tableName}`) are accessed in downstream tooling
- more ergonomic
- removes the need for us to add validation for system names (to ensure they don't have underscores)

PR #2160, opened 2024-01-18
feat(world): remove system name from function signatures/selectors [M-05]

holic:
closes https://github.com/latticexyz/mud/issues/1980


.changeset/bright-kangaroos-battle.md
```diff
@@ -0,0 +1,8 @@
+---
+"@latticexyz/world-modules": patch
+"@latticexyz/world": major
+---
+
+Functions within namespaced systems are now registered on the world as `{namespace}_{functionName}` rather than `{namespace}_{systemName}_{functionName}`. This is more ergonomic and is more consistent with namespaced resources in other parts of the codebase (e.g. accessing tables in schemaful indexer).
+
+If you have a project using the `namespace` key in your `mud.config.ts` or are manually registering systems and function selectors on a namespace, you will likely need to codegen your system interfaces (`pnpm build`) and update any calls to these systems through the world's namespaced function signatures.
```
holic, line 8: Is there a way to bubble this up as a breaking change aside from `major` tag?

yonadaaa:
No suggestions yet but I wanna experiment with this a bit, think about what happens when you register multiple systems with the same name etc.

Maybe document everything in tests if not already covered

holic:
> think about what happens when you register multiple systems with the same name

@alvrs and I chatted a bit about this IRL and we think it'll just end up with a compile error, because codegen will create interfaces for each system and those will get combined into a `IWorld` interface, but that won't compile if there are function signature/selector overlaps

alvrs:
@qbzzt there might be a couple places in the docs we need to update with this!

.changeset/bright-kangaroos-battle.md
alvrs: major should be only breaking changes, no other way make it more prominent unfortunately

alvrs (dismissed):

yonadaaa:
>  it'll just end up with a compile error, because codegen will create interfaces for each system and those will get combined into a IWorld interface, but that won't compile if there are function signature/selector overlaps

The compiler error is unfortunate, I wonder if we can surface something more helpful in codegen? Trying it locally I get:

image

holic:
> The compiler error is unfortunate, I wonder if we can surface something more helpful in codegen? Trying it locally I get:
> 
> image

Hmm, those don't appear to be namespaced? But the compiler does communicate well enough that there's an overlap.

We could def add more validation+errors to codegen.

yonadaaa:
This makes me wonder if we can/should hide the concept of Systems, instead focusing on functions and having "one contract interface per namespace". 

Instead of `registerSystem`, it could be `registerFunction`. Each function is registered separately along with the address of it's bytecode.

Need to think on this more but maybe it would clear things up here. Definitely out of scope for this PR!

yonadaaa:
> Hmm, those don't appear to be namespaced? 

Oh, oops, they weren't. And you're right that the compiler error is pretty decent.

yonadaaa (dismissed):

alvrs (approved):

PR #2161, opened 2024-01-18
refactor(world-modules): simplify getUniqueEntity call

holic:
Spotted this while working on https://github.com/latticexyz/mud/pull/2160

Using `world.call` means we don't need to maintain this manual system interface and it's cheaper on gas.


yonadaaa (approved):
Neat. Using that previous function name was super awkward. 

Makes you wonder if there are other places we can do this

PR #2163, opened 2024-01-19
refactor(store-indexer): clean up middleware

holic:
Moved common routes/behavior to middleware and cleaned up Sentry middleware

(Pulled out of #2013)


yonadaaa (approved):
Works locally!

PR #2164, opened 2024-01-19
fix(store): reorder core table registration

### LINKED ISSUE #2061: `store:ResourcesIds` getting records before it's schema is registered in `store:Tables`
# Issue
Records are being set in the `ResourceIds` table before emitting the event to register that table in `Tables`. 
Without that event indexers are not able to read the `ResourcesIds` schema from the chain, forcing the dev to manually add the schema in the code.

When deploying a world, the first five events are always the same:

1. SetRecord in {tb store Tables} with key: 0x746273746f72650000000000000000005461626c657300000000000000000000 `tb store Tables`. 
- _This one creates the tbstoreTables table_

2. Splice static record in {tb store ResourceIds} with key: 0x746273746f72650000000000000000005461626c657300000000000000000000 `tb store Tables`
- _At this point the ResourceIds table schema event was not sent, so it can not be processed by just reading the events_

3. SetRecord in {tb store Tables} with key: 0x746273746f726500000000000000000053746f7265486f6f6b73000000000000 `tb store StoreHooks`
- _This one creates the tbstoreStoreHooks table_

4. Splice static record in {tb store ResourceIds} with key: 0x746273746f726500000000000000000053746f7265486f6f6b73000000000000 `tb store StoreHooks`
- _At this point the ResourceIds table schema event was not sent, so it can not be processed by just reading the events_

5. SetRecord in {tb store Tables} with key: 0x746273746f72650000000000000000005265736f757263654964730000000000 `tb store ResourceIds`

- After this event everything works fine because we got the `ResourceIds` schema

# Temp solution
For my indexer I'm just creating the `ResourcesIds` table as soon I get a `SpliceStaticRecord` on that table with its schema hardcoded in my code.

I don't know if it's posible to change the deployment script to register the `ResourcesIds` table as the second transaction, but that would allow the indexer to read all the information from the chain instead of needing to manually add the schemas to it.

# Env
Node v18.12.0
Mud running on branch main, current top commit: `392c4b88d033d2d175541b974189a3f4da49e335`

# Steps to reproduce it
Clone this repo, checkout to the commit listed on the env section
`pnpm install`
cd templates/react
`pnpm install && pnpm run dev`

With the frontend running, let the process run a couple of seconds and kill the contract one, this will stop the frontend from cleaning the transactions related to the world deployment.

Screenshot 2023-12-12 at 13 16 29

holic:
fixes #2061


yonadaaa:
I think I prefer low-level approach described here: https://github.com/latticexyz/mud/issues/2061#issuecomment-1900154132

It encapsulates this bootstrapping issue in `StoreCore` instead of having to handle it in every client/indexer implementation. I really like if they can "just work". It does make `registerCoreTables` a bit harder to maintain, but ideally `Store` should be ossified soon anyway.

yonadaaa (approved):
All inlined, it's gnarlier than I thought, but these low-level calls are still the best approach IMO

holic:
@hanchon Do you wanna test if this fix works for you?

alvrs (approved):

PR #2165, opened 2024-01-19
fix(store-indexer): remove duplicate middleware

holic:
accidentally duplicated this in #2163


holic:
yolo merging because that is less embarrassing than this bug being on main

PR #2168, opened 2024-01-19
fix(world): identify modules by addresses instead of names [L-08]

yonadaaa:
Modules are now identified by address instead of a single, global name. 
- Modules have a `__self` variable that stores their address, allowing them to check if they have been installed.
- The ERC20 and ERC721 modules store the address of their puppet module in an immutable variable. This is passed into the registration helpers so **will require a change for consumers** (see the Sky Strife [PostDeploy script](https://github.com/latticexyz/skystrife-public/blob/main/packages/contracts/script/PostDeploy.s.sol#L111-L119))


yonadaaa:
TODO:

- Add a `isInitialized` table the `World` to prevent initialising the world multiple times. 
- Declare the address of a puppet module as an immutable variable in the constructor of the ERC20/ERC721 module

packages/world-modules/src/modules/erc20-puppet/registerERC20.sol
```diff
@@ -18,13 +18,14 @@ import { ERC20Registry } from "./tables/ERC20Registry.sol";
  */
 function registerERC20(
   IBaseWorld world,
+  address puppetModule,
```
alvrs, line 21: i feel like this additional params is making it pretty impractical to use since it's currently not easy to find the address of the puppet module. Why don't we add an overload for this function that doesn't require the puppetModule address and instead calls this overload with the CREATE2 default address of the puppet module? That way it's not a breaking change for consumers (on CREATE2 compatible chains) but still allows customisation if necessary.

packages/world/mud.config.ts
```diff
@@ -103,6 +103,12 @@ export default mudConfig({
       },
       offchainOnly: true,
     },
+    WorldInitialized: {
+      keySchema: {},
+      valueSchema: {
+        isInitialized: "bool",
+      },
+    },
```
alvrs, line 111: this one got pulled out to the other PR right?

alvrs: i guess we can merge the other one first and then rebase this one

yonadaaa: Yes exactly, this PR should really not have those changes but it only occurred to me later to break it out

packages/world-modules/src/modules/erc20-puppet/registerERC20.sol
yonadaaa: The puppet module is part of the deployment pipeline right now, so doesn't use `Create2`. For example Sky Strife manually deploys the puppet: https://github.com/latticexyz/skystrife-public/blob/main/packages/contracts/script/PostDeploy.s.sol#L111-L119

alvrs (dismissed):

PR #2170, opened 2024-01-22
fix(world): prevent initialising the world multiple times [L-05]

yonadaaa:
Prevents the world from being initialised multiple times. Pulled out from #2168


packages/world/src/World.sol
```diff
@@ -71,10 +72,12 @@ contract World is StoreData, IWorldKernel {
     }
 
     // The World can only be initialized once
-    if (InstalledModules._get(CORE_MODULE_NAME, keccak256("")) != address(0)) {
+    if (WorldInitialized.get()) {
       revert World_AlreadyInitialized();
     }
 
+    WorldInitialized.set(true);
```
alvrs, line 79: What if we store the core module's address here? Almost the same cost (since it's one storage write) but might be interesting to know which core module was used to init the world

yonadaaa: Sure, that seems like a reasonable thing to keep around.

packages/world/mud.config.ts
```diff
@@ -103,6 +103,12 @@ export default mudConfig({
       },
       offchainOnly: true,
     },
+    CoreModuleAddress: {
+      keySchema: {},
+      valueSchema: {
+        coreModuleAddress: "address",
+      },
```
alvrs, line 110:
```suggestion
      valueSchema: "address",
```

.changeset/neat-tools-check.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/world": patch
+---
+
+Added a table that tracks whether the world has already been initialized.
```
alvrs, line 5:
```suggestion
Added a table to track the `CoreModule` address the world was initialised with.
```

alvrs (approved):
lgtm! 🚢

Issue #1392, opened 2023-09-04
Add module for registering arbitrary events


alvrs:
See #1123 for more context.

* Currently it's only possible to make the World contract emit arbitrary events by modifying the World contract directly or via a root system. Both is only possible with root access to the World, and there is no default tooling to support this.
* We initially designed ephemeral table's as a replacement for what events are used outside MUD, but the compromise between "events" and "offchain tables" resulted in a design that's suboptimal for either use case.
* Events other than the default `Store(Set|Splice|Delete)Record` events and ephemeral table events are useful to implement existing interfaces (e.g. ERC20, ERC721, etc) in a World contract.
* The proposal is to implement an optional module and corresponding client tooling to register arbitrary events in an existing World contract.
  * An event has to be registered (with its signature and an `eventId` consisting of its namespace and name) before it can be emitted. 
  * Each signature can only be registered once per World and only by the owner of the namespace.
  * After an event is registered it can be emitted by anyone with write access to the namespace (or the specific `eventId`).
  * Client tooling can add listeners for the registration of new events, and create decoders based on the event's signature. Then it can provide a stream of decoded events.
* Not everyone might want this feature in their World. By implementing this functionality as a (root) module, we give developers the choice of whether they want this functionality in their World or not.

Since this is an optional additional module, it doesn't necessarily have to be part of the first stable release.

holic:
Are you able to dynamically register/emit event types in pure Solidity? Or will this need some TS/codegen so we can modify the world contract before deploy to include the custom events?

alvrs:
> Are you able to dynamically register/emit event types in pure Solidity? 

It should be possible via assembly (eg `LOG1`, see https://www.evm.codes/#a1?fork=shanghai), but haven't tried yet

holic:
Do we think our [puppet approach](https://github.com/latticexyz/mud/blob/main/packages/world-modules/src/modules/puppet/Puppet.sol) is good enough for this both as an existing implementation folks can use and also an example of how to do this if folks want more control?

holic:
Do we want to allow any event via https://github.com/latticexyz/mud/pull/2171 ?

alvrs:
Marking as done via [puppet approach](https://github.com/latticexyz/mud/blob/main/packages/world-modules/src/modules/puppet/Puppet.sol)

PR #2171, opened 2024-01-22
feat(common,world): add events,structs,enums to system interfaces [L-03]

dk1a:
fixes https://github.com/latticexyz/issues/issues/136


packages/common/src/codegen/utils/contractToInterface.ts
```diff
@@ -32,77 +52,116 @@ export function contractToInterface(
...
+          symbolImports = symbolImports.concat(
+            variableDeclarationsToImports(ast, parameters.concat(returnParameters ?? []))
+          );
         }
+      } catch (error: unknown) {
+        if (error instanceof MUDError) {
+          error.message = `Function "${name}" in contract "${contractName}": ${error.message}`;
+        }
+        throw error;
       }
```
holic, line 107: mutating the original error here seems like a bad idea, why not throw a new error?

also how do we guarantee that the MUDError thrown here is relevant? (this prob points to our lack of structured errors more than what the code here is doing)

```suggestion
      } catch (error: unknown) {
        if (error instanceof MUDError) {
          throw new MUDError(`Function "${name}" in contract "${contractName}": ${error.message}`);
        }
        throw error;
      }
```

.changeset/fast-chefs-invent.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/world": major
```
holic, line 2: I think this can be `minor` since its not breaking?
```suggestion
"@latticexyz/world": minor
```

.changeset/witty-bottles-rescue.md
```diff
@@ -0,0 +1,8 @@
+---
+"@latticexyz/common": major
```
holic, line 2: same here?
```suggestion
"@latticexyz/common": minor
```

holic (commented):
since we don't have tests for this, can we add some example events, structs, enums to system interfaces so that we can see some codegen output?

.changeset/fast-chefs-invent.md
yonadaaa, line 1: Maybe there could be a sentence here about what `renderSystemInterface` does now, like: _"you can now access events from auto-generated System interfaces"_.

Without any codegen output I was confused why this PR didn't seem to have an effect 🙈

yonadaaa (commented):
Agree with Frolic that tests would be a great sanity check here

.changeset/fast-chefs-invent.md
dk1a: `contractToInterface` signature, and `worldgen` output changed, which is kind of breaking, but I have no strong opinion here

packages/common/src/codegen/utils/contractToInterface.ts
dk1a: Maybe to preserve callstack? don't remember, will go with ur suggestion

holic:
This sort of relates to #1392 and worth thinking about the potential downsides of allowing any event to be emitted by the World via a system. Maybe it's already possible and unavoidable and we're just allowing the interfaces here to reflect what is already possible?

In the case of permissionless worlds, wouldn't it be possible for a malicious system to be registered that emits incorrect data for World events that could represent real value (e.g. ERC20/ERC721 transfer)? I guess this would only make offchain indexers confused but doesn't actually manipulate onchain state?

dk1a:
> since we don't have tests for this, can we add some example events, structs, enums to system interfaces so that we can see some codegen output?

added e2e systems similar to `CustomErrorsSystem` for events, structs, enums (structs and enums are only usable internally, which the system also showcases)

.changeset/fast-chefs-invent.md
dk1a: Added the sentence
But also there's codgen changes in e2e

dk1a:
> Maybe it's already possible and unavoidable and we're just allowing the interfaces here to reflect what is already possible?

Yeah this change only reflects what's already possible
Non-root systems wouldn't be able to spoof world events though, and root systems can ruin things without events anyways

holic:
Another thing I am curious about is whether we should discourage events from systems, since they're kind of like writes or side effects and we sort of have an implicit expectation that systems are stateless/pure. Not sure if it's misleading to allow for events but then it not be clear that they can be emitted from the world address or from a system address, depending on the calling context.

e2e/packages/contracts/src/systems/CustomStructEnumSystem.sol
```diff
@@ -0,0 +1,24 @@
...
+    bytes32 val2;
+  }
+
+  enum TestEnum {
+    VAL1,
+    VAL2
+  }
+
+  TestStruct internal testStruct;
+  TestEnum internal testEnum;
```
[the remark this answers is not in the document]
holic: answered my own question by tweaking code: it'll generate the structs and enums in the interface when they're defined, regardless of usage 👍

e2e/packages/contracts/src/systems/CustomStructEnumSystem.sol
```diff
@@ -14,11 +14,8 @@ contract CustomStructEnumSystem is System {
...
 
-  TestStruct internal testStruct;
-  TestEnum internal testEnum;
-
-  function stub3(uint8 val1, bytes32 val2) public {
-    testStruct = TestStruct({ val1: val1, val2: val2 });
-    testEnum = TestEnum.VAL2;
-  }
+  // TODO: this fails to codegen an interface
+  // function stub3(TestStruct memory testStruct, TestEnum testEnum) public {
```
holic, line 18: Running codegen with this line throws this error:
```
Function "stub3" in contract "CustomStructEnumSystem": Symbol "TestStruct" has no explicit import
```

We require things like structs to be defined outside of systems and imported, to ensure they're the same object/type. Defining them in multiple places means different types. And we have no way in the current codegen pipeline for a system to also inherit from the generated interface (and it would be awkward anyway, since we'd overwrite the file you're meant to be editing).

PR #2175, opened 2024-01-23
fix(store): do not render push and pop for static arrays, use static length [M-02]

yonadaa:
Static arrays are now treated differently to dynamic arrays:
- The `push` and `pop` methods should not be used in static arrays so they are not rendered. 
- The `length` method now returns the hardcoded known length.

### DIFF
--- a/.changeset/honest-geckos-refuse.md
+++ b/.changeset/honest-geckos-refuse.md
@@ -1,0 +1,5 @@
+---
+"@latticexyz/store": patch
+---
+
+Updated codegen to not render `push` and `pop` methods for static arrays. The `length` method now returns the hardcoded known length instead of calculating it like with a dynamic array.


packages/store/ts/codegen/field.ts
```diff
@@ -82,27 +82,47 @@ export function renderFieldMethods(options: RenderTableOptions) {
       const dynamicSchemaIndex = schemaIndex - options.staticFields.length;
 
       if (options.withGetters) {
-        result += renderWithFieldSuffix(options.withSuffixlessFieldMethods, field.name, (_methodNameSuffix) =>
-          renderWithStore(
-            storeArgument,
-            ({ _typedStore, _store, _commentSuffix, _methodNamePrefix }) => `
+        if (field.typeWrappingData && field.typeWrappingData.kind === "staticArray") {
```
yonadaaa, line 85: I don't love this check as is, we repeat it for `pop` and `push`, then again for `length`.

It's also weird to [treat static arrays as dynamic](https://github.com/latticexyz/mud/blob/f6133591a86eb169a7b1b2b8d342733a887af610/packages/store/ts/codegen/userType.ts#L169) with the `if (field.isDynamic)` check, but then differentiate them using this.

holic: curious if @dk1a has any insight into this

packages/store/ts/codegen/field.ts
```diff
@@ -82,27 +82,47 @@ export function renderFieldMethods(options: RenderTableOptions) {
...
-                ${_keyTupleDefinition}
-                uint256 _byteLength = ${_store}.getDynamicFieldLength(_tableId, _keyTuple, ${dynamicSchemaIndex});
-                unchecked {
-                  return _byteLength / ${portionData.elementLength};
-                }
+                _typedStore,
+                _typedTableId,
+                _typedKeyArgs,
+              ])}) internal pure returns (uint256) {
+                return ${field.typeWrappingData?.staticLength};
```
holic, line 98:
```suggestion
                return ${field.typeWrappingData.staticLength};
```

packages/store/ts/codegen/field.ts
```diff
@@ -82,27 +82,47 @@ export function renderFieldMethods(options: RenderTableOptions) {
...
+                _typedKeyArgs,
+              ])}) internal view returns (uint256) {
+              ${_keyTupleDefinition}
+              uint256 _byteLength = ${_store}.getDynamicFieldLength(_tableId, _keyTuple, ${dynamicSchemaIndex});
+              unchecked {
+                return _byteLength / ${portionData.elementLength};
+              }
+            }
+        `
+            )
```
holic, line 123:
```suggestion
              ({ _typedStore, _store, _commentSuffix, _methodNamePrefix }) => `
                /**
                 * @notice Get the length of ${field.name}${_commentSuffix}.
                 */
                function ${_methodNamePrefix}length${_methodNameSuffix}(${renderArguments([
                    _typedStore,
                    _typedTableId,
                    _typedKeyArgs,
                  ])}) internal view returns (uint256) {
                  ${_keyTupleDefinition}
                  uint256 _byteLength = ${_store}.getDynamicFieldLength(_tableId, _keyTuple, ${dynamicSchemaIndex});
                  unchecked {
                    return _byteLength / ${portionData.elementLength};
                  }
                }
            `
            )
```
just aligning whitespace

packages/store/ts/codegen/field.ts
yonadaaa: This gives a type error! I have no idea why the `if` statement isn't "in scope"

image

packages/store/ts/codegen/field.ts
```diff
@@ -135,44 +155,46 @@ export function renderFieldMethods(options: RenderTableOptions) {
...
               /**
                * @notice Push ${portionData.title} to ${field.name}${_commentSuffix}.
                */
               function ${_methodNamePrefix}push${_methodNameSuffix}(${renderArguments([
-            _typedStore,
-            _typedTableId,
-            _typedKeyArgs,
-            `${portionData.typeWithLocation} ${portionData.name}`,
-          ])}) internal {
+              _typedStore,
```
holic, line 167: the whitespace here needs better alignment too

packages/store/ts/codegen/field.ts
holic: that's annoying but I think it makes sense, since the we're passing in a function and aren't in control of when that function is called

holic: could be worth caching it in the conditional and using the cached var, just to make sure we're gonna render what we expected

holic (approved):
good enough imo!

only thing I can think of that might help make this easier to understand is separate how we codegen dynamic vs static fields, but not sure if it's worth it here (we'll do a bigger refactor pass later)

holic (approved):
did a quick indent change to match other parts of the codebase, because it was hard to explain over text

need to add a changeset but otherwise lgtm!

packages/store/ts/codegen/field.ts
dk1a: terminology is weird here
`isDynamic`, as well as dynamic and static table fields is a mud-specific thing
A better name for `isDynamic` might be `isNonPrimitive` - it indicates any array or struct
The word `static` in static arrays is used in an unrelated context, taken from solidity

dk1a: https://github.com/latticexyz/mud/issues/1668 is related, I also don't like how hacky this check is, especially since now it's used in more places

PR #2176, opened 2024-01-23
test(store): add more codegen tests [M-01]

holic:


packages/store/ts/storeEvents.test.ts
yonadaaa, line 1: Should we add tests to cover more events than just `HelloStore`?

packages/store/test/constants.t.sol
yonadaaa, line 1: Random thought: is it possible to implement this logic in TS (avoiding FFI) and read the constants from the Solidity files somehow?

yonadaaa (changes requested):
Overall good, just I think we should be testing more events beyond `helloStoreEvent`, but maybe I missed that change

packages/store/ts/storeEvents.test.ts
yonadaaa: Oh, I see now. Were we already testing this, so you just moved the test from `storeEventsAbi.test.ts`?

holic: yep! I added the store events test to `storeEvents.test.ts` then found that it was essentially duplicating `storeEventsAbi.test.ts` so I removed it (which OZ also recommended)

packages/store/test/constants.t.sol
holic: That was my initial thought/path but I didn't want to introduce the Solidity parser into this package. We were using the `vm.ffi` approach elsewhere already so I borrowed that:
https://github.com/latticexyz/mud/blob/main/packages/world/test/CoreModule.t.sol#L58-L71

PR #2178, opened 2024-01-23
chore(store-sync): add benchmarks to store-sync

yonadaaa:
Uses Vitest [benchmarks](https://vitest.dev/api/#bench) to find bottlenecks in the sync stack and compare the different storage adapter implementations. Inspired by how Viem [benchmarks itself against Ethers](https://github.com/wevm/viem/blob/main/src/actions/public/call.bench.ts).

Currently it compares the time taken to sync logs between the RECS, SQLite, Zustand storage adapter.

image


packages/store-sync/src/constants.ts
holic, line 1: should prob put this in a top-level test so we don't accidentally export it

yonadaaa:
If I remove the logs (`console.` and `debug`), Zustand is ~9x faster than RECS:
image

Compared to the regular codebase where the RECS is timed as 3x faster than Zustand

holic:
> It seems logging has a big impact on the benchmarks. If I remove `console.` and `debug`, Zustand is ~9x faster than 

Can you tell if it's specifically `debug` or `console.log` adding to the processing time? or are they both doing it equally?

I've been looking to swap `debug` for something else as it's a bit heavy and we only use it for a handful of its features, of which it has some trade-offs that I don't love. Curious if this is a good excuse to swap it for something else.

yonadaaa:
> Can you tell if it's specifically debug or console.log adding to the processing time? or are they both doing it equally?

It seems like a specific `console.log` that prints unknown tables: 
https://github.com/latticexyz/mud/blob/main/packages/store-sync/src/zustand/createStorageAdapter.ts#L33

I'm not even sure why that would be running in these benchmarks 🤔

holic:
> It seems like a specific `console.log` that prints unknown tables: https://github.com/latticexyz/mud/blob/main/packages/store-sync/src/zustand/createStorageAdapter.ts#L33
> 
> I'm not even sure why that would be running in these benchmarks 🤔

Oh interesting! I think that log line may have slipped in accidentally. So feel free to just delete it (maybe in a separate PR?)

yonadaaa:
@holic interesting results! require more in-depth benchmarks though

image

packages/store-sync/package.json
```diff
@@ -48,6 +48,7 @@
     }
   },
   "scripts": {
+    "bench": "vitest bench",
```
holic, line 51: 🏋️ 💪

packages/store-sync/src/getRecords.bench.ts
```diff
@@ -0,0 +1,36 @@
...
+  });
+
+  bench("zustand: `getRecords`", async () => {
+    useStore.getState().getRecords(tables.NumberList);
+  });
+
+  bench("sqlite: `select`", async () => {
+    db.select().from(mudStoreTables).where(eq(mudStoreTables.name, "NumberList")).all();
+  });
+});
```
holic, line 36: If the benchmark results output are still nice, might be better to have e.g. RECS bench tests in `recsStorage.bench.ts` that does has a `getRecord` and `getRecords` benchmark test

packages/store-sync/test/utils.ts
```diff
@@ -0,0 +1,26 @@
...
+export const tables = resolveConfig(mudConfig).tables;
+
+export const { components, storageAdapter: recsStorageAdapter } = recsStorage({ world: createWorld(), tables });
+
+export const useStore = createStore({ tables });
+export const zustandStorageAdapter = createStorageAdapter({ store: useStore });
+
+const SqlJs = await initSqlJs();
+export const db = drizzle(new SqlJs.Database(), {});
+export const sqliteStorageAdapter = await sqliteStorage({ database: db, publicClient });
```
holic, line 26: prob fine here since this is test specific but as a general thing, it's good practice to put side effect creating behavior inside a function that gets called when needed, rather than it executing on import

packages/store-sync/src/getRecord.bench.ts
```diff
@@ -0,0 +1,31 @@
...
+
+describe("Get Record", () => {
+  bench("recs: `getComponentValue`", async () => {
+    getComponentValue(components.NumberList, singletonEntity);
+  });
+
+  bench("zustand: `getRecord`", async () => {
+    useStore.getState().getRecord(tables.NumberList, {});
+  });
+});
```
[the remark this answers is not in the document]
holic: Scratch that - I misunderstood how benchmarking tests are set up and comparisons/output are done

packages/store-sync/src/getRecord.bench.ts
```diff
@@ -0,0 +1,31 @@
...
+} from "../test/utils";
+import { singletonEntity } from "./recs";
+
+for (const block of blocks) {
+  await recsStorageAdapter(block);
+  await zustandStorageAdapter(block);
+  await sqliteStorageAdapter(block);
+}
+
+describe("Get Record", () => {
```
holic, line 23:
```suggestion
describe("Get single record", () => {
```

packages/store-sync/src/getRecords.bench.ts
```diff
@@ -0,0 +1,36 @@
...
+  zustandStorageAdapter,
+} from "../test/utils";
+
+for (const block of blocks) {
+  await recsStorageAdapter(block);
+  await zustandStorageAdapter(block);
+  await sqliteStorageAdapter(block);
+}
+
+describe("Get Records", () => {
```
holic, line 22:
```suggestion
describe("Get all records for a table/component", () => {
```

packages/store-sync/test/utils.ts
yonadaaa: Makes sense. I was curious about this, viem seems to follow that pattern for their test utils: https://github.com/wevm/viem/blob/main/test/src/utils.ts#L274

I'll try implement it here

holic (approved):

packages/store-sync/src/getRecord.bench.ts
```diff
@@ -0,0 +1,31 @@
...
+  useStore,
+  zustandStorageAdapter,
+} from "../test/utils";
+import { singletonEntity } from "./recs";
+
+for (const block of blocks) {
+  await recsStorageAdapter(block);
+  await zustandStorageAdapter(block);
+  await sqliteStorageAdapter(block);
+}
```
holic, line 21: Not blocking since the impact is minimal but probably faster to do these in parallel. Could even shorten this to
```ts
await Promise.all(blocks.flatMap((block) => [
  recsStorageAdapter(block),
  zustandStorageAdapter(block),
  sqliteStorageAdapter(block)
]));
```

holic: wait wait, I was wrong on this - storage adapters are meant to run each block serially, waiting for the previous block's storage adapter call to complete before running the next

so it'd look like
```ts
for (const block of blocks) {
  await Promise.all([recsStorageAdapter(block), zustandStorageAdapter(block), sqliteStorageAdapter(block)]);
}
```

yonadaaa: Fixed!

PR #2169, opened 2024-01-19
feat(world): prevent invalid namespace strings [M-05]

alvrs:

PR #2182, opened 2024-01-24
fix(world): prevent namespace from ending with underscore [M-05]

alvrs:
follows https://github.com/latticexyz/mud/pull/2169


holic:
I know I suggested otherwise before, but given the gas report, I wonder if we should split `validateNamespace` from `requireNamespace` where
- `requireNamespace` checks if a resource ID is a namespace type and has an empty name
- `validateNamespace` checks if the namespace does not have invalid chars
We only need to `validateNamespace` on registration, the rest of the functions can just use `requireNamespace` and save gas

packages/world/src/requireNamespace.sol
```diff
@@ -0,0 +1,20 @@
...
+
+using WorldResourceIdInstance for ResourceId;
+
+/**
+ * @notice Checks if a given `resourceId` is a namespace.
+ * @dev Reverts with `IWorldErrors.World_InvalidResourceType` if the ID does not have the correct components.
+ * @param resourceId The resource ID to check.
+ */
+function requireNamespace(ResourceId resourceId) pure {
+  // Require the resourceId to have the namespace type
```
yonadaaa, line 16: This check with `resourceId.getNamespaceId()` feels a little "magical" since we optimised it in https://github.com/latticexyz/mud/pull/2142#discussion_r1457697540.

How about:

```suggestion
  // Require the resourceId to have the namespace type and the root name
  // (the resource ID is identical to the resource ID of its namespace)
```

packages/world/src/requireValidNamespace.sol
```diff
@@ -0,0 +1,33 @@
...
+
+import { Bytes } from "@latticexyz/store/src/Bytes.sol";
+
+import { ResourceId, WorldResourceIdInstance, WorldResourceIdLib } from "./WorldResourceId.sol";
+import { IWorldErrors } from "./IWorldErrors.sol";
+
+using WorldResourceIdInstance for ResourceId;
+
+/**
+ * @notice Checks if a given `resourceId` is a valid namespace.
```
yonadaaa, line 12: `requireNamespace` and `requireValidNamespace` have the same `notice` comment, maybe they could be more specific?

holic: whoops, that was a copy pasta issue on my part

alvrs (commented):
✅ approved

PR #2184, opened 2024-01-24
chore(store-sync): delete console log

yonadaaa:
This log was found to be [slowing down](https://github.com/latticexyz/mud/pull/2178#issuecomment-1907918017) the Zustand storage adapter, and is not needed because the preceding `debug` call logs the same information.


holic (approved):
🙏

PR #2187, opened 2024-01-24
fix(cli): fix namespaced function signatures

holic:
@alvrs pointed this out and I thought we were only relying on the onchain registration for this but apparently not! missed this in #1980


yonadaaa (approved):
LGTM, I would suggest a helper function but we don't seem to do this anywhere else in TS

holic:
> I would suggest a helper function but we don't seem to do this anywhere else in TS

if only we could call solidity from TS

PR #2189, opened 2024-01-24
chore(e2e): add number test data

yonadaaa:
Benchmarking the client libraries (https://github.com/latticexyz/mud/pull/2178) requires test data for multiple records, but the current script only sets one value in a singleton table.

This creates a system to set values in a `Number` table and calls it multiple times in `generate-test-data.ts`.


holic (approved):

yonadaaa:
Thinking about this more, the benchmarks will need a different approach to be able to test ingesting ~thousands of different types of records. 

Given that the benchmarks are run sparingly, we could dynamically generate their test data every time they run, instead of saving to file. The data would be deterministic, but this is more flexible and we don't have to track the file in git. We could also move the benchmarks from `store-sync` to `e2e` as it has all the setup.

holic:
> Thinking about this more, the benchmarks will need a different approach to be able to test ingesting ~thousands of different types of records.
> 
> Given that the benchmarks are run sparingly, we could dynamically generate their test data every time they run, instead of saving to file. The data would be deterministic, but this is more flexible and we don't have to track the file in git. We could also move the benchmarks from `store-sync` to `e2e` as it has all the setup.

I think it could make sense to have benchmarks in both places, one for lower level testing (closer to "metal") and one for the end to end (e.g. getting closer to time to first contentful paint)

PR #2190, opened 2024-01-25
chore: upgrade vitest to ^1.0.4

yonadaaa:

PR #2193, opened 2024-01-25
chore: upgrade vitest to 0.34.6

yonadaaa:


packages/store/ts/config/storeConfig.test-d.ts
yonadaaa, line 1: It seems `toEqualTypeOf` is bugged with zod: https://github.com/mmkal/expect-type/issues/33

e2e/packages/sync-test/package.json
```diff
@@ -30,7 +30,7 @@
     "typescript": "5.1.6",
     "viem": "1.14.0",
     "vite": "^4.2.1",
-    "vitest": "^0.31.0",
+    "vitest": "0.31.4",
```
holic, line 33: intentionally not bumping this one?

packages/store/ts/config/storeConfig.test-d.ts
holic: just upgrading store-sync seems fine to me!

could also try bumping zod to see if this was fixed there?

tash-2s:
[The errors](https://github.com/latticexyz/mud/actions/runs/7655220959/job/20860792277?pr=2190#step:6:2028) in [the previous PR #2190](https://github.com/latticexyz/mud/pull/2190) appear to be caused by the following reasons, though not confirmed:

1. Error: `Definitions of the following identifiers conflict with those in another file`
	- vitest v1 [requires vite v5](https://github.com/vitest-dev/vitest/blob/v1.2.1/docs/guide/migration.md?plain=1#L14), but mud uses vite v4.
2. Error: `Cannot find module 'rollup/parseAst' or its corresponding type declarations.`
	- vite v5, installed via vitest v1, [doesn't support `moduleResolution: node10`](https://github.com/vitejs/vite/discussions/12466#discussioncomment-7596995), which mud uses. This is because vite v5's rollup [doesn't fully support `node10`](https://github.com/rollup/rollup/issues/5199#issuecomment-1762594280).
	- mud's `moduleResolution` issue is tracked at https://github.com/latticexyz/mud/issues/2051.
3. Error: `Cannot find name 'WeakKey'`
	- Recent versions of vitest [require typescript v5.2 or newer](https://github.com/vitest-dev/vitest/issues/4921#issuecomment-1888484277), but mud uses v5.1.

Regarding the `toEqualTypeOf` problem, it seems there is no easy fix at the moment. `toMatchTypeOf` is a looser alternative than `toEqualTypeOf`, as [it allows extra properties](https://github.com/mmkal/expect-type/blob/v0.16.0/README.md?plain=1#L84-L103), but it works.

e2e/packages/sync-test/package.json
yonadaaa: woops, missed that in my ctrl-f

holic (approved):
:shipit:

PR #2194, opened 2024-01-25
fix(common): include only errors defined in the contract

dk1a:
pulled out of #2171

### DIFF
--- a/.changeset/nine-plants-carry.md
+++ b/.changeset/nine-plants-carry.md
@@ -1,0 +1,5 @@
+---
+"@latticexyz/common": patch
+---
+
+Prevented errors not included in the contract (but present in the file) from being included in the interface by `contractToInterface`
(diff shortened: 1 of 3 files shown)


holic (dismissed):
Is there some way to show the outcome of this? In a test or in some example codegen interface?

dk1a:
> Is there some way to show the outcome of this? In a test or in some example codegen interface?

Yeah, we never ran into the issue but I can make an example, test would be trickier (btw do we have tests that compare solidity file contents to expected output?)

holic:
> btw do we have tests that compare solidity file contents to expected output?

not really but a snapshot test would be fine for this I think (grep for `toMatchInlineSnapshot`)! I am fine with either that or an example that shows output

dk1a:
Added a simple example (it incorrectly generates the error in main branch)

holic (approved):

PR #2195, opened 2024-01-25
fix(world): check table exists for register store and system hook [L-09]

yonadaaa:
Adding two checks that we missed in https://github.com/latticexyz/mud/pull/2142

### DIFF
--- a/packages/store/src/StoreCore.sol
+++ b/packages/store/src/StoreCore.sol
@@ -259,6 +259,11 @@ library StoreCore {
       revert IStoreErrors.Store_InvalidResourceType(RESOURCE_TABLE, tableId, string(abi.encodePacked(tableId)));
     }
 
+    // Require the table to exist
+    if (!ResourceIds._getExists(tableId)) {
+      revert IStoreErrors.Store_TableNotFound(tableId, string(abi.encodePacked(tableId)));
+    }
+
     StoreHooks.push(tableId, Hook.unwrap(HookLib.encode(address(hookAddress), enabledHooksBitmap)));
   }
 
--- a/packages/world-modules/test/KeysWithValueModule.t.sol
+++ b/packages/world-modules/test/KeysWithValueModule.t.sol
@@ -105,6 +105,16 @@ contract KeysWithValueModuleTest is Test, GasReporter {
   }
 
   function testInstallTwice() public {
+    // Register source table
+    world.registerTable(
+      sourceTableId,
+      sourceTableFieldLayout,
+      sourceTableKeySchema,
+      sourceTableSchema,
+      new string[](1),
+      new string[](1)
+    );
+
     world.installRootModule(keysWithValueModule, abi.encode(sourceTableId));
     vm.expectRevert(IModule.Module_AlreadyInstalled.selector);
     world.installRootModule(keysWithValueModule, abi.encode(sourceTableId));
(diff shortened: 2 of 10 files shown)


packages/world/src/modules/core/implementations/StoreRegistrationSystem.sol
```diff
@@ -81,6 +81,9 @@ contract StoreRegistrationSystem is System, IWorldErrors, LimitedCallContext {
     // Require the table's namespace to exist
     AccessControl.requireExistence(tableId.getNamespaceId());
 
+    // Require the table to exist
+    AccessControl.requireExistence(tableId);
```
holic, line 85: I wonder if we should put this in `StoreCore.registerStoreHook` and revert with `IStoreErrors.Store_TableNotFound`?

packages/world-modules/test/KeysWithValueModule.t.sol
```diff
@@ -105,6 +105,16 @@ contract KeysWithValueModuleTest is Test, GasReporter {
...
   function testInstallTwice() public {
+    // Register source table
+    world.registerTable(
+      sourceTableId,
+      sourceTableFieldLayout,
+      sourceTableKeySchema,
+      sourceTableSchema,
+      new string[](1),
+      new string[](1)
+    );
```
holic, line 116: I assume before this was added, the test was failing because of the new reverts?

holic (commented):
Are there tests we should add for these cases?

packages/world/src/modules/core/implementations/StoreRegistrationSystem.sol
yonadaaa: I kinda like that, but generally `StoreCore` doesn't have explicit "does this exist" checks - I think you can `get` and `set` records without the table actually being registered.

So for consistency we would have to add it every Store function.

packages/world-modules/test/KeysWithValueModule.t.sol
yonadaaa: Yes! The test was setup incorrectly by not registering the table - this catches that

packages/world/src/modules/core/implementations/StoreRegistrationSystem.sol
holic: `registerTable` does lots of validation, including checking if a table already exists! I think we can safely add these sorts of validations to registration/install steps (i.e. not in the hot path).

packages/store/src/StoreCore.sol
```diff
@@ -259,6 +259,11 @@ library StoreCore {
       revert IStoreErrors.Store_InvalidResourceType(RESOURCE_TABLE, tableId, string(abi.encodePacked(tableId)));
     }
 
+    // Require the table to exist
+    if (!ResourceIds._getExists(tableId)) {
```
holic, line 263: are we sure we can use `_getExists` here over `getExists`?

if so, could the `StoreHooks.push` be `StoreHooks._push`?

holic (commented):
had a question around underscore method vs not, since its unclear which is the right one to use here

also need a changeset for `world` package!

packages/store/src/StoreCore.sol
yonadaaa: I think we can use the `_` because that calls `StoreCore`, which is already the contract we are in. [The other](https://github.com/latticexyz/mud/blob/main/packages/store/src/StoreCore.sol#L139) functions use it like this so I would follow this? It's not like `StoreCore` is delegatecalled.

yonadaaa: it feels out of scope of in this audit fix to change `StoreHooks._push` though

holic: let's make another PR for it?

holic (dismissed):
sorry, missed that store should prob also get a changeset!

packages/store/src/StoreCore.sol
yonadaaa: Here ya go: https://github.com/latticexyz/mud/pull/2201

PR #2198, opened 2024-01-26
feat: erc1155-puppet world-module

MerkleBoy:
building a rough-on-the-edge 1st iteration of what it would look like to have an ERC1155 world-module. Respects OZ/Enjin standards, with only small deviations from it so far (mostly massaging the ERC to make it fit into a MUD context)

Please disregard ERC1155/System.sol for now, as it's just there as a filler for now. ERC1155Module.sol is also highly susceptible to change in later commits.

No unit test files are present as I don't want to break the CI/CD pipeline just yet.


packages/world-modules/mud.config.ts
```diff
@@ -190,7 +190,7 @@ export default mudConfig({
       },
       tableIdArgument: true,
     },
-    TokenURI: {
+    ERC721TokenURI: {
```
yonadaaa, line 193: Maybe ERC721 and ERC1155 could share tables like `TokenURI` or `OperatorApproval` that have the same schema? 

Just like `Balances`, in the `modules/tokens/tables` directory: https://github.com/MerkleBoy/mud/blob/merkleboy/erc1155-puppet-module/packages/world-modules/mud.config.ts#L121

yonadaaa, line 193: Maybe ERC721 and ERC1155 could share tables like `TokenURI` or `OperatorApproval` that have the same schema? 

Just like `Balances`, in the `modules/tokens/tables` directory: https://github.com/MerkleBoy/mud/blob/merkleboy/erc1155-puppet-module/packages/world-modules/mud.config.ts#L121

yonadaaa (commented):
Looking good so far! 

For testing ERC721 we looked at solmates coverage, they also have tests for ERC1155: https://github.com/transmissions11/solmate/blob/main/src/test/ERC1155.t.sol

packages/world-modules/mud.config.ts
MerkleBoy: Thanks ! It's a really rough draft but I thought it'd be wise to make a draft now (even if it doesn't quite compile haha)

I wasn't sure about the data structures, but you're completely right that I should merge the tables that are constructed the same way. I'll include those changes in the next commit

packages/world-modules/mud.config.ts
```diff
@@ -131,42 +131,39 @@ export default mudConfig({
...
       },
       valueSchema: {
-        value: "uint256",
+        approved: "bool",
       },
       tableIdArgument: true,
     },
-    TotalSupply: {
-      directory: "modules/erc20-puppet/tables",
+    Metadata: {
```
holic, line 155: thoughts on these tables using a more specific name like `TokenMetadata` and `TokenRegistry`?

also would be nice to pull these non-ERC1155 changes/refactors out into a separate PR that preps for this one so that the diff here is smaller

MerkleBoy: Sounds good, I'll revert the changes made to ERC20 and ERC721 puppets and prep the way for the refactor PR

MerkleBoy: fixed in [eba27be](https://github.com/latticexyz/mud/pull/2198/commits/eba27beb1823a7bec70f362bb330f5a0781ccc87) 👍

MerkleBoy:
Full disclosure, I think I completely botched the events fired up by the ERC1155System ; more specifically, I forgot to re-route them to the puppet contract

So please don't merge this just yet, I need to fix this first

holic:
No worries! I converted this to a draft for now until you're ready :)

MerkleBoy:
All good now, it was a rather small fix :)

PR #2199, opened 2024-01-26
refactor(world,world-modules): rename module args to encodedArgs

yonadaaa:
Fixes https://github.com/latticexyz/mud/issues/1757

This name better reflects that the `bytes` argument is encoded module arguments, not the actual typed arguments.

### DIFF
--- a/packages/world-modules/src/modules/erc20-puppet/ERC20Module.sol
+++ b/packages/world-modules/src/modules/erc20-puppet/ERC20Module.sol
@@ -25,12 +25,12 @@ contract ERC20Module is Module {
 
   address immutable registrationLibrary = address(new ERC20ModuleRegistrationLibrary());
 
-  function install(bytes memory args) public {
+  function install(bytes memory encodedArgs) public {
     // Require the module to not be installed with these args yet
-    requireNotInstalled(__self, args);
+    requireNotInstalled(__self, encodedArgs);
 
     // Extract args
-    (bytes14 namespace, ERC20MetadataData memory metadata) = abi.decode(args, (bytes14, ERC20MetadataData));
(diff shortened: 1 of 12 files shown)


packages/world-modules/src/modules/erc20-puppet/ERC20Module.sol
```diff
@@ -25,12 +25,12 @@ contract ERC20Module is Module {
 
   address immutable registrationLibrary = address(new ERC20ModuleRegistrationLibrary());
 
-  function install(bytes memory args) public {
+  function install(bytes memory encodedArgs) public {
     // Require the module to not be installed with these args yet
-    requireNotInstalled(__self, args);
+    requireNotInstalled(__self, encodedArgs);
 
     // Extract args
```
yonadaaa, line 32: Thoughts on being more specific here?

```suggestion
    // Decode args
```

packages/world/src/IModule.sol
```diff
@@ -20,16 +20,16 @@ interface IModule is IERC165 {
    * @notice Installs the module as a root module.
    * @dev This function is invoked by the World contract during `installRootModule` process.
    * The module expects to be called via the World contract and thus installs itself on the `msg.sender`.
-   * @param args Arguments that may be needed during the installation process.
+   * @param encodedArgs Arguments that may be needed during the installation process.
```
yonadaaa, line 23: Should this reflect that `encodedArgs` are not the typed arguments?

```suggestion
   * @param encodedArgs The ABI encoded arguments that may be needed during the installation process.
```

holic (approved):
need a changeset but otherwise 👍

yonadaaa:
> need a changeset but otherwise 👍

Wasn't sure on this one because it was just renaming, will do!

PR #2200, opened 2024-01-26
docs(guide/extending-world): add how to add a table

qbzzt:
1. Turn guide/extending-world into a multipage
2. Add highlight for resolveConfig's importation in the old guide (the one about doing it permissionlessly)


alvrs:
looks good but need to resolve the merge conflict before merging

PR #2201, opened 2024-01-29
refactor(store): push to StoreHooks with StoreCore method

yonadaaa:
this function is called in `StoreCore` so we can avoid using `StoreSwitch` to write to the table.

### DIFF
--- a/.changeset/rare-dragons-walk.md
+++ b/.changeset/rare-dragons-walk.md
@@ -1,0 +1,5 @@
+---
+"@latticexyz/store": patch
+---
+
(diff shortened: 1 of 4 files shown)


holic (approved):

holic:
unclear if this needs a changeset but if we're following "it changed the bytecode", then probably

yonadaaa:
> unclear if this needs a changeset but if we're following "it changed the bytecode", then probably

it has one!

holic:
> it has one!

oops 🙈

Issue #2089, opened 2024-01-05
`SystemSwitch` does not replicate `World` behavior when calling a non-root `System` from a root one


qbzzt:
When we call a `System` that is *not* in the root namespace, we expect it to be a `call` and run under the `System`'s address.

However, when we use `SystemSwitch` to call from a root `System`, [it *always* uses `delegatecall`](https://github.com/latticexyz/mud/blob/f6f402896d8256da3b868f865a960db68393caf4/packages/world-modules/src/utils/SystemSwitch.sol#L42-L61). This results in confusing behavior, for example that you need to put ERC-20 tokens to be distributed by the non root `System` on the `World` address.

PR #2205, opened 2024-01-29
fix(world-modules): `SystemSwitch` properly calls systems from root

yonadaaa:
Fixes https://github.com/latticexyz/mud/issues/2089

Added a test case that fails without this code change


packages/world-modules/src/utils/SystemSwitch.sol
```diff
@@ -39,19 +39,27 @@ library SystemSwitch {
   function call(ResourceId systemId, bytes memory callData) internal returns (bytes memory returnData) {
     address worldAddress = WorldContextConsumerLib._world();
 
-    // If we're in the World context, call the system directly via delegatecall
+    // If we're in the World context
     if (address(this) == worldAddress) {
       (address systemAddress, ) = Systems.get(systemId);
```
alvrs, line 44:
```suggestion
      address systemAddress = Systems.getSystem(systemId);
```
while we're here

alvrs, line 44: looking at this I wonder if we should just use the `SystemCall.call` library here


```solidity
  function call(ResourceId systemId, bytes memory callData) internal returns (bytes memory returnData) {
    address worldAddress = WorldContextConsumerLib._world();

    // If we're in the World context, call via the internal library
    if (address(this) == worldAddress) {
      return SystemCall.call(systemId, callData);
    }

    // Otherwise, call the system via world.call
    returnData = IWorldKernel(worldAddress).call(systemId, callData);
  }
```

yonadaaa: I like this! It adds a small amount of gas for the `publicAccess` and `value` checks but I like that we can reuse the code.

holic: should we change all the other places to use the same lib?

yonadaaa: Are there other places that do this? Searching for `delegatecallWithContext` or `== ROOT_NAMESPACE` doesn't bring up much for me

yonadaaa: I wonder if this line could cause an error because it updates the balances expecting a root

`if (value > 0) {`

yonadaaa:
We should merge https://github.com/latticexyz/mud/pull/2216 into this

packages/world-modules/test/SystemSwitch.t.sol
```diff
@@ -194,6 +212,20 @@ contract SystemSwitchTest is Test, GasReporter {
...
+
+    // Call reverts because the non-root system storage does not have table schemas
+    vm.expectRevert(
+      abi.encodeWithSelector(
+        IStoreErrors.Store_TableNotFound.selector,
+        ResourceIdsTableId,
+        string(abi.encodePacked(ResourceIdsTableId))
+      )
+    );
+    _executeFromRootSystemA(systemBId, abi.encodeCall(EchoSystem.readTable, ()));
```
yonadaaa, line 226: This isn't catching the revert properly:

image

Removing it prints the error we want:
image

alvrs (approved):

PR #2206, opened 2024-01-29
test(store-sync): benchmark storage adapters

yonadaaa:
Adds benchmarks comparing RECS, Zustand and SQLite when: 
- hydrating storage adapters
- fetching a single record
- fetching all records in a table

Each benchmark is run for 10, 100, 1000 records. The logs for these records are generated in `e2e` and saved in files as it is awkward/impractical to regenerate them in the benchmark environment.

In future we can add more benchmarks (like comparing `runQuery`), and polish the benchmarking suites with Vitest [hooks](https://github.com/vitest-dev/vitest/pull/5076).

### Hydration
image

### Fetching a single record
image

### Fetching all records in a table
image


e2e/packages/test-data/generate-test-data-100.ts
```diff
@@ -0,0 +1,112 @@
...
+  isHex,
+  numberToHex,
+} from "viem";
+import { mudFoundry } from "@latticexyz/common/chains";
+import { getContract } from "@latticexyz/common";
+import { storeEventsAbi } from "@latticexyz/store";
+import { privateKeyToAccount } from "viem/accounts";
+import IWorldAbi from "../contracts/out/IWorld.sol/IWorld.abi.json";
+
+const LIMIT = 100;
```
alvrs, line 21: could this be a parameter in a function instead of a separate script per limit?

test-data/world-logs-10000.json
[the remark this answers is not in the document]
yonadaaa: Nope. We won't run benchmarks often so I think we can expect everyone to generate it themselves

e2e/packages/test-data/generate-test-data-100.ts
yonadaaa: Ideally, yes! But it'll take some tweaking, the current script writes to one file then kills the whole process (seems this is the only way to shut down `anvil/viem`).

yonadaaa: Or closes slowly: https://github.com/wevm/anvil.js/issues/41

packages/store-sync/src/storageAdapter100.bench.ts
```diff
@@ -0,0 +1,26 @@
+import { bench, describe } from "vitest";
+import worldRpcLogs from "../../../test-data/world-logs-100.json";
+import { recsStorageAdapter, sqliteStorageAdapter, zustandStorageAdapter } from "../test/utils";
+import { logsToBlocks } from "../test/logsToBlocks";
+
+const blocks = logsToBlocks(worldRpcLogs.slice(0, 100));
```
yonadaaa, line 6: This is not completely accurate because the first dozen records will be logs for the World internals, which are processed differently.

yonadaaa:
Hooks (`beforeAll` etc.) are being added for Vitest benchmarks which will make our test setup better: https://github.com/vitest-dev/vitest/pull/5076

holic (approved):
nice! `describe.each` feels a lot nicer

packages/store-sync/src/benchmarks/getRecord.bench.ts
holic, line 1: I wonder if it makes sense to move these to top-level as a sibling of `test` and `src` rather than inside `src` (since we can't colocate these with their respective files they're benchmarking

yonadaaa: Done!

Issue #1850, opened 2023-11-01
WorldFactory: add sender to salt


holic:
https://github.com/latticexyz/mud/blob/52182f70d350bb99cdfa6054cd6d181e58a91aa6/packages/world/src/WorldFactory.sol#L36

We should add `msg.sender` here as part of the salt, so the address is derived from the creator of the world. Even better might be a per address salt or a custom salt in addition to the sender, so folks can attempt to deploy worlds across chains with the same address, regardless of how many they've deployed before.

PR #2210, opened 2024-01-30
feat(world): deployment salt by msg.sender

yonadaaa:
Fixes [#1850](https://github.com/latticexyz/mud/issues/1850)


packages/world/src/WorldFactory.sol
```diff
@@ -17,8 +17,8 @@ contract WorldFactory is IWorldFactory {
   /// @notice Address of the core module to be set in the World instances.
   IModule public immutable coreModule;
 
-  /// @notice Counter to keep track of the number of World instances deployed.
-  uint256 public worldCount;
+  /// @notice Counters to keep track of the number of World instances deployed per address.
+  mapping(address => uint256) public accountCount;
```
holic, line 21: we can name mapping params now, which helps to self-document
```suggestion
  mapping(address creator => uint256 worldCount) public accountCount;
```
not blocking, but I kinda feel like `worldCounts` might be clearer than `accountCount`?

.changeset/selfish-pears-marry.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/world": minor
+---
+
+Added `msg.sender` to the `WorldFactory` deployment salt, so the World address is derived from the creator of the world.
```
holic, line 5:
```suggestion
`WorldFactory` now derives a salt based on number of worlds deployed by `msg.sender`, which should help with deterministic world deployments across chains.
```

holic (dismissed):

holic:
do we need to regenerate test data after this change?

packages/world/src/WorldFactory.sol
yonadaaa: I agree, then it's like "world counts per account"

yonadaaa:
> do we need to regenerate test data after this change?

It might not be strictly necessary (if nothing relies on addresses) but I'll do it anyway

holic (approved):

PR #2211, opened 2024-01-30
docs(world): reference

alvrs:
To generate API docs run `pnpm docs:generate:api` at the MUD root.

The relevant script is in `scripts/render-api-docs.ts`. See the config for store as an example. We need to add all relevant `World` input files there, and then post-process the result via `processContent` and helper functions.


qbzzt:
Replaced by #2232

PR #2219, opened 2024-01-31
feat: add user defined salt in WorldFactory.deployWorld()

### LINKED ISSUE #2214: WorldFactory: allow creator to pass in salt
follow up to https://github.com/latticexyz/mud/pull/2210

> could we allow the sender to pass the salt themselves? so final salt is msg.sender + their salt 
reason is: we might want some world address for the same world on multiple chains, and it might be the case that worldCount is different on different chains

> not sure i have time right now, but it's pretty simple: instead of worldCounts[msg.sender]++ being part of the final salt calculation, it's a value passed by the caller as a function arg.
reason is: worldCount is already a form of value controlled by the sender, but it's annoying to "set it" because the only way to change it is to deploy worlds through that contract.
CREATE2 already reverts if a contract currently exists at that address, so I think we should let people pass whatever salt they want; both preventing them from being frontran by including the msg.sender in the salt and letting them pass whatever extra value they want (enabling the same world address on multiple chains)

partylikeits1983:
This PR adds user-defined salt functionality to the `WorldFactory.deployWorld()` function.

Resolves #2214

The modified `WorldFactory.deployWorld()` function allows the world creator to pass in an arbitrary salt value. This makes it easier for developers to have deterministic addresses across chains. 

The edited function still increments worldCounts when creating a new world, but the worldCount nonce is not used to create a unique salt. 

The salt value still uses msg.sender as a parameter to generate a salt. If msg.sender calls deployWorld() a second time with the same bytes _salt value, the transaction will revert.

solidity
    function deployWorld(bytes memory _salt) public returns (address worldAddress) {
        // Deploy a new World and increase the WorldCount
        bytes memory bytecode = type(World).creationCode;
        uint256 salt = uint256(keccak256(abi.encode(msg.sender, _salt)));
        worldCounts[msg.sender]++;
        
        worldAddress = Create2.deploy(bytecode, salt);
        IBaseWorld world = IBaseWorld(worldAddress);

        // Initialize the World and transfer ownership to the caller
        world.initialize(coreModule);
        world.transferOwnership(ROOT_NAMESPACE_ID, msg.sender);

        emit WorldDeployed(worldAddress);
    }

This PR also updates the Factories.t.sol test. 

In the test, the calculated address matches the deployed address via Create2.

Additionally the test expects revert when attempting to deploy a second world with the same _salt value. 

Please let me know if I need to edit the PR.

### DIFF
--- a/packages/world/src/WorldFactory.sol
+++ b/packages/world/src/WorldFactory.sol
@@ -8,7 +8,9 @@ import { IBaseWorld } from "./codegen/interfaces/IBaseWorld.sol";
 import { IModule } from "./IModule.sol";
 import { ROOT_NAMESPACE_ID } from "./constants.sol";
 
-/**
+import { Test, console } from "forge-std/Test.sol";
+
+/*
  * @title WorldFactory
  * @notice A factory contract to deploy new World instances.
  * @dev This contract allows users to deploy a new World, install the CoreModule, and transfer the ownership.
@@ -25,15 +27,18 @@ contract WorldFactory is IWorldFactory {
     coreModule = _coreModule;
   }
 
-  /**
+  /*
    * @notice Deploys a new World instance, installs the CoreModule and transfers ownership to the caller.
    * @dev Uses the Create2 for deterministic deployment.
+   * @param _salt User defined salt for deterministic world addresses across chains
    * @return worldAddress The address of the newly deployed World contract.
    */
-  function deployWorld() public returns (address worldAddress) {
+  function deployWorld(bytes memory _salt) public returns (address worldAddress) {
     // Deploy a new World and increase the WorldCount
     bytes memory bytecode = type(World).creationCode;
-    uint256 salt = uint256(keccak256(abi.encode(msg.sender, worldCounts[msg.sender]++)));
+    uint256 salt = uint256(keccak256(abi.encode(msg.sender, _salt)));
+    worldCounts[msg.sender]++;
+
     worldAddress = Create2.deploy(bytecode, salt);
     IBaseWorld world = IBaseWorld(worldAddress);
 
--- a/packages/world/test/Factories.t.sol
+++ b/packages/world/test/Factories.t.sol
@@ -62,10 +62,13 @@ contract FactoriesTest is Test, GasReporter {
     address worldFactoryAddress = address(new WorldFactory(coreModule));
     IWorldFactory worldFactory = IWorldFactory(worldFactoryAddress);
 
+    // User defined bytes for create2
+    bytes memory _salt0 = abi.encode(0);
+
     // Address we expect for World
     address calculatedAddress = calculateAddress(
       worldFactoryAddress,
-      keccak256(abi.encode(account, 0)),
+      keccak256(abi.encode(account, _salt0)),
       type(World).creationCode
     );
 
@@ -77,7 +80,7 @@ contract FactoriesTest is Test, GasReporter {
     vm.expectEmit(true, false, false, false);
     emit WorldDeployed(calculatedAddress);
     startGasReport("deploy world via WorldFactory");
-    worldFactory.deployWorld();
+    worldFactory.deployWorld(_salt0);
     endGasReport();
 
     // Set the store address manually
@@ -94,10 +97,13 @@ contract FactoriesTest is Test, GasReporter {
 
     // Deploy another world
 
+    // User defined bytes for create2
+    bytes memory _salt1 = abi.encode(1);
+
     // Address we expect for World
     calculatedAddress = calculateAddress(
       worldFactoryAddress,
-      keccak256(abi.encode(account, 1)),
+      keccak256(abi.encode(account, _salt1)),
       type(World).creationCode
     );
 
@@ -108,7 +114,7 @@ contract FactoriesTest is Test, GasReporter {
     // Check for WorldDeployed event from Factory
     vm.expectEmit(true, false, false, false);
     emit WorldDeployed(calculatedAddress);
-    worldFactory.deployWorld();
+    worldFactory.deployWorld(_salt1);
 
     // Confirm accountCount (which is salt) has incremented
     assertEq(uint256(worldFactory.worldCounts(account)), uint256(2));
@@ -121,6 +127,10 @@ contract FactoriesTest is Test, GasReporter {
 
     // Confirm the msg.sender is owner of the root namespace of the new world
     assertEq(NamespaceOwner.get(ROOT_NAMESPACE_ID), account);
+
+    // Expect revert when deploying world with same bytes salt as already deployed world
+    vm.expectRevert();
+    worldFactory.deployWorld(_salt1);
   }
 
   function testWorldFactoryGas() public {
--- a/packages/world/src/IWorldFactory.sol
+++ b/packages/world/src/IWorldFactory.sol
@@ -25,5 +25,5 @@ interface IWorldFactory {
    * @dev The deployment of the World contract will result in the `WorldDeployed` event being emitted.
    * @return worldAddress The address of the newly deployed World contract.
    */
-  function deployWorld() external returns (address worldAddress);
+  function deployWorld(bytes memory _salt) external returns (address worldAddress);
 }


packages/world/src/WorldFactory.sol
```diff
@@ -8,7 +8,9 @@ import { IBaseWorld } from "./codegen/interfaces/IBaseWorld.sol";
 import { IModule } from "./IModule.sol";
 import { ROOT_NAMESPACE_ID } from "./constants.sol";
 
-/**
+import { Test, console } from "forge-std/Test.sol";
```
yonadaaa, line 11: Do we need these imports?

packages/world/test/Factories.t.sol
```diff
@@ -62,10 +62,13 @@ contract FactoriesTest is Test, GasReporter {
     address worldFactoryAddress = address(new WorldFactory(coreModule));
     IWorldFactory worldFactory = IWorldFactory(worldFactoryAddress);
 
+    // User defined bytes for create2
+    bytes memory _salt0 = abi.encode(0);
```
yonadaaa, line 66: Maybe the salt parameters could be [fuzzed](https://book.getfoundry.sh/forge/fuzz-testing) like `account` is? 

They both need to be hardcoded when called in `testWorldFactoryGas` though (just for our gas reporter).

yonadaaa (commented):
Looking good! Just a couple comments.

packages/world/src/WorldFactory.sol
```diff
@@ -25,15 +27,18 @@ contract WorldFactory is IWorldFactory {
...
+   * @param _salt User defined salt for deterministic world addresses across chains
    * @return worldAddress The address of the newly deployed World contract.
    */
-  function deployWorld() public returns (address worldAddress) {
+  function deployWorld(bytes memory _salt) public returns (address worldAddress) {
     // Deploy a new World and increase the WorldCount
     bytes memory bytecode = type(World).creationCode;
-    uint256 salt = uint256(keccak256(abi.encode(msg.sender, worldCounts[msg.sender]++)));
+    uint256 salt = uint256(keccak256(abi.encode(msg.sender, _salt)));
+    worldCounts[msg.sender]++;
```
yonadaaa, line 40: Maybe we still need `worldCounts` now?

holic: I don't think we need world counts at all anymore?

partylikeits1983: Would it be useful to store the deployed world addresses in an array? 

For example:
```solidity
mapping(address => address[]) public worldAddresses; 
```

This way anyone could see the number of deployed addresses via `worldAddresses.length` and would be able to get the respective addresses of the deployed worlds. 

However, if it is not needed, it would save gas to completely remove `worldCounts`

packages/world/test/Factories.t.sol
```diff
@@ -134,6 +137,10 @@ contract FactoriesTest is Test, GasReporter {
   }
 
   function testWorldFactoryGas() public {
-    testWorldFactory(address(this));
+    testWorldFactory(address(this), 0);
+  }
+
+  function testFuzzWorldDeploy(address account, uint _salt) public {
```
partylikeits1983, line 143: Added fuzzing test for salt.

packages/world/test/Factories.t.sol
```diff
@@ -95,12 +95,15 @@ contract FactoriesTest is Test, GasReporter {
...
-    // Deploy another world
+    // Deploy a second world
 
     // User defined bytes for create2
-    bytes memory _salt1 = abi.encode(1);
-
-    // Address we expect for World
+    // unchecked for the fuzzing test
+    bytes memory _salt1;
+    unchecked {
```
partylikeits1983, line 103: Added unchecked here so that the test doesn't revert on overflow/underflow during the fuzzing test.

partylikeits1983 (commented):
Resolved issues brought up in review.

packages/world/src/WorldFactory.sol
partylikeits1983: You're right, that was on accident. Fixed.

packages/world/test/Factories.t.sol
partylikeits1983: I added the fuzzing test.

packages/world/test/Factories.t.sol
yonadaaa: You actually don't need a separate `testFuzzWorldDeploy` test, because `testWorldFactory` takes the `address account` and `uint salt` parameters, Foundry will automatically run it in Fuzz.

yonadaaa: You actually don't need a separate `testFuzzWorldDeploy` test, Foundry will automatically run `testWorldFactory` as a property-based test because it has parameters.

So we only need `testWorldFactory(address account, uint256 salt)` and `testWorldFactoryGas()`

image

packages/world/test/Factories.t.sol
```diff
@@ -54,18 +54,21 @@ contract FactoriesTest is Test, GasReporter {
     assertEq(uint256(worldFactory.worldCounts(address(0))), uint256(0));
   }
 
-  function testWorldFactory(address account) public {
+  function testWorldFactory(address account, uint salt) public {
```
yonadaaa, line 57: For consistency:

```suggestion
  function testWorldFactory(address account, uint256 salt) public {
```

packages/world/src/WorldFactory.sol
yonadaaa: We can remove it entirely, no need to maintain code or on-chain state if there isn't a proven use case

packages/world/test/Factories.t.sol
yonadaaa: Is the overall intention here just to generate two different salts?

A more clear way to do this would be to have two salt parameters in the fuzz test, and discard them if the fuzzer generates two that are equal. You can do this with a Foundry [assume](https://book.getfoundry.sh/cheatcodes/assume) at the start of the test: 

```solidity
function testWorldFactory(address account, uint256 salt1, uint256 salt2) public {
  vm.assume(salt1 != salt2);
  ...
)

**Review by yonadaaa — COMMENTED** [review]
Almost there, just remove `worldCounts` and a couple nits 🙌

**Comment by partylikeits1983** [comment]
Removed worldCounts from WorldFactory and simplified the `testWorldFactory` test.

File: packages/world/src/WorldFactory.sol [hunk 7]
```diff
@@ -17,23 +17,22 @@ contract WorldFactory is IWorldFactory {
...
 
-  /// @notice Counters to keep track of the number of World instances deployed per address.
-  mapping(address creator => uint256 worldCount) public worldCounts;
-
-  /// @param _initModule The address of the init module.
-  constructor(IModule _initModule) {
-    initModule = _initModule;
+  /// @param _coreModule The address of the core module.
+  constructor(IModule _coreModule) {
+    coreModule = _coreModule;
```
holic, line 22:
```suggestion
  /// @param _initModule The address of the init module.
  constructor(IModule _initModule) {
    initModule = _initModule;
```
(sorry this is my bad, just merged a change that renamed this)

packages/world/src/WorldFactory.sol
```diff
@@ -28,12 +25,14 @@ contract WorldFactory is IWorldFactory {
   /**
    * @notice Deploys a new World instance, installs the InitModule and transfers ownership to the caller.
    * @dev Uses the Create2 for deterministic deployment.
+   * @param _salt User defined salt for deterministic world addresses across chains
    * @return worldAddress The address of the newly deployed World contract.
    */
-  function deployWorld() public returns (address worldAddress) {
+  function deployWorld(bytes memory _salt) public returns (address worldAddress) {
```
holic, line 31: small API thing: because Solidity supports calling functions with named parameters now, it might make sense to call this `salt` and internally we can call the combined one `_salt`

yonadaaa:
@partylikeits1983 are you good to fix the last suggestion (and get CI to pass)? Otherwise I'm happy can get this over the finish line, but you'll need to allow commits on branches you've created so I can push to your fork: 
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

packages/world/src/WorldFactory.sol
partylikeits1983: Fixed!

partylikeits1983:
> @partylikeits1983 are you good to fix the last suggestion (and get CI to pass)? Otherwise I'm happy to get this over the finish line, but you'll need to enable commits on your branches so I can push to your fork: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork
Screenshot 2024-02-02 at 9 31 27 PM

Edits by maintainers are turned on for this branch. 

Is there something specific for repositories with GitHub actions that have secrets?

yonadaaa:
> Edits by maintainers are turned on for this branch.

@partylikeits1983 thanks for pointing that out, seems my `gh` CLI was misconfigured but I was able to push by cloning your repo instead.

This PR is almost done - I will take this over and get it over the finish line!

packages/cli/src/runDeploy.ts
yonadaaa, line 1: @alvrs @holic should/do we test CLI commands? 

For example, we have contract tests for this salt but we don't know if the CLI option actually follows that.

partylikeits1983:
> > Edits by maintainers are turned on for this branch.
> 
> @partylikeits1983 thanks for pointing that out, seems my `gh` CLI was misconfigured but I was able to push by cloning your repo instead.
> 
> This PR is almost done - I will take this over and get it over the finish line!

Lmk if there is anything else I should fix.

packages/cli/src/runDeploy.ts
holic: we sort of do in that our e2e tests lean on deploying a world with the mud cli (I think)

packages/cli/src/deploy/deploy.ts
```diff
@@ -58,7 +61,7 @@ export async function deploy<configInput extends ConfigInput>({
 
   const worldDeploy = existingWorldAddress
     ? await getWorldDeploy(client, existingWorldAddress)
-    : await deployWorld(client);
+    : await deployWorld(client, salt ? salt : `0x${randomBytes(32).toString("hex")}`);
```
holic, line 64: we might want to check `isHex` here

packages/cli/src/commands/dev-contracts.ts
```diff
@@ -89,6 +89,7 @@ const commandModule: CommandModule<typeof devOptions, InferredOptionTypes<typeof
             saveDeployment: true,
             worldAddress,
             srcDir,
+            salt: undefined,
```
holic, line 92: I think it might be nice to do our own internally incrementing counter for this process here, so that restarting the dev process always results in the same contract address.

holic: Nevermind, this might be harder than I initially thought it would be. Since we already pass the `worldAddress` in to upgrade worlds, this should be fine.

Just a note that this means that our tests that generate data will be non-deterministic for now (unless we update them to pass in a consistent salt)

PR #2225, opened 2024-02-01
feat(store): rename StoreCore.registerCoreTables to registerInternalTables

holic:
to clarify and clear the "core" term for other uses

### DIFF
--- a/packages/world/src/modules/core/CoreModule.sol
+++ b/packages/world/src/modules/core/CoreModule.sol
@@ -62,8 +62,8 @@ contract CoreModule is Module {
    * @dev Registers core tables, systems, and function selectors in the World.
    */
   function installRoot(bytes memory) public override {
-    _registerCoreTables();
-    _registerCoreSystems();
+    _registerInternalTables();
+    _registerInternalSystems();
     _registerFunctionSelectors();
   }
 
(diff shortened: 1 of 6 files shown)


yonadaaa (approved):
I like this a lot more, "Core Module" made it seem like some overly important piece

packages/world/src/modules/core/CoreModule.sol
```diff
@@ -108,20 +108,20 @@ contract CoreModule is Module {
...
   /**
-   * @notice Register the core systems in the World.
+   * @notice Register the internal systems in the World.
    */
-  function _registerCoreSystems() internal {
-    _registerSystem(accessManagementSystem, ACCESS_MANAGEMENT_SYSTEM_ID);
-    _registerSystem(balanceTransferSystem, BALANCE_TRANSFER_SYSTEM_ID);
-    _registerSystem(batchCallSystem, BATCH_CALL_SYSTEM_ID);
-    _registerSystem(coreRegistrationSystem, CORE_REGISTRATION_SYSTEM_ID);
+  function _registerInternalSystems() internal {
```
alvrs, line 113: feels a little bit off to call these "internal systems" - with that name I'd think of internal functions, ie something that's not externally accessible, but these systems become part of the "core" API 🙈

holic: since we're _in_ the `CoreModule` (soon `InitModule`), maybe we can just name these `_registerTables`, `_registerSystems`, and `_registerSystem`?

holic: went ahead and did this in https://github.com/latticexyz/mud/pull/2225/commits/e58b76fb6940d59327a763b09b847481b40f3b32

alvrs (approved):

PR #2228, opened 2024-02-01
test(store-sync): add query benchmark, refactor test data scripts

yonadaaa:
Compares RECS `runQuery` with Zustand filtering.

I've also added more complex test data to properly test query joins.

image


e2e/packages/test-data/generate-test-data-query.ts
yonadaaa, line 1: `test-data` is getting messy with these different scripts and outputs. 

Ideally we would generate in the benchmarks at run-time

yonadaaa: Refactored the test setup to use one helper

e2e/packages/test-data/generateLogs.ts
```diff
@@ -1,20 +1,38 @@
...
-export async function generateLogs(numRecords: number, rpc: string) {
+type WorldAbi = typeof IWorldAbi;
+
+type WorldContract = GetContractReturnType<
+  WorldAbi,
+  PublicClient<Transport, Chain>,
+  WalletClient<Transport, Chain, Account>
+>;
+
+export async function generateLogs(
```
yonadaaa, line 32: Is this helper overkill? The idea is that you can pass a list of systems to execute and receive the logs back. 

Ideally we would have a typed list of calls instead of the "hook" that executes the systems. Something like:

```Typescript
const calls: [keyof WorldContract["write"], any>][] = [
  ["set", [[69]]],
  ["push", [420]],
];
```

e2e/packages/test-data/generateLogs.ts
yonadaaa, line 1: Is this helper overkill? We have multiple sets of test data, so with this you can pass a list of systems to execute and receive the logs back. 

Ideally we would have a typed list of calls instead of the "hook" that executes the systems. Something like:

```Typescript
const calls: [keyof WorldContract["write"], any>][] = [
  ["set", [[69]]],
  ["push", [420]],
];
```

holic: That seems fine, but gets hard if the input of one call depends on the result of another. How would you model that?

I think at the time, it was easier types-wise to use the callback approach as well.

holic: Oh, I misreading this. _You_ wrote the callback approach. I like it! It's basically "give me the logs that are emitted as a result of this block of code".

e2e/packages/test-data/generateLogs.ts
```diff
@@ -55,12 +73,10 @@ export async function generateLogs(numRecords: number, rpc: string) {
...
-  console.log("calling setNumber");
-  for (let i = 0; i < numRecords - 1; i++) {
-    await worldContract.write.setNumber([i, i]);
-  }
-
-  const lastTx = await worldContract.write.setNumber([numRecords, numRecords]);
+  console.log("calling set");
+  await worldContract.write.set([[420]]);
+  console.log("calling push");
+  const lastTx = await transactionHook(worldContract);
```
holic, line 79: any reason we're still doing the `.set` above? instead of just letting the logic live within `transactionHook`?

e2e/packages/test-data/generate-test-data-query.ts
holic, line 1: is there a better set of names we can use here?

unclear when I should use `generate-test-data` vs `generate-test-data-records` vs `generate-test-data-query`

packages/store-sync/benchmarks/query.bench.ts
```diff
@@ -0,0 +1,36 @@
+import { bench, describe } from "vitest";
+import { Has, runQuery } from "@latticexyz/recs";
+import { createRecsStorage, createSqliteStorage, createZustandStorage, tables } from "../test/utils";
+import { logsToBlocks } from "../test/logsToBlocks";
+import worldRpcLogs from "../../../test-data/world-logs-query.json";
+import { buildTable, getTables } from "../src/sqlite";
+import { eq } from "drizzle-orm";
+
+const blocks = logsToBlocks(worldRpcLogs);
```
holic, line 9: can we use the `generateLogs` directly in here? or do we not have the wiring yet for running e.g. anvil etc?

e2e/packages/test-data/generateLogs.ts
yonadaaa: Nice! Yes I added this helper with just the callback for now. 

I explored the alternative approach, similar to [encodeSystemCall](https://github.com/latticexyz/mud/blob/main/packages/world/ts/encodeSystemCall.ts): 
https://github.com/latticexyz/mud/compare/f0e9dfb877fde0b55dd788883e8b6e8d22e147b2...5f47bb7a1dba

(interestingly it is not as strongly typed as I'd expect, gonna investigate):

e2e/packages/test-data/generateLogs.ts
yonadaaa: That was a mistake, ta

packages/store-sync/benchmarks/query.bench.ts
yonadaaa: That's exactly the answer: `e2e` has the scaffolding, `store-sync` doesn't. There could also be weirdness if the benchmarking tool runs a process like anvil.

The scaffold project is a great idea, I shall also give it a think

holic (approved):

Issue #2166, opened 2024-01-19
consider moving codegen table ID constant inside library


holic:
Instead of

```solidity
ResourceId constant _tableId = ResourceId.wrap(
  bytes32(abi.encodePacked(RESOURCE_TABLE, bytes14("store"), bytes16("Tables")))
);
ResourceId constant TablesTableId = _tableId;

library Tables {
  ...
}
```

```solidity
library Tables {
  ResourceId constant _tableId = ResourceId.wrap(
    bytes32(abi.encodePacked(RESOURCE_TABLE, bytes14("store"), bytes16("Tables")))
  );

  ...
}
```

Then we can reference it via `Tables._tableId` (unfortunately we can't use `tableId` because it shadows args in codegen methods)

holic:
Just tried this in a few places and it doesn't seem to have any negative gas consequences and I quite like how this further encapsulates tables and reduces imports.

holic:
We could potentially do this with field layouts too.

holic:
@dk1a @alvrs Can you recall if there was a reason to define these constants outside vs inside the library? Maybe this wasn't a feature in Solidity when we started? :)

dk1a:
> Can you recall if there was a reason to define these constants outside vs inside the library? Maybe accessing library constants wasn't a feature in Solidity when we started? :)

As I remember I just didn't think about library constants there, later it was audit time and this didn't seem important
`Tables._tableId` is nice, and event `Tables.tableId` could work, if we add a config check to avoid `tableId` in arg names

yonadaaa:
@holic this is maybe comparable to issue https://github.com/latticexyz/mud/issues/2158, which proposes moving errors and events _outside_ of libraries. 

But isn't this suggestion, moving constants _inside_ libraries, the exact opposite line of thinking? 

PS: I'm having doubts about the file-level approach. For example, `World` currently inherits `IWorldErrors` and can just use the errors without an import. If the errors are declared at the file-level it might be unwieldy to import each of them manually.

holic:
decided to go with `_tableId` as the name for now, since some of our internal tables use `tableId` for field names and this would conflict without renaming those too

thought that renaming both these fields and table ID argument was too big a change for now

PR #2229, opened 2024-02-01
feat(cli): hardcode table ID with codegen

holic:
doing this as a precursor to #2166 and found that this had a gas improvement in the hot path!


e2e/packages/contracts/src/codegen/tables/Multi.sol
```diff
@@ -18,11 +18,8 @@ import { FieldLayout, FieldLayoutLib } from "@latticexyz/store/src/FieldLayout.s
 import { Schema, SchemaLib } from "@latticexyz/store/src/Schema.sol";
 import { PackedCounter, PackedCounterLib } from "@latticexyz/store/src/PackedCounter.sol";
 import { ResourceId } from "@latticexyz/store/src/ResourceId.sol";
-import { RESOURCE_TABLE, RESOURCE_OFFCHAIN_TABLE } from "@latticexyz/store/src/storeResourceTypes.sol";
 
-ResourceId constant _tableId = ResourceId.wrap(
-  bytes32(abi.encodePacked(RESOURCE_TABLE, bytes14(""), bytes16("Multi")))
-);
+ResourceId constant _tableId = ResourceId.wrap(0x746200000000000000000000000000004d756c74690000000000000000000000);
```
alvrs, line 22: can we add a comment above these that explains this hex value?

Something like

```solidity
// Hex below is the result of `WorldResourceIdLib.encode({ namespace: "", name: "Multi", typeId: RESOURCE_TABLE });`
```

PR #2232, opened 2024-02-02
docs(world/reference): initial version

qbzzt:


qbzzt:
Closing: #2233

docs/pages/store/reference/store-core.mdx
```diff
@@ -844,3 +844,1328 @@ event Store_DeleteRecord(ResourceId indexed tableId, bytes32[] keyTuple);
...
+
+```solidity
+bytes32 internal constant DYNAMIC_DATA_LENGTH_SLOT = keccak256("mud.store.dynamicDataLength");
+```
+
+### Functions
+
+#### \_spliceDynamicData
+
+SET DATA
```
alvrs, line 879: This is coming from the code separator comments, we might have to filter this out in the post processing.

docs/pages/store/reference/store-core.mdx
```diff
@@ -844,3 +844,1328 @@ event Store_DeleteRecord(ResourceId indexed tableId, bytes32[] keyTuple);
...
+| `dynamicFieldIndex`      | `uint8`         | The index of the dynamic field to splice data, relative to the start of the dynamic fields. (Dynamic field index = field index - number of static fields) |
+| `startWithinField`       | `uint40`        | The start index within the field for the splice operation.                                                                                                |
+| `deleteCount`            | `uint40`        | The number of bytes to delete in the splice operation.                                                                                                    |
+| `data`                   | `bytes`         | The data to insert into the dynamic data of the record at the start byte.                                                                                 |
+| `previousEncodedLengths` | `PackedCounter` | The previous encoded lengths of the dynamic data of the record.                                                                                           |
+
+#### \_getStaticData
+
+GET DATA
```
alvrs, line 917: same with this, need to filter out during post processing

docs/pages/store/reference/store-core.mdx
```diff
@@ -844,3 +844,1328 @@ event Store_DeleteRecord(ResourceId indexed tableId, bytes32[] keyTuple);
...
+
+**Returns**
+
+| Name     | Type    | Description                                                      |
+| -------- | ------- | ---------------------------------------------------------------- |
+| `<none>` | `bytes` | The static field data as dynamic bytes in the size of the field. |
+
+#### \_getStaticDataLocation
+
+HELPER FUNCTIONS
```
alvrs, line 973: filter in post processing

docs/pages/world/reference/init-module-implementation.mdx
alvrs, line 1: For consumers it doesn't really matter that these functions are added to the world by the init module, what matters most is that the resulting contract has the [`IBaseWorld.sol`](https://github.com/latticexyz/mud/blob/main/packages/world/src/codegen/interfaces/IBaseWorld.sol) interface (combining all function selectors of the World contract itself and those registered during initialization). I wonder if we should not list the init module and world contract separately, but instead just the `BaseWorld` interface.

docs/pages/world/reference/access-control.mdx
alvrs, line 1: This library is only intended for internal use, I wouldn't put it on the docs to not overwhelm readers with unnecessary information.

docs/pages/world/reference/create.mdx
alvrs, line 1: /internal

docs/pages/world/reference/delegation.mdx
alvrs, line 1: /internal

docs/pages/world/reference/erc165-external.mdx
alvrs, line 1: /internal

docs/pages/world/reference/erc165.mdx
alvrs, line 1: /internal

docs/pages/world/reference/init-module.mdx
alvrs, line 1: /internal

alvrs (commented):
looks great! overall main feedback is that i think we should move the non-public apis to a `/internal` folder within the reference, and i think we should combine `World.sol` and `InitModule` to the `IBaseWorld.sol` interface

docs/pages/store/reference/store-core.mdx
qbzzt: You mean the "SET DATA" line, right?

docs/pages/world/reference/delegation.mdx
qbzzt: Sanity check - you want `delegation.mdx` in `/internal`, but `delegation-external.mdx` needs to stay where it is, right?

qbzzt:
> looks great! 

🙇 

> overall main feedback is that i think we should move the non-public apis to a `/internal` folder within the reference, 

Are the ones you marked above a comprehensive list, or are there more?

> and i think we should combine `World.sol` and `InitModule` to the `IBaseWorld.sol` interface

Is `initModule` internal or external? `IBaseWorld` is external (and I forgot to add it, shame on me - I'll do it now).

docs/pages/store/reference/store-core.mdx
alvrs: yep!

docs/pages/world/reference/delegation.mdx
alvrs: yep exactly! `delegation.mdx` is just used by the world itself to check delegations, while `delegation-external.mdx` is the stuff a developer a would have to implement to get a custom type of delegation

qbzzt:
> If `BaseWorld` includes the entire interface including the natspec of each function (that is defined on the initModule) then initModule can go into internal. (InitModule itself is just the code that installs all these functions on the world, IBaseWorld is where the user interacts with it)

Unfortunately, it only shows what it inherits, and it inherits from autogenerated interfaces that lack the NatSpec of the original `System`s. I fixed it as well as I could, given the current system (see https://mud-docs-feaz67yeu-latticexyz.vercel.app/world/reference/world-external), but it's not ideal.

Any chance we can have `mud codegen` copy NatSpec from the `System` to the interface?

docs/pages/world/reference/system.mdx
```diff
@@ -0,0 +1,176 @@
...
+
+[Git Source](https://github.com/latticexyz/mud/blob/main/packages/world/src/System.sol)
+
+**Inherits:**
+[WorldContextConsumer](/world/reference/world-context#worldcontextconsumer)
+
+_The System contract currently acts as an alias for `WorldContextConsumer`.
+This structure is chosen for potential extensions in the future, where default functionality might be added to the System._
+
+## SystemCall
```
alvrs, line 13: `SystemCall` is an internal util, should go into `/internals` since it's not recommended to be used directly in user code

docs/pages/world/reference/world.mdx
```diff
@@ -0,0 +1,461 @@
...
+```solidity
+function _installRootModule(IModule module, bytes memory encodedArgs) internal;
+```
+
+**Parameters**
+
+| Name          | Type      | Description                                        |
+| ------------- | --------- | -------------------------------------------------- |
+| `module`      | `IModule` | The module to be installed.                        |
+| `encodedArgs` | `bytes`   | The ABI encoded arguments for module installation. |
```
alvrs, line 99: can we filter this function out since it's an `internal` function? Probably has to be done in the post-processing

docs/pages/world/reference/world-external.mdx
alvrs, line 1: what if we include the source files here, ie `WorldRegistrationSystem` (which have the natspec), but in the post-processing remove all the headings and just have an `IBaseWorld` heading at the top? So in the end the reference is for how you can interact with `IBaseWorld` without going into detail on where the functions are implemented (which is less relevant for users)

alvrs:
Getting there! Just a couple more suggestions for things we can fix via the post-processing script

qbzzt:
IBaseWorld page - add all interface data.

alvrs:
this is what i had in mind, great!

alvrs (approved):

PR #2234, opened 2024-02-05
test: add benchmark GitHub workflow

yonadaaa:
Adds a GitHub Action where we can run benchmarks on a remote machine and inspect the output.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -31,6 +31,14 @@ jobs:
         working-directory: ./e2e
         run: pnpm install
 
+      - name: Clean end-to-end project
+        working-directory: ./e2e
+        run: pnpm run clean
+
+      - name: Build end-to-end project
+        working-directory: ./e2e
+        run: pnpm run build
+
       - name: Generate test data
         working-directory: ./e2e
         run: pnpm run generate-test-data-records
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -41,4 +41,4 @@ jobs:
         run: pnpm run generate-test-data-records
 
       - name: Run benchmarks
-        run: pnpm run bench --silent
+        run: pnpm run bench
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -1,10 +1,7 @@
 name: Benchmarks
 
 on:
-  push:
-    branches:
-      - main
-  pull_request:
+  workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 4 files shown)


yonadaaa:
> Is there a way for us to configure the benchmark suite to report absolute numbers instead of relative ones? ("x seconds" vs "x times faster than y")

Yes, it logs a full table of stats before "x times faster than y". I'll look into what the numbers mean and see if we can compress them down to something more legible: 
image

yonadaaa:
> Also curious if you looked into actions like this one https://github.com/benchmark-action/github-action-benchmark that have built-in support for reporting the benchmark results to a github-pages branch

My impression is that most Actions only support certain benchmark tools (ie. not Vitest/tinybench). However [github-action-benchmark](https://github.com/benchmark-action/github-action-benchmark) does mention some custom `biggerIsBetter` that we can use as an adapter - let me try that too :)

alvrs:
> Yes, it logs a full table of stats before "x times faster than y". I'll look into what the numbers mean and see if we can compress them down to something more legible:

Ahh that's great. Somehow didn't find it in the action logs

yonadaaa:
> Ahh that's great. Somehow didn't find it in the action logs 

I had `--silent` on some of the runs which I think only prints the "x > y" summary.

.github/workflows/bench.yml
```diff
@@ -0,0 +1,37 @@
+name: Benchmarks
+
+on:
+  workflow_dispatch:
+
+jobs:
+  test:
```
holic, line 7:
```suggestion
  benchmarks:
```

holic (approved):

package.json
```diff
@@ -10,6 +10,7 @@
   "scripts": {
     "all-codegen": "for dir in packages/store packages/world packages/world-modules packages/cli e2e/packages/contracts examples/*/packages/contracts templates/*/packages/contracts; do (cd \"$dir\" && pwd && pnpm build); done",
     "all-install": "for dir in . docs e2e examples/* templates/*; do (cd \"$dir\" && pwd && pnpm install); done",
+    "bench": "pnpm run --filter=store-sync bench",
```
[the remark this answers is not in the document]
holic: I think if we do `pnpm run --recursive bench`, it'll only run packages in the workspace with the bench script

holic, line 13:
```suggestion
    "bench": "pnpm run --recursive bench",
```

holic (approved):

PR #2238, opened 2024-02-06
feat(store-sync): zustand ECS component helpers

yonadaaa:
Adds ECS-style helpers for reading the Zustand datastore:

- `hasComponent(table, entity)`
- `getComponentValue(table, entity)`
- `getComponentValueStrict(table, entity)`
- `getEntitiesWithValue(table, value`
- `getComponentEntities(table)`

Next, we can add a `Query` interface like RECS.


packages/store-sync/src/zustand/createStore.ts
yonadaaa, line 1: It would be good to declare these utilities in an external file but they use Store's `get()` function. Are there good patterns around this?

One idea is to declare them as "higher order" functions that are instantiated:

```typescript
function getComponentValue<Table extends Table>(
  get: () => ZustandState<Tables>
): (table: Table, entity: Entity) => TableRecord<Table>["value"] | undefined {
  ...
}
```

yonadaaa:
Closing as it was an experiment

PR #2251, opened 2024-02-10
feat(store-sync): byte and boolean array type for in decoded indexer

yonadaaa:
`byteX` and `bool` arrays now have array column types instead of JSON blobs. 

Seems like the drivers are incorrectly set up as the indexer errors when trying to store a byte array:
`PostgresError: column "hooks" is of type bytea[] but expression is of type bytea`

### DIFF
--- a/packages/store-sync/src/postgres/columnTypes.ts
+++ b/packages/store-sync/src/postgres/columnTypes.ts
@@ -1,21 +1,5 @@
 import { customType } from "drizzle-orm/pg-core";
-import superjson from "superjson";
 import { Address, ByteArray, bytesToHex, getAddress, Hex, hexToBytes } from "viem";
-
-// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
-export const asJson = <TData>(name: string) =>
-  customType<{ data: TData; driverData: string }>({
-    dataType() {
-      // TODO: move to json column type? if we do, we'll prob wanna choose something other than superjson since it adds one level of depth (json/meta keys)
-      return "text";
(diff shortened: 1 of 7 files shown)


packages/store-sync/src/postgres/columnTypes.ts
```diff
@@ -73,6 +57,20 @@ export const asAddress = (name: string) =>
...
+  customType<{ data: boolean[]; driverData: string[] }>({
+    dataType() {
+      return "bool[]";
+    },
+    toDriver(data: boolean[]): string[] {
+      return data.map((datum) => String(datum));
+    },
+    fromDriver(driverData: string[]): boolean[] {
+      return driverData.map((datum) => Boolean(datum));
+    },
```
yonadaaa, line 71: @holic do you have any intuition why these drivers don't work?

As `Uint8Array` is interpreted as `bytea`, I would expect `Uint8Array[]` to be interpreted as `bytea[]`. Yet this errors with `PostgresError: column "value" is of type bytea[] but expression is of type bytea`. Test run:

https://github.com/latticexyz/mud/actions/runs/7889740453/job/21530302035?pr=2251#step:6:1813

holic: I'm not sure! I do recall seeing some weird behavior with drizzle driver and the way it translates between types.

yonadaaa: fyi to future readers: I meant to highlight the bytes drivers, not boolean.

PR #2254, opened 2024-02-12
feat(store-sync): wait for idle after each chunk of logs in a block

Kooshaba:
don't ask me why or how, but adding `sleep(1)` after calling the `storageAdapter` made hydration updates show up in the client. i was going crazy trying to figure out why i wasn't receiving them in the main thread, but then remembered having a similar problem in DF with how async calls were being handled. if someone actually knows the mechanism that is being abused here lmk and we can come up with a less obscure solution.

also once i started receiving updates i realized the % calculation was wrong, so fixed that too.

getting these updates is important so i can properly benchmark how long hydration takes on the sky strife end. my manual calculations on my pc are ~5 seconds.

### DIFF
--- a/packages/store-sync/src/createStoreSync.ts
+++ b/packages/store-sync/src/createStoreSync.ts
@@ -145,7 +145,7 @@ export async function createStoreSync<TConfig extends StoreConfig = StoreConfig>
         await storageAdapter({ blockNumber, logs: chunk });
         onProgress?.({
           step: SyncStep.SNAPSHOT,
-          percentage: (i / chunks.length) * 100,
+          percentage: ((i  + 1) / chunks.length) * 100,
           latestBlockNumber: 0n,
           lastBlockNumberProcessed: blockNumber,
           message: "Hydrating from snapshot",
--- a/packages/store-sync/src/createStoreSync.ts
+++ b/packages/store-sync/src/createStoreSync.ts
@@ -48,6 +48,10 @@ type CreateStoreSyncOptions<TConfig extends StoreConfig = StoreConfig> = SyncOpt
   }) => void;
 };
 
+function sleep<T>(timeout: number, returns?: T): Promise<T> {
+  return new Promise<T>((resolve) => setTimeout(() => resolve(returns as T), timeout));
+}
+
 export async function createStoreSync<TConfig extends StoreConfig = StoreConfig>({
   storageAdapter,
   onProgress,
@@ -145,11 +149,12 @@ export async function createStoreSync<TConfig extends StoreConfig = StoreConfig>
         await storageAdapter({ blockNumber, logs: chunk });
         onProgress?.({
           step: SyncStep.SNAPSHOT,
-          percentage: ((i + chunk.length) / chunks.length) * 100,
+          percentage: Math.floor((i / chunks.length) * 100),
           latestBlockNumber: 0n,
           lastBlockNumberProcessed: blockNumber,
           message: "Hydrating from snapshot",
         });
+        await sleep(1);
       }
 
       onProgress?.({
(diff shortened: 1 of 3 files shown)


packages/store-sync/src/createStoreSync.ts
```diff
@@ -145,11 +149,12 @@ export async function createStoreSync<TConfig extends StoreConfig = StoreConfig>
         await storageAdapter({ blockNumber, logs: chunk });
         onProgress?.({
           step: SyncStep.SNAPSHOT,
-          percentage: ((i + chunk.length) / chunks.length) * 100,
+          percentage: Math.floor((i / chunks.length) * 100),
           latestBlockNumber: 0n,
           lastBlockNumberProcessed: blockNumber,
           message: "Hydrating from snapshot",
         });
+        await sleep(1);
```
holic, line 157: Can we use waitForIdle here from common package instead?

packages/store-sync/src/createStoreSync.ts
```diff
@@ -145,11 +149,12 @@ export async function createStoreSync<TConfig extends StoreConfig = StoreConfig>
         await storageAdapter({ blockNumber, logs: chunk });
         onProgress?.({
           step: SyncStep.SNAPSHOT,
-          percentage: ((i + chunk.length) / chunks.length) * 100,
+          percentage: Math.floor((i / chunks.length) * 100),
```
holic, line 152: I am intentionally not rounding here so frontends can decide how much precision they want to render. Let’s remove this diff.

packages/store-sync/src/createStoreSync.ts
holic: And can we also add a note here around why we’re doing this? Something like

> RECS is a synchronous API so hydrating in a loop like this blocks downstream render cycles that would display the percentage climbing up to 100. We wait for idle callback here to give rendering a chance to complete.

holic:
Wonder if we should wait on every block or some number of iterations. Just thinking about how hydrating thousands of blocks will take significantly longer to complete if we have to wait for idle on each block.

packages/store-sync/src/createStoreSync.ts
```diff
@@ -145,11 +145,16 @@ export async function createStoreSync<TConfig extends StoreConfig = StoreConfig>
         await storageAdapter({ blockNumber, logs: chunk });
         onProgress?.({
           step: SyncStep.SNAPSHOT,
-          percentage: ((i + chunk.length) / chunks.length) * 100,
+          percentage: (i / chunks.length) * 100,
```
holic, line 148:
```suggestion
          percentage: ((i  + 1) / chunks.length) * 100,
```

packages/store-sync/src/createStoreSync.ts
```diff
@@ -145,11 +145,16 @@ export async function createStoreSync<TConfig extends StoreConfig = StoreConfig>
         await storageAdapter({ blockNumber, logs: chunk });
         onProgress?.({
           step: SyncStep.SNAPSHOT,
-          percentage: ((i + chunk.length) / chunks.length) * 100,
+          percentage: ((i  + 1) / chunks.length) * 100,
```
holic, line 148:
```suggestion
          percentage: ((i + 1) / chunks.length) * 100,
```

holic:
waiting on https://github.com/latticexyz/mud/pull/2264 just so we can have a test that this does what we expect

yonadaaa (approved):
I added a spec and tested manually with Sky Strife, works great 👍 


https://github.com/latticexyz/mud/assets/29184158/400d3dff-758b-47e2-a9ce-49d694a776c7

holic (approved):

PR #2259, opened 2024-02-12
docs(world/account-delegation): add code sample

qbzzt:
Tested for MUD 2.0


docs/pages/world/account-delegation.mdx
```diff
@@ -21,9 +23,68 @@ This feature enables multiple use cases.
 
 The most common type of delegation is when a delegator address allows a specific delegatee address to act on its behalf.
 
+### Deploy delegation
+
+The owner of the root namespace (and therefore the `World`) needs to deploy the [`StandardDelegationModule`](https://github.com/latticexyz/mud/blob/main/packages/world-modules/src/modules/std-delegations/StandardDelegationsModule.sol).
```
alvrs, line 28: The `StandardDelegationsModule` is used to register a couple pre-build types of limited delegations, namely the `TimeboundDelegation` and `SystemboundDelegation`. You don't need the `StandardDelegationsModule` for unlimited delegations or if you have a custom delegation type.

docs/pages/world/account-delegation.mdx
```diff
@@ -21,9 +23,68 @@ This feature enables multiple use cases.
...
+
+Other than boilerplate, this script does two things.
+
+```solidity
+StandardDelegationsModule standardDelegationsModule =
+  new StandardDelegationsModule();
+```
+
+Deploy a new `StandardDelegationsModule` contract.
+This contract, similar to a `System` contract, is stateless.
```
alvrs, line 73: I think this line is a bit confusing - we should teach folks to think about Modules as "onchain installation scripts" that install things into the World. This module installs two things - the `TimeboundDelegation` check and the `SystemboundDelegation`, both of which are systems that follow the `DelegationControl` interface.

docs/pages/world/account-delegation.mdx
```diff
@@ -21,9 +23,68 @@ This feature enables multiple use cases.
...
+This means that if there is already a `StandardDelegationsModule` contract deployed on the blockchain you can just use it.
+
+```solidity
+world.installRootModule(standardDelegationsModule, new bytes(0));
+```
+
+Install the module.
+This module can only be installed by the owner of the root namespace.
```
alvrs, line 81: Maybe we should clarify this sentence a bit - this module can only be installed as a root module, and root modules can only be installed by the root owner of the namespace (same result but making it a bit more explicit that `installRootModule` can in general only be used by the owner of the root namespace, and `StandardDelegationsModule` needs to be installed as a root module)

docs/pages/world/account-delegation.mdx
alvrs: I think we should have examples for how to register an unlimited delegation (the simplest type, doesn't require any additional code to be deployed) and the extend to using custom delegations, using the `StandardDelegationsModule` as an example.

qbzzt: I already explain how to do unlimited delegation:

> `delegationControlId`, this is usually the `ResourceId` for a `System` that decides whether an attempt to do something by the delegatee on behalf of the delegator is authorized or not.
  Alternatively, this value can be [`UNLIMITED_DELEGATION`](https://github.com/latticexyz/mud/blob/main/packages/world/src/constants.sol#L23-L25), in which case the delegatee has unlimited authority.
  
But I don't want to do more than that, because it violates least privilege. It's not that installing limited delegation is difficult.

alvrs (approved):

docs/pages/world/account-delegation.mdx
```diff
@@ -33,6 +34,72 @@ With MUD delegation, you keep your existing account, but (temporarily) approve o
 
 The most common type of delegation is when a delegator address allows a specific delegatee address to act on its behalf.
 
+### Delegation types
+
+Delegation can either be unlimited or limited by various factors (time, calldata, etc.).
+To create a limited delegation you need a contract that implements the [`DelegationControl`](https://github.com/latticexyz/mud/blob/main/packages/world/src/DelegationControl.sol) interface to decide whether or not to approve a specific call.
+The owner of the root namespace (and therefore the `World`) needs to deploy the [`StandardDelegationModule`](https://github.com/latticexyz/mud/blob/main/packages/world-modules/src/modules/std-delegations/StandardDelegationsModule.sol), which includes three delegation control contracts:
```
alvrs, line 41: This sounds like the owner needs to deploy the standard delegations module to be able to use limited delegations in general. Maybe slightly rephrase to sth like this:

```suggestion
The [`StandardDelegationModule`](https://github.com/latticexyz/mud/blob/main/packages/world-modules/src/modules/std-delegations/StandardDelegationsModule.sol) includes three limited delegation control contracts. To use them, the owner of the root namespace needs to install this module.
```

alvrs (approved):

Issue #1839, opened 2023-10-29
deploy: fallback for deterministic deployer for chains without legacy tx support


holic:
deterministic deployer doesn't work on chains without legacy tx support (pre eip-155), so it'd be nice to detect+fallback in these cases to the previous approach of non-deterministic deploys

holic:
could also try to use something like https://github.com/safe-global/safe-singleton-factory/tree/main/artifacts

PR #2261, opened 2024-02-14
feat(cli): deterministic deployer fallback

holic:
closes https://github.com/latticexyz/mud/issues/1839


.changeset/tough-dolphins-bathe.md
yonadaaa, line 1: So in the deterministic case, deployed Worlds have the same address for each account + salt, and in the non-deterministic case the address is just different every time?

holic: mostly true, yes!

the non-deterministic case deploys a deployer in a non-deterministic way, and that deployer determines the address of all the downstream (reusable) contracts like world factory, systems, modules

yonadaaa (dismissed):

PR #2264, opened 2024-02-14
test(e2e): test sync percentage

yonadaaa:
Tests that the sync progress is 100 when live


yonadaaa:
It would be nice to test the percentage increase from 0, 50, 100, etc..., but because of [how chunk sizes work](https://github.com/latticexyz/mud/blob/main/packages/store-sync/src/createStoreSync.ts#L142) we'd need to generate over 5000 logs, which is too slow on the test set up.

Tested manually and it does though!

e2e/packages/sync-test/data/waitForInitialSync.ts
```diff
@@ -3,4 +3,6 @@ import { Page, expect } from "@playwright/test";
 export async function waitForInitialSync(page: Page) {
   const syncStep = page.locator("#sync-step");
   await expect(syncStep).toHaveText("live");
+  const syncProgress = page.locator("#sync-percentage");
+  await expect(syncProgress).toHaveText("100");
```
holic, line 7: do we need to test for this?

I kinda wonder about keeping the "live" indicator (caught up) separate from the percentage synced, so you can e.g. hide your load screen (because "live") but potentially show how "up to date" you are (percentage)

e2e/packages/sync-test/createStoreSync.test.ts
```diff
@@ -0,0 +1,43 @@
...
+import { createWorld, getComponentValueStrict } from "@latticexyz/recs";
+import { singletonEntity, syncToRecs } from "@latticexyz/store-sync/recs";
+import mudConfig from "../contracts/mud.config";
+import { transportObserver } from "@latticexyz/common";
+import { ClientConfig, createPublicClient, http } from "viem";
+import { getNetworkConfig } from "../client-vanilla/src/mud/getNetworkConfig";
+
+const address = "0xad97505a508daf984fe60302109e0115e544b267";
+
+describe("createStoreSync", () => {
```
holic, line 14: I assume this is here because we haven't created all the scaffolding (anvil, database, etc.) in store-sync tests yet?

holic (approved):
> It would be nice to test the percentage increase from 0, 50, 100, etc..., but because of [how chunk sizes work](https://github.com/latticexyz/mud/blob/main/packages/store-sync/src/createStoreSync.ts#L142) we'd need to generate over 5000 logs, which is too slow on the test set up.

We could make this configurable so it's easier to test. But no need to do that here!

e2e/packages/sync-test/createStoreSync.test.ts
yonadaaa: Yes - `store-sync` doesn't have a contracts folder, `execa`, etc. I kind of like that it is more pure but we do risk `e2e` getting bloated with tests that are not fully "end to end"

e2e/packages/sync-test/data/waitForInitialSync.ts
yonadaaa: I think we can remove this test. `createStoreSync.test.ts` does it more directly

Issue #2260, opened 2024-02-14
Support for external wallets in templates


tash-2s:
- This issue is part of https://github.com/latticexyz/mud/issues/2256

# Motivation

Currently, the templates generated by the `create-mud` command use internal burner wallets in the clients. To facilitate non-local environments, developers are required to manually integrate external wallet connections.

Having external wallet support into the templates would be beneficial. This issue is aimed at achieving that goal.

# Possible implementation options

Several approaches can be considered.

## External wallet support options

Note: `wagmi` is a React library that internally utilizes the vanilla JavaScript library `@wagmi/core`.

Major libraries supporting wallet connections include:

- [`@rainbow-me/rainbowkit`](https://www.rainbowkit.com/): Styled React components built on `wagmi`. Used by Sky Strife and Redstone websites.
- [`connectkit`](https://docs.family.co/connectkit): Styled React components built on `wagmi`.
- [`@rabby-wallet/rabbykit`](https://rabbykit.rabby.io/): Styled UI built on `@wagmi/core`.
- [`@web3modal/wagmi`](https://docs.walletconnect.com/web3modal/about): Styled UI built on `@wagmi/core`.
- [`wagmi`](https://wagmi.sh/react/guides/connect-wallet#build-your-own): React Hooks built on `@wagmi/core`.
- [`@wagmi/core` (and `@wagmi/connectors`)](https://wagmi.sh/core/api/connectors): A vanilla JavaScript library.
- [`mipd`](https://github.com/wevm/mipd): JavaScript utilities for [EIP-6963](https://eips.ethereum.org/EIPS/eip-6963) by the wevm team. This is used in `@wagmi/core`. The retrieved Ethereum providers can be used for [the `viem`'s client custom transport](https://viem.sh/docs/clients/wallet#json-rpc-accounts). Although the conventional `window.ethereum` ([EIP-1193](https://eips.ethereum.org/EIPS/eip-1193)) still works with viem, it's not recommended due to limitations in multi-wallet support. Most wallets are [compatible](https://github.com/WalletConnect/EIP6963/blob/4754abdef8a30ceb74d1126e131fee2e08dbd611/src/utils/constants.ts) with EIP-6963.

## Options for integrating external wallets into templates

The objective is to maintain the current user experience in existing templates, which are designed to work out-of-the-box without necessitating external wallets.

Options include:

- Creating new templates that use external wallets instead of the existing burner wallets.
- Updating existing templates to support both external wallets and burner wallets.

[...]

PR #2268, opened 2024-02-15
feat(examples): add external wallet support example

tash-2s:
- This issue is part of https://github.com/latticexyz/mud/issues/2260

This React example project uses external wallets instead of burner wallets. It currently employs Wagmi v1, but the same component structure should be usable with Wagmi v2 or other libraries.

The example works as expected, except when multiple wallets are installed, which will be resolved in v2. Users must transfer some balance to external wallet accounts for transaction fees beforehand.

https://github.com/latticexyz/mud/assets/81064017/4e1bd7b1-ff47-4e8d-8e95-a73d740ddb25


examples/minimal/packages/client-react-external-wallet/src/mud/store.tsx
yonadaaa, line 1: OOC why is this example app set up differently from the others? (`setup.tsx`, calling `syncStore` in `App` etc.)

I would expect `/mud` to be mostly unchanged, but maybe the setup has to be different for external wallets.

tash-2s: Thank you for the comment!

While introducing the top-level `WagmiConfig`, the distinction between the synchronous network setup and the asynchronous store setup became apparent, which resulted in the diff. The benefit of this PR structure is it doesn't wait for the sync fn for non-relevant code.

Perhaps I should retain the `setup().then(async (result) => { root.render( ...` code?

yonadaaa: I don't think it needs to follow the other examples _exactly_ (though I defer to @holic for that), but it would be interesting to explain the set up maybe in the PR description 🙌

tash-2s:
In the existing templates, there is a certainty that wallets (burner wallets) are present and do not change throughout the app, at the setup stage. As a result, all underlying components have access to the fixed wallets and transaction interfaces. However, with an external wallet setup, the wallet is not fixed (or empty), and it can change. Also, we don't want to block the entire UI for users who do not have wallets. This contributes to the difference in structure.

Another solution might be preparing a burner wallet at first anyway and then later replacing it with an injected wallet.

examples/minimal/packages/client-react-external-wallet/src/App.tsx
```diff
@@ -0,0 +1,146 @@
...
+      <Connect />
+      {isConnected && <Wallet />}
+      {store && (
+        <StoreProvider store={store}>
+          <Synced />
+        </StoreProvider>
+      )}
+    </>
+  );
+};
```
holic, line 39: The way these various parts are separated is a little hard for me to grok. I wonder if there are other approaches that would make this easier to understand.

One pattern we've used in some apps is a [loading screen that is shown while we're syncing data](https://github.com/latticexyz/emojimon/blob/complete/packages/client/src/App.tsx#L6-L29)  from the chain or indexer. We could use that here while waiting for a public client.

And like mentioned above, we could separate the things that depend on the wallet client so it doesn't block showing the UI, etc.

examples/minimal/packages/client-react-external-wallet/src/App.tsx
```diff
@@ -0,0 +1,146 @@
...
+  const {
+    worldContract: { worldContract, write$ },
+    publicClient,
+    walletClient,
+  } = useWorldContract();
+
+  useEffect(() => {
+    if (import.meta.env.DEV) {
+      import("@latticexyz/dev-tools").then(({ mount }) =>
+        mount({
```
holic, line 68: We'll probably have to rethink how Dev Tools mounts if the wallet client is mutable now.

You can see this in your demo video, where the Dev Tools address no longer matches the connected address:
image

holic (commented):
> Another solution might be preparing a burner wallet at first anyway and then later replacing it with an injected wallet.

I don't love this, especially on real chains where the burner wallet would be empty and not able to pay gas anyway.

I think we're better off changing the overall structure to get folks thinking more about only doing writes when a wallet is connected.

I have a bunch of comments related to this but realizing this is quite a big lift. Maybe we should sketch out some designs for our ideal React component tree and render flow?

examples/minimal/packages/client-react-external-wallet/src/App.tsx
tash-2s: I didn't notice that issue, but you're right. I'll look into it.

tash-2s:
As you both mentioned, the React structure can be much simpler. Should be something like this:

```
const networkConfig = getNetworkConfig();
const wagmiConfig = getWagmiConfig();
const syncStorePromise = syncStore(networkConfig, wagmiConfig); // not awaited

<WagmiConfig config={wagmiConfig}>
  <MUDContext value={{networkConfig, syncStorePromise}}>
    <App> and its children can:
      - useAccount() // from wagmi
      - useNetworkConfig() // from MUDContext
      - useStore() // (Store | null) from MUDContext 
      - ...
```

I'll look into the Viem's "getContract" matter. (custom client)

tash-2s:
I've updated the React structure as follows:

```jsx
<WagmiConfig config={wagmiConfig}> // manages external wallet
  <ExternalWallet /> // displays the external wallet UI
  <MUDReadProvider value={mud}> // manages read and immutable mud states (chain, world, publicClient, syncToZustand's result, etc.)
    <MUDWriteProvider> // manages write mud state (transaction interface, etc.)
      <App /> // and its children:
        mudRead = useMUDRead() // can be used similarly to the former `useMUD().network`
        mudWrite = useMUDWrite() // can be used similarly to the former `useMUD().systemCalls`
```

Also, the following issues have been addressed:

- The `getContract` method of MUD is now correctly used. The returned `worldContract` is managed for each wallet account. (When an active account changes, the `worldContract` is regenerated using the new account.)
- `dev-tools` now accurately displays an active wallet account. It re-mounts when the account changes.

Other considerations:

- I've reintroduced the top-level `setup().then(...` code. That change was not directly relevant to this PR.
- This setup presumes that the chain and world are fixed and immutable, similar to other templates. Therefore, the `publicClient` and `syncToZustand` are also fixed on setup.

I've not yet explored viem v2/custom client and am still using `getContract`. But, the current structure should be applicable to the v2 setup as well since App and its children use the abstracted hooks instead of directly using with wagmi hooks.

I believe I've addressed all comments except for those regarding the v2/custom client. Could you please take a look? @holic

tash-2s:
I've merged main and now have viem v2 (https://github.com/latticexyz/mud/pull/2284) in this branch.

tash-2s:
Closing this in favor of https://github.com/latticexyz/mud/pull/2306

Issue #2077, opened 2023-12-19
postgres indexer: uint16 doesn't fit into postgres smallint column type


rohanabraham:
One other issue with the Indexer, as I was trying it on Redstone; it worked very well until an issue was last_updated_block_number 1539672

PostgresError: value "32896" is out of range for type smallint
    at ErrorResponse (file:///xxxxx/.npm/_npx/fd18aa7d46b32af0/node_modules/postgres/src/connection.js:790:26)
    at handle (file:///xxxxx/.npm/_npx/fd18aa7d46b32af0/node_modules/postgres/src/connection.js:476:6)
    at Socket.data (file:///xxxxx/.npm/_npx/fd18aa7d46b32af0/node_modules/postgres/src/connection.js:315:9)
    at Socket.emit (node:events:515:28)
    at addChunk (node:internal/streams/readable:545:12)
    at readableAddChunkPushByteMode (node:internal/streams/readable:495:3)
    at Readable.push (node:internal/streams/readable:375:5)
    at TCP.onStreamRead (node:internal/stream_base_commons:190:23) {
  severity_local: 'ERROR',
  severity: 'ERROR',
  code: '22003',
  where: "unnamed portal parameter $1 = '...'",
  file: 'numutils.c',
  line: '349',
  routine: 'pg_strtoint16_safe'

Was running npx -y -p @latticexyz/store-indexer@next postgres-indexer

holic:
Are you indexing all worlds or just your own world? Do you have MUD's debug logs turned on? Curious to know where the source of this data is coming from.

holic:
https://github.com/latticexyz/mud/blob/a60c13916c043d010686a9bd4979b6aaa3cd46df/packages/store-sync/src/postgres-decoded/buildColumn.ts#L12-L17

Our `buildColumn` method attempts to use `smallint` (equivalent to signed `int2`) for `uint16`, but overflows because `uint16` can exceed the max value of a 2-byte signed integer.

```
➜ type(int16).max
Type: int16
├ Hex: 0x7fff
├ Hex (full word): 0x0000000000000000000000000000000000000000000000000000000000007fff
└ Decimal: 32767
➜ type(uint16).max
Type: uint16
├ Hex: 0xffff
├ Hex (full word): 0x000000000000000000000000000000000000000000000000000000000000ffff
└ Decimal: 65535
```

Need to either bump up the column type or somehow specify whether the column is signed/unsigned (unclear of pg allows this).

[...]

PR #2270, opened 2024-02-16
fix(store-sync): fix overflowing column types, bump postgres sync version

yonadaaa:
- extended fix for #2077
- bumps postgres sync version

### DIFF
--- a/.changeset/fifty-guests-rescue.md
+++ b/.changeset/fifty-guests-rescue.md
@@ -1,0 +1,5 @@
+---
+"@latticexyz/store-sync": patch
+---
+
+Added numerical array column types to the decoded indexer instead of JSON blobs.
--- a/.changeset/fifty-guests-rescue.md
+++ b/.changeset/fifty-guests-rescue.md
@@ -2,5 +2,4 @@
 "@latticexyz/store-sync": patch
 ---
 
-- Moved numerical array types to use array column types (instead of JSON columns) for the Postgres decoded indexer
-- Bumped the Postgres column size for `int32`, `uint32`, `int64`, and `uint64` types to avoid overflows
+Bumped the Postgres column size for `int32`, `uint32`, `int64`, and `uint64` types to avoid overflows
--- a/.changeset/fifty-guests-rescue.md
+++ b/.changeset/fifty-guests-rescue.md
@@ -2,4 +2,5 @@
 "@latticexyz/store-sync": patch
 ---
 
-Added numerical array column types to the decoded indexer instead of JSON blobs.
+- Moved numerical array types to use array column types (instead of JSON columns) for the Postgres decoded indexer
+- Bumped the Postgres column size for `int32`, `uint32`, `int64`, and `uint64` types to avoid overflows
--- a/packages/store-sync/src/postgres-decoded/buildColumn.ts
+++ b/packages/store-sync/src/postgres-decoded/buildColumn.ts
@@ -27,12 +27,12 @@ export function buildColumn(name: string, schemaAbiType: SchemaAbiType) {
       // bigint = 8 bytes (https://www.postgresql.org/docs/current/datatype-numeric.html#DATATYPE-INT)
       return asNumber(name, "bigint");
 
-    case "uint64":
+    case "uint56":
     case "int56":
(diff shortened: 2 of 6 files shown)


packages/store-sync/src/postgres-decoded/buildColumn.ts
```diff
@@ -14,12 +14,12 @@ export function buildColumn(name: string, schemaAbiType: SchemaAbiType) {
     case "int8":
     case "int16":
     case "uint24":
-    case "uint32":
```
holic, line 17: `int32` as well?

packages/store-sync/src/postgres-decoded/buildColumn.ts
```diff
@@ -14,12 +14,12 @@ export function buildColumn(name: string, schemaAbiType: SchemaAbiType) {
...
     case "int24":
     case "int32":
       // integer = 4 bytes (https://www.postgresql.org/docs/current/datatype-numeric.html#DATATYPE-INT)
       return asNumber(name, "integer");
 
+    case "uint32":
     case "uint40":
     case "uint48":
     case "int40":
```
holic, line 25: can't comment below but does `int48` need to be bumped to the next size up too?

packages/store-sync/src/postgres-decoded/buildColumn.ts
yonadaaa: No I don't think so, `int32` covers the exact same range as postgres `integer` - they are both 32 byte signed integers from -2147483648 to +2147483647.

holic: should we make em the same type for consistency? feels weird if `uint32` is a `bigint` but `int32` is a `integer`?

packages/store-sync/src/postgres-decoded/buildColumn.ts
yonadaaa: > can't comment below 

What does that mean sorry? Anyway I don't think so, int48 is already a postgres `bigint` here?

packages/store-sync/src/postgres-decoded/buildColumn.ts
yonadaaa: I agree it's not obvious and requires thinking a bit about Solidity AND Postgres. I'm going just move it, I like the symmetry as well.

packages/store-sync/src/postgres-decoded/buildColumn.ts
yonadaaa: We chatted IRL and we mixed up the JavaScript and Postgres column types. 

but maybe `uint64` and `int64` need to move down because they may not fit within the `bigint` column type

packages/store-sync/src/postgres-decoded/buildColumn.ts
```diff
@@ -27,13 +27,12 @@ export function buildColumn(name: string, schemaAbiType: SchemaAbiType) {
       // bigint = 8 bytes (https://www.postgresql.org/docs/current/datatype-numeric.html#DATATYPE-INT)
       return asNumber(name, "bigint");
 
-    case "uint56":
```
holic, line 30: 🧐

packages/store-sync/src/postgres-decoded/buildColumn.ts
yonadaaa: @holic it definitely doesn't fit - `uint64` is 8 bytes unsigned, postgres `bigint` is 8 bytes signed:

https://www.postgresql.org/docs/current/datatype-numeric.html

packages/store-sync/src/postgres-decoded/buildColumn.ts
yonadaaa: oh my! thanks

.changeset/fifty-guests-rescue.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/store-sync": patch
+---
+
+Added numerical array column types to the decoded indexer instead of JSON blobs.
```
holic, line 5:
```suggestion
- Added numerical array column types to the decoded indexer instead of JSON blobs
- Bumped the Postgres column size for `int32`, `uint32`, `int64`, and `uint64` types to avoid overflows
```

holic (dismissed):

PR #2274, opened 2024-02-16
docs(guides/extending): add table, and update to current version

qbzzt:
Replacing #2200


docs/pages/guides/_meta.js
```diff
@@ -1,6 +1,6 @@
 export default {
   "replicating-onchain-state": "Replicating onchain state",
   "hello-world": "Hello World",
-  "extending-a-world": "Extending a World",
+  extending: "Extending a World",
```
holic, line 4: can we keep the `extending-a-world` slug to avoid the need for another redirect and to better capture what this guide is about?

qbzzt:
I'm going to close this PR and replace it with a new one that hopefully won't have conflicts I can't resolve.

PR #2278, opened 2024-02-19
feat(e2e,store-sync): benchmark with sky strife data

yonadaaa:
Adds a script to download logs from the [Sky Strife Redstone Holesky world](https://github.com/latticexyz/skystrife-public/blob/main/packages/contracts/worlds.json#L22-L25) and save them to file, and a benchmark that uses the data to compare our client storage adapters.

This contract has roughly one million logs - reading a fraction of that from file exceeds the JavaScript heap size. Currently using a small block range which generates 3844 logs:
```TypeScript
const WORLD_ADDRESS = "0x7203e7adfdf38519e1ff4f8da7dcdc969371f377";
const FROM_BLOCK = 895629;
const TO_BLOCK = 896500;

const logs = await getLogs(WORLD_ADDRESS, FROM_BLOCK, TO_BLOCK);
```


packages/store-sync/src/zustand/storageAdapterSkystrife.test.ts
yonadaaa, line 1: Currently this test fails with some key schema error. I think it's either because the World uses an outdated MUD version, or Sky Strife upgraded one of these tables.
image

ludns:
To deal with the heap issue you can stream the file so you don't read all of it in memory

yonadaaa:
> To deal with the heap issue you can stream the file so you don't read all of it in memory

Good suggestion, this is probably more realistic too

yonadaaa:
Closing this as the current World is on an old MUD version and the data is incompatible

Issue #2222, opened 2024-02-01
consider including field index (and field bytes offset) in Store events


holic:
Store's splice operations on static and dynamic fields use a field index to map to contract storage slots. Without this data expressed in events, offchain clients/indexers are ~forced to model the data as a single bytes blob for static field data and another for dynamic field data.

Since we already have access to the field index in these methods (and the field-specific bytes offset for dynamic fields), we could expose these as part of the event to allow for more flexible ways of modeling data offchain.

We should probably measure the gas of this to make sure it's worth the trade-off. Adding this also means we'd cement the "one splice event = one field update", which seems fine.

holic:
Unclear how to handle the set event - I think we'd have to send down the whole field layout?

holic:
The idea here is that we currently support creating schemaless indexers, but with the constraint that data is stored in two fields: static data and dynamic data.

However, that's not how it's actually stored onchain and I want to explore what additional args we could provide in events to offer indexer developers to have the option to store in a single blob for all like data vs. blob per field without a schema lookup.

You could still build this with the events we have now, but would require an additional schema lookup.

dk1a:
For more schemalessness the indexer devs could use FieldLayout's total static length
It's like how onchain the data is also schemaless and only FieldLayout is used, and dynamic data always comes after the static data, whose length never changes

> How do we feel then about removing start for the static data event?

Some thoughts on making `start` an encapsulated implementation detail, for the more schemaful route:

- `start` in `spliceStaticData` allows editing several (not necessarily all, unlike `setRecord`) static fields at once, which is mostly useless, and table libs don't take advantage of this
- `start` in `spliceDynamicData` allows editing multiple array elements (or a byte slice) at once, which seems more useful, like a js splice, but table libs once again don't take advantage of this (they could, I just never got around to it - table libs would translate an element index to bytes, they already do similar stuff in some methods)

We can replace `start` with `startElementIndex` or the like, without losing functionality, all while making table libs do less translation, but then splice methods in `StoreCore` would need to get FieldLayout, to get the necessary byte lengths. So the cost would be 1 more storage read I guess (or 1 more arg, if we think just sending it is fine)

holic:
- add field index to dynamic splice so folks have the option to store each dynamic field separately (only looking at events without schema lookups)
- maybe could do the same for static field but would need both field index in splice event and potentially field layout in set record event

yonadaaa:
[...]

PR #2279, opened 2024-02-19
feat(store): add field index to Store_SpliceDynamicData event

yonadaaa:
Fixes https://github.com/latticexyz/mud/issues/2222


packages/store/src/StoreCore.sol
```diff
@@ -1086,6 +1088,7 @@ library StoreCoreInternal {
       emit StoreCore.Store_SpliceDynamicData({
         tableId: tableId,
         keyTuple: keyTuple,
+        dynamicFieldIndex: dynamicFieldIndex,
```
holic, line 1091: Do we also need to include `startWithinField`?

yonadaaa: Oh good point - because `dynamicFieldIndex` only tells you which field is being updated, but not the element. I will add `startWithinField` and we can see the gas impact.

packages/store/test/StoreCore.t.sol
```diff
@@ -800,17 +806,6 @@ contract StoreCoreTest is Test, StoreMock {
...
-    vm.expectEmit(true, true, true, true);
-    emit Store_SpliceDynamicData(
-      data.tableId,
-      data.keyTuple,
-      uint48(data.secondDataBytes.length),
-      0,
-      PackedCounterLib.pack(data.newSecondDataBytes.length, data.thirdDataBytes.length),
-      data.secondDataToPush
-    );
-
```
yonadaaa, line 813: ❗ temporarily deleted these tests until I figure out what `startWithinField` should be for push/pop

yonadaaa: Added them back

packages/store/src/StoreCore.sol
yonadaaa: Added it!

holic: looks like ~4-5% increase in dynamic field operations

docs/pages/store/reference/store-core.mdx
```diff
@@ -821,14 +823,16 @@ event Store_SpliceDynamicData(
...
+| Name                | Type            | Description                                                               |
+| ------------------- | --------------- | ------------------------------------------------------------------------- |
+| `tableId`           | `ResourceId`    | The ID of the table where the data is spliced.                            |
+| `keyTuple`          | `bytes32[]`     | An array representing the composite key for the record.                   |
+| `dynamicFieldIndex` | `uint8`         | The index of the dynamic field.                                           |
+| `startWithinField`  | `uint40`        | The start byte position within the field for splicing.                    |
+| `start`             | `uint48`        | The start position in bytes for the splice operation.                     |
+| `deleteCount`       | `uint40`        | The number of bytes to delete in the splice operation.                    |
+| `encodedLengths`    | `PackedCounter` | The encoded lengths of the dynamic data of the record.                    |
+| `data`              | `bytes`         | The data to insert into the dynamic data of the record at the start byte. |
```
holic, line 835: Seeing this all laid out, I am wondering:
- Do we still like the `Store_SpliceDynamicData` name since splice events are now ~guaranteed to operate on one field at a time and it includes both information for splicing within the field and for all of the dynamic data? Wondering if this should be something like `Store_SpliceDynamicField` to clarify the intent.
- With above, I am also wondering if it's worth clarifying some argument names like `startWithinField` to `fieldStart` and `start` -> `dynamicDataStart`.
- "dynamic" in `dynamicFieldIndex` might be redundant since this is a `Store_SpliceDynamicData` event

alvrs: Do we need `startWithinField`? Couldn't that be computed from `dynamicFieldIndex`, `start` and `encodedLengths`? I'd prefer adding that one additional field to the event than refactoring the event name and other field names. The current naming feels very intuitive if you're thinking of the data as one byte blob, while the `Store_SpliceDynamicField` naming would be a little more intuitive if you're storing data in separate locations, but I don't there is a strong enough bias in either direction to change the name at this point.

holic:
```ts
const fieldLengths = getFieldLengths(encodedLengths); // some sort of decoding function
const fieldStart = fieldLengths.slice(0, dynamicFieldIndex).reduce((sum, length) => sum + length, 0);
const startWithinField = start - fieldStart;
```
checks out ✅

packages/store/src/StoreCore.sol
```diff
@@ -55,6 +55,8 @@ library StoreCore {
    * @notice Emitted when dynamic data in the store is spliced.
    * @param tableId The ID of the table where the data is spliced.
    * @param keyTuple An array representing the composite key for the record.
+   * @param dynamicFieldIndex The index of the dynamic field.
```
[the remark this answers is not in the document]
yonadaaa: Done 📖

.changeset/rich-deers-doubt.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/store": minor
```
holic, line 2:
```suggestion
"@latticexyz/store": major
```

.changeset/rich-deers-doubt.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Added `dynamicFieldIndex` to the `Store_SpliceDynamicData` event. This enables indexers to store records as a blob per field without a schema lookup.
```
holic, line 6:
```suggestion
Added `dynamicFieldIndex` to the `Store_SpliceDynamicData` event. This enables indexers to store dynamic data as a blob per dynamic field without a schema lookup.
```

holic (approved):

Issue #2162, opened 2024-01-18
add tests for core system functions that they can only be delegatecalled


holic:
https://github.com/latticexyz/mud/pull/2111#issuecomment-1899029168

PR #2280, opened 2024-02-19
test(world): check core system functions can only be delegatecalled

yonadaaa:
Fixes https://github.com/latticexyz/mud/issues/2162

~Testing this in Solidity is quite cumbersome - I wonder if we can write a JavaScript test that iterates through every system and tries to call them. Or even codegen so we don't have to manually write each selector.~
(found an ok way to do it in Solidity)


holic:
testing this in JS seems reasonable! and then we can just enumerate the system methods via the generated ABI

packages/world/gas-report.json
yonadaaa, line 1: The gas diff comes from the `getFunctionSignaturesX` helpers that were added because Solidity doesn't support `string` arrays as constants. 

It's a frustrating limitation but I think the 0.01% increase is worthwhile for maintainability.

packages/world/src/modules/init/functionSignatures.sol
```diff
@@ -0,0 +1,62 @@
...
+ * @dev Function signatures for access management system
+ */
+function getFunctionSignaturesAccessManagement() pure returns (string[4] memory) {
+  return [
+    // --- AccessManagementSystem ---
+    "grantAccess(bytes32,address)",
+    "revokeAccess(bytes32,address)",
+    "transferOwnership(bytes32,address)",
+    "renounceOwnership(bytes32)"
+  ];
```
holic, line 14: WHAT YOU CAN DO THIS

yonadaaa: We already were! only works for static arrays

packages/world/src/modules/init/functionSignatures.sol
yonadaaa, line 1: In a future PR, we can code generate this file, similarly to system interfaces.

holic (approved):

holic:
the idea here is that we have centralized where the function signatures are defined so that we can both register and test their proper registration using the same source array

in the future, we'd like to automate this via some codegen (#2155) but gonna skip that for now

PR #2281, opened 2024-02-19
docs(config): add enums  🏎

qbzzt:
also add back the singleton explanation in `store/tables`.


docs/pages/store/tables.mdx
```diff
@@ -77,6 +77,16 @@ keySchemaTypes[0] = SchemaType.ADDRESS;
 Schema keySchema = SchemaLib.encode(keySchemaTypes);
 ```
 
+#### Singleton tables
+
+When you just need to store a single value (for example, the world map or the name of an in-game currency), you use a singleton table, a table whose key schema is empty.
+With an empty key schema there is just one row.
```
holic, line 83:
```suggestion
With an empty key schema there is just one record.
```
https://www.notion.so/latticexyz/MUD-naming-conventions-terminology-66e11356a1934465b2aa1a951c61a299

docs/pages/store/tables.mdx
```diff
@@ -77,6 +77,16 @@ keySchemaTypes[0] = SchemaType.ADDRESS;
...
+
+```solidity copy
+SchemaType[] memory _keySchema = new SchemaType[](0);
+Schema keySchema SchemaLib.encode(_keySchema);
+```
```
holic, line 88: No change needed here because this is more consistent with the rest of this page, but I think we should reevaluate having these implementation detail code snippets here. I think what's more useful to demonstrate is how to get data out of Store and the different flavors it comes in (singleton tables vs one key vs multiple keys, tuple of values returned vs. struct returned)

qbzzt: We already have how to use tables in your code in https://mud.dev/store/table-libraries, and a link to it. So I think under **Advanced** we should keep this code.

qbzzt: Also, I decided that singleton tables belong in the "how to use" section, which is https://mud.dev/store/table-libraries, not the theory section, https://mud.dev/store/tables.

docs/pages/store/table-libraries.mdx
```diff
@@ -34,6 +34,18 @@ const Inventory = {
...
+### Singleton tables
+
+When you just need to store a single value (for example, the world map or the name of an in-game currency), you use a singleton table, a table whose key schema is empty.
+With an empty key schema there is just one record.
+
+```ts
+const Counter = {
+  keySchema: {},
+  valueSchema: "uint32",
+};
```
alvrs, line 46: this is outdated

alvrs (approved):

alvrs (approved):

PR #2283, opened 2024-02-20
feat(store-sync): bool array column types for decoded indexer

yonadaaa:
Pulled out from https://github.com/latticexyz/mud/pull/2251

### DIFF
--- a/.changeset/lemon-numbers-design.md
+++ b/.changeset/lemon-numbers-design.md
@@ -1,0 +1,5 @@
+---
+"@latticexyz/store-sync": patch
+---
+
+Added boolean array column types for the Postgres decoded indexer.
--- a/.changeset/lemon-numbers-design.md
+++ b/.changeset/lemon-numbers-design.md
@@ -2,4 +2,4 @@
 "@latticexyz/store-sync": patch
 ---
 
-Added boolean array column types for the Postgres decoded indexer.
+Moved boolean array types to use array column types (instead of JSON columns) for the Postgres decoded indexer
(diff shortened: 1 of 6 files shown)


packages/store-sync/src/postgres-decoded/buildColumn.ts
```diff
@@ -234,7 +243,7 @@ export function buildColumn(name: string, schemaAbiType: SchemaAbiType) {
     case "bytes31[]":
     case "bytes32[]":
     case "bool[]":
-      return asJson(name);
+      return asBoolArray(name);
```
holic, line 246: erm, I think this applying the column type to all the above array types, not _just_ the bool type

.changeset/lemon-numbers-design.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/store-sync": patch
+---
+
+Added boolean array column types for the Postgres decoded indexer.
```
holic, line 5:
```suggestion
Moved boolean array types to use array column types (instead of JSON columns) for the Postgres decoded indexer
```
nit: just to match the changeset I added in https://github.com/latticexyz/mud/pull/2282

packages/store-sync/src/postgres-decoded/buildColumn.ts
holic: since this is a column change, should bump the postgres schema/sync version

yonadaaa: Oh thanks - been looking at these column types for too long 😵‍💫

holic (approved):

Issue #2267, opened 2024-02-15
upgrade viem to v2


holic:
so we can also make use of wagmi v2 and latest rainbowkit/connectkit

related to #2256

PR #2284, opened 2024-02-20
feat: upgrade viem to v2

tash-2s:
closes https://github.com/latticexyz/mud/issues/2267


tash-2s:
warning when executing `pnpm install` at the repository root:

```
 WARN  Issues with peer dependencies found
packages/cli
└─┬ viem 2.7.11
  └─┬ abitype 1.0.0
    └── ✕ unmet peer zod@"^3 >=3.22.0": found 3.21.4

packages/config
└─┬ viem 2.7.11
  └─┬ abitype 1.0.0
    └── ✕ unmet peer zod@"^3 >=3.22.0": found 3.21.4

packages/faucet
└─┬ viem 2.7.11
  └─┬ abitype 1.0.0
    └── ✕ unmet peer zod@"^3 >=3.22.0": found 3.21.4

packages/react
└─┬ @testing-library/react-hooks 8.0.1
  ├── ✕ unmet peer @types/react@"^16.9.0 || ^17.0.0": found 18.2.22
  ├── ✕ unmet peer react@"^16.9.0 || ^17.0.0": found 18.2.0
  └── ✕ unmet peer react-test-renderer@"^16.9.0 || ^17.0.0": found 18.2.0

packages/store
└─┬ viem 2.7.11
  └─┬ abitype 1.0.0
    └── ✕ unmet peer zod@"^3 >=3.22.0": found 3.21.4

packages/store-indexer
└─┬ viem 2.7.11
  └─┬ abitype 1.0.0
    └── ✕ unmet peer zod@"^3 >=3.22.0": found 3.21.4

packages/store-sync
└─┬ viem 2.7.11
  └─┬ abitype 1.0.0
    └── ✕ unmet peer zod@"^3 >=3.22.0": found 3.21.4

packages/world
└─┬ viem 2.7.11
  └─┬ abitype 1.0.0
    └── ✕ unmet peer zod@"^3 >=3.22.0": found 3.21.4
```

holic:
I think we could bump zod here too!

tash-2s:
@holic The `abitype`'s `zod` peer dependency is optional, and we're not using `abitype/zod`, so I believe it doesn't cause an issue. But it's better to resolve the warning later.

tash-2s:
Also, I might be able to pass the typescript test without upgrading the mud's abitype, but also it's better to upgrade it alongside viem.

holic:
> but also it's better to upgrade it alongside viem

agree on aligning our abitype version with the one that viem's version uses

packages/block-logs-stream/tsconfig.json
```diff
@@ -9,6 +9,7 @@
     "isolatedModules": true,
     "esModuleInterop": true,
     "forceConsistentCasingInFileNames": true,
+    "skipLibCheck": true,
```
holic, line 12: what are the consequences of this?

packages/cli/src/deploy/ensureWorldFactory.ts
```diff
@@ -17,7 +17,7 @@ export const accessManagementSystemDeployedBytecodeSize = size(
 );
 export const accessManagementSystemBytecode = encodeDeployData({
   bytecode: accessManagementSystemBuild.bytecode.object as Hex,
-  abi: [],
+  abi: [] as Abi,
```
holic, line 20: why do we have to cast this? I would expect TS to be able to infer this in the call to `encodeDeployData`

packages/common/src/chains/latticeTestnet.ts
```diff
@@ -3,7 +3,6 @@ import type { MUDChain } from "./types";
 export const latticeTestnet = {
   name: "Lattice Testnet",
   id: 4242,
-  network: "lattice-testnet",
```
holic, line 6: for anyone reading: the `network` key got removed from `Chain` in latest viem (I think this was a ethers carry-over)

packages/cli/src/deploy/ensureWorldFactory.ts
tash-2s: `abi: []` is now not allowed for `encodeDeployData`. I think `[]` is not the abi expected here. I'll look into this.

```
src/deploy/ensureWorldFactory.ts:18:64 - error TS2345: Argument of type '{ bytecode: `0x${string}`; abi: never[]; }' is not assignable to parameter of type 'never'.

18 export const accessManagementSystemBytecode = encodeDeployData({
                                                                  ~
19   bytecode: accessManagementSystemBuild.bytecode.object as Hex,
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
20   abi: [],
   ~~~~~~~~~~
21 });
   ~
```

packages/block-logs-stream/tsconfig.json
tash-2s: `"skipLibCheck": true` is necessary for using viem, and even in the viem repository, the same error occurs when removing the `"skipLibCheck": true`. I believe there isn't a good way to maintain `"skipLibCheck": false`, although I don't fully understand the exact cause of the error.

Skipping the check should be fine here because the declaration files are not utilized in the packages where I enabled the flag, making it relevant only for library files.

packages/cli/src/deploy/ensureWorldFactory.ts
tash-2s: The encodeDeployData function does the task only if the contract has constructor args. It appears to have started rejecting abis with no effects at the type level? The logic of the function hasn't changed, so its behavior remains the same as before.

Alternatively, we can bypass this function:

```diff
diff --git a/packages/cli/src/deploy/ensureWorldFactory.ts b/packages/cli/src/deploy/ensureWorldFactory.ts
index 1f2913a8..f45cbed8 100644
--- a/packages/cli/src/deploy/ensureWorldFactory.ts
+++ b/packages/cli/src/deploy/ensureWorldFactory.ts
@@ -15,10 +15,7 @@ import { Contract } from "./ensureContract";
 export const accessManagementSystemDeployedBytecodeSize = size(
   accessManagementSystemBuild.deployedBytecode.object as Hex
 );
-export const accessManagementSystemBytecode = encodeDeployData({
-  bytecode: accessManagementSystemBuild.bytecode.object as Hex,
-  abi: [] as Abi,
-});
+export const accessManagementSystemBytecode = accessManagementSystemBuild.bytecode.object as Hex;
 export const accessManagementSystem = getCreate2Address({
   from: deployer,
   bytecode: accessManagementSystemBytecode,
```

packages/common/src/chains/latticeTestnet.ts
tash-2s: The network field was removed in [this commit](https://github.com/wevm/viem/commit/dd8f0c77011d6b75d4440f033366c9edd3214f77#diff-e342de8b8f37c8c14edefc2e0e74e7a34b156a5022e58564cf7da481db8efeaa). It suggests using the id field instead when necessary.

packages/world/ts/encodeSystemCall.ts
```diff
@@ -1,33 +1,29 @@
...
-export function encodeSystemCall<abi extends Abi, functionName extends string = string>({
+export function encodeSystemCall<abi extends Abi, functionName extends ContractFunctionName<abi>>({
   abi,
   systemId,
   functionName,
   args,
-}: SystemCall<abi, functionName>): GetFunctionArgs<typeof IWorldCallAbi, "call">["args"] {
+}: SystemCall<abi, functionName>): AbiParametersToPrimitiveTypes<
+  ExtractAbiFunction<typeof IWorldCallAbi, "call">["inputs"]
+> {
```
tash-2s, line 20: These encode call functions has been manually tested like this:

```ts
await worldContract.write.call(
  encodeSystemCall({
    abi: worldContract.abi,
    functionName: "addTask",
    args: [label],
    systemId: resourceToHex({ type: "system", namespace: "", name: "TasksSystem" }),
  })
);
await worldContract.write.batchCall([
  encodeSystemCalls(worldContract.abi, [
    {
      functionName: "addTask",
      args: [label],
      systemId: resourceToHex({ type: "system", namespace: "", name: "TasksSystem" }),
    },
  ]).map(([systemId, callData]) => ({ systemId, callData })),
]);
await worldContract.write.callFrom(
  encodeSystemCallFrom({
    from: walletClient.account.address,
    abi: worldContract.abi,
    functionName: "addTask",
    args: [label],
    systemId: resourceToHex({ type: "system", namespace: "", name: "TasksSystem" }),
  })
);
await worldContract.write.batchCallFrom([
  encodeSystemCallsFrom(worldContract.abi, walletClient.account.address, [
    {
      functionName: "addTask",
      args: [label],
      systemId: resourceToHex({ type: "system", namespace: "", name: "TasksSystem" }),
    },
  ]).map(([from, systemId, callData]) => ({ from, systemId, callData })),
]);
```

packages/common/src/getContract.ts
```diff
@@ -69,13 +72,17 @@ export function getContract<
...
-    publicClient,
-    walletClient,
-  }) as unknown as GetContractReturnType<Abi, PublicClient, WalletClient>;
+    client: {
+      public: publicClient,
+      wallet: walletClient,
+    },
+  }) as unknown as GetContractReturnType<TAbi, { public: TPublicClient; wallet: TWalletClient }, TAddress> & {
+    write: unknown;
+  };
```
tash-2s, line 85: The argument structure for viem's getContract have changed slightly, but mud's getContract still uses the same arguments to minimize the effect.

packages/common/src/sendTransaction.ts
```diff
@@ -24,7 +24,7 @@ export async function sendTransaction<
 >(
   client: Client<Transport, TChain, TAccount>,
   request: SendTransactionParameters<TChain, TAccount, TChainOverride>
-): Promise<WriteContractReturnType> {
+): Promise<SendTransactionReturnType> {
```
tash-2s, line 27: This aligns with the viem's function.

packages/common/src/getContract.ts
holic: since `getContract` is meant to align with/mirror viem, I wonder if we should update our arguments/types to math

tash-2s: To ensure it remains a drop-in replacement, we should update the `mud_getContract` arguments to match those of viem v2. While mud users will need to update their code that uses `mud_getContract`, I now feel that this is natural.

OK, I'll go ahead and update the `mud_getContract` args.

```ts
// viem v2
getContract({
  abi,
  address,
  client: {
    public,
    wallet,
  },
})

// current mud
getContract({
  abi,
  address,
  publicClient,
  walletClient,
  onWrite?,
})

// will be...
getContract({
  abi,
  address,
  client: {
    public,
    wallet,
  },
  onWrite?,
})
```

packages/cli/src/deploy/ensureWorldFactory.ts
holic: seems fine to leave the empty ABI approach!

packages/common/src/getContract.ts
```diff
@@ -69,13 +75,17 @@ export function getContract<
...
     address,
-    publicClient,
-    walletClient,
-  }) as unknown as GetContractReturnType<Abi, PublicClient, WalletClient>;
+    client: {
+      public: publicClient,
+      wallet: walletClient,
+    },
+  }) as unknown as GetContractReturnType<TAbi, { public: TPublicClient; wallet: TWalletClient }, TAddress> & {
+    write: unknown;
```
holic, line 87: what does this bit do? do we need it?

packages/common/src/getContract.ts
```diff
@@ -98,11 +108,11 @@ export function getContract<
               args,
               ...options,
               onWrite,
-            } as unknown as WriteContractParameters<TAbi, typeof functionName, TChain, TAccount>;
+            } as unknown as WriteContractParameters;
```
holic, line 111: are we losing type specificity here?

packages/world/ts/encodeSystemCall.ts
```diff
@@ -1,33 +1,29 @@
...
   return [
     systemId,
     encodeFunctionData({
       abi,
       functionName,
       args,
-    } as unknown as EncodeFunctionDataParameters<abi, functionName>),
+    } as unknown as EncodeFunctionDataParameters),
```
holic, line 27: same as above, are we losing any type specificity/safety here?

holic (commented):
this is looking great!

just a few comments and I went ahead and added a changeset

packages/common/src/getContract.ts
tash-2s: This corresponds to [this viem code](https://github.com/wevm/viem/blob/viem%402.7.12/src/actions/getContract.ts#L513-L524). The actual properties attached to the contract object depend on the clients and the abi, and since we don't know the actual abi here, we need to specify this to have the write property.
There might be a better way to handle this?

tash-2s: While calling this MUD's getContract, it is typed the same as Viem's since it uses the same arguments and return type.

packages/common/src/getContract.ts
tash-2s: This loses some type specificity. I'll fix it.

tash-2s:
@holic Thank you so much for the review and the changeset! I've addressed the comments and resolved the conflict.

holic (approved):

PR #2289, opened 2024-02-21
WIP sbg indexer

holic:


alvrs:
wonder if we should merge this as a reference? could replace the hardcoded things with .env values and make it extra easy for anyone to spin up an indexer on render

PR #2290, opened 2024-02-22
feat(store): add types for v2 config resolvers

alvrs:
Note: this only includes types for now, no runtime implementation. Those with be in a followup PR.


packages/store/ts/config/v2/error.ts
```diff
@@ -0,0 +1 @@
+export type Error<reason extends string = string> = `!Error: ${reason}`;
```
holic, line 1: would this conflict at all with the native `Error` class/type (e.g. `throw new Error(...)`)?

packages/config/src/library/v2/common.ts
```diff
@@ -0,0 +1,21 @@
...
+}
+
+// eslint-disable-next-line @typescript-eslint/no-empty-interface, @typescript-eslint/no-unused-vars
+export interface MudResolvedConfigOutput<configInput extends MudConfigInput = MudConfigInput> {
+  /** This interface can be extended by MUD config plugins */
+}
+
+export interface MudConfigOutput<configInput extends MudConfigInput = MudConfigInput> {
+  resolved: MudResolvedConfigOutput<configInput>;
+}
```
holic, line 13: do we still use these interfaces? should they be `type` instead of `interface`?

packages/config/src/library/v2/common.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export type resolveConfig<configInput extends MudConfigInput> = configInput & MudConfigOutput<configInput>;
+
+export const resolveConfig = <configInput extends MudConfigInput>(
```
holic, line 17: thoughts on `resolveConfig` vs `defineConfig`?

maybe `defineConfig` wraps `resolveConfig` among other things if we e.g. have a separate but built-in codegen config, ECS config, etc?

packages/store/ts/config/v2/generics.ts
```diff
@@ -0,0 +1,7 @@
+export type get<input, key> = key extends keyof input ? input[key] : undefined;
+
+export type freeze<T> = T extends Array<infer R>
+  ? ReadonlyArray<freeze<R>>
+  : T extends object
+  ? { readonly [P in keyof T]: freeze<T[P]> }
+  : T;
```
holic, line 7: can prob remove since we're not using this?

packages/store/ts/config/v2/generics.ts
```diff
@@ -0,0 +1,7 @@
+export type get<input, key> = key extends keyof input ? input[key] : undefined;
```
holic, line 1: is `undefined` here desirable?

packages/store/ts/config/v2/schema.ts
```diff
@@ -0,0 +1,23 @@
...
+    /** the Solidity primitive ABI type */
+    type: scope["validTypes"][schema[key]];
+    /** the user defined type or Solidity primitive ABI type */
+    internalType: schema[key];
+  };
+}>;
+
+export function resolveSchema<schema extends SchemaInput<scope>, scope extends AbiTypeScope = AbiTypeScope>(
+  schema: schema,
+  scope?: scope
```
holic, line 19: wondering about naming this arg `abiTypeScope`

packages/store/ts/config/v2/scope.ts
```diff
@@ -0,0 +1,36 @@
...
+export type AbiType = StaticAbiType | DynamicAbiType;
+
+export const EmptyScope = { validTypes: {} } as const satisfies ScopeOptions;
+export type EmptyScope = typeof EmptyScope;
+
+export type AbiTypeScope = ScopeOptions<{ [t in AbiType]: t }>;
+export const AbiTypeScope = { validTypes: {} } as AbiTypeScope; // TODO: runtime implementation
+
+export type ScopeOptions<validTypes extends Dict<string, AbiType> = Dict<string, AbiType>> = {
+  validTypes: validTypes;
```
holic, line 14: would this ever include invalid types? 🤔

packages/store/ts/config/v2/table.ts
```diff
@@ -0,0 +1,145 @@
...
+export type validateTableConfig<
+  input,
+  scope extends AbiTypeScope = AbiTypeScope
+> = input extends TableShorthandInput<scope>
+  ? validateTableShorthand<input, scope>
+  : input extends string
+  ? validateTableShorthand<input, scope>
+  : validateTableFull<input, scope>;
+
+type resolveTableFullConfig<
```
holic, line 105: should prob export ~everything?

templates/vanilla/packages/contracts/package.json
```diff
@@ -16,6 +16,7 @@
   },
   "dependencies": {
     "@latticexyz/cli": "link:../../../../packages/cli",
+    "@latticexyz/config": "link:../../../../packages/config",
```
holic, line 19: I don't think this is used?

templates/vanilla/packages/contracts/v2.mud.config.ts
```diff
@@ -0,0 +1,13 @@
...
+const resolveConfig = <configInput extends WorldConfigInput>(
+  configInput: configInput
+): resolveStoreConfig<withWorldConfig<configInput>> => {
+  return resolveStoreConfig(withWorldConfig(configInput));
+};
+
+const config = resolveConfig({
+  namespaces: { testNamespace: { name: "testNamespaceName" } },
+  tables: { testTable: { name: "testNameTable" } },
+} as const);
```
holic, line 13: should we remove this for now? so it doesn't land in new templtaes

packages/store/ts/config/v2/table.ts
```diff
@@ -0,0 +1,145 @@
...
+
+export type TableInput<
+  schema extends SchemaInput<scope> = SchemaInput,
+  scope extends AbiTypeScope = AbiTypeScope,
+  primaryKey extends ValidKeys<schema, scope> = ValidKeys<schema, scope>
+> = TableFullInput<schema, scope, primaryKey> | TableShorthandInput<scope>;
+
+export type TableShorthandInput<scope extends AbiTypeScope = AbiTypeScope> =
+  | SchemaInput<scope>
+  | keyof scope["validTypes"];
```
holic, line 21: chatted last night about potentially removing this

I am fine keeping this here for now but should make sure we consider what we wanna expose/ship with because it'll be harder to do once this is cemented into the API

packages/config/src/library/v2/common.ts
alvrs: nope, removing this

packages/config/src/library/v2/common.ts
alvrs: just a leftover i missed, removing

packages/store/ts/config/v2/generics.ts
alvrs: yeah, it behaves the same way as accessing a non-existent property in js and allows us to handle that case in typeland too (as opposed to sth like `never`)

packages/store/ts/config/v2/schema.ts
alvrs: `abiTypeScope` is just the default value but the scope can include more than just abi types

alvrs: `scope` is basically "all types that are in scope here"

packages/store/ts/config/v2/table.ts
alvrs: for this function i agree, in general i'm in favor of only exporting things that need to be exported so the non-exported things can be refactored more easily (bc we know nothing else depends on it)

holic: yeah makes sense, but I also know from working with viem that exporting intermediate type shapes is helpful when trying to extend/wrap things

but def wanna make sure we're ready to support whatever is exported

packages/store/ts/config/v2/scope.ts
alvrs: renamed to `types`

.changeset/twelve-jars-pull.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/store": minor
+---
+
+Added the type foundation for the new store config. Not all types are included yet and the runtime implementation is still missing. Those will be added in followup PRs.
```
holic, line 5: do we wanna leave out the changeset for now since we're not wanting folks to consume this quite yet?

packages/store/ts/config/v2/table.ts
alvrs: exported most from this file

holic (approved):
cleannn

.changeset/twelve-jars-pull.md
alvrs: yeah could do!

alvrs: removed

PR #2293, opened 2024-02-22
[WIP] add failing test case for sending value to batchCall

alvrs:
Sending `value` to `batchCall` is currently not supported (as the functions are non-payable). If we add `payable` we need to be careful to manually adjust the Balance table to avoid double accounting (currently the balance would be added to the root namespace and every namespace that is being called as part of the batch call)


yonadaaa:
This is awkward to implement because currently, `call` will increase ALL of the system balances based on the msg.value 🤔  we might need to intercept this

```
(bool success, bytes memory returnData) = address(world).delegatecall(
  abi.encodeCall(world.call, (systemId, systemCalls[i].callData))
);
```

Issue #2285, opened 2024-02-20
add salt to WorldDeployed event


holic:
and maybe also the world version?

PR #2301, opened 2024-02-23
feat(world): emit salt in WorldDeployed event

yonadaaa:
Fixes https://github.com/latticexyz/mud/issues/2285


holic (approved):
need a changelog but otherwise lgtm!

Issue #2158, opened 2024-01-18
consider moving errors, events to file level


holic:
Rather than a contract to inherit from or inside libraries, I think we can now define errors at the file level and import them where we need them?

may depend on #2120 

TODO:
- [x] store errors
  - [x] add natspec like world errors
- [x] store events?
- [x] world errors
- [x] world events?
- [x] world modules errors
- [x] standalone libs (FieldLayout etc) errors
  - [x] add natspec like world errors

holic:
We should also rename `SchemaLib_*` errors to `Schema_*` and `FieldLayoutLib_*` errors to `FieldLayout_` for consistency with `Slice_*`, `PackedCounter_*`, `Store_` etc. errors

holic:
we should check that the events, errors are still included in docs if we move these out

yonadaaa:
My takeaway from this exploration - 

There are some errors used by `Store` and `World` that are not included in `IWorld` interface for clients. Ideally, the `IWorld` ABI should include ALL possible errors that may be emitted, so clients can decode failed transactions.

The **events that are included are defined in interfaces** inherited by `IBaseWorld` (for example, `IStore` has `Store_TableNotFound`). However, the **events that are not included are defined internally in libraries** and are only referenced in the contract implementations (for example, `FieldLayoutLib` has `FieldLayoutLib_Empty`).

Ideally these would still be defined in their libraries, but we can bring them into the interface somehow. For example it's nice that Slice errors are in `Slice.sol`:

```

/**
 * @title Static functions for Slice
 * @author MUD (https://mud.dev) by Lattice (https://lattice.xyz)
 */
library SliceLib {
  error Slice_OutOfBounds(bytes data, uint256 start, uint256 end);
  ...
}
```

yonadaaa:
OpenZeppelin use abstract contracts instead of libraries to solve this: https://github.com/ethereum/solidity/issues/13086#issuecomment-1445324173

yonadaaa:
I think we are done here: 
- https://github.com/latticexyz/mud/pull/2356
- https://github.com/latticexyz/mud/pull/2348
- https://github.com/latticexyz/mud/pull/2357
- https://github.com/latticexyz/mud/pull/2358
- https://github.com/latticexyz/mud/pull/2380

holic:
natspec added in https://github.com/latticexyz/mud/pull/2390 and https://github.com/latticexyz/mud/pull/2388

PR #2302, opened 2024-02-23
refactor(store): move events to file level

yonadaaa:
Store events are now declared at the file-level instead of an `IStoreEvents` interface. They are imported individually like:

```solidity
import { Store_SetRecord, Store_SpliceStaticData, Store_SpliceDynamicData } from "../src/StoreEvents.sol";
```


holic:
decided against this approach for now since these don't appear in ABI unless they're part of the contract/interface (see https://github.com/latticexyz/mud/issues/2158)

PR #2303, opened 2024-02-23
chore: upgrade prettier to 3.2.5 and prettier-plugin-solidity to 1.3.1

yonadaa:
Pulled out from https://github.com/latticexyz/mud/pull/2302 (Solidity prettier needs to support events at file level)

PR #2304, opened 2024-02-23
chore(noise): remove noise package

yonadaaa:
We don't use this package so better to stop maintaining it and remove it before stable release.

This was prompted by upgrading prettier in #2303, which caused a build error with `@latticexyz/noise` for some reason (see [failed build](https://github.com/latticexyz/mud/actions/runs/8022570730/job/21917197766#step:6:27)).


alvrs:
Agree with moving it out of the monorepo since it is very standalone and doesn't have any other dependencies in MUD, but [looking at npm stats](https://www.npmjs.com/package/@latticexyz/noise) it is definitely being used so let's keep the code around. Created a new home for it here: https://github.com/latticexyz/noise

.changeset/healthy-items-deliver.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/noise": patch
+---
+
+Removed the @latticexyz/noise package.
```
alvrs, line 5:
```suggestion
Moved the @latticexyz/noise out of the MUD monorepo and into a [new standalone repo](https://github.com/latticexyz/noise).
```

alvrs (approved):

PR #2306, opened 2024-02-23
feat(examples): add external wallet support example

tash-2s:
This PR is created with the exact same commits as https://github.com/latticexyz/mud/pull/2268, but now the source branch is within this latticexyz/mud repository instead of my fork repository. This makes it easier to create other PRs based on these changes.

PR #2294, opened 2024-02-23
WIP: viem custom client within external wallet example

tash-2s:
This pull request is not meant to be merged.

In this PR, I'm experimenting with the viem custom client implementation within the external wallet example project.

---

The target branch is exactly the same as https://github.com/latticexyz/mud/pull/2268. This change will be incorporated into that PR later, but for now, I'm using this PR to easily review the diff.

(Going forward, I'll create branches directly from latticexyz/mud instead of using my fork for PRs so that I can create a PR targeting my other PR within this repository.)

PR #2307, opened 2024-02-23
feat(examples): upgrade wagmi to v2 and implement viem custom client within example

tash-2s:
This PR is created with the exact same commits as https://github.com/latticexyz/mud/pull/2294, but now the source branch is within this latticexyz/mud repository.

---

This PR targets https://github.com/latticexyz/mud/pull/2306, which implements external wallet support using Wagmi v1 in the example project. In this PR, I've updated the code to use Wagmi v2 and the viem custom client. This code doesn't use the `mud_getContract` wrapper, allowing the viem client to be used more straightforwardly. As a result, the structure of the React components has been improved as well.


examples/minimal/packages/client-react-external-wallet/package.json
```diff
@@ -14,12 +14,14 @@
     "@latticexyz/common": "link:../../../../packages/common",
     "@latticexyz/dev-tools": "link:../../../../packages/dev-tools",
     "@latticexyz/store-sync": "link:../../../../packages/store-sync",
+    "@tanstack/react-query": "5.22.2",
```
tash-2s, line 17: Wagmi v2 needs this `@tanstack/react-query` installed alongside.

examples/minimal/packages/client-react-external-wallet/package.json
```diff
@@ -14,12 +14,14 @@
...
     "contracts": "workspace:*",
+    "p-retry": "^5.1.2",
```
tash-2s, line 19: To adapt the transaction sending extension in this example, `p-retry`, which is utilized in the common library, has been installed here. If the direction of this implementation proves to be suitable, I plan to move the custom client implementation to the common library later. Then, I can remove `p-retry` from here.

examples/minimal/packages/client-react-external-wallet/src/ExternalWallet.tsx
tash-2s, line 1: These changes are to adapt Wagmi v2 API changes.

examples/minimal/packages/client-react-external-wallet/src/index.tsx
```diff
@@ -11,17 +11,19 @@ const rootElement = document.getElementById("react-root");
...
+        <ExternalWallet />
+        <MUDNetworkProvider value={network}>
           <App />
           {import.meta.env.DEV && <DevTools />}
-        </MUDWriteProvider>
-      </MUDReadProvider>
-    </WagmiConfig>
+        </MUDNetworkProvider>
+      </QueryClientProvider>
+    </WagmiProvider>
```
tash-2s, line 27: The write context has been removed since we don't have states to hold for that.

examples/minimal/packages/client-react-external-wallet/src/mud/customWalletClient.ts
```diff
@@ -0,0 +1,82 @@
...
+export const useMUD = () => {
+  const network = useMUDNetwork();
+  const { data: connectorWalletClient } = useWalletClient();
+  const { chainId } = useAccount();
+
+  let walletClient;
+  if (network.publicClient.chain.id === chainId && connectorWalletClient?.chain.id === chainId) {
+    // TODO: Should this be memoized?
+    // `walletClient = connectorWalletClient.extend(burnerActions);` is unnecessary for an external wallet
+    walletClient = connectorWalletClient.extend(setupObserverActions(network.onWrite));
```
tash-2s, line 17: Here is the code for creating the custom client.
I've realized that nonce management and retry functionality are unnecessary (though they work) for external wallets. This is because MetaMask ignores the nonce value, and the retry behavior might not be desirable for external wallets.

examples/minimal/packages/client-react-external-wallet/src/mud/customWalletClient.ts
```diff
@@ -0,0 +1,82 @@
...
+
+export type WalletClient = NonNullable<UseWalletClientReturnType["data"]>;
+
+// See @latticexyz/common/src/sendTransaction.ts
+const burnerActions = (client: WalletClient): Pick<WalletActions<Chain, Account>, "sendTransaction"> => {
+  // TODO: Use the `debug` library once this function has been moved to the `common` library.
+  const debug = console.log;
+
+  return {
+    sendTransaction: async (args) => {
```
tash-2s, line 31: This functions similarly to `mud_writeContract` and `mud_sendTransaction` through the viem client interactions, with the exception of transaction simulation. This approach is consistent with the standard viem action behavior, which requires users to initiate the simulation separately when needed.

tash-2s:
This PR is ready for review! 👐 @holic Could you review this when you have time?

examples/minimal/packages/client-react-external-wallet/src/mud/customWalletClient.ts
holic: I think we're still going to ~always have a session wallet as the primary sender of txs (to have ~invisible tx signing) so may make sense to just keep this here.

examples/minimal/packages/client-react-external-wallet/src/mud/customWalletClient.ts
```diff
@@ -0,0 +1,82 @@
...
+        () =>
+          pRetry(
+            async () => {
+              if (!nonceManager.hasNonce()) {
+                await nonceManager.resetNonce();
+              }
+
+              const nonce = nonceManager.nextNonce();
+              debug("sending tx with nonce", nonce, "to", args.to);
+              return sendTransaction(client, { ...args, nonce } as typeof args);
```
holic, line 48: we should prob use `getAction('sendTransaction')` here instead of importing from viem, in case the client has overridden this method already

examples/minimal/packages/client-react-external-wallet/src/mud/customWalletClient.ts
```diff
@@ -0,0 +1,82 @@
...
+    writeContract: async (args) => {
+      const result = writeContract(client, args);
+
+      const id = `${client.chain.id}:${client.account.address}`;
+      onWrite({ id, request: args as WriteContractParameters, result });
+
+      return result;
+    },
+  });
+};
```
holic, line 82: can we pull these actions out into a separate, independent PR that adds them to the common package and import them here?

examples/minimal/packages/client-react-external-wallet/src/mud/setupNetwork.ts
```diff
@@ -24,14 +25,20 @@ export async function setupNetwork() {
...
     worldAddress: networkConfig.worldAddress,
-    mudConfig,
     tables,
     useStore,
     publicClient,
     latestBlock$,
     storedBlockLogs$,
     waitForTransaction,
+    onWrite,
+    write$: write$.asObservable().pipe(share()),
```
holic, line 42: this data flow feels a little weird now, curious if you have ideas for improving this

examples/minimal/packages/client-react-external-wallet/src/mud/supportedChains.ts
```diff
@@ -1,3 +1,12 @@
 import { type MUDChain, mudFoundry } from "@latticexyz/common/chains";
 
-export const supportedChains: MUDChain[] = [mudFoundry];
+export const supportedChains: MUDChain[] = [
+  {
+    ...mudFoundry,
+    // This is not an actual explorer URL; however, Wagmi requires a URL here. Wagmi uses the `blockExplorers` config of this chain to determine the `blockExplorerUrls` value for `wallet_addEthereumChain`. The `blockExplorerUrls` must be either null or an array containing at least one URL; however, Wagmi exclusively employs the array format. Thus, we need to include at least one URL here.
```
holic, line 6: can we pull this out and fix it in the common package instead of here?

examples/minimal/packages/client-react-external-wallet/src/mud/systemCalls.ts
```diff
@@ -0,0 +1,15 @@
...
+export const increment = async (client: WalletClient, network: MUDNetwork) => {
+  const { request } = await network.publicClient.simulateContract({
+    address: network.worldAddress,
+    abi: IWorldAbi,
+    functionName: "increment",
+    args: [],
+    account: client.account,
+  });
+  const tx = await client.writeContract(request);
+  network.waitForTransaction(tx);
```
holic, line 14: would be neat if this was a custom client action too, which means we could simplify the args to potentially just the public client + wallet client

there's a bit of a chicken-and-egg problem though, where functions like `syncToZustand` require the public client and return `waitForTransaction`, which we'd need to extend the public client

maybe we just pass in the plain public client to set up the sync then extend it before returning it to the rest of the app?

```ts
const publicClient = createPublicClient(...);
const syncResult = syncToZustand({ publicClient, ... }) 

return {
  publicClient: publicClient.extend(syncActions(syncResult)),
  ...
}
```

thoughts?

holic (commented):
there's a bunch going on here that we can just upstream into mud directly as improvements that we know we need to make, which will help to simplify the diff of this PR

examples/minimal/packages/client-react-external-wallet/src/mud/customWalletClient.ts
tash-2s: > session wallet

Yes, we should use all custom actions (nonce, retry, write$) for session wallets.

In the upcoming burner wallet PR #2309, all these actions are applied to the burner wallet. When using both external and burner wallets, I believe the external wallet doesn't need to be extended. This is because, we don't require nonce management for MetaMask, and dev-tools, which takes a single wallet client, should be used with the burner wallet.

---

> getAction

Right, using getAction is better since it accounts for more extensions beforehand. I'll switch to getAction.

examples/minimal/packages/client-react-external-wallet/src/mud/customWalletClient.ts
tash-2s: will do! 💪

examples/minimal/packages/client-react-external-wallet/src/mud/supportedChains.ts
tash-2s: will do! 💪

examples/minimal/packages/client-react-external-wallet/src/mud/systemCalls.ts
tash-2s: Having waitForTransaction as a view client action makes sense. I'll look into this.

examples/minimal/packages/client-react-external-wallet/src/mud/setupNetwork.ts
tash-2s: Is it weird? 🤔 I think the app requires a single `write$` for the dev-tool, which can be used for any wallet clients. Hence, it made sense to me to include this in the `setup`.

holic:

Issue #2081, opened 2023-12-23
rethink `sendTransaction`'s internal queue


holic:
I think I have my head wrapped around why our `sendTransaction` can sort of have a "failure cascade" and why this shows up most drastically with optimistic rendering, especially when the next write depends on the previous write.

Our current approach mirrors `sendTransaction` but uses an in-memory nonce counter and parallelizes everything except submitting the prepared tx to the mempool (to ensure nonce ordering). It looks something like:

```
tx1: simulate --> send --> return
tx2: simulate -----------> send --> return
tx3: simulate --------------------> send --> return
```
(the nonce isn't incremented until just before `send`, where we send the simulated/prepared tx to the mempool with the next nonce, to ensure we don't increment the nonce if the simulation fails)

The problem shows up if `tx2` or `tx3` depend on state from `tx1`, where the parallelized simulation step might end up with incorrect results, but we end up submitting it anyway. And our current approach makes it hard to signal that we should re-simulate a given tx and all the ones after it.

To allow for re-simulating a tx after a revert and to avoid doing too much unnecessary work ahead of time, we can restructure the queue to always have the next tx (or two?) in the queue simulated before sending to the mempool, and be able to retry the simulation in case of failures:

```
tx1: simulate --> send --> return
tx2:              simulate --> send --> return
tx3:                           simulate --> send --> return
```

If a `send` call to the mempool reverts, a retry might look like:

```
tx1: simulate --> send --> return
tx2:              simulate --> send --> revert --> simulate --> send --> return
tx3:                           simulate ----------------------> simulate --> send --> return
```

<details>
<summary>

~~If I recall, the `simulate` call is there to ensure we can safely increment the nonce and to help us get nicer errors from reverts (where a `send` doesn't give us this data), so we could also flip this around to only simulate on a revert or increment nonce on success:~~

</summary>

```
tx1: send --> nonce++ and return
tx2:          send --> revert --> send --> revert --> simulate and return error
tx3:                                                  send --> nonce++ and return
```
(the above assumes we only retry a reverted `send` call once)

There's also an open question around whether we should flush the queue if there are consistent reverts after retries, where `tx3` in the above example wouldn't execute at all.

</details>

(Edit: Nevermind, we should still simulate to determine if a tx will revert, because the mem pool will accept a valid tx even if it's expected to revert and then we'd 1) pay gas for the revert and 2) have to wait until the tx is mined before determine its success, negating this benefits of this queue)

holic:
[...]

PR #2309, opened 2024-02-25
feat(examples): support external wallet delegating to burner wallet within example

tash-2s:
This PR targets https://github.com/latticexyz/mud/pull/2307, which implements external wallet support in the example project. In this PR, external wallets delegate to local storage burner wallets, and the burner wallets send transactions on behalf of the external wallets.

I created this as a separate PR to see the diff required for the change. However, please feel free to merge this with the target PR if it simplifies the review process or if it is ready for merging.

### DIFF
--- a/examples/minimal/packages/client-react-external-wallet/src/mud/customClient.ts
+++ b/examples/minimal/packages/client-react-external-wallet/src/mud/customClient.ts
@@ -12,6 +12,7 @@ import type {
 } from "viem";
 import { sendTransaction, writeContract, simulateContract } from "viem/actions";
 import pRetry from "p-retry";
+import { type Subject } from "rxjs";
 import { getNonceManager, type ContractWrite } from "@latticexyz/common";
 import { type SyncResult } from "@latticexyz/store-sync";
 
@@ -77,7 +78,9 @@ export const burnerActions = <transport extends Transport, chain extends Chain, 
 };
 
 // See @latticexyz/common/src/getContract.ts
-export const setupObserverActions = (onWrite: (write: ContractWrite) => void) => {
+export const setupObserverActions = (write$: Subject<ContractWrite>) => {
+  let nextWriteId = 0;
+
   return <transport extends Transport, chain extends Chain, account extends Account>(
     client: WalletClient<transport, chain, account>
   ): Pick<WalletActions<chain, account>, "writeContract"> => ({
@@ -85,8 +88,8 @@ export const setupObserverActions = (onWrite: (write: ContractWrite) => void) =>
     writeContract: async (args) => {
       const result = writeContract(client, args);
 
-      const id = `${client.chain.id}:${client.account.address}`;
-      onWrite({ id, request: args as WriteContractParameters, result });
+      const id = `${client.chain.id}:${client.account.address}:${nextWriteId++}`;
+      write$.next({ id, request: args as WriteContractParameters, result });
 
       return result;
     },
--- a/examples/minimal/packages/client-react-external-wallet/src/mud/customClient.ts
+++ b/examples/minimal/packages/client-react-external-wallet/src/mud/customClient.ts
@@ -13,6 +13,7 @@ import type {
 import { sendTransaction, writeContract, simulateContract } from "viem/actions";
 import pRetry from "p-retry";
 import { getNonceManager, type ContractWrite } from "@latticexyz/common";
+import { type SyncResult } from "@latticexyz/store-sync";
 
 const burnerBlockTag = "pending";
 
@@ -89,5 +90,11 @@ export const setupObserverActions = (onWrite: (write: ContractWrite) => void) =>
 
       return result;
     },
+  });
+};
+
+export const setupStoreSyncActions = (syncResult: SyncResult) => {
+  return () => ({
+    waitForStoreSync: syncResult.waitForTransaction,
   });
 };
--- a/examples/minimal/packages/client-react-external-wallet/src/mud/customClient.ts
+++ b/examples/minimal/packages/client-react-external-wallet/src/mud/customClient.ts
@@ -1,21 +1,44 @@
-import type { WriteContractParameters, Transport, Chain, Account, WalletActions, WalletClient } from "viem";
-import { sendTransaction, writeContract } from "viem/actions";
+import type {
+  WriteContractParameters,
+  Transport,
+  Chain,
+  Account,
+  Client,
+  WalletRpcSchema,
+  WalletActions,
+  PublicRpcSchema,
+  PublicActions,
+  WalletClient,
+} from "viem";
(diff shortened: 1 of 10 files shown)


tash-2s:
This is how it looks:

https://github.com/latticexyz/mud/assets/81064017/8194e573-3ae8-44ff-a1ba-625edb53f620

examples/minimal/packages/client-react-external-wallet/src/mud/setupNetwork.ts
```diff
@@ -30,6 +32,13 @@ export async function setupNetwork() {
   const onWrite = (write: ContractWrite) =>
     write$.next({ id: `${write.id}:${nextWriteId++}`, request: write.request, result: write.result });
 
+  const burnerWalletClient = createWalletClient({
+    ...clientOptions,
+    account: privateKeyToAccount(getBurnerPrivateKey("mud:example:burnerWallet")),
+  })
+    .extend(burnerActions)
+    .extend(setupObserverActions(onWrite));
```
tash-2s, line 40: I think creating a burner wallet at the very beginning like this is the easiest approach to implement. However, generating burner wallets for users without external wallets seems unnecessary, making it less than ideal. If I decide to create burner wallets only when needed, I'm considering introducing a new React context and state for this purpose, which would allow easy access to burner wallets for underlying components.

tash-2s:
I can think of additional nice-to-have features, such as unregistering delegation and sending eth back to external wallets, but I think the current code sufficiently demonstrates the burner delegation.

examples/minimal/packages/client-react-external-wallet/src/mud/systemCalls.ts
```diff
@@ -1,15 +1,53 @@
...
+import { type MUDNetwork, type ExternalWalletClient } from "./NetworkContext";
 import IWorldAbi from "contracts/out/IWorld.sol/IWorld.abi.json";
 
-export const increment = async (client: WalletClient, network: MUDNetwork) => {
+// Executes the `increment` system call on behalf of the external wallet using the burner wallet.
+// This is solely for demonstrating burner wallet delegation, as the `increment` call can actually be made by any sender.
+export const increment = async (externalWalletClient: ExternalWalletClient, network: MUDNetwork) => {
   const { request } = await network.publicClient.simulateContract({
+    account: network.burnerWalletClient.account,
+    blockTag: "pending", // burner uses this tag
```
tash-2s, line 12: I've realized that manually specifying blockTag here is cumbersome. I'll move this to the burner actions.

tash-2s: This has been fixed.

tash-2s:
@holic This should correctly follow what mud_writeContract does, but what is the best way to test and see the nonce manager is functioning as expected?

examples/minimal/packages/client-react-external-wallet/src/mud/setupNetwork.ts
```diff
@@ -18,27 +30,28 @@ export async function setupNetwork() {
...
-  const onWrite = (write: ContractWrite) =>
-    write$.next({ id: `${write.id}:${nextWriteId++}`, request: write.request, result: write.result });
+
+  const burnerClient = createWalletClient({
+    ...clientOptions,
+    account: privateKeyToAccount(getBurnerPrivateKey("mud:example:burnerWallet")),
+  })
+    .extend(publicActions)
+    .extend(burnerActions)
+    .extend(setupObserverActions(write$));
```
tash-2s, line 48: write$ and the burnerClient now look like this.

In this burner wallet setup, it's natural to have write$ here. If we don't have a wallet client at this setup phase, we can create write$ when getting/creating the wallet client, and it should be kept as long as the same wallet client is in use.

I thought it's nice to update dev-tools like these, although they may not be immediately necessary:

- Allow changing the wallet client, or make it easier to unmount it.
- Make the wallet client optional. Even without it, it's still really useful.

examples/minimal/packages/client-react-external-wallet/src/mud/customClient.ts
```diff
@@ -0,0 +1,103 @@
...
+// See @latticexyz/common/src/getContract.ts
+export const setupObserverActions = (write$: Subject<ContractWrite>) => {
+  let nextWriteId = 0;
+
+  return <transport extends Transport, chain extends Chain, account extends Account>(
+    client: WalletClient<transport, chain, account>
+  ): Pick<WalletActions<chain, account>, "writeContract"> => ({
+    // Applies to: `client.writeContract`, `getContract(client, ...).write`
+    writeContract: async (args) => {
+      const result = writeContract(client, args);
```
tash-2s, line 89: It turned out I couldn't use `getAction` here. This is because if I use `getAction(client, writeContract, "writeContract")(args)`, it uses the normal `walletActions`' `writeContract`, which in turn uses the normal `sendTransaction`. As a result, `burnerActions`' `sendTransaction` isn't invoked. When extending a client, the `client` argument in the action function is captured at that point, making `getAction` unsuitable for this use case.

https://github.com/wevm/viem/blob/viem%402.7.12/src/clients/decorators/wallet.ts#L684-L710

examples/minimal/packages/client-react-external-wallet/src/mud/systemCalls.ts
```diff
@@ -1,15 +1,63 @@
...
+) => {
+  const { request } = await publicClient.simulateContract({
+    account: externalWalletClient.account,
+    address: worldAddress,
+    abi: IWorldAbi,
+    functionName: "registerDelegation",
+    args: [burnerClient.account.address, UNLIMITED_DELEGATION, "0x0"],
+  });
+
+  const tx1 = await externalWalletClient.writeContract(request);
```
tash-2s, line 54: I think we shouldn't include a transaction simulation within the standard viem client action so that users can have control. If we want to do something advanced like #2081, we can build the functionality outside the standard viem action.

tash-2s (commented):
I've addressed all comments added in #2307, except for pulling out some code outside, which I'll do in separate PRs. Could you take a look?

examples/minimal/packages/client-react-external-wallet/src/mud/customClient.ts
holic: internally, viem's `writeContract` does call `getAction('sendTransaction')`:
https://github.com/wevm/viem/blob/main/src/actions/wallet/writeContract.ts#L163-L167

so I am wondering if I am misunderstanding the intent

tash-2s: Thanks for the comment! I realized that we first need to separate the custom client to move forward, so I’ll address this comment in the independent custom client PR, which I’ll be creating shortly.

holic:

PR #2311, opened 2024-02-26
refactor(store): errors at file-level

yonadaaa:
Exploring defining errors at the file-level. See https://github.com/latticexyz/mud/pull/2302 for the corresponding refactor for events.


packages/store/src/StoreErrors.sol
holic, line 1: can we call this just `errors.sol`? (intentionally lower case, since this doesn't export a `StoreErrors` contract/struct)

packages/store/src/StoreErrors.sol
```diff
@@ -0,0 +1,19 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.24;
+
+import { ResourceId } from "./ResourceId.sol";
+
+// Errors include a stringified version of the tableId for easier debugging if cleartext tableIds are used
```
holic, line 6: I know this wasn't part of your PR, but thought we could improve/clarify this one:

```suggestion
// Errors include a human readable resource/table ID for better logging and debugging
```

packages/store/test/StoreCore.t.sol
```diff
@@ -11,7 +11,7 @@ import { FieldLayout, FieldLayoutLib } from "../src/FieldLayout.sol";
 import { Schema } from "../src/Schema.sol";
 import { PackedCounter, PackedCounterLib } from "../src/PackedCounter.sol";
 import { StoreMock } from "../test/StoreMock.sol";
-import { IStoreErrors } from "../src/IStoreErrors.sol";
+import { Store_TableAlreadyExists, Store_InvalidSplice, Store_IndexOutOfBounds, Store_InvalidBounds, Store_TableNotFound, Store_InvalidStaticDataLength, Store_InvalidValueSchemaStaticLength, Store_InvalidResourceType, Store_InvalidValueSchemaDynamicLength, Store_InvalidFieldNamesLength, Store_InvalidKeyNamesLength, Store_InvalidValueSchemaLength } from "../src/StoreErrors.sol";
```
holic, line 14: is it important to be explicit about imports when we're in a test?

wondering about:
```solidity
import "../src/StoreErrors.sol"
```

holic (approved):
this seems like a much clearer win to me than file-level events

just wanna rename the file but otherwise lgtm!

don't forget a changeset

packages/store/src/StoreErrors.sol
yonadaaa: Yes, I like this pattern for constants and other free stuff

packages/store/test/StoreCore.t.sol
yonadaaa: Good shout, would clean up all these imports!

.changeset/pink-pandas-fetch.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/store": patch
```
holic, line 2:
```suggestion
"@latticexyz/store": minor
```
prob actually breaking but 🤷

.changeset/pink-pandas-fetch.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Removed the `IStoreErrors` interface, instead Store errors are defined at the file-level and imported individually.
```
holic, line 5:
```suggestion
Moved Solidity custom errors (e.g. `Store_TableNotFound`) from `IStoreErrors` interface to file-level errors in `errors.sol`. If you were using these errors before, you may need to update your imports.
```

packages/store/src/errors.sol
```diff
@@ -0,0 +1,19 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.24;
+
+import { ResourceId } from "./ResourceId.sol";
+
+// Errors include a human readable resource/table ID for better logging and debugging
```
holic, line 6:
```suggestion
// Errors include a human-readable resource/table ID string for better logging and debugging
```

holic (approved):

Issue #2106, opened 2024-01-10
add comments for `prohibitDirectCallback` modifier and why it should be used


holic:
see https://github.com/latticexyz/mud/pull/2091#discussion_r1447181668

PR #2312, opened 2024-02-26
docs(world): add comments for prohibitDirectCallback modifier

yonadaaa:
Closes https://github.com/latticexyz/mud/issues/2106 by explaining why we use the `prohibitDirectCallback` modifier.

The comment is a rewording of the PR description from the original PR that added it https://github.com/latticexyz/mud/pull/1563

### DIFF
--- a/docs/pages/world/reference/world.mdx
+++ b/docs/pages/world/reference/world.mdx
@@ -45,9 +45,9 @@ constructor();
 #### prohibitDirectCallback
 
 _Prevents the World contract from calling itself.
-The world is not able to call itself; all operations to internal tables happen as internal library calls, and all calls to root system happen as a delegatecall to the system.
-However, since this is an important invariant, we make it explicit by reverting if `msg.sender` is `address(this)` in all `World` methods.
-If it was possible to make the `World` call itself, it would be possible to access internal tables that only the `World` should have access to._
+If the World is able to call itself via delegatecall from a system, the system would have root access to context like internal tables, causing a potential vulnerability.
+Ideally this should not happen because all operations to internal tables happen as internal library calls, and all calls to root systems happen as a `delegatecall` to the system.
+However, since this is an important invariant, we make it explicit by reverting if `msg.sender` is `address(this)` in all `World` methods._
 
 ```solidity
 modifier prohibitDirectCallback();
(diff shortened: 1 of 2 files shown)


packages/world/src/World.sol
```diff
@@ -51,6 +51,9 @@ contract World is StoreData, IWorldKernel {
 
   /**
    * @dev Prevents the World contract from calling itself.
+   * The world is not able to call itself; all operations to internal tables happen as internal library calls, and all calls to root system happen as a delegatecall to the system.
```
holic, line 54:
```suggestion
   * The world is not able to call itself; all operations to internal tables happen as internal library calls, and all calls to root systems happen as a delegatecall to the system.
```

packages/world/src/World.sol
```diff
@@ -51,6 +51,9 @@ contract World is StoreData, IWorldKernel {
...
+   * However, since this is an important invariant, we make it explicit by reverting if `msg.sender` is `address(this)` in all `World` methods.
+   * If it was possible to make the `World` call itself, it would be possible to access internal tables that only the `World` should have access to.
```
holic, line 56: this feels a bit incomplete, like we should either expand on how/why or just remove

> If the `World` is able to call itself via `delegatecall` from a system, this would be a backdoor where that system would have "root" access to things like internal tables.

(maybe could workshop this exact sentence a bit more, but just providing a bit more context)

docs/pages/world/reference/world.mdx
```diff
@@ -44,7 +44,10 @@ constructor();
 
 #### prohibitDirectCallback
 
-_Prevents the World contract from calling itself._
+_Prevents the World contract from calling itself.
+If the World is able to call itself via delegatecall from a system, the system would have root access to context like internal tables, causing a potential vulnerability.
```
holic, line 48:
```suggestion
If the World is able to call itself via `delegatecall` from a system, the system would have root access to context like internal tables, causing a potential vulnerability.
```

packages/world/src/World.sol
```diff
@@ -51,6 +51,9 @@ contract World is StoreData, IWorldKernel {
 
   /**
    * @dev Prevents the World contract from calling itself.
+   * If the World is able to call itself via delegatecall from a system, the system would have root access to context like internal tables, causing a potential vulnerability.
```
holic, line 54:
```suggestion
   * If the World is able to call itself via `delegatecall` from a system, the system would have root access to context like internal tables, causing a potential vulnerability.
```

PR #2315, opened 2024-02-27
feat(store-sync,store-indexer): add followBlockTag option

holic:
- adds `followBlockTag` option, defaults to `latest` (current behavior)
- uses `followBlockTag: safe` for indexers

### DIFF
--- a/.changeset/brave-ghosts-walk.md
+++ b/.changeset/brave-ghosts-walk.md
@@ -5,4 +5,4 @@
 
 Added a `followBlockTag` option to configure which block number to follow when running `createStoreSync`. It defaults to `latest` (current behavior), which is recommended for individual clients so that you always have the latest chain state.
 
-Indexers now default to `safe` to avoid issues with reorgs and load-balanced RPCs being out of sync. This means indexers will be slightly behind the latest block number, but clients can quickly catch up.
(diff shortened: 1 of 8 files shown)


alvrs (approved):

alvrs (dismissed):

.changeset/brave-ghosts-walk.md
```diff
@@ -0,0 +1,8 @@
+---
+"@latticexyz/store-indexer": minor
+"@latticexyz/store-sync": minor
+---
+
+Added a `followBlockTag` option to configure which block number to follow when running `createStoreSync`. It defaults to `latest` (current behavior), which is recommended for individual clients so that you always have the latest chain state.
```
yonadaaa, line 6: It sticks out to me that clients and indexers have a different default setting. What might I run into if I use `latest` on the client?

holic: Client uses `latest` by default, indexers use `safe` by default.

This just means the indexer is ~32 blocks behind (or whatever the safe setting is for a given chain). So you'll get a snapshot that is slightly behind, then the client fetches from RPC to catch up.

This helps us sidestep both reorg issues and load balanced RPC issues.

alvrs: Might miss events if the RPC that notifies you of the new block number is ahead of the RPC you request logs from. We can fix this in a followup PR by doing a batch RPC call to check the latest block number of the RPC you're requesting logs from and retry if it doesn't have the latest number yet.

holic: see https://github.com/latticexyz/mud/issues/2316

alvrs: #2316

Issue #2236, opened 2024-02-06
Add RPC call concurrency restrictions to deploy command options


hourglasshoro:
Errors often occur in RPC call quotes while executing `mud deploy` command, so I want to limit the number of calls at a time to suppress errors.

For example
- Option: `--rpcConcurrency`
- Meaning: Limit the concurrency of RPC calls
- Type: number
- Default value: 1000

https://mud.dev/cli/deploy

holic:
great suggestion!

prob best to do this at the viem client level, maybe with a custom transport, and could lean on something like `p-queue` for it

hourglasshoro:
Thanks, it does indeed look like a good idea to use custom transport. I already have one implemented in a naive way, can I submit it to PR once I have it? If it is not a good implementation, I will re-implement it with custom transport.

holic:
Thanks for taking a stab at this! And sorry to jostle you around, but I do think a custom transport with throttling built in would be better and general purpose for ~any viem user!

hourglasshoro:
Understood.
I made it once because I needed it right at hand, but I'll fix the PR later when I have more time👍

holic:
@hourglasshoro A question for you: https://twitter.com/indexsupply/status/1754924224375095349

PR #2322, opened 2024-02-27
feat(cli): add a RPC batch option to cli

hourglasshoro:
This PR adds a RPC batch size option to the deploy command.

Resolve: https://github.com/latticexyz/mud/issues/2236


hourglasshoro:
@holic 

> https://github.com/latticexyz/mud/issues/2236#issuecomment-1929947276

There was no corresponding throttling, but there was one that did batch processing instead, so this was substituted. 
https://viem.sh/docs/clients/transports/http#batch-json-rpc

In my environment, setting batchSize to 1000 allowed me to deploy without any errors.

I have closed PRs I made in the past and made new ones
https://github.com/latticexyz/mud/pull/2322

holic:
Out of curiosity, does `batch: true` do the same thing for you?

hourglasshoro:
I don't think just `batch:true` will work, because if you don't specify `batch.wait`, the batch control will not work for 1 second(default wait 0 second).
https://viem.sh/docs/clients/transports/http#batchwait-optional

hourglasshoro:
Do you usually update the document side with the same PR?

holic:
If you read further up, it'll still batch within the same "tick" (see also [zero delay](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Event_loop#zero_delays)):

image

This makes me suspect most of our operations would be batched because they're often sent simultaneously.

holic:
> Do you usually update the document side with the same PR?

Yep!

hourglasshoro:
The same effect as throttling could only be achieved if a sufficiently large batchSize and a sufficiently long wait were set

My hypothesis is this.
For example, each `ensureHoge` calls `Promise.all` internally, and `Promise.all` does multiple etc_calls in one step
https://github.com/latticexyz/mud/blob/main/packages/cli/src/deploy/deploy.ts#L93-L114

1) If batchSize is small, more batches are split, more requests are sent per second, and quote is limited.
2) If the wait time is short, requests are sent at short intervals even if batchSize is large, and quote is limited.

I am not sure what should be taken as the argument of cli

hourglasshoro:
I'm guessing based on the actual results I've tried, so I could be wrong.

hourglasshoro:
Of course, I've tried `batch: true` and still hit RPC rate limits

Since batchSize and wait are defaults when batchTrue, respectively
batchSize: [1000](https://viem.sh/docs/clients/transports/http#batchbatchsize-optional), wait: [0(ms)](https://viem.sh/docs/clients/transports/http#batchwait-optional)

My environment is this. 
- 1 system
- 20 tables
- 1 module
- RPC Endpoint: Alchemy Growth Plan (https://www.alchemy.com/pricing)
---

Here are the results of the search with each parameter fixed
- batchSize: 10, wait: 1000(ms): hit RPC rate limits
- batchSize: 100, wait: 1000(ms): no error
- batchSize: 100, wait: 0(ms): hit RPC rate limits
- batchSize: 100, wait: 10(ms): hit RPC rate limits
- batchSize: 100, wait: 100(ms): hit RPC rate limits

holic:
asked if we could enable this by default but sounds like there's no good way to detect if batching is supported so keeping it as an option makes sense: https://discord.com/channels/1156791276818157609/1212120089219043369

holic (approved):

PR #2324, opened 2024-02-27
docs(world/modules): add code sample for module installation

qbzzt:


qbzzt:
No longer relevant in light of `mud.config.ts` changes.

PR #2325, opened 2024-02-27
fix(cli): throw warning for duplicate systems

### LINKED ISSUE #2263: Throw warning when systems have same prefixes
When two system names have identical prefixes up to index 16, there is an occasional bug in the deploy script where one of the systems doesn't register properly to the world. The revert thrown when calling the system excludes a message and the bug occurs at different rates depending on the environment (20-50%).

vmaark:
Resolves #2263 

Throw error when systems with the same ids occur.

### DIFF
--- a/packages/cli/src/deploy/resolveConfig.ts
+++ b/packages/cli/src/deploy/resolveConfig.ts
@@ -70,6 +70,24 @@ export function resolveConfig<config extends ConfigInput>({
     };
   });
 
+  systems.forEach((item, index) => {
+    let duplicateName = "";
+    if (
+      systems.some((elem, idx) => {
+        const found = elem.systemId === item.systemId && idx !== index;
+        if (found) {
+          duplicateName = elem.name;
+        }
+        return found;
+      })
+    ) {
+      throw new Error(`Found two systems with the same system ID: ${item.name} ${duplicateName}. 
+      
+      System IDs are generated from the first 14 bytes of the name, so you may need to rename one of these systems to avoid the overlap.
+      `);
+    }
+  });
+
   // ugh (https://github.com/latticexyz/mud/issues/1668)
   const resolveContext = {
     tableIds: Object.fromEntries(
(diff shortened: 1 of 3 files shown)


packages/cli/src/deploy/resolveConfig.ts
```diff
@@ -79,6 +79,12 @@ export function resolveConfig<config extends ConfigInput>({
     };
   });
 
+  systems.forEach((item, index) => {
+    if (systems.some((elem, idx) => elem.systemId === item.systemId && idx !== index)) {
+      throw new Error(`Duplicate system ID ${item.systemId} found in the config.`);
+    }
+  });
```
holic, line 86: Since we are working with objects, we could probably simplify this a bit by just checking if each `item` and `elem` are not the same (rather than their indices).

I think we could improve this error message too, to something like:

> Found two systems with the same system ID: {name1}, {name2}
>
> System IDs are generated from the first 14 bytes of the name, so you may need to rename one of these systems to avoid the overlap.

holic:
Looks like there's a merge conflict, which is keeping CI from running.

packages/cli/src/deploy/resolveConfig.ts
vmaark: improved the error message. 
I think it's safer to use the idx in the context of objects, a change in the logic could easily introduce a bug where you'd be checking e.q. the equality of different objects but with the same id (and they would never equal on the object reference level)

holic: in this case, we're just trying to make sure we're comparing system IDs, but we don't want to compare to itself, so an object reference

holic: re: above code, could potentially clarify with

```
const overlappingSystem = systems.find(...);
if (overlappingSystem) {
  throw new Error(`Found two systems with the same system ID: ${system.name} ${overlappingSystem.name}.\n\nSystem IDs are generated from the first 14 bytes of the name, so you may need to rename one of these systems to avoid the overlap.`);
}
```

PR #2327, opened 2024-02-27
refactor: move table ID and field layout constants into table library

holic:
fixes https://github.com/latticexyz/mud/issues/2166

no gas changes but feels cleaner


yonadaaa (approved):
nice! interested to see this with the hardcoded schema too

PR #2328, opened 2024-02-27
refactor: hardcode key/value schema in table libraries

holic:
similar to https://github.com/latticexyz/mud/issues/2166

looks like decent gas savings but only on register/install methods and some bytecode savings when using table libs


e2e/packages/contracts/src/codegen/tables/Multi.sol
```diff
@@ -46,25 +46,15 @@ library Multi {
...
    */
   function getKeySchema() internal pure returns (Schema) {
-    SchemaType[] memory _keySchema = new SchemaType[](4);
-    _keySchema[0] = SchemaType.UINT32;
-    _keySchema[1] = SchemaType.BOOL;
-    _keySchema[2] = SchemaType.UINT256;
-    _keySchema[3] = SchemaType.INT120;
-
-    return SchemaLib.encode(_keySchema);
+    return Schema.wrap(0x0034040003601f2e000000000000000000000000000000000000000000000000);
```
yonadaaa, line 49: Neat! Is there any gas impact to declaring these as constants, ie. `Schema constant KEY_SCHEMA = ...`?

Wondering if we can go even further and remove the functions (`getKeySchema`), leaving only the constants. Seems more fitting if all the functions do is return a hardcoded value.

yonadaaa (commented):
I like this - the more logic we have in TS, the better.

e2e/packages/contracts/src/codegen/tables/Multi.sol
holic: yep, that'd be the next step (see OP) but wanted to wait for the other PR to land first

packages/world-modules/src/modules/erc20-puppet/tables/ERC20Metadata.sol
```diff
@@ -29,28 +26,10 @@ library ERC20Metadata {
...
-   */
-  function getValueSchema() internal pure returns (Schema) {
-    SchemaType[] memory _valueSchema = new SchemaType[](3);
-    _valueSchema[0] = SchemaType.UINT8;
-    _valueSchema[1] = SchemaType.STRING;
-    _valueSchema[2] = SchemaType.STRING;
-
-    return SchemaLib.encode(_valueSchema);
-  }
+  // Hex-encoded key schema of ()
```
yonadaaa, line 29: kinda cursed but what can ya do

packages/world-modules/src/modules/erc20-puppet/tables/ERC20Metadata.sol
```diff
@@ -29,28 +26,10 @@ library ERC20Metadata {
   FieldLayout constant _fieldLayout =
```
yonadaaa, line 26: Would be interesting to have a similar `// Hex-encoded ....` comment explaining how the field layout is generated!

yonadaaa (changes requested):
What about a changeset? Seems like a breaking change to remove `getKeySchema`

packages/world-modules/src/modules/erc20-puppet/tables/ERC20Metadata.sol
holic: it's accurate at least!

holic: I could add a conditional for this but seems messy

packages/world-modules/src/modules/erc20-puppet/tables/ERC20Metadata.sol
holic: I considered that but didn't have an obvious way to generate a human-readable comment for it so I left it (it never had one)

packages/store/ts/codegen/renderTable.ts
```diff
@@ -80,27 +78,14 @@ export function renderTable(options: RenderTableOptions) {
...
-        return SchemaLib.encode(_valueSchema);
-      }
+      // Hex-encoded key schema of (${keyTuple.map((field) => field.internalTypeId).join(", ")})
+      Schema constant _keySchema = Schema.wrap(${keySchemaToHex(
+        Object.fromEntries(keyTuple.map((field) => [field.name, field.internalTypeId])) as KeySchema
+      )});
+      // Hex-encoded value schema of (${fields.map((field) => field.internalTypeId).join(", ")})
+      Schema constant _valueSchema = Schema.wrap(${valueSchemaToHex(
+        Object.fromEntries(fields.map((field) => [field.name, field.internalTypeId])) as ValueSchema
+      )});
```
alvrs, line 88: I generally agree with this change but am a little wary of this being a change to an audited code path. Are there any tests we can add to really make sure the typescript encoding matches the expected encoding? maybe a couple forge test cases where we encode with the new method and confirm it matches what would be produced by `SchemaLib.encode`?

holic: how about this? https://github.com/latticexyz/mud/pull/2328/commits/39fa54244aa141416953e6a32e698a34aa7854a9

alvrs: Perfect!

packages/world/gas-report.json
```diff
@@ -93,7 +93,7 @@
     "file": "test/World.t.sol",
     "test": "testDeleteRecord",
     "name": "Delete record",
-    "gasUsed": 8957
+    "gasUsed": 8958
```
yonadaaa, line 96: 🙀

yonadaaa (approved):

PR #2330, opened 2024-02-28
feat(common): add `blockExplorers` field in mudFoundry for Wagmi

tash-2s:
Pulling out from https://github.com/latticexyz/mud/pull/2309.

Without this, we encounter the error below when adding the mudFoundry chain into external wallets using Wagmi. Ideally, this issue should be addressed in Wagmi. I might attempt to fix it later.

```
MetaMask - RPC Error: Expected null or array with at least one valid string HTTPS URL 'blockExplorerUrl'.
```

### DIFF
--- a/packages/common/src/chains/mudFoundry.ts
+++ b/packages/common/src/chains/mudFoundry.ts
@@ -7,4 +7,12 @@ export const mudFoundry = {
     // This is intentionally defined as a function as a workaround for https://github.com/wagmi-dev/viem/pull/1280
     defaultPriorityFee: () => 0n,
   },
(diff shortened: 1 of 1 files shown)


holic:
I couldn't find anything in the wagmi repo about this and I'm surprised we're the first to run into it. I also couldn't find any types that would expect/enforce at least one block explorer URL. 🤔

tash-2s:
I've made an issue on Wagmi https://github.com/wevm/wagmi/issues/3652

tash-2s:
I've fixed this issue upstream. https://github.com/wevm/wagmi/pull/3653

Issue #631, opened 2023-04-15
Add support for inheritance to `worldgen`


alvrs:

holic:
Adding this to stable release milestone so we can consider if we want to include this.

PR #2332, opened 2024-02-28
chore(world): generate init system function signatures

yonadaaa:


yonadaaa:
This issue is a little difficult because `RegistrationSystem` inherits from multiple Systems, which is not supported yet https://github.com/latticexyz/mud/issues/631

PR #2334, opened 2024-02-28
refactor(store,world-modules): clean up error names/locations

holic:
part of #2158


yonadaaa (approved):

packages/store/src/errors.sol
```diff
@@ -17,3 +17,6 @@ error Store_InvalidValueSchemaLength(uint256 expected, uint256 received);
 error Store_InvalidValueSchemaStaticLength(uint256 expected, uint256 received);
 error Store_InvalidValueSchemaDynamicLength(uint256 expected, uint256 received);
 error Store_InvalidSplice(uint40 startWithinField, uint40 deleteCount, uint40 fieldLength);
+
+/// @notice Error emitted when a function is not implemented.
+error StoreHook_NotImplemented();
```
holic, line 22: kinda undecided on this being here

also considered making this a generic `error NotImplemented()` but wasn't sure if that would be clear enough when the error is thrown somewhere in code

yonadaaa: I prefer the specific error name, but would not be against it being defined at the top of `StoreHook`

PR #2340, opened 2024-02-28
chore: test workflow runs with ubuntu and windows OS

yonadaaa:
Fixes https://github.com/latticexyz/mud/issues/2197

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,10 +29,3 @@ jobs:
 
       - name: Outdated files, run `pnpm build` and commit them
         uses: ./.github/actions/require-empty-diff
-
-      - name: Build API docs
-        shell: bash
-        run: pnpm docs:generate:api
-
-      - name: Outdated files, run `pnpm docs:generate:api` and commit them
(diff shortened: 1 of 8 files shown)


yonadaaa:
Running with `windows-latest` is failing:

https://github.com/latticexyz/mud/actions/runs/8082956911/job/22084918032?pr=2340

This is because we use a container for Postgres. Containers are not supported: 

https://github.com/actions/runner/issues/904#issuecomment-760484077

yonadaaa:
The GitHub action does it's job, but we found the MUD monorepo has some issues with Windows:

- in `prepare` we use `mkdir -p`, which doesn't work for Windows. We could use [mkdirp](https://www.npmjs.com/package/mkdirp) instead
- we use Bun to generate docs, [which doesn't support Windows](https://github.com/oven-sh/bun/issues/43#issuecomment-1951349647)
- the `copy-templates.sh` bash script relies on Unix commands https://github.com/latticexyz/mud/issues/2344

holic:
replaced by https://github.com/latticexyz/mud/pull/3078

PR #2347, opened 2024-02-28
feat(common): add viem actions that work the same as the current wrappers

tash-2s:
This code is pulled out from https://github.com/latticexyz/mud/pull/2309 and got some modifications.

This PR introduces viem client actions into the common library. These actions work the same as MUD's `getContract`, `writeContract`, and `sendTransaction`, but this enables the use of the functions through client (e.g., `client.writeContract(...)`).

We've decided to include the simulation code in these overridden viem actions, so now the same code as the wrappers is used. If we decide to deprecate the wrapper functions, I can reorganize the code for these actions. (While `getContract` might be deprecated, do we want to use the `writeContract` and `sendTransaction` wrappers independently?)

Including simulation in transaction sending actions is especially necessary for [further enhancements](https://github.com/latticexyz/mud/issues/2081) that require precise control over simulation.

Additionally, I've updated a template for testing/demonstration purposes, which can be removed prior to merging.


templates/react/packages/client/src/mud/setupNetwork.ts
```diff
@@ -3,12 +3,28 @@
...
+  createPublicClient,
+  fallback,
+  webSocket,
+  http,
+  createWalletClient,
+  Hex,
+  parseEther,
+  ClientConfig,
+  getContract,
+} from "viem";
```
tash-2s, line 16: Now this uses `viem_getContract` instead of `mud_getContract`.

With this custom client setup, there's an opportunity to rethink this template structure (e.g., not doing `getContract` at the setup phase), but I've not explored it here.

packages/common/src/viemActions.ts
```diff
@@ -0,0 +1,43 @@
...
+  WalletClient,
+  WriteContractReturnType,
+} from "viem";
+import { getAction } from "viem/utils";
+import { writeContract } from "viem/actions";
+import { writeContract as mud_writeContract } from "./writeContract";
+import { sendTransaction as mud_sendTransaction } from "./sendTransaction";
+import { type ContractWrite } from "./getContract";
+
+export const sessionActions = <TChain extends Chain, TAccount extends Account>(
```
holic, line 16: I tend to prefer one definition/export per file when we've got functions like this, as it makes it easier to navigate around

if we feel like we want more granularity, we can drop all these files into an `actions` folder to mirror viem but I tend to defer this sort of organization until things get messy rather than trying to organize too early (easier to add organization later than it is to undo/redo organization)

holic: Also thinking about the naming of this, I wonder if this should be something like `transactionQueueActions`

packages/common/src/viemActions.ts
```diff
@@ -0,0 +1,43 @@
...
+): Pick<WalletActions<TChain, TAccount>, "writeContract" | "sendTransaction"> => {
+  return {
+    // Applies to: `client.writeContract`, `getContract(client, ...).write`
+    writeContract: (args) => mud_writeContract(client, args),
+    // Applies to: `client.sendTransaction`
+    sendTransaction: (args) => mud_sendTransaction(client, args),
+  };
+};
+
+export const setupObserverActions = <TChain extends Chain, TAccount extends Account>(
```
holic, line 27: maybe `writeObserverActions`?

templates/react/packages/client/src/mud/setupNetwork.ts
holic: this is great!

we could consider moving away from the contract instance approach in favor of calling functions with pure viem actions (I think this is recommended by viem team anyway as it saves on bundle sizes)

but happy to defer this to a follow up PR

templates/react/packages/client/src/mud/setupNetwork.ts
```diff
@@ -38,6 +54,12 @@ export async function setupNetwork() {
 
   const publicClient = createPublicClient(clientOptions);
 
+  /*
+   * Create an observable for contract writes that we can
+   * pass into MUD dev tools for transaction observability.
+   */
+  const write$ = new Subject<ContractWrite>();
```
holic, line 61: just a bit of context on this approach:

my ideal would be that dev tools would export something that we could "write to" instead of passing in a write stream to dev tools, but when I initially tried this, it got a bit hairy with bundling/code splitting, where an instance of a script running on the page may not have context on the variables created by another part of the code using the same-ish variables

would be curious if you have other ideas for approaching this

templates/react/packages/client/src/mud/setupNetwork.ts
tash-2s: > defer this to a follow up PR

I'll make a follow-up PR for the template improvements!

> consider moving away from the contract instance approach

We can consider this in that PR. If the bundle size difference isn't significant for our app, the contract instance approach might be better for convenience, as we don't need to pass IWorldAbi and worldAddress every time.

packages/common/src/viemActions.ts
tash-2s: I'll make a file for each function with the name of the function.

> transactionQueueActions

Considering the content, I like the name. I'll use that name for this.

packages/common/src/viemActions.ts
tash-2s: I like `writeObserverActions`, but I feel we want something like `setup` for distinguishability.

Given this naming convention:
```ts
client.extend(walletActions).extend(transactionQueueActions)
```

This feels natural:
```ts
client.extend(writeObserverActions)
```

But this looks different:
```ts
client.extend(writeObserverActions(onWrite))
```

templates/react/packages/client/src/mud/setupNetwork.ts
tash-2s: I don't have a good idea for now. 🤔 

Is the `transportObserver` approach closer to your ideal?

holic: I think it might be tricky to use the transport observer approach because we'd have to somehow tie the simulation request (`eth_call`) to the transaction (`eth_sendTransaction` and `eth_sendRawTransaction`).

packages/common/src/viemActions.ts
holic: I see what you mean now. I'd recommend we use the pattern that [viem does](https://viem.sh/op-stack#3-consume-op-stack-actions) (and generally mirrors "middleware" approaches of libraries [like express.js](https://expressjs.com/en/resources/middleware.html)) by making them all functions, so we can eventually support passing in options to each as needed.

So they'd become

```
client
  .extend(transactionQueueActions())
  .extend(writeObserverActions())
```

I could also see moving this into an `actions` dir and naming them each `transactionQueue` and `writeObserver` (to avoid/clean up the "actions" prefix). But I don't feel as strongly there.

tash-2s: > making them all functions, so we can eventually support passing in options to each as needed

I agree with this approach for future extendability. I'll employ the `writeObserver({ onWrite })` structure here.

I didn't realize viem already have this function style actions.

templates/react/packages/client/src/mud/setupNetwork.ts
```diff
@@ -46,13 +63,9 @@ export async function setupNetwork() {
...
-  });
-
-  /*
-   * Create an observable for contract writes that we can
-   * pass into MUD dev tools for transaction observability.
-   */
-  const write$ = new Subject<ContractWrite>();
+  })
+    .extend(transactionQueue())
+    .extend(writeObserver({ onWrite: (write) => write$.next(write) }));
```
tash-2s, line 68: I've applied the renames.

packages/common/src/actions/transactionQueue.ts
```diff
@@ -0,0 +1,14 @@
+import type { Transport, Chain, Account, WalletActions, WalletClient } from "viem";
+import { writeContract as mud_writeContract } from "../writeContract";
+import { sendTransaction as mud_sendTransaction } from "../sendTransaction";
+
+export const transactionQueue = <TChain extends Chain, TAccount extends Account>(): ((
```
holic, line 5: nit: slight preference for the `export function transactionQueue` definition style

packages/common/src/actions/transactionQueue.ts
```diff
@@ -0,0 +1,14 @@
...
+  client: WalletClient<Transport, TChain, TAccount>
+) => Pick<WalletActions<TChain, TAccount>, "writeContract" | "sendTransaction">) => {
+  return (client) => ({
+    // Applies to: `client.writeContract`, `getContract(client, ...).write`
+    writeContract: (args) => mud_writeContract(client, args),
+    // Applies to: `client.sendTransaction`
+    sendTransaction: (args) => mud_sendTransaction(client, args),
```
holic, line 12: If I said we wanted to eventually get rid of `writeContract` and `sendTransaction` "wrappers" in favor of these actions, would you think it better to move the logic in here now?

In that case, I _think_ we'd only need to implement `sendTransaction`? Since `writeContract` internally wraps `sendTransaction`.

tash-2s: > would you think it better to move the logic in here

We don't need the logic exactly here, but by not exposing `mud_sendTransaction`, I think I can remove a few lines of code:

https://github.com/latticexyz/mud/blob/db314a7490b7797322fd0568cbeec0066c231666/packages/common/src/sendTransaction.ts#L28-L33

> we'd only need to implement sendTransaction?

I wouldn't do so. Since we include simulation, we want to use `call` for sendTransaction and `simulateContract` for writeContract for the best error revert message.

holic: oh right! I forgot that's why they were done separately

holic (commented):
Looking great!

A few more requests:
- add `// @deprecated Use walletClient.extend(transactionQueue()) instead.` to each of `getContract`, `writeContract`, `sendTransaction`
- set up the new actions in the rest of the templates
- add a changeset (`pnpm changeset` from root, or via the link in the bot comment above)

tash-2s:
Thanks for the reviews! I'll address these requests.

> set up the new actions in the rest of the templates

OK, I'll apply the same changes as the current react template got to other templates in this PR.

PR #2348, opened 2024-02-29
refactor(store): event interfaces for Store libraries 

yonadaaa:
Exploring my comment on https://github.com/latticexyz/mud/issues/2158#issuecomment-1971071251

Basically some errors (`FieldLayoutLib_*`, `PackedCounter_*`, etc.), are not included in the `IStore` ABI because they are not defined in that interface or anything it inherits. Instead, they are defined in libraries and are referenced only in the _implementation_ of Store.

Some solutions:
- Define error interfaces for each library, eg. `IPackedCounterErrors` (this PR)
- Move every error to a single `IStoreErrors`
- Use abstract contracts [like OpenZeppelin](https://github.com/ethereum/solidity/issues/13086#issuecomment-1445324173) (I'm not sure how that works)
- Stop using interfaces and have clients import the `Store` implementation

### DIFF
--- a/packages/store/src/IFieldLayoutErrors.sol
+++ b/packages/store/src/IFieldLayoutErrors.sol
@@ -5,6 +5,8 @@ pragma solidity >=0.8.24;
  * @title IFieldLayoutErrors
  * @author MUD (https://mud.dev) by Lattice (https://lattice.xyz)
  * @notice This interface includes errors for the FieldLayout library.
+ * @dev We bundle these errors in an interface (instead of at the file-level or in their corresponding libraries) so they can be inherited by IStore.
+ * This ensures that all possible errors are included in the IStore ABI for proper decoding in the frontend.
  */
 interface IFieldLayoutErrors {
   error FieldLayout_TooManyFields(uint256 numFields, uint256 maxFields);
--- a/packages/store/src/IFieldLayoutErrors.sol
+++ b/packages/store/src/IFieldLayoutErrors.sol
@@ -1,0 +1,12 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.24;
+
+interface IFieldLayoutErrors {
+  error FieldLayoutLib_TooManyFields(uint256 numFields, uint256 maxFields);
+  error FieldLayoutLib_TooManyDynamicFields(uint256 numFields, uint256 maxFields);
+  error FieldLayoutLib_Empty();
+  error FieldLayoutLib_InvalidStaticDataLength(uint256 staticDataLength, uint256 computedStaticDataLength);
+  error FieldLayoutLib_StaticLengthIsZero(uint256 index);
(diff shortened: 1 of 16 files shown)


packages/store/src/ISliceErrors.sol
```diff
@@ -0,0 +1,6 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.24;
+
+library ISliceErrors {
```
holic, line 4:
```suggestion
interface ISliceErrors {
```

packages/store/src/IFieldLayoutErrors.sol
```diff
@@ -0,0 +1,17 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.24;
+
+/**
+ * @title IFieldLayoutErrors
+ * @author MUD (https://mud.dev) by Lattice (https://lattice.xyz)
+ * @notice This interface includes errors for the FieldLayout library.
+ */
```
holic, line 8: Can we add the _why_ to each of these?

```suggestion
/**
 * @title IFieldLayoutErrors
 * @author MUD (https://mud.dev) by Lattice (https://lattice.xyz)
 * @notice This interface includes errors for the FieldLayout library.
 * @dev We centralize and reference these errors in an interface so that we can ensure they're included in the resulting Store ABI for proper decoding in the frontend. This is a workaround until the Solidity compiler will do this for us.
 */
```

holic (dismissed):
looking good! just want to include a comment in each interface about _why_ we're doing this approach, for future spelunkers (including ourselves)

did we double check we got all error types?

yonadaaa:
I agree we can improve the comments! afaik this covers all events in Store, then we tackle World later.

> This is a workaround until the Solidity compiler does this for us.

Gonna reword this because I don't think this is actually a bug - we just didn't include those events in `IStore`. It's not obvious to the compiler that the events that `StoreCore` uses should be included in the `IStore` ABI, because it's just one implementation of the interface.

holic:
wondering if we should include some/all errors in the other interfaces as well: https://github.com/latticexyz/mud/blob/main/packages/store/src/IStoreRead.sol#L13

packages/store/src/IStoreErrors.sol
```diff
@@ -3,6 +3,11 @@ pragma solidity >=0.8.24;
 
 import { ResourceId } from "./ResourceId.sol";
 
+/**
+ * @title IStoreErrors
+ * @author MUD (https://mud.dev) by Lattice (https://lattice.xyz)
+ * @notice This interface includes errors for Store.
+ */
```
holic, line 10: can we add the dev comment here too?

holic (approved):
one more comment for `IStoreErrors` imo (since it's the same pattern as our other error interfaces) but otherwise lgtm

PR #2352, opened 2024-02-29
fix(world-modules): token modules always register namespace

yonadaaa:
Updates the ERC20 and ERC721 implementations to always register the token namespace, instead of checking if it has already been registered. This prevents issues with registering the namespace beforehand, namely that only the owner of a system can create a puppet for it.

Fixes https://github.com/latticexyz/mud/issues/2354

### DIFF
--- a/.changeset/dull-otters-smash.md
+++ b/.changeset/dull-otters-smash.md
@@ -2,4 +2,4 @@
 "@latticexyz/world-modules": patch
 ---
 
-Updated the ERC20 and ERC721 implementations to always register the token namespace.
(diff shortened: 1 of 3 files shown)


.changeset/dull-otters-smash.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/world-modules": patch
+---
+
+Updated the ERC20 and ERC721 implementations to always register the token namespace.
```
holic, line 5:
```suggestion
ERC20 and ERC721 implementations now always register the token namespace, instead of checking if it has already been registered. This prevents issues with registering the namespace beforehand, namely that only the owner of a system can create a puppet for it.
```

yonadaaa: Thank you, changelog 👑

alvrs (approved):

PR #2353, opened 2024-02-29
refactor(world): batch call as world function instead of system

yonadaaa:
Exploring this because it makes the balance accounting for batch call with value easier

bytecode size is okay

```
| World                      | 21.835    | 2.741       |
```


holic:
Decided not to do this for v2. We can add a "batch call with value" code path as a non-breaking change, and we can also move from it being a system to a set of first-class functions on the world as a non-breaking change as well.

PR #2357, opened 2024-03-01
refactor(store): HelloStore in IStoreEvents

yonadaaa:


packages/store/src/IStoreEvents.sol
```diff
@@ -63,4 +63,10 @@ interface IStoreEvents {
    * @param keyTuple An array representing the composite key for the record.
    */
   event Store_DeleteRecord(ResourceId indexed tableId, bytes32[] keyTuple);
+
+  /**
+   * @notice Emitted when the store is initialized.
+   * @param storeVersion The version of the Store contract.
+   */
+  event HelloStore(bytes32 indexed storeVersion);
```
holic, line 71: nit: would put this at the top since it's the first event you'll see

yonadaaa: i was also wondering that haha

holic (dismissed):
don't forget a changeset!

should we make it so `StoreData.sol` uses `emit IStoreEvents.HelloStore` rather than `emit HelloStore` for consistency?

~~and maybe `IStore` inherits `IStoreEvents` directly (rather than transitively) so that we get an error in `StoreData.sol` if we don't use the interface reference for emitting the event?~~
nevermind, don't think this is possible with how we've set up store abstractions

yonadaaa:
> should we make it so StoreData.sol uses emit IStoreEvents.HelloStore rather than emit HelloStore for consistency?

sure, seems more flexible to me!

holic (approved):

PR #2358, opened 2024-03-01
refactor(world): add IWorldEvents with HelloWorld

yonadaaa:


docs/pages/world/reference/world-external.mdx
yonadaaa, line 1: I guess `IWorldEvents` needs to be in a docs script?

packages/world/src/IWorldEvents.sol
```diff
@@ -0,0 +1,18 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.24;
+
+import { IWorldErrors } from "./IWorldErrors.sol";
+import { IModule } from "./IModule.sol";
+import { ResourceId } from "./WorldResourceId.sol";
+
+/**
+ * @title IWorldEvents
+ * @author MUD (https://mud.dev) by Lattice (https://lattice.xyz)
```
holic, line 10: should we have a dev comment like with errors?

packages/world/src/IWorldKernel.sol
```diff
@@ -60,12 +60,6 @@ interface IWorldCall {
  * registered functions selectors from the `InitModule`.
  */
 interface IWorldKernel is IWorldModuleInstallation, IWorldCall, IWorldErrors {
```
holic, line 62: Does this need `IWorldEvents`?

packages/world/src/IWorldEvents.sol
yonadaaa: just realised that!

packages/world/src/IWorldKernel.sol
yonadaaa: just realised that!

yonadaaa:
TODO: add `IModulesEvents`

packages/world/src/IWorldEvents.sol
yonadaaa: added one :)

holic (approved):

PR #2363, opened 2024-03-04
test: investigate warning when parsing some logs

yonadaaa:
Parsing some logs from this `BattleResult` table prints a warning, but generating the logs from scratch doesn't seem to.


yonadaaa:
Closed as issue solved

PR #2365, opened 2024-03-04
docs(quickstart): add versions and the need for git

qbzzt:


docs/pages/quickstart.mdx
```diff
@@ -2,13 +2,16 @@
...
-1. Ensure you have [node](https://nodejs.org/en/download/package-manager#macos) 18+, [pnpm](https://pnpm.io/installation) 8+ and [foundry](https://getfoundry.sh/) installed.
+1. Ensure you have [node](https://nodejs.org/en/download/package-manager#macos) 18+, [pnpm](https://pnpm.io/installation) 8+, [git](https://git-scm.com/downloads), and [foundry](https://getfoundry.sh/) 0.2.0+ installed.
+
 1. Create a new app from a template:
 
    ```sh copy
    pnpm create mud@next my-project
    ```
 
+1. Select the template to use.
```
holic, line 13: I don't think we need to document this - the `pnpm create mud` prompts for this and we may change the prompts over time and don't think we need to keep this doc in sync with those prompts. The prompts themselves will be self-documenting.

PR #2366, opened 2024-03-04
feat(world): add viem actions for delegation

tash-2s:
This code has been pulled out from https://github.com/latticexyz/mud/pull/2314.

By extending viem clients with this actions function after delegation, the delegation is automatically applied to World contract writes. Internally, it transforms the `writeContract` arguments into one that wraps it with `callFrom`.

It can be used like this:

```ts
walletClient.extend(
  delegation({
    worldAddress,
    delegatorAddress,
    getSystemId: (functionSelector) =>
      network.useStore.getState().getValue(network.tables.FunctionSelectors, { functionSelector })!.systemId,
  }),
);
```


packages/world/package.json
```diff
@@ -11,6 +11,7 @@
   "type": "module",
   "exports": {
     ".": "./dist/ts/index.js",
+    "./actions": "./dist/ts/actions/index.js",
```
tash-2s, line 14: Following the common library's actions path, I've made a new path for this. But we only have this function now, so I can move the function into `./ts/delegationActions.ts` if preferred.

tash-2s:
@holic Could you review this when you have a moment?

packages/world/ts/actions/delegation.ts
```diff
@@ -0,0 +1,56 @@
...
+export function delegation<TChain extends Chain, TAccount extends Account>({
+  worldAddress,
+  delegatorAddress,
+  getSystemId,
+}: DelegationParameters): (
+  client: WalletClient<Transport, TChain, TAccount>,
+) => Pick<WalletActions<TChain, TAccount>, "writeContract"> {
+  return (client) => ({
+    // Applies to: `client.writeContract`, `getContract(client, ...).write`
+    writeContract: (originalArgs): Promise<WriteContractReturnType> => {
```
holic, line 23:
```suggestion
    writeContract: (writeArgs): Promise<WriteContractReturnType> => {
```
`writeArgs` or `writeContractArgs` might be a little more clear when we're juggling with `callFromArgs`

packages/world/ts/actions/delegation.ts
```diff
@@ -0,0 +1,56 @@
+import type { WalletClient, Transport, Chain, Account, Hex, WalletActions, WriteContractReturnType } from "viem";
+import { getAction, getAbiItem, formatAbiItem, toFunctionSelector, type GetAbiItemParameters } from "viem/utils";
+import { writeContract } from "viem/actions";
+import { encodeSystemCallFrom, type SystemCallFrom } from "../encodeSystemCallFrom";
+
+type DelegationParameters = {
+  worldAddress: Hex;
+  delegatorAddress: Hex;
+  getSystemId: (functionSelector: Hex) => Hex;
```
holic, line 9:
```suggestion
  getSystemId: (functionSelector: Hex) => Promise<Hex>;
```

We should probably make this async because in the future 1) our client query APIs will be async and 2) we may wanna automatically retrieve this from the world

packages/world/ts/actions/delegation.ts
```diff
@@ -0,0 +1,56 @@
...
+};
+
+// By extending clients with this function after delegation, the delegation automatically applies
+// to the World contract writes, meaning these calls are made on behalf of the delegator.
+export function delegation<TChain extends Chain, TAccount extends Account>({
```
holic, line 14:
```suggestion
export function delegatedCalls<TChain extends Chain, TAccount extends Account>({
```
Since "delegation" is a bigger concept, how about calling this `delegatedCalls` or similar?

packages/world/ts/actions/delegation.ts
```diff
@@ -0,0 +1,56 @@
...
+      const callFromArgs: typeof originalArgs = {
+        ...originalArgs,
+        functionName: "callFrom",
+        args: encodeSystemCallFrom({
+          abi: originalArgs.abi,
+          from: delegatorAddress,
+          systemId,
+          functionName: originalArgs.functionName,
+          args: originalArgs.args,
+        } as unknown as SystemCallFrom),
```
holic, line 50: ooc why do we need to cast this?

.changeset/nervous-nails-notice.md
```diff
@@ -0,0 +1,18 @@
...
+Added viem custom client actions for delegation. By extending viem clients with this function after delegation, the delegation is automatically applied to World contract writes. Internally, it transforms the `writeContract` arguments to incorporate `callFrom`.
+
+Usage example:
+
+```ts
+walletClient.extend(
+  delegation({
+    worldAddress,
+    delegatorAddress,
+    getSystemId: (functionSelector) =>
```
holic, line 14:
```suggestion
    getSystemId: async (functionSelector) =>
```
similar as the other comment

packages/world/ts/actions/delegation.ts
```diff
@@ -0,0 +1,56 @@
...
+      // Ensured not to be `undefined` because of the `writeContract` args type.
+      const functionAbiItem = getAbiItem({
+        abi: originalArgs.abi,
+        name: originalArgs.functionName,
+        args: originalArgs.args,
+      } as unknown as GetAbiItemParameters)!;
+
+      // `callFrom` requires `systemId`.
+      const functionSelector = toFunctionSelector(formatAbiItem(functionAbiItem));
+      const systemId = getSystemId(functionSelector);
```
holic, line 38: I think we might be missing some nuance around world function selectors vs system function selectors here.

When we call a namespaced system, the `originalArgs.functionName` will be something like `namespace__functionName` and the ABI will likely be the `IWorld` ABI (all of the world functions).

But when we translate that to a system function call, we need to pass in the system ID + system function selector (i.e. just `functionName` without the namespace, since the namespace is part of the system ID).

Maybe `getSystemId` needs to be something like `worldFunctionToSystemFunction` or `getSystemCall` or similar? It takes in a world function selector and returns both the system ID and system function selector. The args will be the same, but we may have to manually encode the args instead of using `encodeSystemCallFrom`.

holic (commented):
This is a great first pass!

A few notes around naming, sync vs async interface, and how we're translating between world function selectors and system function selectors.

packages/world/ts/actions/delegation.ts
tash-2s: Right, `writeArgs` has more info in the name, thus it is better.

packages/world/ts/actions/delegation.ts
tash-2s: The change makes sense!

packages/world/ts/actions/delegation.ts
tash-2s: When wrapping this kind of viem functions that take an abi, we often need this kind of type cast. This is because it requires the actual abi types to provide strong typing. (in this case, to type functionName and args)

E.g., https://github.com/wevm/viem/blob/viem%402.7.19/src/actions/wallet/writeContract.ts#L156-L162

packages/world/ts/actions/delegation.ts
tash-2s: I was misunderstanding this point. As you mentioned, it only worked for root namespace systems.

packages/dev-tools/src/actions/WriteSummary.tsx
```diff
@@ -62,9 +68,19 @@ export function WriteSummary({ write }: Props) {
...
+    // result in an error.
+    try {
+      const functionData = decodeFunctionData({ abi: worldAbi, data: functionSelectorAndArgs });
+      functionName = functionData.functionName;
+      functionArgs = functionData.args;
+    } catch (error) {
+      if (!(error instanceof AbiFunctionSignatureNotFoundError)) {
+        throw error;
+      }
+    }
```
tash-2s, line 83: I've found an error in dev-tools. I've made a workaround, but I'm not sure what's the best way to address this, so leaving this as TODO now.

For root namespaced calls:
Screenshot 2024-03-05 at 21 39 55

For non-root namespaced calls (This was an error before this change):
Screenshot 2024-03-05 at 21 35 54

tash-2s (commented):
Thank you so much for your comments! 😸 I've addressed all issues. 

The `callFrom` function has been tested with calls for root/non-root systems, with and without call args.

packages/world/ts/actions/callFrom.ts
```diff
@@ -0,0 +1,72 @@
...
+
+// By extending viem clients with this function after delegation, the delegation is automatically
+// applied to the World contract writes, meaning these writes are made on behalf of the delegator.
+export function callFrom<TChain extends Chain, TAccount extends Account>({
+  worldAddress,
+  delegatorAddress,
+  worldFunctionToSystemFunction,
+}: CallFromParameters): (
+  client: WalletClient<Transport, TChain, TAccount>,
+) => Pick<WalletActions<TChain, TAccount>, "writeContract"> {
```
holic, line 32: since we already have an assumption here about this being a world with `callFrom`, we could add an async check that the client account has a proper delegation set up for delegator address and warn if not

packages/world/ts/actions/callFrom.ts
```diff
@@ -0,0 +1,72 @@
...
+    worldFunctionSelector: Hex,
+  ) => Promise<{ systemId: Hex; systemFunctionSelector: Hex }>;
+};
+
+// By extending viem clients with this function after delegation, the delegation is automatically
+// applied to the World contract writes, meaning these writes are made on behalf of the delegator.
+export function callFrom<TChain extends Chain, TAccount extends Account>({
+  worldAddress,
+  delegatorAddress,
+  worldFunctionToSystemFunction,
```
holic, line 29: thought about this some more and since we already assume you're operating on a world and function selectors never change once set (i.e. we can safely cache them here), I think we can get rid of this option in favor of something like:

```ts
type SystemFunction = { readonly systemId: Hex, readonly systemFunctionSelector: Hex };
const worldFunctionSelectorCache = new Map<Hex, SystemFunction>();
async function worldFunctionToSystemFunction(client: PublicClient, worldAddress: Address, worldFunctionSelector: Hex): Promise<SystemFunction> {
  const cached = worldFunctionSelectorCache.get(worldFunctionSelector);
  if (cached != null) return cached;

  const systemFunction = await readContract(client, {
    address: worldAddress,
    abi: worldAbi,
    functionName: 'getRecord',
    args: [worldConfig.tables.FunctionSelectors.tableId, worldFunctionSelector]
  });
  worldFunctionSelectorCache.set(worldFunctionSelector, systemFunction);
  return systemFunction;
}
```
(Note that this is pseudocode and may require parsing the record from the `getRecord` call, and I'm not sure how best to expect the public client, wether we expect to extend a public client or take it as an argument)

holic (commented):
had another thought about making this API easier to use by creating a default method for retrieving the system function selectors

curious what you think of the approach?

packages/dev-tools/src/actions/WriteSummary.tsx
holic: we could try to do the inverse, where we resolve the system function selector form the world one and assume the same args?

packages/world/ts/actions/callFrom.ts
tash-2s: Right, it's better to have that automatic retrieving, so users without this info sync can use this function more easily.

Getting a public client as an argument is more flexible, as it doesn't ask users to extend their wallet clients with public actions, so they have options. I'll implement this.

packages/dev-tools/src/actions/WriteSummary.tsx
tash-2s: > we don't have the system signature

right 🤔

packages/world/ts/actions/callFrom.ts
holic: makes sense!

packages/dev-tools/src/actions/WriteSummary.tsx
holic: added an issue to track this here: https://github.com/latticexyz/mud/issues/2382

tash-2s:
@holic I've added the default function that reads data from the world contract. Could you take a look? 🥷

tash-2s:
Regarding the dev-tools issue, thanks to https://github.com/latticexyz/mud/pull/2392, we can now use `FunctionSignatures` to display human-readable `call` / `callFrom` actions. Since it's an off-chain table, we can use zustand's `useStore` if provided.

packages/world/ts/actions/callFrom.ts
```diff
@@ -0,0 +1,148 @@
...
+  worldAddress: Hex,
+  worldFunctionSelector: Hex,
+): Promise<SystemFunction> {
+  const table = worldConfig.tables.world__FunctionSelectors;
+
+  const _keySchema = getKeySchema(table);
+  const keySchema = mapObject<typeof _keySchema, { [K in keyof typeof _keySchema]: (typeof _keySchema)[K]["type"] }>(
+    _keySchema,
+    ({ type }) => type,
+  );
```
tash-2s, line 125: This `keySchema`'s type is `{ worldFunctionSelector: "bytes4" }`. Without specifying the generics type argument (using `mapObject(_keySchema, ...)`), the type results in `{ worldFunctionSelector: unknown }`. There might be a better way to manage this with the new config.

tash-2s (commented):
I've merged main and aligned with the changes in main.

packages/world/ts/actions/callFrom.ts
tash-2s: ah, `getSchemaTypes`

packages/world/package.json
```diff
@@ -11,6 +11,7 @@
   "type": "module",
   "exports": {
     ".": "./dist/index.js",
+    "./actions": "./dist/actions.js",
```
alvrs, line 14: I don't think we need to make this a separate export now that we don't have global side effects on the global export. We can just export this action via `/ts/exports/index.ts` (or `ts/exports/internal.ts` for now until we want to make it "public")

packages/world/ts/actions/callFrom.ts
```diff
@@ -0,0 +1,144 @@
...
+      // Get the systemId and System's function selector.
+      const { systemId, systemFunctionSelector } = await worldFunctionToSystemFunction(params, worldFunctionSelector);
+
+      // Construct the System's calldata.
+      // If there's no args, use the System's function selector as calldata.
+      // Otherwise, use the World's calldata, replacing the World's function selector with the System's.
+      const systemCalldata =
+        worldCalldata === worldFunctionSelector
+          ? systemFunctionSelector
+          : concat([systemFunctionSelector, slice(worldCalldata, 4)]);
```
alvrs, line 82: why not always `concat([systemFunctionSelector, slice(worldCalldata, 4)]);`? if ` worldCalldata === worldFunctionSelector` then that means `slice(worldCalldata, 4)` is empty, so concat would be a noop, so we don't need to handle it with a special case?

alvrs (commented):
This is great! just some minor nits, but aside from those i think this is ready to be merged

.changeset/nervous-nails-notice.md
```diff
@@ -0,0 +1,23 @@
+---
+"@latticexyz/world": minor
```
alvrs, line 2: I would opt for not exporting this from `world/external` yet to test it a bit and bundle the external feature release with a couple other libraries for the wallet connection flow. That means i would declare this as patch for now and add a note to the changeset that it's not ready for "stable" consumption yet.

tash-2s: makes sense!

packages/world/ts/actions/callFrom.ts
tash-2s: Using `const systemCalldata = concat([systemFunctionSelector, slice(worldCalldata, 4)]);` could result in an out-of-bounds error in `slice()` if there's no arg.
(`SliceOffsetOutOfBoundsError: Slice starting at offset "4" is out-of-bounds (size: 4).`)

I could alternatively use `size(worldCalldata) === 4` as a condition. Would that be better?

tash-2s:
@alvrs Thanks for your comments! I've made the function internal and noted it in the changeset.

packages/world/ts/actions/callFrom.ts
alvrs: could we use `readHex` from `@latticexyz/common` here? That should handle the case where the start would be out of bounds

tash-2s: Maybe having a helper function that splits the function calldata into selector and argument data would simplify this function a bit.

tash-2s: @alvrs I'll look into readHex. thanks!

alvrs (approved):

PR #2376, opened 2024-03-05
build: run prepare before install in Dockerfile

holic:
had this fix/workaround in CI but not in Dockerfile


holic:
confirmed that this got past the install/build step:
https://github.com/latticexyz/mud/actions/runs/8160408709/job/22306926234

merging to fix main's docker release step

PR #2377, opened 2024-03-06
test(e2e): run contracts test

yonadaaa:
We missed a failing contract test (`testMemoryCorruption`, now fixed in https://github.com/latticexyz/mud/pull/2378) because only the e2e sync tests were running


.github/workflows/e2e.yml
```diff
@@ -53,3 +53,7 @@ jobs:
           DATABASE_URL: "postgres://postgres@localhost:5432/postgres"
         working-directory: ./e2e/packages/sync-test
         run: pnpm test:ci
+
+      - name: Run contracts tests
+        working-directory: ./e2e/packages/contracts
+        run: pnpm test:ci
```
holic, line 59:
```suggestion
      - name: Run tests
        working-directory: ./e2e
        run: pnpm run test:ci
```
what if we just run em all?

.github/workflows/e2e.yml
```diff
@@ -53,3 +53,7 @@ jobs:
           DATABASE_URL: "postgres://postgres@localhost:5432/postgres"
         working-directory: ./e2e/packages/sync-test
         run: pnpm test:ci
+
+      - name: Run tests
+        working-directory: ./e2e
+        run: pnpm run test:ci
```
holic, line 59: we should combine this with the above so we don't run sync tests twice

yonadaaa: done

PR #1880, opened 2023-11-08
test(e2e): test memory corruption in dynamic-to-static array conversion

dk1a:
fixes https://github.com/latticexyz/mud/issues/1603

PR #2378, opened 2024-03-06
test(e2e): fix memory corruption test

yonadaaa:
Fixes the test created in https://github.com/latticexyz/mud/pull/1880

I'm not 100% sure what the intention of this test is, when I run it the memory it the elements of `result` are the same as `data` (ie. not corrupted). Maybe this test added _before_ the fix, to prove it was broken?

### DIFF
--- a/e2e/packages/contracts/mud.config.ts
+++ b/e2e/packages/contracts/mud.config.ts
@@ -49,7 +49,7 @@ export default mudConfig({
     },
     StaticArray: {
       keySchema: {},
-      valueSchema: "uint256[2]",
+      valueSchema: "uint256[3]",
     },
   },
 });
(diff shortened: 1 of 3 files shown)


dk1a:
Memory corruption would occur when the dynamic length is smaller than the static length, so this test doesn't even make sense
I don't remember why I made it this way, but it looks like a mistake
Rather than fixing asserts, remaking the test completely would probably be better

dk1a:
On the other hand 2 tests are also fine, but they would test for different kinds of memory corruption
The current one seems much less important

yonadaaa:
Thanks for the clarification @dk1a! I added tests for different dynamic array lengths, but maybe we can test memory corruption more thoroughly

dk1a:
> but maybe we can test memory corruption more thoroughly

You can add `memoryAfterResult` for the others, especially shorterDynamic - you'd want to allocate something right after `data`, and make sure it's not corrupted (probably would require manual assembly allocation, you can't rely on solidity to not reorder allocations)

holic:
ooc any reason this lives in e2e and not in store?

dk1a:
> ooc any reason this lives in e2e and not in store?

to avoid a test-only table in store

holic (approved):

Issue #2173, opened 2024-01-23
make namespace usage consistent with double underscore


frolic:
We're now using a double underscore delimiter for namespaced function signatures (`{namespace}__{functionName}`) and may need to update other areas to be consistent with this change:
- SQL table names/schemas
- MUD config output and TS types
- human-readable resource IDs (currently uses colon as a delimiter)

holic:
I think this is good enough. The rest will come with the config refactor (#1668)

PR #2379, opened 2024-03-06
feat(store-sync): adjust DB schema/table names for consistency

holic:
part of #2173 

- moves postgres from `{address}__{namespace}.{tableName}` to `{address}.{namespace}__{tableName}`
- moves sqlite to use snake case for table/column names to match postgres


yonadaaa (commented):
The reasoning makes sense! I worry this might be less useable though, because the most useful "divisions" (ie. pg namespaces) we can have are MUD namespaces. 

For example, with the Sky Strife analytics indexer, the indexer already has a specific world address, and it's useful to `SEARCH_PATH` a specific namespace to save rewriting long table names: https://github.com/latticexyz/skystrife-public/blob/main/sql-queries/WalletMatchesJoined.sql#L1-L2

basically does this best take advantage of Postgres namespaces?

holic:
since sky strife uses root namespace, you won't need to do anything differently here except remove the `__` suffix from the schema

PR #2381, opened 2024-03-06
chore(phaserx): remove phaserx from monorepo

yonadaaa:
Removing the `phaserx` package from the MUD monorepo as it is no longer actively maintained. 

Previously we removed the noise package in https://github.com/latticexyz/mud/pull/2304


examples/minimal/packages/client-phaser/package.json
```diff
@@ -13,7 +13,7 @@
     "@improbable-eng/grpc-web": "^0.15.0",
     "@latticexyz/common": "link:../../../../packages/common",
     "@latticexyz/dev-tools": "link:../../../../packages/dev-tools",
-    "@latticexyz/phaserx": "link:../../../../packages/phaserx",
+    "@latticexyz/phaserx": "2.0.0-next.17",
```
yonadaaa, line 16: Curious which version is best to use here

holic: seems fine! surprised we were still publishing it with the rest

holic (approved):

PR #2387, opened 2024-03-07
fix(store-sync): track changed records together in zustand

holic:
when playing with this in #2355, I noticed a possible edge case where tracking these updated IDs separately meant that an update -> delete or delete -> update for the same record in the same block could give the wrong results in the end (out of sync from chain state)


yonadaaa (dismissed):
changeset?

PR #2389, opened 2024-03-07
feat(world): add types for v2 config resolvers

alvrs:


.eslintrc
```diff
@@ -5,9 +5,18 @@
...
+    "max-len": ["error", { "code": 180, "comments": 180 }],
+    "@typescript-eslint/ban-types": [
+      "error",
+      {
+        "extendDefaults": true,
+        "types": {
+          "{}": false,
+        },
+      },
+    ],
```
alvrs, line 20: this is disabling the error when defining a type as "empty object" (= `{}`)

.vscode/settings.json
```diff
@@ -1 +1,3 @@
-{}
+{
+  "typescript.tsdk": "node_modules/typescript/lib"
+}
```
alvrs, line 3: This makes sure vscode uses the repo version of ts instead of a different one in vscode. This is important because some of the type inference only works with the newer versions of typescript, so the dev experience would be much worse if there is accidentally an old version configured

packages/world/tsconfig.json
```diff
@@ -13,7 +13,8 @@
     "esModuleInterop": true,
     "forceConsistentCasingInFileNames": true,
     "strict": true,
-    "skipLibCheck": true
+    "skipLibCheck": true,
+    "noErrorTruncation": true
```
alvrs, line 17: this makes increases the length of types before being truncated, which is useful for development

.eslintrc
holic: can we add a comment here? I think eslintrc supports it and, if not, we could switch this to `.eslintrc.js`
https://eslint.org/docs/latest/use/configure/configuration-files#configuration-file-formats

.vscode/settings.json
holic: we should add this to the templates too?

packages/store/ts/config/v2/store.test.ts
```diff
@@ -5,151 +5,166 @@ import { attest } from "@arktype/attest";
...
+            },
+          },
+          primaryKey: ["key"],
+        },
+      },
+      userTypes: {},
+      enums: {},
+      namespace: "",
+    } as const;
+    attest<typeof expected>(config);
```
holic, line 40: beautiful

packages/store/ts/config/v2/store.test.ts
```diff
@@ -563,46 +595,53 @@ describe("resolveStoreConfig", () => {
...
+        ValidNames: ["first", "second"],
+      },
+      namespace: "",
+    } as const;
+    attest<typeof expected>(config);
+  });
+
+  it("should use the root namespace as default namespace", () => {
+    const config = resolveStoreConfig({ tables: { Example: {} } });
+    attest<"">(config.namespace);
```
holic, line 645: should we have a test for when namespace is set?

holic: also I thought namespaces were meant to be a world thing, not a store thing?

packages/store/ts/config/v2/store.ts
```diff
@@ -22,44 +23,43 @@ export type validateStoreTablesConfig<input, scope extends AbiTypeScope = AbiTyp
...
-      >
+    ? validateStoreTablesConfig<input[key], extendedScope<input>>
     : key extends "userTypes"
       ? UserTypes
       : key extends "enums"
         ? narrow<input[key]>
         : input[key];
 };
 
+export type resolveEnums<enums> = { [key in keyof enums]: Readonly<enums[key]> };
```
holic, line 53: do we have a rule for when we use `readonly` keys vs not?

packages/store/ts/config/v2/store.ts
```diff
@@ -22,44 +23,43 @@ export type validateStoreTablesConfig<input, scope extends AbiTypeScope = AbiTyp
...
-      >
-    : input[key];
+  readonly tables: "tables" extends keyof input ? resolveStoreTablesConfig<input["tables"], extendedScope<input>> : {};
+  readonly userTypes: "userTypes" extends keyof input ? input["userTypes"] : {};
+  readonly enums: "enums" extends keyof input ? resolveEnums<input["enums"]> : {};
+  readonly namespace: "namespace" extends keyof input ? input["namespace"] : "";
 }>;
 
-export function resolveStoreConfig<input>(input: validateStoreConfig<input>): resolveStoreConfig<input> {
+export function resolveStoreConfig<const input>(input: validateStoreConfig<input>): resolveStoreConfig<input> {
```
holic, line 62: does this make it so a regular object input gets used as if it were a const/strongly typed?

packages/world/ts/config/v2/world.test.ts
```diff
@@ -0,0 +1,879 @@
...
+      attest(
+        resolveWorldConfig({
+          tables: {
+            First: {
+              schema: { firstKey: "address", firstName: "string", firstAge: "uint256" },
+              primaryKey: ["firstKey", "firstAge"],
+            },
+            Second: {
+              schema: { secondKey: "address", secondName: "string", secondAge: "uint256" },
+              // @ts-expect-error Type '"firstKey"' is not assignable to type '"secondKey" | "secondAge"'
```
holic, line 776: not at all blocking, but where this condition lives, do we have room to expand on/improve this error type?

like a branch that outputs one of the following:
> `firstKey` field was not found in the schema
> `firstKey` field is not a static ABI type

packages/world/ts/config/v2/world.test.ts
```diff
@@ -0,0 +1,879 @@
...
+              primaryKey: ["firstKey", "secondAge"],
```
holic, line 777: do we have/want/need any checks around "unique fields" defined here so we don't allow for `["firstKey", "firstKey"]`?

packages/world/ts/config/v2/world.ts
```diff
@@ -0,0 +1,77 @@
...
+
+export type WorldConfigInput<userTypes extends UserTypes = UserTypes, enums extends Enums = Enums> = evaluate<
+  StoreConfigInput<userTypes, enums> & {
+    namespaces?: NamespacesInput;
+  }
+>;
+
+export type NamespacesInput = { [key: string]: NamespaceInput };
+
+export type NamespaceInput = Omit<StoreConfigInput, "userTypes" | "enums">;
```
holic, line 23: do you think this will be annoying to maintain vs the inverse? `Pick<StoreConfigInput, "tables">`?

packages/world/ts/config/v2/world.ts
```diff
@@ -0,0 +1,77 @@
...
+    : never
+  : never;
+
+export type resolveWorldConfig<input> = evaluate<
+  resolveStoreConfig<input> & {
+    readonly tables: "namespaces" extends keyof input
+      ? {
+          readonly [key in namespacedTableKeys<input>]: key extends `${infer namespace}__${infer table}`
+            ? resolveTableConfig<
+                get<get<get<get<input, "namespaces">, namespace>, "tables">, table>,
```
holic, line 58: 🙈

holic (dismissed):
looks great! mostly just questions/curiosities, nothing blocking

packages/store/ts/config/v2/store.test.ts
alvrs: protocol wise yes, but tablegen also needs namespaces, so it feels fine to add it to the store config (so it can be part of the expected resolved table config)

packages/store/ts/config/v2/store.ts
alvrs: Should probably always be `readonly`. I think the test is passing in this case because of the `const input` parameter, so the incoming enum config is already readonly, which makes the enum object on the output readonly too. Will still add it here for consistency.

packages/store/ts/config/v2/store.ts
alvrs: yes basically! https://xebia.com/blog/typescript-5-0-and-the-new-const-modifier-on-type-parameters/

packages/world/ts/config/v2/world.test.ts
alvrs: this one is a native typescript error, it's probably possible to create a validator function that routes this condition to a custom string error, but would leave that for later

packages/world/ts/config/v2/world.test.ts
alvrs: Don't have it yet, unclear if it's possible at a type level, but will ask david about ideas. Maybe we can map the tuple to an object that has the tuple values as keys, then convert it back to a tuple and check the length? Would also keep this one for later though.

packages/world/ts/config/v2/world.ts
alvrs: agree it might be nicer to explicitly pick the things we want there than adding a store config option and forgetting about it here

packages/world/ts/config/v2/world.ts
alvrs: yeah haha, there might be a prettier way to do this with some kind of `deepGet` type that takes a tuple of keys but i opted against it for the moment bc 1. don't have a `deepGet` type yet and 2. it might be the case that this manual nesting here takes better advantage of the typescript cache (can only really know that once we do some type benchmarking)

PR #2392, opened 2024-03-07
feat(world): add system signatures to FunctionSignatures

yonadaaa:


packages/world/src/modules/init/implementations/WorldRegistrationSystem.sol
yonadaaa, line 1: We also set the world function selector in `registerRootFunctionSelector`. Do we need to do set the system one there or are they the same?

holic: they're the same!

.changeset/poor-fishes-pretend.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/world": patch
+---
+
+Added system signatures to the `FunctionSignatures` table, so they can be decoded by syncing the table.
```
holic, line 5:
```suggestion
Added system signatures to the `FunctionSignatures` table, so they can be used to generate system ABIs and decode system calls made via the world.
```

holic (approved):

packages/world/src/modules/init/implementations/WorldRegistrationSystem.sol
```diff
@@ -214,7 +214,8 @@ contract WorldRegistrationSystem is System, IWorldErrors, LimitedCallContext {
     bytes4 systemFunctionSelector = bytes4(keccak256(bytes(systemFunctionSignature)));
     FunctionSelectors._set(worldFunctionSelector, systemId, systemFunctionSelector);
 
-    // Register the function signature for offchain use
+    // Register the function signatures for offchain use
+    FunctionSignatures._set(systemFunctionSelector, systemFunctionSignature);
```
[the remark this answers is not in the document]
yonadaaa: I like that! I definitely think `string` is the "canonical" form.

PR #2393, opened 2024-03-07
refactor(world): `registerRootFunctionSelector` does not take selector

yonadaaa:
The signature and selector provided to this function are equivalent - we can derive the selector instead of passing it in.

We should actually rename this function to `registerRootFunctionSignature`!

### DIFF
--- a/packages/world/src/modules/init/implementations/WorldRegistrationSystem.sol
+++ b/packages/world/src/modules/init/implementations/WorldRegistrationSystem.sol
@@ -223,31 +223,29 @@ contract WorldRegistrationSystem is System, IWorldErrors, LimitedCallContext {
    * @notice Registers a root World function selector
    * @dev Creates a mapping for a root World function without namespace or name prefix
    * @param systemId The system ID
-   * @param worldFunctionSignature The signature of the World function
-   * @param systemFunctionSelector The selector of the system function
-   * @return worldFunctionSelector The selector of the World function
+   * @param systemFunctionSignature The signature of the World function
+   * @return systemFunctionSelector The selector of the World function
    */
   function registerRootFunctionSelector(
     ResourceId systemId,
-    string memory worldFunctionSignature,
-    bytes4 systemFunctionSelector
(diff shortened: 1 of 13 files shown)


packages/world/test/World.t.sol
```diff
@@ -1645,12 +1645,11 @@ contract WorldTest is Test, GasReporter {
     world.registerNamespace(systemId.getNamespaceId());
     world.registerSystem(systemId, system, true);
 
-    string memory worldFunc = "testSelector()";
-    bytes4 sysFunc = WorldTestSystem.msgSender.selector;
+    string memory worldFunc = "msgSender()";
```
holic, line 1648: should we call this `systemFunc`?

holic: or maybe just use the hardcoded string everywhere instead of a variable

yonadaaa: hardcoded strings would be more consistent i think

yonadaaa: but this variable is used more so will keep it for now

packages/world/src/modules/init/implementations/WorldRegistrationSystem.sol
```diff
@@ -223,31 +223,29 @@ contract WorldRegistrationSystem is System, IWorldErrors, LimitedCallContext {
    * @notice Registers a root World function selector
    * @dev Creates a mapping for a root World function without namespace or name prefix
    * @param systemId The system ID
-   * @param worldFunctionSignature The signature of the World function
-   * @param systemFunctionSelector The selector of the system function
-   * @return worldFunctionSelector The selector of the World function
+   * @param systemFunctionSignature The signature of the World function
+   * @return systemFunctionSelector The selector of the World function
    */
   function registerRootFunctionSelector(
```
yonadaaa, line 229: We could rename this to `registerRootFunctionSignature`

yonadaaa:
Closing in favour of https://github.com/latticexyz/mud/pull/2395 and also this was intended behaviour

PR #2394, opened 2024-03-07
WIP config v2 runtime resolving

holic:


packages/store/package.json
```diff
@@ -47,7 +47,7 @@
     "dev": "tsup --watch",
     "gas-report": "gas-report --save gas-report.json",
     "lint": "solhint --config ./.solhint.json 'src/**/*.sol'",
-    "test": "vitest typecheck --run && vitest --run --passWithNoTests && forge test",
+    "test": "vitest --run --passWithNoTests && forge test",
```
holic, line 50: do we intentionally not run typecheck here?

packages/store/ts/config/v2/schema.ts
```diff
@@ -16,8 +17,27 @@ export type resolveSchema<schema extends SchemaInput<scope>, scope extends AbiTy
 
 export function resolveSchema<schema extends SchemaInput<scope>, scope extends AbiTypeScope = AbiTypeScope>(
   schema: schema,
-  scope?: scope,
+  scope: scope = AbiTypeScope as scope,
 ): resolveSchema<schema, scope> {
-  // TODO: runtime implementation
-  return {} as never;
+  return Object.fromEntries(
+    Object.entries(schema).map(([key, internalType]) => {
```
holic, line 23: just remembered we have a `mapObject` helper if that could clean this up

packages/store/ts/config/v2/store.test.ts
```diff
@@ -557,11 +563,13 @@ describe("resolveStoreConfig", () => {
           },
         },
       }),
-    ).type.errors(`Type '"firstKey"' is not assignable to type '"secondKey" | "secondAge"'`);
+    )
+      .throws("Invalid primary key. Expected (secondKey|secondAge)[], received [firstKey,secondAge]")
```
holic, line 567: while we're here, can we wrap these key names in double quotes? and backticks around each expression?

> Invalid primary key. Expected `("secondKey" | "secondAge")[]`, received `["firstKey", "secondAge"]`.

I wanna get to a place where our errors rival viem!

holic: and would like to come back to this later and add more branches like 
> `Invalid primary key. `"firstKey"` is not a static ABI type and cannot be used as part of a key.
> `Invalid primary key. No field named `"firstKey"` in schema.

packages/store/ts/config/v2/store.ts
```diff
@@ -26,20 +27,71 @@ export type resolveStoreTablesConfig<input, scope extends AbiTypeScope = AbiType
...
   ? scope
   : enums extends Enums
     ? extendScope<scope, { [key in keyof enums]: "uint8" }>
     : scope;
 
+export function scopeWithEnums<enums, scope extends AbiTypeScope = AbiTypeScope>(
+  enums: enums,
+  scope: scope = AbiTypeScope as scope,
+): scopeWithEnums<enums, scope> {
+  if (isEnums(enums)) {
```
holic, line 82: I reason I was using mostly things like `validateEnums` where it returns `asserts enums is Enums` was so we could both type narrow and throw runtime errors for e.g. invalid enums

not blocking but any reason this approach changed?

packages/store/ts/config/v2/table.test.ts
```diff
@@ -177,7 +32,7 @@ describe("resolveTableConfig", () => {
       primaryKey: ["key"],
     } as const;
 
-    attest<typeof expected>(table);
+    attest<typeof expected>(table).snap(expected);
```
holic, line 35:
```suggestion
    attest<typeof expected>(table).equals(expected);
```
wasn't sure if `.snap` was intended here

packages/store/ts/config/v2/table.ts
```diff
@@ -1,116 +1,30 @@
...
-    },
-    scope
-  >;
-  readonly valueSchema: resolveSchema<
-    {
-      readonly [key in Exclude<keyof input["schema"], input["primaryKey"][number]>]: input["schema"][key];
-    },
-    scope
-  >;
-}>;
```
holic, line 112: thanks for moving this out! 🙏

packages/store/ts/config/v2/tableFull.ts
```diff
@@ -0,0 +1,131 @@
...
+
+export type ValidKeys<schema extends SchemaInput<scope>, scope extends AbiTypeScope> = readonly [
+  getStaticAbiTypeKeys<schema, scope>,
+  ...getStaticAbiTypeKeys<schema, scope>[],
+];
+
+function getValidKeys<schema extends SchemaInput<scope>, scope extends AbiTypeScope>(
+  schema: schema,
+  scope: scope = AbiTypeScope as scope,
+) {
```
holic, line 24: does this need a return type?

packages/store/ts/config/v2/tableFull.ts
```diff
@@ -0,0 +1,131 @@
...
+    ? validateKeys<getStaticAbiTypeKeys<conform<get<input, "schema">, SchemaInput<scope>>, scope>, input[key]>
+    : key extends "schema"
+      ? conform<input[key], SchemaInput<scope>>
+      : input[key];
+};
+
+export function validateTableFull<input, scope extends AbiTypeScope = AbiTypeScope>(
+  input: input,
+  scope: scope = AbiTypeScope as scope,
+) {
```
holic, line 69: likewise does this need a return type?

packages/store/package.json
alvrs: Vitest somehow doesn't like the `attest(config).equals(expected)` type, unclear why. Types of all new libs are tested by attest though.

holic (commented):
can't approve but nothing particularly blocking!

packages/store/ts/config/v2/schema.ts
alvrs: Tried this but the types on `mapObject` are sadly not good enough, makes it such that we'd have to cast more. Punting on it for now, can revisit this later.

packages/store/ts/config/v2/store.ts
alvrs: in this case i wanted to mirror the type behavior which extends the scope with enums if enums are provided and just returns the old scope if none are provided, instead of throwing an error

packages/store/ts/config/v2/tableFull.ts
```diff
@@ -0,0 +1,137 @@
...
+    ? validateKeys<getStaticAbiTypeKeys<conform<get<input, "schema">, SchemaInput<scope>>, scope>, input[key]>
+    : key extends "schema"
+      ? conform<input[key], SchemaInput<scope>>
+      : input[key];
+};
+
+export function validateTableFull<input, scope extends AbiTypeScope = AbiTypeScope>(
+  input: input,
+  scope: scope = AbiTypeScope as scope,
+): void {
```
holic, line 69: would it be useful to have this narrow type type as well?
```suggestion
): asserts input is TableFullInput {
```

alvrs: updated

PR #2395, opened 2024-03-07
refactor(world): registerRootFunctionSelector takes system signature

yonadaaa:


.changeset/funny-countries-hang.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/world": patch
+---
+
+Replaced the `systemFunctionSelector` argument in `registerRootFunctionSelector` with `systemFunctionSignature`, so it can be registered onchain.
```
holic, line 5:
```suggestion
`registerRootFunctionSelector` now expects a `systemFunctionSignature` instead of a `systemFunctionSelector`. Internally, we compute the selector from the signature. This allows us to track system function signatures that are registered at the root so we can later generate ABIs for these systems.
```

.changeset/funny-countries-hang.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/world": patch
```
holic, line 2:
```suggestion
"@latticexyz/world": minor
```

this is _technically_ breaking but kinda more of an internal API and don't expect many downstream users to be affected, so maybe at least make this `minor`?

holic (approved):

.changeset/funny-countries-hang.md
yonadaaa: I agree!

PR #2398, opened 2024-03-08
chore: remove some unused files

yonadaaa:
These files are not referenced elsewhere in the codebase and cause type errors (surfaced in https://github.com/latticexyz/mud/pull/2398), so we can remove them and simplify the codebase a tiny bit.


holic (approved):

packages/common/src/utils/identity.ts
```diff
@@ -1,3 +0,0 @@
-export function identity<T>(value: T): T {
-  return value;
-}
```
holic, line 3: can we leave this one?

packages/common/src/utils/index.ts
```diff
@@ -3,9 +3,7 @@ export * from "./bigIntMax";
 export * from "./bigIntMin";
 export * from "./bigIntSort";
 export * from "./chunk";
-export * from "./curry";
 export * from "./groupBy";
-export * from "./identity";
```
holic, line 8: prob need a changeset for removing exports

yonadaaa: True

packages/common/src/utils/identity.ts
yonadaaa: sure! no issues

.changeset/dry-flowers-shave.md
```diff
@@ -0,0 +1,6 @@
+---
+"@latticexyz/common": patch
+"@latticexyz/react": patch
+---
+
+Removed some unused files, namely `curry` in `@latticexyz/common` and `useDeprecatedComputedValue` from `@latticexyz/react`.
```
holic, line 6: for future reference, you can create a changeset for each package with the package-specific changes so that we don't need the irrelevant change in another package's changelog

PR #2399, opened 2024-03-09
fix(dev-tools): fix `callFrom` transaction display errors for non-root namespace functions

tash-2s:
This PR is a spin-off from #2366.

Currently, in dev-tools, errors occur when `world.call()` / `world.callFrom()` transactions containing non-root namespace System functions are submitted. (Error: `decodeFunctionData.ts:67 Uncaught AbiFunctionSignatureNotFoundError: Encoded function signature "0x67238562" not found on ABI.`) This problem arises because dev-tools uses the World ABI to decode System function calldata for displaying actions (e.g., 'addTask("something") via callFrom').

This PR resolves the error. Now, dev-tools uses the `FunctionSignatures` off-chain table through Zustand if available for proper calldata decoding, as enabled by #2392. Otherwise, it attempts decoding with the World ABI, then proceeds without decoding if unsuccessful.

For the React component structure, I extracted the section displaying the function name and args from `WriteSummary` into `WriteFunction` and made the implementation.

All paths have been tested:

- root/non-root namespace
- `world.doSomething(...)` (fallback call) / `world.callFrom(...)`
- with/without `useStore`
- with/without System function args

### DIFF
--- a/.changeset/kind-rats-compare.md
+++ b/.changeset/kind-rats-compare.md
@@ -1,0 +1,5 @@
+---
+"@latticexyz/dev-tools": patch
+---
+
+Fixed `call`/`callFrom` transaction display errors for non-root namespace functions in dev-tools.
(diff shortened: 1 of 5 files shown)


tash-2s:
@holic I've fixed the dev-tools error. Could you take a look when you have a moment? 🙏

PR #2402, opened 2024-03-10
fix: add bytes12 to Bytes.sol

transmissions11:
To support packed uint96 types in tables.


holic:
closing in favor of https://github.com/latticexyz/mud/pull/2403

PR #2403, opened 2024-03-10
fix(store): restore bytesN helpers

holic:
These were removed in #2090 as part of audit fixes because they appeared unused, but we missed that they are used by codegen (likely because the function signature is generated by a template string):
https://github.com/latticexyz/mud/blob/1a82c278dd8037155b6449e383b3a00d6453ea3e/packages/store/ts/codegen/field.ts#L269

Shoutout to @transmissions11 for spotting this and kicking this off in #2402

While I was here, I renamed the helpers from `Bytes.sliceN` to `Bytes.getBytesN` to both match the `Bytes.setBytesN` methods and to better differentiate this from `Slice` and `SliceLib` utils.

### DIFF
--- a/.changeset/lemon-falcons-arrive.md
+++ b/.changeset/lemon-falcons-arrive.md
@@ -1,5 +1,5 @@
 ---
-"@latticexyz/store": minor
+"@latticexyz/store": patch
 ---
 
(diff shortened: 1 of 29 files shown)


yonadaaa (approved):
Great spot! I wish there was a way we could not get the compiler warning

PR #2408, opened 2024-03-11
test(cli): ignore foundry compiler warnings

holic:
inspired by https://github.com/latticexyz/mud/pull/2403#pullrequestreview-1927451556

enabled by https://twitter.com/PaulRBerg/status/1625131427737088001


yonadaaa:
We should not disable these warnings, they are caused by an unused param fixed in https://github.com/latticexyz/mud/pull/2410

Issue #2369, opened 2024-03-05
revisit `abstract Store` contract


holic:
iirc we didn't make `Store.sol` a thing to inherit from because we expect folks to implement `IStoreWrite.sol` interface
wondering why we don't have e.g. `StoreWrite.sol` as an abstract contract, then a `abstract Store is StoreRead, StoreWrite` ? 
(we use ^ this approach for store hooks)

holic:
@yonadaaa did you start on this?

Issue #2351, opened 2024-02-29
clear caches before generating docs


yonadaaa:
run into this issue of docs generating incorrectly until some cache was cleared https://github.com/latticexyz/mud/pull/2303#discussion_r1507441983

PR #2411, opened 2024-03-11
refactor(store): add StoreWrite and Store abstract contracts

yonadaaa:
Fixes https://github.com/latticexyz/mud/issues/2369

Added an `abstract` `StoreKernel` contract, which includes all Store interfaces except for registration, and implements write methods, `protocolVersion` and initializes `StoreCore`. `Store` extends `StoreKernel` with the `IStoreRegistration` interface. `StoreData` is removed as a separate interface/contract. `World` now extends `StoreKernel` (since the registration methods are added via the `InitModule`).


packages/store/src/StoreData.sol
```diff
@@ -14,7 +14,7 @@ import { IStoreEvents } from "./IStoreEvents.sol";
  * @dev This abstract contract initializes `StoreCore`, implements `storeVersion`, and read methods,
  * but not write methods.
  */
-abstract contract StoreData is IStoreData, StoreRead {
+abstract contract StoreData is StoreWrite, StoreRead {
```
holic, line 17:
```suggestion
abstract contract Store is StoreWrite, StoreRead {
```
🥺🙏

packages/store/src/IStore.sol
```diff
@@ -15,11 +16,18 @@ import { ISliceErrors } from "./ISliceErrors.sol";
...
   IPackedCounterErrors,
   ISchemaErrors,
   ISliceErrors
-{}
+{
+  /**
+   * @notice Returns the version of the Store contract.
+   * @return version The version of the Store contract.
+   */
+  function storeVersion() external view returns (bytes32 version);
```
holic, line 32: wonder if we should separate this into its own interface and have both `IStoreRead` and `IStoreWrite` inherit it as part of the protocol, or if it's fine just being here to signal "`IStore` is the protocol" and the rest are just "incomplete parts" of it

holic: basically wondering: is it valid to be able to read or write to a partial implementation of a store without also exposing a `storeVersion` (since the encoding may change between versions)?

holic:
docs need to be regenerated

yonadaaa:
> docs need to be regenerated (and prob an issue of https://github.com/latticexyz/mud/issues/2351 again)

It wasn't actually, it was because the docs tried to read `StoreData`

> 
I am now wondering if my protocol version tests are incorrect and should use IStore and IBaseWorld instead of the current approach (per our chat in Discord about "incomplete" implementations due to "diamond")

Interesting, so the tests should read different interfaces? As opposed to what?

packages/store/src/Store.sol
```diff
@@ -2,21 +2,22 @@
...
 import { IStoreEvents } from "./IStoreEvents.sol";
 
 /**
- * @title Store Data Contract
+ * @title Store Contract
  * @author MUD (https://mud.dev) by Lattice (https://lattice.xyz)
  * @notice This contract integrates the core storage functionalities and provides an interface for data storage.
  * @dev This abstract contract initializes `StoreCore`, implements `storeVersion`, and read methods,
  * but not write methods.
+ * Note that `Store` does not implement `IStore`, because we want `IWorldKernel` to inherit `IStore` but it does not implement all the Store methods.
```
alvrs, line 16: This comment is a bit confusing to me. Can you elaborate? Would the following be accurate:
```suggestion
 * Note that `Store` does not implement `IStore`, because the public write methods require access control that needs to be implemented by the consumer.
```

docs/pages/store/reference/store-hook.mdx
```diff
@@ -1,5 +1,22 @@
...
+```solidity
+bytes4 constant STORE_HOOK_INTERFACE_ID = IStoreHook.onBeforeSetRecord.selector ^
+  IStoreHook.onAfterSetRecord.selector ^
+  IStoreHook.onBeforeSpliceStaticData.selector ^
+  IStoreHook.onAfterSpliceStaticData.selector ^
+  IStoreHook.onBeforeSpliceDynamicData.selector ^
+  IStoreHook.onAfterSpliceDynamicData.selector ^
+  IStoreHook.onBeforeDeleteRecord.selector ^
+  IStoreHook.onAfterDeleteRecord.selector;
+```
```
holic, line 18: I dunno why this is showing up suddenly

.changeset/silly-mirrors-marry.md
```diff
@@ -0,0 +1,6 @@
+---
+"@latticexyz/store": minor
+"@latticexyz/world": patch
+---
+
+Added an `abstract` `StoreWrite` contract, and renamed `StoreData` to `Store`.
```
alvrs, line 6:
```suggestion
Added an `abstract` `StoreKernel` contract, which includes all Store interfaces except for registration, and implements write methods, `protocolVersion` and initializes `StoreCore`. `Store` extends `StoreKernel` with the `IStoreRegistration` interface. `StoreData` is removed as a separate interface/contract. `World` now extends `StoreKernel` (since the registration methods are added via the `InitModule`).
```

packages/world/src/World.sol
```diff
@@ -33,8 +33,11 @@ import { Balances } from "./codegen/tables/Balances.sol";
  * @author MUD (https://mud.dev) by Lattice (https://lattice.xyz)
  * @dev This contract is the core "World" contract containing various methods for
  * data manipulation, system calls, and dynamic function selector handling.
+ *
+ * Note World unfortunately cannot inherit Store because some Store methods are
+ * registered via modules instead of directly on this contract.
```
alvrs, line 38:
```suggestion
 * @dev World doesn't inherit `Store` because the `IStoreRegistration` methods are added via the `InitModule`.
```

packages/store/src/IStoreKernel.sol
```diff
@@ -0,0 +1,31 @@
...
+import { IStoreErrors } from "./IStoreErrors.sol";
+import { IFieldLayoutErrors } from "./IFieldLayoutErrors.sol";
+import { IEncodedLengthsErrors } from "./IEncodedLengthsErrors.sol";
+import { ISchemaErrors } from "./ISchemaErrors.sol";
+import { ISliceErrors } from "./ISliceErrors.sol";
+
+/**
+ * @title IStoreKernel
+ * @author MUD (https://mud.dev) by Lattice (https://lattice.xyz)
+ * @notice IStore includes the error interfaces for each library that it uses.
```
holic, line 15:
```suggestion
 * @notice IStoreKernel includes the error interfaces for each library that it uses.
```

packages/store/src/IStoreKernel.sol
```diff
@@ -0,0 +1,31 @@
...
+  IStoreWrite,
+  IStoreErrors,
+  IFieldLayoutErrors,
+  IEncodedLengthsErrors,
+  ISchemaErrors,
+  ISliceErrors
+{
+  /**
+   * @notice Returns the version of the Store contract.
+   * @return version The version of the Store contract.
```
holic, line 28:
```suggestion
   * @notice Returns the protocol version of the Store contract.
   * @return version The protocol version of the Store contract.
```

packages/store/src/IStore.sol
```diff
@@ -1,25 +1,12 @@
...
-import { IFieldLayoutErrors } from "./IFieldLayoutErrors.sol";
-import { IEncodedLengthsErrors } from "./IEncodedLengthsErrors.sol";
-import { ISchemaErrors } from "./ISchemaErrors.sol";
-import { ISliceErrors } from "./ISliceErrors.sol";
 
 /**
  * @title IStore
  * @author MUD (https://mud.dev) by Lattice (https://lattice.xyz)
- * @notice IStore implements the error interfaces for each library that it uses.
+ * @notice IStore includes the error interfaces for each library that it uses.
```
holic, line 10:
```suggestion
```

holic (approved):
small docs changes but otherwise good to go

PR #2413, opened 2024-03-11
feat: transactionContext mutex applied to system calls

MerkleBoy:
`call()`, `callFrom()` and `fallback()` will now log the initial `msg.sender` that made the first call to MUD; the values stored in `TransactionContext` are cleared once the transaction effectively exits the World.

To make sure that this context is properly initialized, directly calling a System instead of using `world.call()`, `world.callFrom()` or World's fallback() sub-routine has been explicitly forbidden, but _only if_ that direct call to a System result in an internal call to World. This is to prevent anyone from impersonating a System further down the call stack. Trying to do that will make the transaction revert with a custom `World_DirectCallToSystemForbidden` error. ...It's a bit overkill, but you shouldn't do it anyway so 🤷


ludns:
We could implement this using TSTORE and TLOAD

holic:
closing because no longer relevant (this code is now in user land)

Issue #1641, opened 2023-09-28
better recs support for composite keys


alvrs:

holic:
thought from our discussion this week: can we build this support on top of zustand instead of into recs?

PR #2416, opened 2024-03-12
feat(store-sync): recs queries on query-cache

yonadaaa:
does https://github.com/latticexyz/mud/issues/1641


packages/store-sync/src/query-cache/queryRECS.ts
```diff
@@ -0,0 +1,117 @@
...
+    .map(fragmentToQueryConditions)
+    .flat();
+
+  const result = await query(store, {
+    from,
+    except,
+    where,
+  });
+
+  return result.map((result) => result.join(":"));
```
holic, line 116: what's this do?

packages/store-sync/src/query-cache/queryRECS.ts
```diff
@@ -0,0 +1,118 @@
...
+    .map(fragmentToQueryConditions)
+    .flat();
+
+  const result = await query(store, {
+    from,
+    except,
+    where,
+  });
+
+  const entities = result.map((record) => record.join(":"));
```
holic, line 116: could use `encodeEntity`

but we might consider not using any sort of encoding and how we can just better support composite keys everywhere

packages/store-sync/src/query-cache/queryRECS.ts
```diff
@@ -0,0 +1,118 @@
...
+  value: SchemaToPrimitives<T["valueSchema"]>;
+};
+
+type NotValueQueryFragment<T extends Table> = {
+  type: QueryFragmentType.NotValue;
+  table: T;
+  value: SchemaToPrimitives<T["valueSchema"]>;
+};
+
+export function Has<T extends Table>(table: T): HasQueryFragment<T> {
```
holic, line 35: might be nice to think about how we can support customizing the "subject" (defaulting to primary key)

if you have a composite key on one table and you want to combine with another and normalize their "subjects" across the query (so that the query is valid/possible)

packages/store-sync/src/query-cache/queryRECS.ts
yonadaaa: `encodeEntity` is possible here but a little awkward because the records don't come with their `table` object

How would you support composite keys in RECS without breaking the API?

packages/store-sync/src/query-cache/queryRECS.ts
yonadaaa: If I understand, I'm not sure how to do this in RECS land - the assumption is that you operate over the entire key tuple so you can't specify subjects more finely

packages/store-sync/src/query-cache/queryRECS.ts
yonadaaa: done

yonadaaa:
Closing in favor of https://github.com/latticexyz/mud/pull/2438

PR #2421, opened 2024-03-13
feat(world): add config resolution at runtime

alvrs:
- adds the runtime implementation corresponding to #2389

### DIFF
--- a/packages/store/ts/config/index.ts
+++ b/packages/store/ts/config/index.ts
@@ -2,4 +2,4 @@ export * from "./defaults";
 export * from "./storeConfig";
 
 export * from "./experimental/resolveConfig";
-export * from "./v2/store";
+export * from "./v2";
--- a/packages/world/ts/config/v2/world.ts
+++ b/packages/world/ts/config/v2/world.ts
@@ -73,13 +73,13 @@ export type resolveWorldConfig<input> = evaluate<
 >;
 
 export function resolveWorldConfig<const input>(input: validateWorldConfig<input>): resolveWorldConfig<input> {
-  const namespaces = get(input, "namespaces", {});
+  const namespaces = get(input, "namespaces") ?? {};
   const scope = extendedScope(input);
 
   const namespacedTables = Object.fromEntries(
     Object.entries(namespaces)
       .map(([namespaceKey, namespace]) => {
-        const tables = get(namespace, "tables", {});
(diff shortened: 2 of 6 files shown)


packages/store/ts/config/index.ts
```diff
@@ -2,4 +2,4 @@ export * from "./defaults";
 export * from "./storeConfig";
 
 export * from "./experimental/resolveConfig";
-export * from "./v2/store";
+export * from "./v2";
```
holic, line 5: thoughts on a separate config entry point for now?

packages/world/ts/config/v2/world.ts
```diff
@@ -72,6 +73,30 @@ export type resolveWorldConfig<input> = evaluate<
...
-  // TODO: runtime implementation
-  return {} as never;
+  const namespaces = get(input, "namespaces") ?? {};
+  const scope = extendedScope(input);
+
+  const namespacedTables = Object.fromEntries(
+    Object.entries(namespaces)
+      .map(([namespaceKey, namespace]) => {
+        const tables = get(namespace, "tables") ?? {};
+        return Object.entries(tables).map(([tableKey, table]) => [`${namespaceKey}__${tableKey}`, table]);
```
holic, line 83: not sure if we wanna use this here but we have `resourceLabel` in `@latticexyz/common`

packages/world/ts/config/v2/world.ts
```diff
@@ -72,6 +73,30 @@ export type resolveWorldConfig<input> = evaluate<
...
+
+  const resolvedNamespaces = Object.fromEntries(
+    Object.entries(namespaces).map(([namespaceKey, namespace]) => [
+      namespaceKey,
+      { tables: resolveStoreTablesConfig(get(namespace, "tables") ?? {}, scope) },
+    ]),
+  );
+
+  return {
+    tables: resolveStoreTablesConfig({ ...(get(input, "tables") ?? {}), ...namespacedTables }, scope),
```
holic, line 96: do we need to do any validation that there are no overlaps?

holic (approved):

packages/store/ts/config/index.ts
holic: doin it here https://github.com/latticexyz/mud/pull/2422

PR #2423, opened 2024-03-13
feat(query): add query package

holic:
pulled out of https://github.com/latticexyz/mud/pull/2415


packages/common/src/type-utils/evaluate.ts
```diff
@@ -0,0 +1,2 @@
+// https://github.com/arktypeio/arktype/blob/93e79fa3d28567b7547e8e2df9a84e3c5b86e8e1/src/utils/generics.ts#L23
```
yonadaaa, line 1: thank you for linking this! i don't know what most of these helpers do

yonadaaa (approved):

packages/common/src/type-utils/evaluate.ts
holic: they're @ssalbdivad magic ✨ 

this one takes a complicated type like a union of a bunch of diff objects and named types and "collapses" it into a an object with properties, making it a bit easier to read as an output type

PR #2434, opened 2024-03-14
fix(store-sync): fix query-cache type

yonadaaa:
this fix causes a type error though 
image


holic:
ended up doing a diff approach in https://github.com/latticexyz/mud/pull/2436

PR #2437, opened 2024-03-14
feat(store): add transformer from v2 config output to v1 config output

alvrs:


test/mock-game-contracts/mud.config.ts
```diff
@@ -114,3 +114,5 @@ export const configV2 = resolveStoreConfig({
     },
   },
 });
+
+export default config;
```
yonadaaa, line 118: OOC what's the reason for this change? just a style thing?

alvrs: oh this was just a temporary thing to be able to inspect individual fields of the config output better

PR #2438, opened 2024-03-14
feat(store-sync): recs queries on query-cache

yonadaaa:


packages/store-sync/src/query-cache/common.ts
```diff
@@ -41,13 +41,13 @@ type queryConditions<tables extends Tables> = {
   [tableName in keyof tables]: tableConditions<tableName & string, tables[tableName]>;
 }[keyof tables];
 
+export type QuerySubjects<tables extends Tables = Tables> = {
+  readonly [k in keyof tables]?: readonly [keyof tables[k]["schema"], ...(keyof tables[k]["schema"])[]];
+};
```
holic, line 46: not sure if this is any more readable:
```suggestion
export type QuerySubject<table extends Table = Table> = readonly [keyof table["schema"], ...(keyof table["schema"])[]];

export type QuerySubjects<tables extends Tables = Tables> = {
  readonly [k in keyof tables]?: QuerySubject<tables[k]>;
};
```

packages/store-sync/src/query-cache/recs/Query.ts
```diff
@@ -0,0 +1,141 @@
...
+
+type NotQueryFragment<T extends Table> = {
+  type: QueryFragmentType.Not;
+  table: T;
+};
+
+type HasValueQueryFragment<T extends Table> = {
+  type: QueryFragmentType.HasValue;
+  table: T;
+  value: SchemaToPrimitives<T["valueSchema"]>;
```
holic, line 24: should switch these to `schema`

packages/store-sync/src/query-cache/common.ts
yonadaaa: done!

packages/store-sync/src/query-cache/recs/Query.ts
yonadaaa: interestinggg. really differs from RECS but seems better?

holic:
will be replaced by #2944

PR #2429, opened 2024-03-14
feat: devcontainer setup improvements

roaminro:
This PR improves the devcontainer setup for a quick dev environment setup:

- installs vscode extensions to improve developer experience (eslint, prettier and solidity)
- spins up a postgres instance
- forwards all needed ports to host machine

Happy to make any additional changes to improve it even more ;)

PR #2440, opened 2024-03-14
feat(create-mud): add additional recommended vscode extensions

roaminro:
This PR adds additional recommended vscode extensions for the templates and the multi-accounts example.

continued from #2429


holic (approved):
🙏

PR #2442, opened 2024-03-14
docs(world): much better illustrations from @turbo

qbzzt:


docs/pages/world/world-table.svg
holic, line 1: access control is done at the world level (based on namespace)

then the function on the system is called

qbzzt: Good point, we need to move the 2 square in the introduction. I'll do that in a bit.

alvrs (approved):

PR #2446, opened 2024-03-15
feat(world): add transformer from v2 config output to v1 config output

alvrs:


packages/world/ts/config/v2/output.ts
```diff
@@ -1,4 +1,47 @@
 import { Config as StoreConfig, Table } from "@latticexyz/store/config/v2";
+import { ModuleConfig } from "./input";
+
+export type SystemConfig = {
+  /** The full resource selector consists of namespace and name */
```
holic, line 5: is this an old comment? not sure what it's referring to

packages/world/ts/config/v2/output.ts
```diff
@@ -1,4 +1,47 @@
...
+  readonly openAccess: boolean;
+  /** An array of addresses or system names that can access the system */
+  readonly accessList: string[];
+};
+
+export type SystemsConfig = { readonly [key: string]: SystemConfig };
+
+export type DeploymentConfig = {
+  /** The name of the World contract to deploy. If no name is provided, a vanilla World is deployed */
+  readonly worldContractName: string | undefined;
```
holic, line 25: thoughts on naming this something like `customWorldContract`?

packages/world/ts/config/v2/output.ts
```diff
@@ -1,4 +1,47 @@
 import { Config as StoreConfig, Table } from "@latticexyz/store/config/v2";
+import { ModuleConfig } from "./input";
+
+export type SystemConfig = {
+  /** The full resource selector consists of namespace and name */
+  readonly name: string | undefined;
```
holic, line 6: when would this be undefined?

packages/world/ts/config/v2/output.ts
```diff
@@ -8,4 +51,13 @@ export type Config = StoreConfig & {
       };
     };
   };
+  readonly systems: SystemsConfig;
+  /** Systems to exclude from automatic deployment */
```
holic, line 55: system names?

holic (approved):
nothing blocking but would like to be mindful of naming+shapes of output types if we expect downstream users of em

packages/world/ts/config/v2/output.ts
alvrs: are you referring to the comment for `excludeSystems`?

holic: yeah, I think this is system contract names to exclude?

packages/world/ts/config/v2/output.ts
alvrs: yeah was an old comment, updated

packages/world/ts/config/v2/output.ts
alvrs: good catch, never undefined

PR #2454, opened 2024-03-16
docs(templates/getting-started): remove phaser link

qbzzt:


qbzzt:
Already done when quickstart was updated

PR #2456, opened 2024-03-17
chore: remove gitHead in package.json

holic:
Just cleaning up before release. I think this is an artifact of an old build pipeline (yarn + lerna?).


yonadaaa (approved):
sounds like a lerna artifact! https://github.com/lerna/lerna/issues/1880#issuecomment-456160357

PR #2460, opened 2024-03-18
feat(templates): add the foundry path if necessary

qbzzt:
`pnpm create` now installs foundry if necessary. That is great, but we can improve usability even more if we make sure it's in the path when we need it.

### DIFF
--- a/templates/react-ecs/package.json
+++ b/templates/react-ecs/package.json
@@ -2,14 +2,14 @@
   "name": "mud-template-react-ecs",
   "private": true,
   "scripts": {
-    "build": "pnpm recursive run build",
-    "dev": "mprocs",
+    "build": "forge || export PATH=$PATH:~/.foundry/bin; pnpm recursive run build",
+    "dev": "forge || export PATH=$PATH:~/.foundry/bin; mprocs",
     "dev:client": "pnpm --filter 'client' run dev",
     "dev:contracts": "pnpm --filter 'contracts' dev",
     "foundry:up": "curl -L https://foundry.paradigm.xyz | bash && bash $HOME/.foundry/bin/foundryup",
     "mud:up": "pnpm mud set-version --tag main && pnpm install",
     "prepare": "(forge --version || pnpm foundry:up)",
-    "test": "pnpm recursive run test"
+    "test": "forge || export PATH=$PATH:~/.foundry/bin; pnpm recursive run test"
   },
   "devDependencies": {
     "@latticexyz/cli": "link:../../packages/cli",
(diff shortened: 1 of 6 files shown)


alvrs:
Can you point me to where foundry is installed now? I think i missed that change

templates/react-ecs/package.json
```diff
@@ -2,14 +2,14 @@
   "name": "mud-template-react-ecs",
   "private": true,
   "scripts": {
-    "build": "pnpm recursive run build",
-    "dev": "mprocs",
+    "build": "forge || export PATH=$PATH:~/.foundry/bin; pnpm recursive run build",
+    "dev": "forge || export PATH=$PATH:~/.foundry/bin; mprocs",
```
alvrs, line 6: i think this would feel cleaner if we moved `forge || export PATH=$PATH:~/.foundry/bin;` into a separate pnpm script that we then use in the places where we need forge

qbzzt:
> Can you point me to where foundry is installed now? I think i missed that change

[It's in the templates](https://github.com/latticexyz/mud/blob/main/templates/react/package.json#L11). I don't know why `pnpm run prepare` happens, but I used virgin VMs and I saw it does auto-install foundry.

templates/react-ecs/package.json
qbzzt: It doesn't work like that. We need to change `$PATH`, but if we run a separate pnpm script, it runs in a different shell, so the changes to `$PATH` don't propagate to where we need them. We could create a separate script file and just do `". <new script>.sh; mprocs"` if you think that's cleaner, the `.` (or `source`, depending on the shell used) runs the script in the existing shell - but you need a shell script for that, not a pnpm script.

alvrs:
I'm not a fan of adding this snippet to all the scripts, makes it look much more complex than it has to be. What if we add something to the `prepare` script that verifies that `forge` is available after `foundry:up` ran, and if it isn't we print a hint for how to make forge available in the shell?

qbzzt:
> I'm not a fan of adding this snippet to all the scripts, makes it look much more complex than it has to be. What if we add something to the `prepare` script that verifies that `forge` is available after `foundry:up` ran, and if it isn't we print a hint for how to make forge available in the shell?

There's no point. When the `prepare` script runs `foundryup`, that program tells them how to add to the path if necessary:

```
qbzzt1@instance-20240327-182132:~$ curl -L https://foundry.paradigm.xyz | bash && bash $HOME/.foundry/bin/foundryup
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100  2189  100  2189    0     0  11066      0 --:--:-- --:--:-- --:--:-- 11066
Installing foundryup...

Detected your preferred shell is bash and added foundryup to PATH.
Run 'source /home/qbzzt1/.bashrc' or start a new terminal session to use foundryup.
Then, simply run 'foundryup' to install Foundry.
```

The whole point is to help people where they need the help, when they run `pnpm dev` or `pnpm build`. If you think that usability improvement is not worth the extra complexity just close this PR.

Issue #2319, opened 2024-02-27
Library auto-generation for non-root Modules


MerkleBoy:
### Below, a proposal for a steamlined way to call non-root Systems in a way that abstracts away the namespace it's deployed onto:

One of the main concern I have with registering some System methods as World-function selectors are the following: 

- First, registering those methods as World-function selectors has limits since the available space is only bytes4 (~4,294,967,295 possible function selectors), and so by the time we reach 65536 registered method in the World, the collision chance will be about 50% (birthday paradox). Even though this number seems high, it's really not *that* high once a given World's "builder" scene takes off.
So, if we had a way that makes *not* registering function selectors at the World level a viable alternative to build and interact with it, e.g., libraries to inline `world.call(SystemId, abi.encodeCall(...))` for them, it would be a preferable approach.

- Secondly, calling methods through World-function selectors implies that, in the event a Module is registered more than once, it will break that interface. Right now, MUD's CLI auto-generates name-spaced method ( `myNamespace__foo()` ), so if someone was building on top of those methods, and the module ends up being redeployed somewhere else, it will break the interface.

Transforming a System interface into it's `world.call(..)` equivalent is pretty straightforward, right now it needs to be done by hand, which could lead to mismatching errors if said interfaces have to be updated; so following that logic we need something like `mud worldgen` or `mud tablegen`, so that libraries can be auto-generated in one command line.

`mud modulegen` perhaps ?

Let's study a dummy system `DummySystem` to see how it plays out : 

```ts
export default mudConfig({
    namespace: "MySystem_v0",
    systems: {
        DummySystem: {
            name: "DummySystem,
            openAccess: true,
        },
    },
    tables: {
        MyTable: {
            keySchema: {
                a: "uint256",
            }
            valueSchema: {
                b: "uint256",
            },
            storeArgument: true,
            tableIdArgument: true,
        },
    },
});
```

```solidity
// DummySystem.sol
contract DummySystem is System {
  function foo(uint256 a, uint256 b) public {
    MyTable.set(_namespace().myTableTableId(), a, b);
  }

  function bar(uint256 a) public returns (uint256 b) {
    return MyTable.get(_namespace().myTableId(), a);
  }
}
```
```solidity
// Utils.sol
library Utils {
  using WorldResourceIdInstance for ResourceId;

  function dummySystemId(bytes14 namespace) internal pure returns (ResourceId) {
[...]

PR #2461, opened 2024-03-18
WIP world call library generation

dk1a:
fixes https://github.com/latticexyz/mud/issues/2319

some issues/TODOs:
- namespace argument
- world argument (these can even be options, like `storeArgument`)
- separation of worldgen and libgen(?), if it's even necessary - they both share the `contractToInterface` call
- all lib methods have highest mutability (not view/pure), because of how World's `call` works
- not sure `encodeCall` will work correctly with overloaded signatures


vdrg:
Closing in favor of #3374

PR #2464, opened 2024-03-18
refactor(store, world): separate shorthand config from full config resolvers and cleanup

alvrs:
- Separating the shorthand resolvers from the main resolvers to improve maintainability
- Cleaning up all types
- Reduce complexity of internal types for performance improvement (8s -> 2.8s for the world config, 3.55s -> 2.1s for the store config)


packages/store/ts/config/v2/README.md
```diff
@@ -0,0 +1,47 @@
+# Config conventions
+
+These are the types and functions that should be placed in a file called `x.ts`
+
+```ts
+/**
+ * validateX returns input if the input is valid and expected type otherwise.
```
holic, line 7: do we wanna add anything about _why_ this is the approach we use? returning valid input makes sense, returning expected type is less obvious to me

packages/store/ts/config/v2/generics.ts
```diff
@@ -18,3 +20,24 @@ export function hasOwnKey<obj, const key extends PropertyKey>(
 export function isObject<input>(input: input): input is input & object {
   return input != null && typeof input === "object";
 }
+
+export type mergeIfUndefined<base, merged> = merge<
```
holic, line 24: what about `defaults` or `applyDefaults`?

packages/store/ts/config/v2/input.ts
holic, line 1: nice to see this all in one place!

packages/store/ts/config/v2/table.ts
```diff
@@ -1,99 +1,184 @@
...
+    Array.isArray(key) &&
+    key.every(
+      (key) =>
+        hasOwnKey(schema, key) && hasOwnKey(scope.types, schema[key]) && isStaticAbiType(scope.types[schema[key]]),
+    )
+  );
+}
+
+/** @deprecated */
+export function isTableInput(input: unknown): input is TableInput {
```
holic, line 41: should we get rid of this if it's deprecated?

packages/store/ts/config/v2/table.ts
```diff
@@ -1,99 +1,184 @@
...
+      ? boolean
+      : key extends keyof typeof TABLE_CODEGEN_DEFAULTS
+        ? (typeof TABLE_CODEGEN_DEFAULTS)[key]
+        : never;
+};
+
+export function resolveTableCodegen<input extends TableInput>(input: input): resolveTableCodegen<input> {
+  const options = input.codegen;
+  return {
+    directory: get(options, "directory") ?? TABLE_CODEGEN_DEFAULTS.directory,
```
holic, line 113: I wonder if we should name this something like `output` or `outputDirectory` or `outDir`

packages/store/ts/config/v2/input.ts
holic: we should prob add tsdoc everywhere so as you type, you get hints as to what we expect each value to be, esp for things like codegen where the config keys are a bit nuanced (like why you'd wanna enable a table ID argument)

packages/world/ts/config/v2/codegen.ts
```diff
@@ -0,0 +1,12 @@
+import { isObject, mergeIfUndefined } from "@latticexyz/store/config/v2";
+import { CODEGEN_DEFAULTS } from "./defaults";
```
holic, line 2: we should decide if we wanna SHOUT our constants or not, feels not necessary in today's world with `const`

holic (approved):
this is looking _much_ easier to reason about and love the conventions doc

nothing blocking! feel free to add notes to tracker if you wanna do later

I wonder if we can isolate the "shorthand plugin" even more, because it feels a bit deeply tied into everything right now. Like `expandTableShorthand` that takes in store input with optional shorthand tables and gives back store input with regular table defs? Can also iterate towards this too, not needed here.

packages/store/ts/config/v2/README.md
alvrs: returning the expected type so that typescript compares the input to the expected type!

if the function is `defineConfig<input>(input: validateInput<input>)` then what you're passing in has to match `validateInput<input>`. If `validateInput<input>` is happy it returns `input`, so it matches; if it's not happy it returns what it expected, so the input param basically becomes `defineConfig<input>(input: Expected)`, typescript will tell you why `input` doesn't match `Expected`

packages/world/ts/config/v2/codegen.ts
alvrs: no strong opinion here

alvrs:
> I wonder if we can isolate the "shorthand plugin" even more, because it feels a bit deeply tied into everything right now. Like expandTableShorthand that takes in store input with optional shorthand tables and gives back store input with regular table defs, all contained/isolated in a shorthand dir, then defineStore/defineWorld can use that as an inner wrapper? Can also iterate towards this too, not needed here.

That's essentially what `storeWithShorthands` does no? Turning the shorthand inputs into regular inputs. The reason why it's not _just_ that is so we still get the type errors in the right places (rather than a big fat red underline across the entire output of the inner resolver). So what it does is apply the store validation, but replace the table validation with the `tableWithShorthand` validation

packages/store/ts/config/v2/README.md
holic: great, do we wanna document that here?

PR #2470, opened 2024-03-19
feat: use new config

holic:


packages/store-sync/src/postgres/createStorageAdapter.ts
```diff
@@ -92,7 +92,7 @@ export async function createStorageAdapter<TConfig extends StoreConfig = StoreCo
           const previousStaticData = previousValue?.staticData ?? "0x";
           const newStaticData = spliceHex(previousStaticData, log.args.start, size(log.args.data), log.args.data);
 
-          debug("upserting record via splice static", {
+          debug("upcserting record via splice static", {
```
yonadaaa, line 95:
```suggestion
          debug("upserting record via splice static", {
```

holic:
starting over in #2483 with minimal changeset

PR #2472, opened 2024-03-19
feat(store/world): disable `namespace`/`name` override on store/world table inputs, move validation to define

alvrs:
- Adds an explicit type error to validators when trying to override `namespace` or `name` when defining tables in a store/world config. `name`/`namespace` is still on the `TableInput` type so we can use it internally.
- Slightly refactored the validator pattern to move runtime validation to `defineX` and expect a fully validated input in `resolveX`
- Updates the store resolver to map tables to a prefixed key if a `namespace` option is provided in the config

### DIFF
--- a/packages/store/ts/config/v2/input.ts
+++ b/packages/store/ts/config/v2/input.ts
@@ -26,7 +26,7 @@ export type TablesInput = {
 
 export type StoreInput = {
   readonly namespace?: string;
-  readonly tables: TablesInput;
+  readonly tables: Omit<TablesInput, "name" | "namespace">;
   readonly userTypes?: UserTypes;
   readonly enums?: Enums;
   readonly codegen?: Partial<Codegen>;
--- a/packages/store/ts/config/v2/input.ts
+++ b/packages/store/ts/config/v2/input.ts
@@ -21,12 +21,12 @@ export type TableInput = {
 };
 
 export type TablesInput = {
-  readonly [key: string]: TableInput;
+  readonly [key: string]: Omit<TableInput, "namespace" | "name">;
 };
 
 export type StoreInput = {
   readonly namespace?: string;
-  readonly tables: Omit<TablesInput, "name" | "namespace">;
+  readonly tables: TablesInput;
   readonly userTypes?: UserTypes;
   readonly enums?: Enums;
   readonly codegen?: Partial<Codegen>;
--- a/packages/store/ts/config/v2/namespaces.ts
+++ b/packages/store/ts/config/v2/namespaces.ts
@@ -1,9 +1,9 @@
 export type namespacedTableKeys<store> = "tables" extends keyof store
-  ? "namespace" extends keyof store
-    ? store["namespace"] extends string
-      ? "" extends store["namespace"]
-        ? keyof store["tables"]
-        : `${store["namespace"] & string}__${keyof store["tables"] & string}`
-      : keyof store["tables"]
-    : keyof store["tables"]
+  ? keyof {
+      [key in keyof store["tables"] as "namespace" extends keyof store
+        ? store["namespace"] extends ""
+          ? key
+          : `${store["namespace"] & string}__${key & string}`
+        : key]: void;
+    }
   : never;
--- a/packages/store/ts/config/v2/namespaces.ts
+++ b/packages/store/ts/config/v2/namespaces.ts
@@ -1,0 +1,9 @@
+export type namespacedTableKeys<store> = "tables" extends keyof store
+  ? "namespace" extends keyof store
+    ? store["namespace"] extends string
+      ? "" extends store["namespace"]
+        ? keyof store["tables"]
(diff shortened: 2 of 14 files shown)


packages/store/ts/config/v2/input.ts
```diff
@@ -26,7 +26,7 @@ export type TablesInput = {
 
 export type StoreInput = {
   readonly namespace?: string;
-  readonly tables: TablesInput;
+  readonly tables: Omit<TablesInput, "name" | "namespace">;
```
holic, line 29: this is `TablesInput` so it's an object with tables, so how does this work? don't we have to map over it and omit? or put this in `TableInput`?

holic: also tiny nit: I tend to order these as namespace then name

packages/store/ts/config/v2/namespaces.ts
```diff
@@ -0,0 +1,9 @@
+export type namespacedTableKeys<store> = "tables" extends keyof store
+  ? keyof {
+      [key in keyof store["tables"] as "namespace" extends keyof store
```
holic, line 3:
```suggestion
      [tableName in keyof store["tables"] as "namespace" extends keyof store
```
I've been trying to use named keys rather than just `key` for clarity

packages/store/ts/config/v2/store.ts
```diff
@@ -42,10 +42,9 @@ export type resolveStore<store> = evaluate<{
   readonly tables: "tables" extends keyof store
     ? resolveTables<
         {
-          [key in keyof store["tables"]]: mergeIfUndefined<
-            store["tables"][key],
-            { namespace: get<store, "namespace"> }
-          >;
+          [key in namespacedTableKeys<store>]: key extends `${string}__${infer table}`
+            ? mergeIfUndefined<get<get<store, "tables">, table>, { namespace: get<store, "namespace">; name: table }>
+            : mergeIfUndefined<get<get<store, "tables">, key>, { namespace: get<store, "namespace">; name: key }>;
```
holic, line 47: why do we do this in reverse (namespace the table keys then extract the table name)? vs something like

```ts
{
  [tableName in keyof store['tables'] as resourceToLabel<store['namespace'], tableName>: ...
}
```

packages/store/ts/config/v2/input.ts
alvrs: ah yes you're right, this doesn't do anything

packages/store/ts/config/v2/store.ts
```diff
@@ -56,12 +55,15 @@ export type resolveStore<store> = evaluate<{
...
-export function resolveStore<const store>(store: store): resolveStore<store> {
-  validateStore(store);
-
+export function resolveStore<const store extends StoreInput>(store: store): resolveStore<store> {
   return {
     tables: resolveTables(
-      mapObject(store.tables ?? {}, (table) => mergeIfUndefined(table, { namespace: store.namespace })),
+      Object.fromEntries(
+        Object.entries(store.tables ?? {}).map(([tableKey, table]) => {
+          const key = store.namespace ? `${store.namespace}__${tableKey}` : tableKey;
```
holic, line 63: not sure if we wanna lean on this here but we have `resourceToLabel` helper

packages/store/ts/config/v2/store.ts
alvrs: true, this is the better approach here! had started from the world version where it's not possible because we change the shape of the object, but here it is possible

packages/store/ts/config/v2/table.ts
```diff
@@ -37,28 +37,25 @@ export function isValidPrimaryKey<schema extends SchemaInput, scope extends Scop
...
   [i in keyof keys]: keys[i] extends validKeys ? keys[i] : validKeys;
 };
 
-export type validateTable<input, scope extends Scope = AbiTypeScope> = {
+export type ValidateTableOptions = { inStoreContext: boolean };
+
+export type validateTable<
+  input,
+  scope extends Scope = AbiTypeScope,
+  options extends ValidateTableOptions = { inStoreContext: false },
```
holic, line 49: 😬 don't love this but it's prob a big refactor otherwise?

holic: ooc why wouldn't this be on scope?

holic (dismissed):
maybe a bug with `StoreInput` type and using `Omit` on `TablesInput`, some nits, and mostly questions

packages/store/ts/config/v2/store.ts
alvrs: normally i would be down for that but in this particular case i feel like keeping the type and runtime implementation as similar as possible would be better

packages/store/ts/config/v2/table.ts
alvrs: yeah could be on scope!

Issue #2469, opened 2024-03-19
update threejs template


yonadaaa:
The example used in the threejs template is a bit obscure and could be clearer. 

Also the player character doesn't render until their Position is set for the first time by moving the arrow keys.
![image](https://github.com/latticexyz/mud/assets/29184158/ee2f1d58-c256-41da-acf8-1e7306003179)

yonadaaa:
done in https://github.com/latticexyz/mud/pull/2473

PR #2473, opened 2024-03-19
feat(templates/threejs): easier to use template

qbzzt:
Closing: #2469


templates/threejs/packages/client/src/App.tsx
```diff
@@ -5,23 +5,106 @@ import { getComponentValueStrict, Has } from "@latticexyz/recs";
...
+const Plane = () => {
   return (
-    <mesh {...props}>
+    <>
       {/* eslint-disable-next-line react/no-unknown-property */}
-      <boxGeometry args={[10, 5, 10]} />
-      <meshStandardMaterial color="green" />
-    </mesh>
+      <mesh position={[0, -1, 0]}>
+        {/* eslint-disable-next-line react/no-unknown-property */}
```
holic, line 13: can we figure out how to get react+eslint to allow this? I assume there's a three.js eslint config to include here

qbzzt: Done.

qbzzt: And the solution I found online doesn't work. Can we *please* just not worry about it? Prettier here is more trouble than it is worth, IMAO.

yonadaaa: I've always run into this issue with ThreeJS and TypeScript, I'll play around and try fix it. Also I get errors for all the elements so we are going to need a lot of these comments otherwise!
image

yonadaaa: Sounds like an open issue and the [@react-three/eslint-plugin](https://socket.dev/npm/package/@react-three/eslint-plugin) didn't work for me: https://github.com/pmndrs/react-three-fiber/discussions/2487

For now I added a comment at the top of the file but we could also add a `eslintrc` rule like 
```
  ts"react/no-unknown-property": ["off", { "ignore": ["JSX"] }],
```

yonadaaa (approved):

yonadaaa:
Tweaked the topbar colors: 
image

holic:
@yonadaaa need a changeset for "create-mud"

PR #2474, opened 2024-03-19
feat(store): support fixed length arrays in v2 config

holic:


packages/schema-type/src/typescript/arrayAbiTypes.ts
```diff
@@ -1,17 +1,42 @@
...
   ? StaticAbiType
   : never;
 
-export function isArrayAbiType<T extends `${StaticAbiType}[]`>(abiType: string): abiType is T {
-  return arrayAbiTypePattern.test(abiType);
+export type FixedArrayAbiTypeToStaticAbiType<T extends string> = T extends `${infer StaticAbiType}[${number}]`
+  ? StaticAbiType
+  : never;
+
+export function isArrayAbiType<T extends ArrayAbiType>(abiType: string): abiType is T {
```
alvrs, line 18: i would expect a type guard like this to accept an `unknown` input, otherwise typescript will complain if your input might not be an `ArrayAbiType`, which is what the type guard is supposed to check

packages/schema-type/src/typescript/arrayAbiTypes.ts
```diff
@@ -1,17 +1,42 @@
...
+  return arrayPattern.test(abiType) && isStaticAbiType(abiType.replace(arrayPattern, ""));
+}
+
+export function isFixedArrayAbiType<T extends FixedArrayAbiType>(abiType: string): abiType is T {
```
alvrs, line 22: same here, would expect something like this:
```suggestion
export function isFixedArrayAbiType<T>(abiType: T): abiType is T & FixedArrayAbiType {
```

packages/schema-type/src/typescript/arrayAbiTypes.ts
```diff
@@ -1,17 +1,42 @@
...
+import { isStaticAbiType } from "./staticAbiTypes";
 
-const arrayAbiTypePattern = /\[\]$/;
+export type ArrayAbiType = `${StaticAbiType}[]`;
+export type FixedArrayAbiType = `${StaticAbiType}[${number}]`;
+
+const arrayPattern = /\[\]$/;
+const fixedArrayPattern = /\[\d+\]$/;
 
 export type ArrayAbiTypeToStaticAbiType<T extends string> = T extends `${infer StaticAbiType}[]`
```
alvrs, line 10: this was already here so might not be worth refactoring, but according to our new implicit typescript style guide this should be lowercase

packages/schema-type/src/typescript/arrayAbiTypes.ts
holic: this was already like this but yeah, can expand to match the other `isStaticAbiType`, etc helpers in this package

holic: oh I see what you mean now with the comment below

alvrs (approved):

PR #2477, opened 2024-03-20
fix: store-sync uses `getKeySchema` helper

yonadaaa:
Looks like a couple areas were missed in https://github.com/latticexyz/mud/pull/2475

### DIFF
--- a/packages/store-sync/tsup.config.ts
+++ b/packages/store-sync/tsup.config.ts
@@ -13,7 +13,7 @@ export default defineConfig({
   ],
   target: "esnext",
   format: ["esm"],
-  dts: false,
+  dts: true,
   sourcemap: true,
   clean: true,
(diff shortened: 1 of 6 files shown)


packages/store-sync/tsup.config.ts
```diff
@@ -13,7 +13,7 @@ export default defineConfig({
   ],
   target: "esnext",
   format: ["esm"],
-  dts: true,
+  dts: false,
```
holic, line 16: prob should leave this on for now

holic, line 16:
```suggestion
  dts: true,
```

yonadaaa: oops! good spot ty

holic (approved):

PR #2478, opened 2024-03-20
feat(protocol-parser): add new schema helpers

holic:
pulled out of https://github.com/latticexyz/mud/pull/2470


packages/store/ts/config/v2/table.test.ts
```diff
@@ -186,6 +187,7 @@ describe("resolveTable", () => {
   });
 });
 
+// TODO: move tests to protocol parser after we add arktype
```
holic, line 190: https://github.com/latticexyz/mud/issues/2479

yonadaaa (dismissed):
lgtm, we can move tests later

Issue #2383, opened 2024-03-06
Register/unregister delegates via delegate call + delegator signature


0xhank:
## Proposal
Allow the following flow for initializing an account delegation:
1. User controls a low risk account (A) and wants to interface with MUD using a high-risk 'session account' (B)
2. User signs a message with A that states it is giving B delegation access
3. B submits a tx registering/unregistering B as a delegate to A. A's signature is attached. 
4. MUD confirms the signature is valid and grants execution of the delegation function.
5. Delegation proceeds as normal.

## Why
Primodium's biggest onboarding hurdle has been players running into every bug imaginable to submit the first tx to spawn and delegate (in our case these calls are batched). Some issues include:
- Chain id not recognized (WalletConnect/Coinbase)
- CoinId not recognized (okx)
- Unknown Ethereum Address when clicking play
- Transaction failed
ETC.

It would greatly improve UX and streamline our user flow.

yonadaaa:
Adding signatures is getting closer full-blown account abstraction, IMO. We should see if we can generalise this and/or integrate with EIP's like [ERC4337](https://www.erc4337.io/)

alvrs:
Added in #2480 (for now in an optional module, moving to the core world contract once audited)

PR #2480, opened 2024-03-20
feat(world-modules): register delegation with signature

yonadaaa:
Addresses https://github.com/latticexyz/mud/issues/2383

This PR adds a new `registerDelegationWithSignature` function to the World:

```solidity
function registerDelegationWithSignature(
  address delegatee,
  ResourceId delegationControlId,
  bytes memory initCallData,
  address delegator,
  bytes memory signature
) public onlyDelegatecall;
```

and a new `UserDelegationNonces` table, to prevent reply attacks:

```ts
UserDelegationNonces: {
  schema: { delegator: "address", nonce: "uint256" },
  key: ["delegator"],
}
  ```
Anyone can call `registerDelegationWithSignature` to register a delegation for `delegator`, provided that the given [EIP-1271](https://eips.ethereum.org/EIPS/eip-1271) signature is valid (ie. was signed for the same contract address, chain ID, nonce, delegation details etc.)

If the signature is valid, the delegation is created and the nonce is incremented.

### CODE REVIEW

File: docs/pages/world/reference/internal/init-module-implementation.mdx [hunk 1]
```diff
@@ -475,6 +475,18 @@ function registerDelegation(
...
 
+#### registerDelegationWithSignature
+
+```solidity
+function registerDelegationWithSignature(
+  address delegatee,
+  ResourceId delegationControlId,
+  bytes memory initCallData,
+  address delegator,
+  bytes memory signature
```
karooolis, line 486:
```suggestion
  bytes calldata signature
```

Uses slightly less gas and doesn't change the functionality. `calldata` can be used on params that do not get altered inside the function.

packages/world/src/modules/init/implementations/WorldRegistrationSystem.sol
```diff
@@ -264,27 +268,28 @@ contract WorldRegistrationSystem is System, IWorldErrors, LimitedCallContext {
...
-      });
+    createDelegation(_msgSender(), delegatee, delegationControlId, initCallData);
+  }
+
+  function registerDelegationWithSignature(
+    address delegatee,
+    ResourceId delegationControlId,
+    bytes memory initCallData,
+    address delegator,
+    bytes memory signature
```
karooolis, line 279:
```suggestion
    bytes calldata signature
```

Uses slightly less gas and doesn't change the functionality. calldata can be used on params that do not get altered inside the function.

yonadaaa: Thanks! for some reason I thought we didn't do this in MUD but it totally works

packages/world/src/modules/init/implementations/getSignedMessageHash.sol
```diff
@@ -0,0 +1,29 @@
...
+);
+
+function getSignedMessageHash(
+  address delegatee,
+  ResourceId delegationControlId,
+  bytes memory initCallData,
+  uint256 nonce,
+  address worldAddress
+) view returns (bytes32) {
+  bytes32 domainSeperator = keccak256(
```
karooolis, line 17: You may cache the `domainSeparator` for some gas savings. It'd be something like this:

```
constructor() {
       ....

        cachedChainId = block.chainid;
        cachedThis = address(this);
        cachedDomainSeparator = _constructDomainSeparator();
}

function _getDomainSeparator() internal view returns (bytes32) {
    if (address(this) == cachedThis && block.chainid == cachedChainId) {
        return cachedDomainSeparator;
    } else {
        return _constructDomainSeparator();
    }
}

/// @notice Construct the domain separator
function _constructDomainSeparator() internal view returns (bytes32) {
    return keccak256(abi.encode(DOMAIN_TYPEHASH, block.chainid, address(this)));
}
```

holic: is keccak so expensive that its worth the sstore/sread?

yonadaaa: Yup I saw this pattern in [OpenZeppelin's implementation](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography/EIP712.sol). It's awkward here with MUD's proxy pattern (we can't use the constructor), but definitely want to optimize this after the first pass on the PR

karooolis: You're right. It might have worked if `DOMAIN_TYPEHASH` was made of more arguments, good call.

e2e/packages/sync-test/data/getWorld.ts
yonadaaa, line 1: Added this utility to get the world address (as it is part of the signature)

e2e/packages/sync-test/registerDelegationWithSignature.test.ts
yonadaaa, line 1: Testing that we can generate a signature with Viem and verify it in Solidity onchain

packages/world/ts/data.ts
yonadaaa, line 1: This `types` are the same for anyone generating a registration signature. We could even build a more complete `signDelegation` wrapper that takes a wallet client and generates a signature (kinda like `encodeSystemCall`)

copied the naming from https://viem.sh/docs/actions/wallet/signTypedData.html#usage

packages/world/src/modules/init/implementations/getSignedMessageHash.sol
```diff
@@ -0,0 +1,29 @@
...
+    abi.encode(keccak256("EIP712Domain(uint256 chainId,address verifyingContract)"), block.chainid, worldAddress)
```
yonadaaa, line 18: With this domain separator, signatures are unique across world addresses and chains. The EIP has optional `name` and `version` fields but i think they only makes sense if we were upgrading the signature scheme? https://eips.ethereum.org/EIPS/eip-712#definition-of-domainseparator

e2e/packages/contracts/mud.config.ts
```diff
@@ -50,4 +50,11 @@ export default defineWorld({
       key: [],
     },
   },
+  modules: [
+    {
+      name: "DelegationModule",
```
alvrs, line 55: We should call this something more specific, maybe `DelegationWithSignatureModule`

e2e/packages/sync-test/registerDelegationWithSignature.test.ts
```diff
@@ -0,0 +1,110 @@
...
+import { mudFoundry } from "@latticexyz/common/chains";
+import { encodeEntity } from "@latticexyz/store-sync/recs";
+import { callPageFunction } from "./data/callPageFunction";
+import worldConfig from "@latticexyz/world/mud.config";
+import { worldToV1 } from "@latticexyz/world/config/v2";
+import { types } from "@latticexyz/world/internal";
+import { getWorld } from "./data/getWorld";
+import { callRegisterDelegationWithSignature } from "./data/callRegisterDelegationWithSignature";
+
+const PRIVATE_KEY = "0x67bbd1575ecc79b3247c7d7b87a5bc533ccb6a63955a9fefdfaf75853f7cd543";
```
alvrs, line 19: should be more specific with the naming for what this constant is used for, maybe `BURNER_WALLET_PRIVATE_KEY`?

e2e/packages/sync-test/registerDelegationWithSignature.test.ts
```diff
@@ -0,0 +1,110 @@
...
+    const account = createBurnerAccount(PRIVATE_KEY);
+    const walletClient = createWalletClient({
+      ...clientOptions,
+      account,
+    });
+
+    const worldContract = await getWorld(page);
+
+    // Sign registration message
+    const delegatee = "0x7203e7ADfDF38519e1ff4f8Da7DCdC969371f377";
```
alvrs, line 65: where is this address coming from? is it `account.address`?

packages/world/ts/data.ts
```diff
@@ -0,0 +1,10 @@
+// Follows https://viem.sh/docs/actions/wallet/signTypedData#usage
+
+export const types = {
```
alvrs, line 3: A constant called `types` (as a global export from the package) is way to generic of a name here. This is the type of the message that is used as payload for `registerDelegationWithSignature`? We should reflect that in the variable name. My preference would be to call the constant `DelegationSignature` and use it as `{ DelegationSignature }` in `signTypedData`. I think we should also rename the file from `data.ts` to something more specific.

e2e/packages/sync-test/registerDelegationWithSignature.test.ts
```diff
@@ -0,0 +1,110 @@
...
+      encodeEntity(worldConfigV1.tables.UserDelegationControl.keySchema, {
+        delegator: account.address,
+        delegatee,
+      }),
+    ]);
+
+    expect(value).toMatchInlineSnapshot(`
+        {
+          "__staticData": "0x73790000000000000000000000000000756e6c696d6974656400000000000000",
+          "delegationControlId": "0x73790000000000000000000000000000756e6c696d6974656400000000000000",
```
alvrs, line 106: It's hard for me to confirm that this is the expected data. Could we use something like `expect(value).toEqual({ __staticData: deletationControlId })` or something more explicit here?

e2e/packages/contracts/mud.config.ts
yonadaaa: Agreed, trying to find a balance bc `registerDelegationWithSignature` is so long to write

packages/world/ts/data.ts
yonadaaa: Agreed! Was just being lazy and copying the naming from the viem example

e2e/packages/sync-test/registerDelegationWithSignature.test.ts
yonadaaa: It's just a made up address! it doesn't matter who the delegatee is. 

I'll make this more clear

packages/world-modules/src/modules/delegation/DelegationSystem.sol
```diff
@@ -0,0 +1,47 @@
...
+   */
+  function registerDelegationWithSignature(
+    address delegatee,
+    ResourceId delegationControlId,
+    bytes memory initCallData,
+    address delegator,
+    bytes memory signature
+  ) public {
+    uint256 nonce = UserDelegationNonces.get(delegator);
+    bytes32 hash = getSignedMessageHash(delegatee, delegationControlId, initCallData, nonce, _world());
```
alvrs, line 35: should the `delegator` be part of the signed message?

e2e/packages/sync-test/registerDelegationWithSignature.test.ts
yonadaaa: This example is confusing because there are three accounts:
- `delegator`, which is generated during the test and signs the message
- `delegatee`, which is just generated during the test to fill in the delegatee value
- `walletClient.account.address`, which actually registers the delegation onchain but is otherwise not involved

packages/world-modules/src/modules/delegation/registerDelegationWithSignature.sol
alvrs, line 1: do we need this wrapper?

packages/world-modules/test/DelegationModule.t.sol.sol
```diff
@@ -0,0 +1,106 @@
...
+    vm.prank(delegator);
+    world.unregisterDelegation(delegatee);
+
+    // Expect a revert when attempting to perform a call via callFrom after a delegation was unregistered
+    vm.expectRevert(abi.encodeWithSelector(IWorldErrors.World_DelegationNotFound.selector, delegator, delegatee));
+    vm.prank(delegatee);
+    world.callFrom(delegator, systemId, abi.encodeCall(WorldTestSystem.msgSender, ()));
+
+    // Attempt to register a limited delegation using an old signature
+    vm.expectRevert(abi.encodeWithSelector(DelegationSystem.InvalidSigner.selector, delegator, delegatee));
```
alvrs, line 86: I guess this error is more like `InvalidSignature`? (ie in this case the signer the same but the nonce is outdated, so the signature is not correct anymore?)

packages/world-modules/src/modules/delegation/DelegationSystem.sol
yonadaaa: Good question. My intuition is no because the `delegator` is always the signer. 

If someone submits a signature that was signed by an different account to the provided`delegator`, `ecrecover` will fail. We don't ALSO need to put it in the signature.

packages/world-modules/test/DelegationModule.t.sol.sol
```diff
@@ -0,0 +1,106 @@
...
+    hash = getSignedMessageHash(delegatee, UNLIMITED_DELEGATION, new bytes(0), 1, address(world));
+    (v, r, s) = vm.sign(delegatorPk, hash);
+    signature = abi.encodePacked(r, s, v);
+
+    registerDelegationWithSignature(world, delegatee, UNLIMITED_DELEGATION, new bytes(0), delegator, signature);
+
+    // Call a system from the delegatee on behalf of the delegator
+    vm.prank(delegatee);
+    returnData = world.callFrom(delegator, systemId, abi.encodeCall(WorldTestSystem.msgSender, ()));
+    returnedAddress = abi.decode(returnData, (address));
```
alvrs, line 104: for completeness, should we have an `assertEq` here to check it's the expected address?

packages/world/src/modules/init/implementations/WorldRegistrationSystem.sol
```diff
@@ -264,27 +265,7 @@ contract WorldRegistrationSystem is System, IWorldErrors, LimitedCallContext {
...
-
-      // Call the delegation control contract's init function
-      SystemCall.callWithHooksOrRevert({
-        caller: _msgSender(),
-        systemId: delegationControlId,
-        callData: initCallData,
-        value: 0
-      });
-    }
+    createDelegation(_msgSender(), delegatee, delegationControlId, initCallData);
```
alvrs, line 268: what do you think about calling this helper function `registerDelegation`?

packages/world-modules/src/modules/delegation/registerDelegationWithSignature.sol
yonadaaa: Not really! But because the method is not on the World interface, people need to do this obscure wrapping `DelegationSystem(address(world))`

packages/world-modules/src/interfaces/IRegisterDelegationWithSignatureSystem.sol
alvrs, line 1: is this file outdated? doesn't seem to correspond to the name of the system contract. (I do prefer the more explicit name of the system though)

packages/world/src/modules/init/implementations/WorldRegistrationSystem.sol
yonadaaa: I considered that, but it might be confusing and because this is internal, a godlike "create" makes sense to me.

Plus the compiler doesn't let you! 
image

e2e/packages/contracts/mud.config.ts
alvrs: I think it'd be ok since it's just temporary until we'll include it in the `WorldRegistrationModule` by default. We could even name it in some way that makes it explicit that this is temporary.

e2e/packages/sync-test/registerDelegationWithSignature.test.ts
alvrs: ohh i see

packages/world-modules/src/modules/delegation/registerDelegationWithSignature.sol
alvrs: I think explicitly wrapping in the interface is fine because it's meant to be temporary, and the wrapper function obscures some of the complexity if someone would try to call this from a root system (it's not clear from the outside that that would fail)

packages/world-modules/src/interfaces/IRegisterDelegationWithSignatureSystem.sol
yonadaaa: Yup this is stale. Interesting that it isn't cleared 👀

packages/world-modules/test/DelegationModule.t.sol.sol
alvrs, line 1: one `.sol` too many in the filename (`.t.sol.sol`)

packages/world-modules/src/modules/delegation/DelegationSystem.sol
alvrs: Do we need `delegator` as an additional argument in the function signature then? Couldn't we just always use the recovered `signer`, and just revert if it's `address(0)`?

Btw I think right now it's possible to submit `address(0)` as the `delegator` param and successfully register a signature for `address(0)` because `ECDSA.tryRecover` would return `address(0)` if the signature is invalid.

packages/world-modules/src/interfaces/IRegisterDelegationWithSignatureSystem.sol
alvrs: I think it can't be cleared because the codegen directory is `src` here

packages/world-modules/src/modules/delegation/DelegationSystem.sol
karooolis: I'd say it's good practice to include `delegator` param in this case to make sure that the function is behaving as intended i.e. signature resolves to the expected delegator, or otherwise reverts. Besides that can't think of other good reasons right now though so don't think it's strictly necessary

yonadaaa: We should really use `recover`, the reason I didn't before is because I wanted to emit a `World_` error instead of the `ECDSA` one

alvrs: > signature resolves to the expected delegator

My understanding is it's impossible to create a valid signature that resolves to a different delegator? If it's possible to accidentally create a signature that resolves to some signer who didn't actually sign it, then i feel like we should include the `delegator` in the signed message to prevent that from accidentally happening. (_If_ this was possible, it would be really easy to locally recover the "signer" and pass it as `delegator` param to register a delegation on behalf of someone who didn't actually sign this)

karooolis: Yeah, we can assume it's not possible. The comment about "expected delegator" is related more to the consumer side to make sure that their message signing implementation is correct, and the signatures they produce resolve to the delegator that is expected.

packages/world-modules/test/DelegationModule.t.sol.sol
yonadaaa: We are now using `recover`, so I'm not sure what this outputs if the nonce is wrong 🤔 technically the recovered "signer" is incorrect either way

packages/world-modules/src/modules/delegation/DelegationSystem.sol
alvrs: We synced offline, putting it here for context:
- an "invalid signature" would "recover" _some_ address, probably one without a known private key
- it would be possible to pass this address as the delegator and thereby register a delegation for an address that never signed the delegation
- to avoid that we can make the delegator part of the message that is being signed, so it's not possible to create an invalid signature, recover a signer, and use the signer as delegator.

packages/world-modules/src/modules/delegation/DelegationWithSignatureModule.sol
```diff
@@ -0,0 +1,48 @@
...
+          "registerDelegationWithSignature(address,bytes32,bytes,address,bytes)",
+          "registerDelegationWithSignature(address,bytes32,bytes,address,bytes)"
+        )
+      )
+    );
+    if (!success) revertWithBytes(data);
+  }
+
+  function install(bytes memory) public pure {
+    revert Module_RootInstallNotSupported();
```
alvrs, line 46:
```suggestion
    revert Module_NonRootInstallNotSupported();
```

packages/world-modules/src/modules/delegation/DelegationWithSignatureSystem.sol
```diff
@@ -0,0 +1,47 @@
...
+    address delegator,
+    bytes memory signature
+  ) public {
+    uint256 nonce = UserDelegationNonces.get(delegator);
+    bytes32 hash = getSignedMessageHash(delegatee, delegationControlId, initCallData, delegator, nonce, _world());
+
+    // If the message was not signed by the delegator or is invalid, revert
+    address signer = ECDSA.recover(hash, signature);
+    if (signer != delegator) {
+      revert InvalidSigner(delegator, delegatee);
```
alvrs, line 40: I think `InvalidSignature` would be more descriptive here, since the signer would not match the delegator if any piece of the signature doesn't match (delegatee, nonce, etc)

```suggestion
      revert InvalidSignature(signer);
```

packages/world/ts/delegationWithSignatureData.ts
alvrs, line 1: since this file exports a single constant called `delegationWithSignatureTypes` we should call the file the same

packages/world-modules/src/modules/delegation/DelegationWithSignatureSystem.sol
alvrs, line 1: I want to really make sure it's clear this module is not meant for stable usage yet and we will remove it in the near future once it's audited and moved into the World. Should we rename it to `Unstable_DelegationWithSignatureSystem` for now?

packages/world-modules/src/interfaces/IBaseWorld.sol
```diff
@@ -6,11 +6,13 @@ pragma solidity >=0.8.24;
...
+
 /**
  * @title IBaseWorld
  * @author MUD (https://mud.dev) by Lattice (https://lattice.xyz)
  * @notice This interface integrates all systems and associated function selectors
  * that are dynamically registered in the World during deployment.
  * @dev This is an autogenerated file; do not edit manually.
  */
-interface IBaseWorld is IStore, IWorldKernel {}
+interface IBaseWorld is IStore, IWorldKernel, IUnstable_DelegationWithSignatureSystem {}
```
alvrs, line 18: need to add this system to the excludes of the mud.config

.changeset/cold-apes-play.md
```diff
@@ -0,0 +1,7 @@
+---
+"@latticexyz/cli": patch
+"@latticexyz/world-modules": patch
+"@latticexyz/world": patch
+---
+
+Added a new module, `DelegationWithSignatureModule`, which allows registering delegations with a signature.
```
alvrs, line 7:
```suggestion
Added a new preview module, `Unstable_DelegationWithSignatureModule`, which allows registering delegations with a signature.

Note: this module is marked as `Unstable`, because it will be removed and included in the default `World` deployment once it is audited.
```

packages/world-modules/src/interfaces/IBaseWorld.sol
yonadaaa: done!

alvrs (approved):

PR #2481, opened 2024-03-20
feat(store, world): throw if required keys are not provided, more validations

alvrs:
Add more type errors for missing required keys or invalid inputs

### DIFF
--- a/packages/store/ts/config/v2/input.ts
+++ b/packages/store/ts/config/v2/input.ts
@@ -26,7 +26,7 @@ export type TablesInput = {
 
 export type StoreInput = {
   readonly namespace?: string;
-  readonly tables: TablesInput;
+  readonly tables?: TablesInput;
   readonly userTypes?: UserTypes;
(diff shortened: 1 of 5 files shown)


packages/store/ts/config/v2/input.ts
```diff
@@ -26,7 +26,7 @@ export type TablesInput = {
 
 export type StoreInput = {
   readonly namespace?: string;
-  readonly tables: TablesInput;
+  readonly tables?: TablesInput;
```
holic, line 29: is it necessary for this to be optional?

holic (approved):

packages/store/ts/config/v2/input.ts
alvrs: this was an oversight, you don't have to provide tables in your config

holic: when would you want to specify a config without tables?

PR #2482, opened 2024-03-20
feat(store, world): add table deploy config, add v2 to v1 compat config

alvrs:
- rename `deployment` to `deploy`
- add `deploy` config option to `table` options
- add `v2` key to output of `v1` compat layer


packages/world/ts/config/v2/world.ts
```diff
@@ -19,7 +19,7 @@ import { Tables } from "@latticexyz/store";
 import { resolveSystems } from "./systems";
 import { resolveNamespacedTables, validateNamespaces } from "./namespaces";
 import { resolveCodegen } from "./codegen";
-import { resolveDeployment } from "./deployment";
+import { resolveDeploy as resolveDeploy } from "./deploy";
```
alvrs, line 22:
```suggestion
import { resolveDeploy } from "./deploy";
```

holic (approved):
only question I have is `disable` vs `disabled` 🤔

holic (approved):

PR #2483, opened 2024-03-20
feat: use new config

holic:
taking another stab at #2470


packages/store-sync/src/zustand/logToTable.ts
holic, line 1: this isn't used, was moved up a directory but this never got removed

templates/react/packages/contracts/mud.config.ts
```diff
@@ -1,13 +1,15 @@
...
   tables: {
     Tasks: {
-      valueSchema: {
+      schema: {
+        id: "bytes32",
         createdAt: "uint256",
         completedAt: "uint256",
         description: "string",
       },
+      key: ["id"],
```
yonadaaa, line 12: OOC it is possible to use shorthand here? 

Or alternatively you could reduce the diff by keeping the `key` name

yonadaaa (approved):

templates/react/packages/contracts/mud.config.ts
alvrs, line 12:
```suggestion
      id: "bytes32",
      createdAt: "uint256",
      completedAt: "uint256",
      description: "string",
```

alvrs: not blocking at all though

alvrs (approved):

PR #2493, opened 2024-03-21
test(store,world): update protocol version tests

### LINKED ISSUE #2492: revisit interface used for protocol version snapshots
I renamed a thing in #2490 and it got picked up by world protocol version tests but not store (but the change was in store), so we're prob using the wrong interface(s) to represent the protocol in tests

holic:
closes #2492 

We were using the wrong representation of the protocol before and weren't capturing things like errors, etc. inside store protocol version tests.


yonadaaa (approved):
Nice! I appreciate the comments, it was unclear to me how these snapshots were generated before

PR #2506, opened 2024-03-21
build: ignore docs for most CI things

holic:
we don't need to run any of these when only docs changes

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -2,8 +2,6 @@ name: Benchmarks
 
 on:
   workflow_dispatch:
-    paths-ignore:
-      - "docs/**"
 
 jobs:
   benchmarks:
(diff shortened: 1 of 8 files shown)


.github/workflows/bench.yml
```diff
@@ -2,6 +2,8 @@ name: Benchmarks
 
 on:
   workflow_dispatch:
```
yonadaaa, line 4: This isn't needed for `workflow_dispatch` runs

yonadaaa (changes requested):

.github/workflows/bench.yml
```diff
@@ -2,6 +2,8 @@ name: Benchmarks
...
+    paths-ignore:
+      - "docs/**"
```
holic, line 6:
```suggestion
```

yonadaaa (approved):

PR #2519, opened 2024-03-21
docs(contrib): clarify that we mean a path, not a link

qbzzt:


docs/pages/contribute.mdx
```diff
@@ -37,7 +37,7 @@ The following steps are only necessary if you want to contribute to MUD. To use
 To check the changes in an application, run these commands at the application's root.
 
 ```sh copy
-pnpm mud set-version --link <link to MUD repository>
+pnpm mud set-version --link /path/to/local/mud
```
alvrs, line 40:
```suggestion
pnpm mud set-version --link /relative/path/to/local/mud
```

alvrs: just noticed that this actually doesn't work if an absolute path is provided

alvrs (approved):

PR #2520, opened 2024-03-21
docs(templates/typescript): add link to quickstart

qbzzt:


docs/pages/templates/typescript/link2quickstart.mdx
```diff
@@ -0,0 +1,7 @@
+You can install a copy of this template using:
+
+```sh copy
+pnpm create mud tutorial
+```
+
+For more information, [see the quickstart](/quickstart).
```
holic, line 7: can we make this a regular component somewhere rather than an MDX "partial" that we have to ignore in `_meta.json`?

alvrs: i think it would be even nicer if in each of the template's pages we show the command as

```sh copy
pnpm create mud tutorial --template <templateType>
```

qbzzt: > can we make this a regular component somewhere rather than an MDX "partial" that we have to ignore in `_meta.json`?

I'd rather not, because logically it's not a generic component, it's something we only use here. Anyway, I don't know how to add parameters, so I'll remove it anyway.

qbzzt: Anyway, `--template <templateType>` has been added.

docs/pages/templates/typescript/_meta.js
```diff
@@ -3,4 +3,7 @@ export default {
   vanilla: "Vanilla",
   "react-ecs": "React-ECS",
   three: "Three.js",
+  "link2quickstart": {
+    display: "hidden",
+  },
```
alvrs, line 8: don't need this anymore

holic (approved):

PR #2525, opened 2024-03-22
docs: add CI action to check links

holic:


holic:
not working so giving up for now

PR #2529, opened 2024-03-22
chore: threejs template changeset

yonadaaa:
Adding changeset missed in https://github.com/latticexyz/mud/pull/2473

### DIFF
--- a/.changeset/cold-tools-admire.md
+++ b/.changeset/cold-tools-admire.md
@@ -1,0 +1,5 @@
+---
+"create-mud": patch
+---
+
+Updated the react-three template.
--- a/.changeset/cold-tools-admire.md
+++ b/.changeset/cold-tools-admire.md
@@ -2,4 +2,4 @@
 "create-mud": patch
 ---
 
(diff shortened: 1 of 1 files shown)


.changeset/cold-tools-admire.md
```diff
@@ -0,0 +1,5 @@
+---
+"create-mud": patch
+---
+
+Updated the react-three template.
```
holic, line 5: can we describe what changed? this is going into the changelog and it kinda defeats the purpose of this entry if it doesn't actually explain anything

holic: also template name is `threejs`

.changeset/cold-tools-admire.md
```diff
@@ -0,0 +1,5 @@
+---
+"create-mud": patch
+---
+
+Changing the controls in the `threejs` template from arrow keys to WASD, and added text to explain what the app does.
```
holic, line 5:
```suggestion
Changed the controls in the `threejs` template from arrow keys to WASD, and added text to explain what the app does.
```

holic (approved):

PR #2532, opened 2024-03-22
docs(store/tables): update for MUD 2.0

qbzzt:
The biggest change is that this syntax doesn't compile for some reason:

```
FieldLayout fieldLayout = FieldLayoutLib.encode({
  staticFields: staticFields,
  numDynamicFields: 0
});
```

### DIFF
--- a/docs/pages/store/tables.mdx
+++ b/docs/pages/store/tables.mdx
@@ -46,9 +46,7 @@ The table ID encodes the table's type and name in a 32-byte `ResourceId` value a
 import { ResourceId, ResourceIdLib } from "@latticexyz/store/src/ResourceId.sol";
 import { RESOURCE_TABLE, RESOURCE_OFFCHAIN_TABLE } from "@latticexyz/store/src/storeResourceTypes.sol";
 
-.
-.
-.
+...
 
 // Using the `RESOURCE_TABLE` type makes it an onchain table
 ResourceId tableId = ResourceIdLib.encode({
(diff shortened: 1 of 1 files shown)


docs/pages/store/tables.mdx
```diff
@@ -57,6 +61,9 @@ ResourceId offchainTableId = ResourceIdLib.encode({
   typeId: RESOURCE_OFFCHAIN_TABLE,
   name: "Balance"
 });
+
+console.log("Onchain table id:%x", uint256(ResourceId.unwrap(tableId)));
+console.log("Offchain table id:%x", uint256(ResourceId.unwrap(offchainTableId)));
```
holic, line 66: is this leftover?

docs/pages/store/tables.mdx
```diff
@@ -46,6 +46,10 @@ The table ID encodes the table's type and name in a 32-byte `ResourceId` value a
 import { ResourceId, ResourceIdLib } from "@latticexyz/store/src/ResourceId.sol";
 import { RESOURCE_TABLE, RESOURCE_OFFCHAIN_TABLE } from "@latticexyz/store/src/storeResourceTypes.sol";
 
+.
+.
+.
+
```
holic, line 52: I don't think we need this?

docs/pages/store/tables.mdx
qbzzt: No, it's something I thought would be useful for people so I left it. I'll make that clearer.

docs/pages/store/tables.mdx
qbzzt: It signifies a change of context, from the header matter to inside a function inside the contract. I changed it to horizontal dots so it'll take less space.

qbzzt: It signifies a change of context from the header stuff to inside a function inside the contract. I changed it to a horizontal ... so it'll take less space.

alvrs (approved):

PR #2533, opened 2024-03-22
fix: threejs template style handlebars

yonadaaa:
Template could not be published because the style used the template engine handlebars 
https://github.com/latticexyz/mud/actions/runs/8391273439/job/22986718489


holic:
there's a `{{ padding: ... }}` in there still

holic (approved):

PR #2535, opened 2024-03-23
docs(store/encoding): update and clarify 🏎

qbzzt:
Also, expand on how to interpret events.


docs/pages/store/encoding.mdx
```diff
@@ -7,381 +7,427 @@ It is comparable to `abi.encodePacked`, but with some notable differences:
...
-**For events it reduces the payload size by ~80%** compared to the Solidity default of `abi.encode` (depending on the table schema).
+**For events it reduces the payload size by over 80%** compared to the Solidity default of `abi.encode` (depending on the table schema).
 Note that since the encoding happens at runtime and does not have access to compiler internals, there is some overhead for encoding and decoding that is not present in vanilla Solidity.
 
-## Schema
+## Schemas
 
-Each table has a key schema and a value schema, which define the data types of each column.
+Each table has two schemas: a key schema and a value schema.
+The schemas define the data type of each column.
```
alvrs, line 22: This paragraph is a but confusing with the new config where the schema is defined as one schema and some columns are marked as key. Onchain we still distinguish between key schema and value schema though. How can we rephrase this to make it more clear? One schema but it's stored separately for value fields and key fields onchain (because onchain the table is more like a key/value store than a table, ie we can't look up by value, only by key)

alvrs:
> remove the storage explanation.

Instead of removing it, can we move it to a new docs page? I'd like to still be able to refer to the docs for how data in MUD is represented in Solidity storage

qbzzt:
> > remove the storage explanation.
> 
> Instead of removing it, can we move it to a new docs page? I'd like to still be able to refer to the docs for how data in MUD is represented in Solidity storage

I added it back in. If we are to have it, store/encoding is probably the right place.

docs/pages/store/encoding.mdx
qbzzt: Can you check the new explanation? I think it's clearer.

PR #2539, opened 2024-03-24
refactor(store-sync): update recs sync to use new config

holic:
update RECS to use v2 config under the hood rather than the v1 compat layer

I initially added some compat for previous users of the underlying `recsStorage` method, but its untested and may be incomplete, so I removed for now


yonadaaa (approved):
clean :shipit:

alvrs (approved):

PR #2555, opened 2024-03-27
refactor(store,world): refactor types, remove redundant casts

ssalbdivad:
This refactors some existing types in world and store, adding some useful patterns for sharing logic between types and values and removing redundant casts.

I plan to review with @alvrs and @holic on a call if possible so we can cover some of the changes as efficiently as possible. There should not be any impact on the type or value-level behavior.

### DIFF
--- a/packages/store/ts/config/v2/compat.ts
+++ b/packages/store/ts/config/v2/compat.ts
@@ -71,5 +71,5 @@ export function storeToV1<store>(store: conform<store, Store>): storeToV1<store>
     codegenIndexFilename: store.codegen.indexFilename,
     tables: resolvedTables,
     v2: store,
-  } as unknown as storeToV1<store>;
+  } as never;
 }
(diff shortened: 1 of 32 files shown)


packages/store/ts/config/v2/compat.ts
```diff
@@ -71,5 +71,5 @@ export function storeToV1<store>(store: conform<store, Store>): storeToV1<store>
     codegenIndexFilename: store.codegen.indexFilename,
     tables: resolvedTables,
     v2: store,
-  } as unknown as storeToV1<store>;
+  } as never;
```
holic, line 74: ooc why `as never`?

ssalbdivad: Especially if you have to cast to `unknown` first, the cast itself provides no type-safety.

The return type is already explicitly specified on the function's definition, so `as never` is just a convenient way to allow an assignment to a predefined type (in this case `storeToV1<store>`) without duplicating the signature.

It would also be fine to use `as any` in this position, but `as never` is more likely to surface downstream if the explicit signature is ever accidentally removed.

Also less likely to trigger linting rules 😄

alvrs (dismissed):

Issue #2552, opened 2024-03-26
callWithDelegation should support ERC1271 


yonadaaa:
We can use this `SignatureChecker` instead of `ECDSA.recover` to also support smart contract wallets https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography/SignatureChecker.sol

PR #2559, opened 2024-03-27
feat(world-modules): call with signature supports ERC1271

yonadaaa:
fixes https://github.com/latticexyz/mud/issues/2552

We now emit a generic `InvalidSignature` error with no arguments, because we are agnostic about signature schemes and the `signer` (if any) derived by signing contracts.


yonadaaa:
Do we want to add a test that sets up a smart contract and registers a delegation for it? or do we trust the OZ signature checker

holic (approved):

PR #2568, opened 2024-03-28
test: reproduce accessing uninitialisated indices in static arrays bug  

yonadaaa:
This shouldn't happen because static array values are all initialised. Probably related to them being dynamic arrays under the hood


yonadaaa:
Closing as we have fix https://github.com/latticexyz/mud/pull/2613

PR #2576, opened 2024-03-31
refactor(cli): remove forge cache workaround

holic:
this was fixed upstream! 🎉


alvrs (approved):
Did you confirm locally that the IWorld Abi is now regenerated when a system contract interface is changed?

holic:
Yep! It does appear to update `IWorld.abi.json` when I change a system (e.g. adding or removing a function)

PR #2458, opened 2024-03-18
feat(template/react): setup AA infra

mouseless0x:

PR #2578, opened 2024-04-01
feat(common,world): improvements for smart accounts

holic:
pulled out of https://github.com/latticexyz/mud/pull/2458


packages/common/src/writeContract.ts
```diff
@@ -31,6 +32,7 @@ export async function writeContract<
 >(
   client: Client<Transport, chain, account>,
   request: WriteContractParameters<abi, functionName, args, chain, account, chainOverride>,
+  publicClient?: PublicClient<Transport, chain>,
```
holic, line 35: this method is deprecated, so I didn't document this change in the changelog

.changeset/smooth-ducks-sell.md
```diff
@@ -0,0 +1,9 @@
+---
+"@latticexyz/common": minor
```
holic, line 2: so according to semver, this should be `minor` because its an extra feature, not a fix

but maybe we don't want to trigger a bump to 2.1 yet?

holic, line 2:
```suggestion
"@latticexyz/common": patch
```

.changeset/smooth-ducks-sell.md
```diff
@@ -0,0 +1,9 @@
+---
+"@latticexyz/common": patch
```
holic, line 2: this should prob be minor but we're trying to batch a few minor version changes and don't want to block patch changes

packages/world/ts/actions/callFrom.ts
```diff
@@ -52,12 +52,17 @@ type SystemFunction = { systemId: Hex; systemFunctionSelector: Hex };
...
     // Applies to: `client.writeContract`, `getContract(client, ...).write`
     writeContract: async (writeArgs): Promise<WriteContractReturnType> => {
-      // Skip if the contract isn't the World.
-      if (writeArgs.address !== params.worldAddress) {
+      // Skip if the contract isn't the World or the function called should not be redirected through `callFrom`.
+      if (
+        writeArgs.address !== params.worldAddress ||
+        writeArgs.functionName === "call" ||
+        writeArgs.functionName === "callFrom" ||
+        writeArgs.functionName === "registerDelegationWithSignature"
```
tash-2s, line 64: Is this `registerDelegationWithSignature` condition necessary? The `registerDelegationWithSignature` function should be called by a delegator, not a delegatee. This `callFrom` extension action is intended for the use of the delegatee's wallet client (after delegation), and the delegator's wallet client does not need to use this.

holic: `registerDelegationWithSignature` _is_ called by the delegatee, using a signed message by the delegator!

alvrs (approved):

PR #2589, opened 2024-04-02
feat(common): allow specifying concurrency in transactionQueue

holic:
After hearing from @kooshaba about Sky Strife latency, specifically when playing from Japan, I did a bit of testing with a fresh template deployed to Redstone Holesky to find out where some of the latency was coming from. There's quite a few factors at play:
- each tx requires gas estimation, and viem does two RPC round trips: one to get max priority fee per gas and one to get the gas estimate of the call
  - the former can likely be cached but I am unclear how safe it is to do so (is this a hardcoded chain value? or something that fluctuates based on demand?)
  - the latter can possibly be cached for recurring calls to the same methods with same/similar args (Sky Strife does some variant of this) but would need some thought to get this right, if not just leave this to user land to figure out
- the transaction queue defaults to processing a single item at a time, to ensure no nonce errors, but with all the rpc round trips, that means 1-2s before a tx is even submitted (given Japan latency), then waiting for the next block to be mined, and this is super apparent once you queue up more than a couple txs

Even though increasing the concurrency will lead to nonce errors, it may be an acceptable trade-off for certain games/applications where its safe to retry after nonce failure and its fine to have somewhat out-of-order operations.

### DIFF
--- a/.changeset/poor-maps-itch.md
+++ b/.changeset/poor-maps-itch.md
@@ -1,0 +1,5 @@
+---
+"@latticexyz/common": patch
+---
+
+`transactionQueue` now accepts a `queueConcurrency` to allow adjusting the number of concurrent calls to the mempool. This defaults to `1` to ensure transactions are ordered and nonces are handled properly. Any number greater than that is likely to see nonce errors and transactions arriving out of order, but this may be an acceptable trade-off for some applications that can safely retry.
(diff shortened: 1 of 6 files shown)


Kooshaba (approved):
makes sense to me. i think even a concurrency of 2 will already be a huge win for UX in sky strife.

PR #2592, opened 2024-04-02
feat(world-modules): add `callWithSignature`

yonadaaa:
Replace `registerDelegationWithSignature` with a more generic `callWithSignature`


e2e/packages/sync-test/data/callWithSignature.ts
```diff
@@ -2,51 +2,46 @@ import { Page } from "@playwright/test";
 import { GetContractReturnType, PublicClient, WalletClient } from "viem";
 import { AbiParametersToPrimitiveTypes, ExtractAbiFunction, ExtractAbiFunctionNames } from "abitype";
 
-const DelegationAbi = [
+const Abi = [
```
holic, line 5: any reason we can't import this from world-modules?

packages/world/ts/callWithSignatureTypes.ts
```diff
@@ -0,0 +1,10 @@
+// Follows https://viem.sh/docs/actions/wallet/signTypedData#usage
+
+export const callWithSignatureTypes = {
+  Call: [
+    { name: "delegator", type: "address" },
+    { name: "systemId", type: "bytes32" },
+    { name: "callData", type: "bytes" },
+    { name: "nonce", type: "uint256" },
+  ],
+} as const;
```
holic, line 10: I wonder if there's a way we can express this as a struct in solidity then import the ABI here instead of manually defining in both spots

yonadaaa: I agree this would be great, I'm surprised I haven't seen it anyone do it. The problem is we need to derive the EIP712 typehash:
```solidity
bytes32 constant CALL_TYPEHASH = keccak256("Call(address delegator,bytes32 systemId,bytes callData,uint256 nonce)");
```

We could define a struct:

```solidity
struct Call {
  address delegator,
  bytes32 systemId,
  bytes memory callData,
  uint256 nonce
}
``` 

But we'd have to somehow get the names and datatypes of each member as strings, then concat them. It's like we need `.selector` for structs!

Maybe we could define a function selector instead? 🤔  I'll keep thinking about it

e2e/packages/sync-test/data/callWithSignature.ts
yonadaaa: I guess I didn't think to try it 🤯

packages/world-modules/src/modules/delegation/getSignedMessageHash.sol
```diff
@@ -3,27 +3,23 @@ pragma solidity >=0.8.24;
...
-
-bytes32 constant DELEGATION_TYPEHASH = keccak256(
-  "Delegation(address delegatee,bytes32 delegationControlId,bytes initCallData,address delegator,uint256 nonce)"
-);
+bytes32 constant CALL_TYPEHASH = keccak256("Call(address delegator,bytes32 systemId,bytes callData,uint256 nonce)");
 
 /**
  * @notice Generate the message hash for a given delegation signature.
+ * For EIP712 signatures https://eips.ethereum.org/EIPS/eip-712
  * @dev We include the delegator to prevent generating a garbage signature and registering a delegation for the corresponding recovered address.
```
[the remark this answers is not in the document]
yonadaaa: That's better!

packages/world-modules/src/modules/delegation/Unstable_CallWithSignatureSystem.sol
```diff
@@ -0,0 +1,46 @@
...
+  /**
+   * @notice Calls a system with a given system ID using the given signature.
+   * @param delegator The address on whose behalf the system is called.
+   * @param systemId The ID of the system to be called.
+   * @param callData The ABI data for the system call.
+   * @param signature The EIP712 signature.
+   * @return Return data from the system call.
+   */
+  function callWithSignature(
+    address delegator,
```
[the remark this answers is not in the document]
yonadaaa: I agree `delegator` is confusing in this context. `signer` is also what OpenZeppelin calls it in the [`SignatureChecker`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography/SignatureChecker.sol#L22) so i'm good with that

packages/world/ts/callWithSignatureTypes.ts
yonadaaa: I see what you mean, see we define it twice, but in the same place? As struct and hash.

How do you include a struct in the ABI though? also we'd need to transform the ABI item into the proper viem type

holic: ahh I think I assumed structs would be part of the ABI but they aren't, unless they're used as a function arg but even then it's an ABI for the function args with an `internalType` named `struct Call` (not just `Call`)

close but not quite, but maybe close enough?

alvrs (approved):

PR #2598, opened 2024-04-03
fix(world-modules): explicitly export mud config

holic:
I was seeing issues when importing from `@latticexyz/world-modules/mud.config` and an explicit export seems to fix it.


yonadaaa (approved):

packages/world-modules/package.json
```diff
@@ -10,8 +10,15 @@
...
   "exports": {
-    "./out/*": "./out/*",
-    "./*": "./dist/*"
+    "./internal/mud.config": "./dist/mud.config.js",
+    "./out/*": "./out/*"
+  },
+  "typesVersions": {
+    "*": {
+      "internal/mud.config": [
+        "./mud.config.ts"
```
holic, line 19: I am surprised I needed this, but I guess its because the `.ts` file is at the same path as the import for world/store packages.

PR #2611, opened 2024-04-04
docs(guide/extending): update for MUD 2.0

qbzzt:
The `resolveConfig` part could be made more elegant once https://github.com/latticexyz/mud/pull/2539 is merged.


docs/pages/guides/_meta.js
```diff
@@ -1,7 +1,7 @@
 export default {
   "replicating-onchain-state": "Replicating onchain state",
   "hello-world": "Hello World",
-  "extending-a-world": "Extending a World",
+  extending: "Extending a World",
```
holic, line 4: Reminder to set up redirects if we’re changing URLs

docs/pages/guides/extending/_meta.js
```diff
@@ -0,0 +1,3 @@
+export default {
+  "extending-a-world": "Extending a World Permissionlessly",
```
holic, line 2: hmm now that I am looking more closely at this, do we expect more than one page in this section?

otherwise we can continue calling this section/dir `extending-a-world` and just use `index.mdx` for the page name

docs/pages/guides/_meta.js
qbzzt: Oops, you're right. Fixed.

docs/pages/guides/extending/_meta.js
qbzzt: Yes, that's the reason I moved it like this. "Extending" is for when you have an existing world and need to change it. Coming soon, a guide on how to add a table to such a `World`.

holic: it's still "extending-a-world" so I would still opt to keep the directory name to match the old path, use index.mdx for the base page, then can add other pages to the section for additional details

qbzzt: Fixed.

qbzzt: Done, please rereview.

holic (approved):
didn't take too close a look at the code to confirm but I see you working through some help threads so I assume this is working!

PR #2613, opened 2024-04-04
fix(store): return zero for uninitialised static array elements

yonadaaa:
Changes the behaviour of static arrays to return default values on "uninitialised" values. If an array index is less than the static length, but greater than the underlying dynamic length, we return zero.


packages/store/ts/codegen/field.ts
```diff
@@ -137,6 +140,15 @@ export function renderFieldMethods(options: RenderTableOptions): string {
...
                 ${_keyTupleDefinition}
+
+                uint256 _byteLength = ${_store}.getDynamicFieldLength(_tableId, _keyTuple, ${dynamicSchemaIndex});
+                uint256 dynamicLength = _byteLength / ${portionData.elementLength};
+                uint256 staticLength = ${typeWrappingData.staticLength};
+
+                if (_index < staticLength && _index >= dynamicLength) {
+                  return 0;
+                }
+
```
alvrs, line 151: could we just have the `if (typeWrappingData && typeWrappingData.kind === "staticArray")` check around this small block?

another question, is `0` implicitly convertible to the expected return type or do we need to do some decoding/casting here?

yonadaaa: 1) Good point, we don't need this big diff

2) We need casting, I'm still figuring out which helpers to use 💪

yonadaaa: Done both these!

e2e/packages/contracts/test/StaticArrayLength.t.sol
yonadaaa, line 1: Wonder if there is somewhere more appropriate than E2E for this

e2e/packages/contracts/test/StaticArrayLength.t.sol
```diff
@@ -0,0 +1,47 @@
...
+
+    StaticArray.getItemValue(_index);
+  }
+}
+
+contract StaticArrayLengthTest is MudTest {
+  function testLength() public {
+    assertEq(StaticArray.lengthValue, 3);
+    assertEq(StaticArray.getValue().length, 3);
+    assertEq(StaticArray.getValue()[0], 0);
```
alvrs, line 27:
```suggestion
```
looks like we're testing the same thing below

e2e/packages/contracts/test/StaticArrayLength.t.sol
```diff
@@ -0,0 +1,47 @@
...
+contract Wrapper {
+  function getItemValue(address worldAddress, uint256 _index) public {
+    StoreSwitch.setStoreAddress(worldAddress);
+
+    StaticArray.getItemValue(_index);
+  }
+}
+
+contract StaticArrayLengthTest is MudTest {
+  function testLength() public {
```
alvrs, line 24: is there a test somewhere that tests that if an element is set that value is returned instead of the default `0`? (maybe somewhere in the existing test suite?)

packages/store/ts/codegen/field.ts
```diff
@@ -137,6 +137,22 @@ export function renderFieldMethods(options: RenderTableOptions): string {
...
+                    ? `
+                uint256 _byteLength = ${_store}.getDynamicFieldLength(_tableId, _keyTuple, ${dynamicSchemaIndex});
+                uint256 dynamicLength = _byteLength / ${portionData.elementLength};
+                uint256 staticLength = ${typeWrappingData.staticLength};
+
+                if (_index < staticLength && _index >= dynamicLength) {
+                  return ${renderCastStaticBytesToType(field.arrayElement, `bytes${field.arrayElement.staticByteLength}(new bytes(0))`)};
+                }`
+                    : ``
+                }
```
alvrs, line 154: this is being passed through formatting afterwards, so we can adjust the spacing here to match the indentation of the source file (= add 7 spaces or something like that to lines 146-152 so the indentation matches the ``)

e2e/packages/contracts/test/StaticArrayLength.t.sol
yonadaaa: There some some in CLI, maybe we should move these here? https://github.com/latticexyz/mud/blob/main/packages/cli/contracts/test/Tablegen.t.sol#L70-L77

.changeset/popular-oranges-end.md
```diff
@@ -0,0 +1,6 @@
+---
+"@latticexyz/store": patch
+"@latticexyz/cli": patch
+---
+
+Fixed the behaviour of static arrays, so that they always return zero for uninitialised values. Previously they would revert with `Store_IndexOutOfBounds` if the index had not been set yet.
```
holic, line 6: Should we make a note about this matching native solidity behavior?

alvrs, line 6:
```suggestion
Fixed the behaviour of static arrays, so that they return zero for uninitialised values, to mirror the native Solidity behavior. Previously they reverted with `Store_IndexOutOfBounds` if the index had not been set yet.
```

alvrs (dismissed):

.changeset/popular-oranges-end.md
```diff
@@ -0,0 +1,6 @@
+---
+"@latticexyz/store": patch
+"@latticexyz/cli": patch
```
holic, line 3:
```suggestion
```

the changes here are just internal/for dev, so no need for a changelog

packages/store/ts/codegen/field.ts
```diff
@@ -137,6 +137,22 @@ export function renderFieldMethods(options: RenderTableOptions): string {
                 "uint256 _index",
               ])}) internal view returns (${portionData.typeWithLocation}) {
                 ${_keyTupleDefinition}
+
+                ${
+                  // If the index is within the static length,
+                  // but ahead of the dynamic length, return zero
+                  typeWrappingData && typeWrappingData.kind === "staticArray" && field.arrayElement
+                    ? `
+                    uint256 _byteLength = ${_store}.getDynamicFieldLength(_tableId, _keyTuple, ${dynamicSchemaIndex});
```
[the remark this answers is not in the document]
yonadaaa: I wondered that also. It seems like we don't report table libraries for this (or at all?)

.changeset/popular-oranges-end.md
yonadaaa: I like that! lemme think

PR #2614, opened 2024-04-04
feat(world-modules): add `validateCallWithSignature` to `Unstable_CallWithSignatureModule`

alvrs:
this is what using it would look like: https://github.com/latticexyz/gas-tank/pull/5


yonadaaa (approved):
seems like a useful helper to have around!

PR #2617, opened 2024-04-05
fix(create-mud): add test for threejs template

holic:
something recently changed in foundry to cause CI to start failing for threejs template because there are no tests to run

I didn't wanna remove the `mud test` call because it's useful to make sure the template's world/MUD config can be deployed, so I just added a really basic test instead (copied from other templates)


holic:
merging to unblock main

PR #2623, opened 2024-04-09
feat: account kit

holic:


holic:
replaced by https://github.com/latticexyz/mud/pull/3240

PR #2629, opened 2024-04-10
feat(account-kit): bridging

karooolis:


packages/account-kit/src/abis/OptimismPortal.json
alvrs, line 1: we can import this from https://www.npmjs.com/package/@eth-optimism/contracts-ts

packages/account-kit/src/abis/PaymasterSystem.json
alvrs, line 1: this is now in mud, we can import from `@latticexyz/gas-tank`

packages/account-kit/src/steps/gas-tank/consts.ts
alvrs, line 1: nit, we usually try to avoid abbreviations in code and would call this file `constants.ts`

packages/mud-login/dist/index.js
alvrs, line 1: this is outdated

packages/mud-login/dist/index.js.map
alvrs, line 1: outdated too

templates/react/packages/account-abstraction/dist/index.js
alvrs, line 1: dist should not be checked in

PR #2632, opened 2024-04-10
feat(world): add WorldProxy and WorldProxyFactory

yonadaaa:
Adds a new `WorldProxy` contract that uses `World` as it's implementation. Constructors are not run for proxies, so `WorldProxy` has it's own constructor that replicates the logic of the `World` and `StoreKernel` constructors.


packages/world/src/World.sol
```diff
@@ -40,7 +40,7 @@ contract World is StoreKernel, IWorldKernel {
   using WorldResourceIdInstance for ResourceId;
 
   /// @notice Address of the contract's creator.
-  address public immutable creator;
+  address public creator;
```
holic, line 43: 😭

yonadaaa: I know! We need this as a storage variable so it can be read/written by the proxy

packages/cli/src/deploy/deploy.ts
```diff
@@ -28,6 +29,7 @@ type DeployOptions<configInput extends ConfigInput> = {
    * not have a deterministic address.
    */
   deployerAddress?: Hex;
+  deployAsProxy: boolean;
```
alvrs, line 32: nit, should we call this something like `withWorldProxy`? or `deployWorldAsProxy` but that feels a bit too verbose 🤔

packages/cli/src/deploy/ensureWorldProxyFactory.ts
alvrs, line 1: seems like lots of overlap with `ensureWorldFactory` here, can we consolidate somehow? ie having a common `ensureWorldDependencies`, or having the `worldFactoryBuild` and `worldFactoryAbi` as input?

packages/cli/src/runDeploy.ts
```diff
@@ -41,6 +41,11 @@ export const deployOptions = {
     type: "string",
     desc: "The deployment salt to use. Defaults to a random salt.",
   },
+  deployAsProxy: {
+    type: "boolean",
+    desc: "Whether to deploy the world as an EIP1967 proxy.",
```
alvrs, line 46: are we fully EIP1967 compliant?

packages/world/src/Proxy.sol
```diff
@@ -0,0 +1,69 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts (last updated v5.0.0) (proxy/Proxy.sol)
```
alvrs, line 2: we could link the full github permalink here

packages/world/src/WorldProxy.sol
```diff
@@ -0,0 +1,69 @@
...
+import { AccessControl } from "./AccessControl.sol";
+import { ROOT_NAMESPACE_ID } from "./constants.sol";
+import { Proxy } from "./Proxy.sol";
+import { StorageSlot } from "./StorageSlot.sol";
+
+/**
+ * @dev Storage slot with the address of the current implementation.
+ * This is the keccak-256 hash of "eip1967.proxy.implementation" subtracted by 1.
+ */
+// solhint-disable-next-line private-vars-leading-underscore
```
alvrs, line 18: why do we need this solhint comment?

packages/world/src/WorldProxy.sol
```diff
@@ -0,0 +1,69 @@
...
+bytes32 constant IMPLEMENTATION_SLOT = 0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc;
+
+/**
+ * @dev Stores a new address in the EIP1967 implementation slot.
+ */
+function _setImplementation(address newImplementation) {
+  StorageSlot.getAddressSlot(IMPLEMENTATION_SLOT).value = newImplementation;
+}
```
alvrs, line 26: Is there some standard around whether to emit an event when the implementation changed?

packages/world/src/WorldProxyFactory.sol
```diff
@@ -0,0 +1,52 @@
...
+   */
+  function deployWorld(bytes memory salt) public returns (address worldAddress) {
+    // Deploy a new World and increase the WorldCount
+    bytes memory bytecode = type(World).creationCode;
+    uint256 _salt = uint256(keccak256(abi.encode(msg.sender, salt)));
+
+    // Deploy the world implementation
+    address worldImplementationAddress = Create2.deploy(bytecode, _salt);
+    // Deploy the world proxy
+    // TODO: use Create2 for the proxy
```
alvrs, line 41: This is pretty important I think, otherwise it's not possible to have deterministic addresses when deploying the world via a proxy

packages/world/test/WorldProxyFactory.t.sol
alvrs, line 1: Can we add all World tests by having a contract that extends `WorldTest` and overrides the `setUp` function to deploy the world behind a proxy?

packages/world/src/WorldProxy.sol
yonadaaa: There is, but I think we should move away from the standards altogether (ie. not referring to this as EIP1967)

packages/world/src/WorldProxy.sol
yonadaaa: nope, just a leftover from OZ

packages/world/test/WorldProxyFactory.t.sol
yonadaaa: done!

packages/world/src/Proxy.sol
yonadaaa: I think for these imports it's cleaner to copy-paste as is. Plus we'd have to change other ones (ERC20, ECDSA etc.)

packages/cli/src/runDeploy.ts
yonadaaa: nope! removed this

packages/world/src/Proxy.sol
alvrs: I just meant putting the permalink to github in the comment on the top

packages/world/src/IERC1967.sol
```diff
@@ -0,0 +1,24 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts (last updated v5.0.0) (interfaces/IERC1967.sol)
+
+pragma solidity ^0.8.20;
```
alvrs, line 4: Should we align this with the solidity version in other files?

packages/world/src/WorldProxyFactory.sol
```diff
@@ -0,0 +1,56 @@
...
+   * @param salt User defined salt for deterministic world addresses across chains
+   * @return worldAddress The address of the newly deployed WorldProxy contract.
+   */
+  function deployWorld(bytes memory salt) public returns (address worldAddress) {
+    // Deploy a new World and increase the WorldCount
+    uint256 _salt = uint256(keccak256(abi.encode(msg.sender, salt)));
+
+    // Deploy the world implementation
+    bytes memory worldBytecode = type(World).creationCode;
+    address worldImplementationAddress = Create2.deploy(worldBytecode, _salt);
```
alvrs, line 39: Could this be deployed once per world factory? And if it already exists we use the existing implementation?

Thinking about this, we could technically also undo the change of making `creator` not immutable and just require the deployer of the world proxy also to be the deployer of the implementation (which is the case when deploying via the world factory, and could be implemented this way by whoever wants to make a different world factory)

packages/world/test/WorldProxyFactory.t.sol
```diff
@@ -0,0 +1,146 @@
...
+
+    vm.startPrank(account1);
+
+    // Set proxy implementation to new world
+    startGasReport("set WorldProxy implementation");
+    WorldProxy(payable(worldAddress)).setImplementation(newWorldImplementationAddress);
+    endGasReport();
+
+    worldImplementationAddress = address(uint160(uint256(vm.load(worldAddress, IMPLEMENTATION_SLOT))));
+    assertEq(worldImplementationAddress, newWorldImplementationAddress);
```
alvrs, line 140: can we add a step after this where we test executing a world method (ie writing to a table) and confirm it still works?

packages/world/src/IERC1967.sol
yonadaaa: Yes, we do that with the other OZ files

packages/world/test/WorldProxyFactory.t.sol
yonadaaa: done!

packages/world/src/WorldProxyFactory.sol
yonadaaa: Arguably we shouldn't even deploy the implementation in the factory, it's more of a parameter like `initModule`. I think for convenience it's fine but we could restructure it is a constructor argument and there is a `ensureWorldImplementation` in the deploy pipeline. 

I'm not sure the `creator` idea would work. The immutable variable is a constant on the implementation. If we `delegatecall` the implementation from the proxy, how would we read it?

yonadaaa: Because it's a constant you would have expose a read method or even read the bytecode 😬

packages/cli/src/deploy/deployWorld.ts
```diff
@@ -11,8 +11,9 @@ export async function deployWorld(
   client: Client<Transport, Chain | undefined, Account>,
   deployerAddress: Hex,
   salt: Hex,
+  withWorldProxy: boolean,
```
alvrs, line 14:
```suggestion
  withWorldProxy?: boolean,
```

packages/cli/src/deploy/ensureWorldFactory.ts
```diff
@@ -14,6 +16,7 @@ import { Contract } from "./ensureContract";
 export async function ensureWorldFactory(
   client: Client<Transport, Chain | undefined, Account>,
   deployerAddress: Hex,
+  withWorldProxy: boolean,
```
alvrs, line 19:
```suggestion
  withWorldProxy?: boolean,
```

packages/world/src/WorldProxyFactory.sol
```diff
@@ -0,0 +1,54 @@
...
+  IModule public immutable initModule;
+  address public immutable worldImplementation;
+
+  /// @param _initModule The address of the init module.
+  constructor(IModule _initModule) {
+    initModule = _initModule;
+
+    // Deploy a world implementation
+    bytes memory worldBytecode = type(World).creationCode;
+    worldImplementation = Create2.deploy(worldBytecode, 0);
```
alvrs, line 29: would this revert if there is already a world deployed at this address? ~~i guess it should revert because otherwise the `creator` doesn't match, but maybe we should use the factory address as a salt here to get around the case where this world address already exists? Although it's still possible that someone else deployed with this salt.~~ Could we instead just deploy a world via `new World()` so we don't run into any Create2 conflicts?

packages/world/test/WorldProxyFactory.t.sol
```diff
@@ -0,0 +1,156 @@
...
+
+    // Confirm correct Core is installed
+    assertTrue(InstalledModules.get(address(initModule), keccak256(new bytes(0))));
+
+    // Confirm the msg.sender is owner of the root namespace of the new world
+    assertEq(NamespaceOwner.get(ROOT_NAMESPACE_ID), account1);
+
+    // Test we can execute a world function
+    IBaseWorld(address(worldAddress)).registerDelegation(account2, UNLIMITED_DELEGATION, new bytes(0));
+    assertEq(ResourceId.unwrap(UserDelegationControl.get(account1, account2)), ResourceId.unwrap(UNLIMITED_DELEGATION));
```
alvrs, line 150: i was a little confused for a sec where `account1` comes from here; as a general best practice we should use `vm.prank` over `vm.startPrank` to make it easier to follow

packages/cli/src/deploy/deploy.ts
```diff
@@ -28,6 +28,7 @@ type DeployOptions<configInput extends ConfigInput> = {
    * not have a deterministic address.
    */
   deployerAddress?: Hex;
+  withWorldProxy: boolean;
```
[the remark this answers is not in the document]
yonadaaa: Agreed!

packages/world/src/WorldProxyFactory.sol
yonadaaa: Yes I agree, that would be simpler also

alvrs (dismissed):

packages/world/src/WorldProxy.sol
```diff
@@ -0,0 +1,68 @@
...
+bytes32 constant IMPLEMENTATION_SLOT = 0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc;
+
+/**
+ * @dev Stores a new address in the EIP1967 implementation slot.
+ */
+function _setImplementation(address newImplementation) {
+  StorageSlot.getAddressSlot(IMPLEMENTATION_SLOT).value = newImplementation;
+
+  emit IERC1967.Upgraded(newImplementation);
+}
```
alvrs, line 28: why is this a free function rather than an internal function on WorldProxy?

ludns (commented):
ERC1967 Specifies a bunch of storage slot we are not implementing or that are incorrect here (the admin one as an example). this could be confusing.

alvrs:
> ERC1967 Specifies a bunch of storage slot we are not implementing or that are incorrect here (the admin one as an example). this could be confusing.

Are you proposing to store the admin at that address, or to not mention erc1967 here? The former one might also be confusing because we'd be introducing another level of ownership (with very similar security assumptions as the ROOT owner)

ludns:
Could we have a hook or something that copies the root namespace owner to that storage slot and keeps it in sync + emit the event?

alvrs:
Looks like the admin slot is actually optional in the ERC. Wonder if there are any real benefits of storing the admin address at that slot? (It's not like it's a security guarantee for anything since any root system could change the implementation address.) If there aren't any big benefits I would opt for not doing it to reduce complexity.

packages/world/src/WorldProxy.sol
yonadaaa: I generally prefer free functions when possible! but open to moving it

yonadaaa:
Agreeing with [alvrs](https://github.com/alvrs), our implementation is complaint because the [admin](https://eips.ethereum.org/EIPS/eip-1967#admin-address) and [beacon](https://eips.ethereum.org/EIPS/eip-1967#beacon-contract-address) are optional.

Though it is inconvenient how that we have to really scrutinise our custom implementations to make sure they're compliant.

packages/world/src/WorldProxy.sol
alvrs: why do you prefer free functions? have opted for internal functions in MUD so far, don't have a strong preference, but unless there is a reason for free functions i'd say let's move it to an internal one for consistency with the rest of the codebase

yonadaaa: Agreed, done

PR #2634, opened 2024-04-10
fix: pin CI foundry version

yonadaaa:
There is a Foundry issue that only happens in CI: https://github.com/foundry-rs/foundry/issues/7620

So we can pin CI to an old version for now.


alvrs (approved):

CometShock:
Yesterday's fix now is returning 404 errors when trying to download that version
image

PR #2635, opened 2024-04-10
feat(create-mud): change `anvil` to create a block every two seconds

qbzzt:
- Better simulate the production environment
- The indexer works better with this


templates/phaser/mprocs.yaml
```diff
@@ -7,4 +7,4 @@ procs:
     shell: pnpm mud dev-contracts --rpc http://127.0.0.1:8545
   anvil:
     cwd: packages/contracts
-    shell: anvil --base-fee 0
+    shell: anvil --base-fee 0 -b 2
```
holic, line 10: to make this more readable:
```suggestion
    shell: anvil --base-fee 0 --block-time 2
```

templates/react-ecs/mprocs.yaml
```diff
@@ -7,4 +7,5 @@ procs:
     shell: pnpm mud dev-contracts --rpc http://127.0.0.1:8545
   anvil:
     cwd: packages/contracts
-    shell: anvil --base-fee 0
+    shell: anvil --base-fee 0 -b 2
+
```
holic, line 11: no need for the extra line at the end
```suggestion
```

holic:
can we update all the mprocs files with the same? examples, etc.

also need a changeset for this since it impacts the create-mud package

qbzzt:
How do I make a changeset?

qbzzt:
> can we update all the mprocs files with the same? examples, etc.
> 
> also need a changeset for this since it impacts the create-mud package

Done and done

.changeset/polite-dots-tell.md
```diff
@@ -0,0 +1,5 @@
+---
+"create-mud": patch
+---
+
+feat(create-mud): change `anvil` to create a block every two seconds
```
holic, line 5:
```suggestion
Updated `anvil` args with two second block time to better reflect L2s
```

holic (approved):

PR #2641, opened 2024-04-12
fix(common): latency improvements

alvrs:


packages/common/src/writeContract.ts
```diff
@@ -63,43 +71,59 @@ export async function writeContract<
...
-      opts.publicClient ?? client,
+    const { abi, address, args, dataSuffix, functionName } = request;
+    const data = encodeFunctionData({
+      abi,
+      args,
+      functionName,
+    } as EncodeFunctionDataParameters);
+
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const { nonce, maxFeePerGas, maxPriorityFeePerGas, ...preparedTransaction } = await prepareTransactionRequest(
```
holic, line 90: we should aim to use `getAction` in these situations so that other action decorators can override this

packages/common/src/writeContract.ts
```diff
@@ -63,43 +71,59 @@ export async function writeContract<
...
+      client,
       {
+        // The nonce and fee values don't need to be accurate for gas estimation
+        // and we can save a couple rpc calls by providing stubs here
+        nonce: 0,
+        maxFeePerGas: 0n,
+        maxPriorityFeePerGas: 0n,
+        ...defaultParameters,
         ...request,
-        blockTag: "pending",
```
holic, line 79: why did we lose the pending block tag?

packages/common/src/writeContract.ts
```diff
@@ -63,43 +71,59 @@ export async function writeContract<
...
+
+  async function prepare(): Promise<WriteContractParameters<abi, functionName, args, chain, account, chainOverride>> {
     if (request.gas) {
-      debug("gas provided, skipping simulate", request.functionName, request.address);
+      debug("gas provided, skipping preparation", request.functionName, request.address);
       return request;
     }
 
-    debug("simulating", request.functionName, "at", request.address);
-    const result = await simulateContract<chain, account | undefined, abi, functionName, args, chainOverride>(
```
holic, line 75: we're no longer simulating?

iirc there was some nuance between estimating gas vs simulating where estimating gas has less nice errors than simulating, and maybe even didn't surface certain kinds of errors ahead of time?

alvrs: discussed offline: viem is just doing gas estimation internally, and simulate doesn't give you a gas estimate, so it'd be one more rpc call. Using gas estimate for now and can do an additional simulate in case of a revert later if we find that gas estimation doesn't give good errors.

packages/common/src/writeContract.ts
alvrs: discussed offline, this is useful for estimating against other txs in the mempool, bringing it back

packages/common/src/writeContract.ts
holic: to put another way: we're gonna follow viem defaults here and hope for the best 🤞 

we can come back and add simulation again if we find we get better error handling and less reverts

packages/common/src/createNonceManager.ts
```diff
@@ -51,6 +52,11 @@ export function createNonceManager({
     return nonceRef.current >= 0;
   }
 
+  function getNonce(): number {
+    if (!hasNonce()) throw new Error("call resetNonce before using getCurrentNonce");
```
holic, line 56:
```suggestion
    if (!hasNonce()) throw new Error("call resetNonce before using getNonce");
```

packages/common/src/writeContract.ts
```diff
@@ -63,43 +75,63 @@ export async function writeContract<
...
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const { nonce, maxFeePerGas, maxPriorityFeePerGas, ...preparedTransaction } = await getAction(
+      client,
+      viem_prepareTransactionRequest,
+      "prepareTransactionRequest",
+    )({
+      // The fee values don't need to be accurate for gas estimation
+      // and we can save a couple rpc calls by providing stubs here
+      maxFeePerGas: 0n,
+      maxPriorityFeePerGas: 0n,
```
holic, line 102: Can we leave a comment that we "stub" these here but then override them later with fee data from the fee ref?

packages/common/src/createFeeRef.ts
```diff
@@ -0,0 +1,29 @@
...
+  const feeRef: FeeRef = { fees: {}, lastUpdatedTimestamp: 0 };
+
+  async function updateFees(): Promise<void> {
+    const fees = await estimateFeesPerGas(client, args);
+    feeRef.fees = fees;
+    feeRef.lastUpdatedTimestamp = Date.now();
+  }
+
+  updateFees();
+  setInterval(updateFees, refreshInterval);
```
holic, line 26: viem has some internal stuff for doing polling things that can dedupe etc, we should see if those are exported so we can use em?

packages/common/src/createFeeRef.ts
```diff
@@ -0,0 +1,29 @@
...
+export function createFeeRef({ client, args, refreshInterval }: CreateFeeRefOptions): FeeRef {
+  const feeRef: FeeRef = { fees: {}, lastUpdatedTimestamp: 0 };
+
+  async function updateFees(): Promise<void> {
+    const fees = await estimateFeesPerGas(client, args);
+    feeRef.fees = fees;
+    feeRef.lastUpdatedTimestamp = Date.now();
+  }
+
+  updateFees();
```
holic, line 25: should we make this function `async` and then do `await updateFees()` here so we always have fees set before this is used?

packages/common/src/writeContract.ts
```diff
@@ -63,43 +75,63 @@ export async function writeContract<
...
+          const preparedRequest = await prepare();
+
           const nonce = nonceManager.nextNonce();
-          debug("calling", preparedWrite.functionName, "with nonce", nonce, "at", preparedWrite.address);
-          return await viem_writeContract(client, {
-            nonce,
-            ...preparedWrite,
-          } as typeof preparedWrite);
+
+          const fullRequest = { ...preparedRequest, nonce, ...feeRef.fees };
```
holic, line 132: I think there's a chance fee ref data can be unset here if this gets called immediately (see above comment)

packages/store-sync/src/createStoreSync.ts
```diff
@@ -263,7 +268,9 @@ export async function createStoreSync<config extends StoreConfig = StoreConfig>(
     // This currently blocks for async call on each block processed
     // We could potentially speed this up a tiny bit by racing to see if 1) tx exists in processed block or 2) fetch tx receipt for latest block processed
     const hasTransaction$ = recentBlocks$.pipe(
-      concatMap(async (blocks) => {
+      // We use `mergeMap` instead of `concatMap` here to send the fetch request immediately when a new block range appears,
+      // instead of sending the next request only when the previous one completed.
+      mergeMap(async (blocks) => {
```
holic, line 273: can we separate out these changes into its own PR?

packages/common/src/createFeeRef.ts
alvrs: #2664

packages/store-sync/src/createStoreSync.ts
alvrs: pulling out into a separate PR

alvrs: Followup here: https://github.com/latticexyz/mud/pull/2665

packages/common/src/getFeeRef.ts
```diff
@@ -0,0 +1,8 @@
+import { CreateFeeRefOptions, FeeRef, createFeeRef } from "./createFeeRef";
+
+let feeRef: FeeRef | undefined;
```
holic, line 3: this seems to be a global cache regardless of fee args, which means one client could end up with the cached values of another client?

PR #2648, opened 2024-04-12
feat(world-modules): callWithSignature chain id is salt

yonadaaa:
As per the EIP712 spec, wallets attempt to switch networks based on the `chainId` field in the domain: https://eips.ethereum.org/EIPS/eip-712#definition-of-domainseparator

We work around this by including the chain ID in the `salt` field instead. As the salt is still present in the message we still avoid replay attacks.


.changeset/itchy-baboons-camp.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/world-modules": patch
+---
+
+Moved the chain ID in `CallWithSignature` from the `chainId` to the `salt` field, preventing wallets from attempting to switch networks based on the messages chain.
```
holic, line 5:
```suggestion
Moved the chain ID in `CallWithSignature` from the `chainId` to the `salt` field to allow for cross-chain signing without requiring wallets to switch networks.
```

packages/world-modules/src/modules/delegation/getSignedMessageHash.sol
```diff
@@ -24,7 +24,7 @@ function getSignedMessageHash(
   address worldAddress
 ) view returns (bytes32) {
   bytes32 domainSeperator = keccak256(
-    abi.encode(keccak256("EIP712Domain(uint256 chainId,address verifyingContract)"), block.chainid, worldAddress)
+    abi.encode(keccak256("EIP712Domain(address verifyingContract,bytes32 salt)"), worldAddress, bytes32(block.chainid))
```
holic, line 27: can we add some context here to explain why we use the salt, not the chain ID param, for this value?

yonadaaa: agreed, it is quite nonstandard!

packages/world-modules/src/modules/delegation/getSignedMessageHash.sol
holic, line 1: left a comment in Discord but this directory name doesn't make much sense anymore and also kinda confusing alongside `modules/std-delegations`

packages/world-modules/src/modules/callwithsignature/tables/CallWithSignatureNonces.sol
holic, line 1: I can still see the old file, I think it needs to be manually deleted because we're not yet putting these files under a codegen dir
https://github.com/latticexyz/mud/blob/2720170c211fd1eec2502ff851cd68556afaa1bd/packages/world-modules/src/modules/delegation/tables/CallWithSignatureNonces.sol

holic (approved):

PR #2651, opened 2024-04-12
feat(gas-tank): add gas tank

alvrs:
Moving the gas tank contract into the MUD monorepo


.changeset/red-carrots-sip.md
```diff
@@ -0,0 +1,6 @@
+---
+"@latticexyz/gas-tank": patch
+---
+
+Adds the `GasTank` paymaster, which will allow MUD users to use smart account and share gas balance between apps.
```
holic, line 5:
```suggestion
Adds the `GasTank` paymaster, which will allow MUD users to use smart accounts and share gas balance between apps.
```

packages/gas-tank/src/PaymasterSystem.sol
```diff
@@ -0,0 +1,226 @@
...
+      userAccount = _recoverCallWithSignature(userOp);
+      if (userAccount == address(0)) {
+        revert("Account is not registered as spender for any user");
+      }
+    }
+
+    // Require the user account to have sufficient balance
+    uint256 balance = UserBalances.get(userAccount);
+    if (maxCost > balance) {
+      revert("Insufficient user balance");
```
holic, line 73: Should we have custom errors?

packages/gas-tank/worlds.json
```diff
@@ -0,0 +1,9 @@
+{
+  "17069": {
+    "address": "0xba0149de3486935d29b0e50dfcc9e61bd40ae095",
+    "blockNumber": 286789
+  },
+  "31337": {
+    "address": "0xc7b0b46dc0a1131adbdb8762a2773f2769abb527"
+  }
```
holic, line 8: Should we gitignore this?

packages/gas-tank/src/PaymasterSystem.sol
alvrs: the reason i went with string errors here is that unfortunately these errors are not surfaced to users anyway (the entry point contract doesn't forward them) and string errors make it easier to debug what's going wrong in the tx trace if one doesn't have the ABI handy

packages/gas-tank/worlds.json
alvrs: the `31337` part is unnecessary but I think for the other chains it's actually useful to see where the gas tank is deployed to

alvrs:
Merging this into #2653 because it's needed for the full local AA setup

PR #2653, opened 2024-04-13
feat(account-kit): add cli for local bundler setup

alvrs:


packages/account-kit/package.json
```diff
@@ -23,6 +23,10 @@
       ]
     }
   },
+  "bin": {
+    "aa-setup": "./dist/aa-setup.js",
+    "anvil-set-balance": "./dist/anvil-set-balance.js"
```
holic, line 28: I'm not really convinced this does enough to justify its own binary to maintain

packages/account-kit/tsup.config.ts
```diff
@@ -4,6 +4,8 @@ export default defineConfig({
   entry: {
     index: "src/exports/index.ts",
     internal: "src/exports/internal.ts",
+    "anvil-set-balance": "cli/anvil-set-balance.ts",
+    "aa-setup": "cli/aa-setup.ts",
```
holic, line 8:
```suggestion
    "bin/anvil-set-balance": "cli/anvil-set-balance.ts",
    "bin/aa-setup": "cli/aa-setup.ts",
```

packages/account-kit/cli/aa-setup.ts
holic, line 1: should we name this something like `local-aa-setup` or `local-setup`? (maybe "aa" is assumed/inferred by this package being "account kit")

packages/account-kit/cli/utils/deployGasTank.ts
```diff
@@ -0,0 +1,54 @@
+import { existsSync } from "fs";
+import { $, execa } from "execa";
+import { join, dirname } from "path";
+import { deployerClient, devPrivateKey } from "./deployLocal";
+import worlds from "@latticexyz/gas-tank/worlds.json" assert { type: "json" };
+
+const accountKitPath = join(dirname(process.argv[1]), "..");
```
holic, line 7: what does this path represent? how is the arg used?

packages/account-kit/cli/utils/deployLocal.ts
```diff
@@ -0,0 +1,43 @@
...
+export const devPrivateKey = "0x4bbbf85ce3377467afe5d46f804f221813b2bb87f24d81f60f1fcdbf7cbf4356";
+const account = privateKeyToAccount(devPrivateKey);
+
+export const deployerClient = createWalletClient({
+  account,
+  chain: foundry,
+  transport: http("http://127.0.0.1:8545"),
+}).extend(publicActions);
+
+export const deterministicDeployer: Address = "0x4e59b44847b379578588920ca78fbf26c0b4956c";
```
holic, line 15: wonder if we should expose some of our deployment utils from cli package

holic: or expand our deploy logic (maybe via mud config?) to allow including external contracts to deploy alongside

packages/gas-tank/package.json
```diff
@@ -0,0 +1,61 @@
...
+    "@latticexyz/cli": "workspace:*",
+    "@latticexyz/schema-type": "workspace:*",
+    "@latticexyz/store": "workspace:*",
+    "@latticexyz/world": "workspace:*",
+    "@latticexyz/world-modules": "workspace:*"
+  },
+  "devDependencies": {
+    "@openzeppelin/contracts": "^5.0.0",
+    "@types/node": "^18.15.11",
+    "account-abstraction": "https://github.com/eth-infinitism/account-abstraction.git#v0.7.0",
```
holic, line 53: why do we use this git reference here instead of installing this via `"@account-abstraction": "0.7.0"` like we do in account-kit?

packages/account-kit/package.json
alvrs: i feel like we need to provide something that's easier than telling people to copy paste `cast rpc --rpc-url http://127.0.0.1:8545 anvil_setBalance <address> $(cast to-unit 1ether wei)`

holic: I think this is fine, and gives folks exposure to cast + anvil RPC commands. My worry is that folks will start asking for more CLI options like allowing for setting RPC URL or whatever and then soon we'll just have a generic wrapper around `cast rpc anvil_setBalance`.

alvrs: discussed on discord: we'll remove this wrapper but will add a feature to the account kit bridge screen to fund the gas tank balance ("bridge from anvil to anvil") without executing a transaction from metamask

packages/account-kit/cli/utils/deployGasTank.ts
alvrs: `process.argv[1]` is the path to the currently running script (`process.argv[2]` would be the first user provided arg). Added a comment to make this clearer.

packages/account-kit/cli/utils/deployLocal.ts
alvrs: Agree something in the mud config to deploy extra contracts might be helpful (this is also what `PostDeploy.s.sol`'s original intention was). In this specific case where we're creating a "standalone" script to run local account abstraction infra adding it to the mud deployer would probably not help though. Agree exposing `ensureContract` could help deduping some logic here, but would need some more changes (ie accepting a salt, making `deployedBytecodeSize` optional) so i'd rather add a follow up issue for that

packages/gas-tank/package.json
alvrs: an oversight! fixed

PR #2655, opened 2024-04-14
build: disable DTS in dev

### LINKED ISSUE #2418: investigate slow DTS build
- do we need this in dev environments? or just before publishing packages?
- why is it so slow and resource heavy? is it our complex, old config types?

currently store-sync is eating up ram and causing oom issues in some enviroments

holic:
closes #2418

### DIFF
--- a/packages/cli/tsup.config.ts
+++ b/packages/cli/tsup.config.ts
@@ -21,7 +21,7 @@ export default defineConfig({
   entry: ["src/index.ts", "src/mud.ts"],
   target: "esnext",
   format: ["esm"],
-  dts: false,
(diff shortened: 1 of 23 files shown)


packages/cli/tsup.config.ts
```diff
@@ -21,7 +21,7 @@ export default defineConfig({
   entry: ["src/index.ts", "src/mud.ts"],
   target: "esnext",
   format: ["esm"],
-  dts: false,
```
holic, line 24: some of these were `false` and unclear why, so I enabled for now to see if CI passes

alvrs (approved):

PR #2656, opened 2024-04-14
feat(dev-tools): add balance top up button when running on anvil

holic:
image

### DIFF
--- a/packages/dev-tools/src/summary/AccountSummary.tsx
+++ b/packages/dev-tools/src/summary/AccountSummary.tsx
@@ -1,12 +1,15 @@
 import { useEffect, useState } from "react";
-import { formatUnits } from "viem";
+import { formatUnits, parseEther, testActions } from "viem";
 import { useDevToolsContext } from "../DevToolsContext";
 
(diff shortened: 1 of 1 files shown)


packages/dev-tools/src/summary/AccountSummary.tsx
```diff
@@ -1,12 +1,15 @@
 import { useEffect, useState } from "react";
-import { formatUnits } from "viem";
+import { formatUnits, parseEther, testActions } from "viem";
 import { useDevToolsContext } from "../DevToolsContext";
 
 export function AccountSummary() {
   const { publicClient, walletClient } = useDevToolsContext();
+  const walletAddress = walletClient?.account?.address;
+  const testClient = publicClient.chain.id === 31337 ? publicClient.extend(testActions({ mode: "anvil" })) : null;
```
holic, line 8: deriving this from the public client for now because wallet client in account-kit branch gave errors due to it pointing at a bundler instance, which rejected the anvil rpc methods

alvrs (approved):

PR #2657, opened 2024-04-15
docs(guides/best-practices/dividing-into-systems): first version

qbzzt:


docs/pages/guides/best-practices/dividing-into-systems.mdx
```diff
@@ -0,0 +1,161 @@
+import { Callout } from "nextra/components";
+
+# Dividing Code into `System`s
+
+It is usually simplest to keep all the code that needs the same permissions and lives in the same namespace in a single `System`.
+This page discusses situations in which either you need to divide your code into multiple `System`s, or it seems like that would be a good solution.
```
alvrs, line 6: I would say the guiding policy of whether to keep code in the same system or different systems is whether the code depends on each other (ie some public function is also used within the system). If that's not the case I would always separate code that's semantically separate into different systems (ie a `MoveSystem`, an `InventorySystem`, a `HealthSystem` etc instead of one big `GameSystem`)

docs/pages/guides/best-practices/dividing-into-systems.mdx
```diff
@@ -0,0 +1,161 @@
...
+There are several reasons to divide the code in a namespace into multiple `System`s.
+
+### Access Control
+
+Each `System` is either publicly accessible (can be called by anybody) or accessible only from authorized addresses.
+These addresses are can be:
+
+- [Externally owned accounts](https://ethereum.org/en/developers/docs/accounts/#types-of-account)
+- Contracts outside of the `World`
+- `System`s that are _not_ in the root namespace
```
alvrs, line 19: this one is a little confusing because systems in the root namespace can also call other systems (including the private ones), it's just requires a different implementation

docs/pages/guides/best-practices/dividing-into-systems.mdx
```diff
@@ -0,0 +1,161 @@
...
+
+You can move the code for `move` and `eat` into [public libraries](/guides/best-practices/system-best-practices#use-libraries-to-bypass-the-contract-size-limit).
+These libraries can then be called from `LocationSystem`, `EnergySystem`, and `CombinedActionSystem`.
+
+<details>
+  <summary>Illustration</summary>
+  ![](./libraries.svg)
+</details>
+
+Alternatively, you might be able to unite `LocationSystem` and `EnergySystem` into a single `GameSystem`, which turns the calls from `eatAndMove` into standard function calls within a contract.
```
alvrs, line 107: Mhm I don't feel good about this pattern (see comment above) but I understand what you're trying to point out. Maybe we can rephrase to be more like "if the logic is always coupled (ie moving always takes energy or sth) then you should combine them", but then it's basically the same argument as in "Modularity", so I maybe we should just remove this sentence.

docs/pages/guides/best-practices/dividing-into-systems.mdx
```diff
@@ -0,0 +1,161 @@
...
+  ![](./callFrom.svg)
+</details>
+
+#### When root namespace is unavoidable
+
+There are a few cases that require `System` functions to be called from the root namespace.
+
+- The called `System` needs to know the value sent with the transaction.
+- You cannot use `callFrom` because users won't sign `registerDelegation`.
+- The code that the calling `System` has to implement needs to be in the root namespace for some other reason (for example, to modify system tables).
```
alvrs, line 158:
```suggestion
- The code that the calling `System` has to implement needs to be in the root namespace for some other reason (for example, to modify root tables).
```
Do you mean root tables here? If not, what do you mean by system tables?

docs/pages/world/systems.mdx
```diff
@@ -125,6 +130,12 @@ When calling other contracts from a `System`, be aware that if you use `delegate
 
 ### Calling one `System` from another
 
+<Callout type="info" emoji="ℹ️">
+  Sometimes you think you need to call a `System` from another, when there is no reason to have separate `System`s at
+  all. [See here](/guides/best-practices/dividing-into-systems) for a discussion of when it makes sense to have a single
+  `System` and when splitting the code makes sense.
```
alvrs, line 136:
```suggestion
Sometimes there are [easier alternatives than calling a `System` from another](/guides/best-practices/dividing-into-systems).
```

docs/pages/guides/best-practices/batchCall.svg
alvrs, line 1: Can we get @achalvs to prettify these graphics before merging?

achal-ritual: Yes, will have this ready shortly! @alvrs

achalvs: Yes, will have this ready by EOD @alvrs

docs/pages/guides/best-practices/dividing-into-systems.mdx
qbzzt: Also, does access control happen in those calls, or are root `System`s allowed to call any function on any `System`, public or private?

alvrs (approved):

PR #2660, opened 2024-04-15
feat(world): store function signatures onchain

yonadaaa:
If we store function signature onchain, we can decode calldata onchain, for example in `callWithSignature` (and making signed typed data more readable)


holic:
closing for now due to too much gas and maybe not worth the trade off (esp because our use case would add lots more calldata)

PR #2661, opened 2024-04-15
fix(cli): stack too deep error when too many fields

yonadaaa:
adding test case for now


yonadaaa:
Closing as this issue is unavoidable when you have many fields

Issue #2615, opened 2024-04-04
add contract verification to mud deployer


alvrs:

yonadaaa:
done in https://github.com/latticexyz/mud/pull/2662

PR #2662, opened 2024-04-15
feat(cli): verify command

yonadaaa:
Adds a `mud verify --worldAddress` command

Solves https://github.com/latticexyz/mud/issues/2615


packages/cli/src/commands/verify.ts
```diff
@@ -0,0 +1,84 @@
...
+const commandModule: CommandModule<Options, Options> = {
+  command: "verify",
+
+  describe: "Verify contracts",
+
+  builder(yargs) {
+    return yargs.options({
+      configPath: { type: "string", desc: "Path to the config file" },
+      profile: { type: "string", desc: "The foundry profile to use" },
+      worldAddress: { type: "string", desc: "Deploy to an existing World at the given address" },
```
holic, line 29: desc seems wrong here?

packages/cli/src/commands/verify.ts
```diff
@@ -0,0 +1,84 @@
...
+    });
+  },
+
+  async handler(opts) {
+    const profile = opts.profile ?? process.env.FOUNDRY_PROFILE;
+
+    const configV2 = (await loadConfig(opts.configPath)) as WorldConfig;
+    const config = worldToV1(configV2);
+
+    const srcDir = opts.srcDir ?? (await getSrcDirectory(profile));
```
holic, line 40: do we need the source dir for verification?

packages/cli/src/deploy/ensureWorldFactory.ts
```diff
@@ -59,40 +56,85 @@ export async function ensureWorldFactory(
...
+    args: [accessManagementSystem, balanceTransferSystem, batchCallSystem, registration],
+  });
+
+  const initModule = getCreate2Address({ from: deployerAddress, bytecode: initModuleBytecode, salt });
+
+  const worldFactoryBytecode = encodeDeployData({
+    bytecode: worldFactoryBuild.bytecode.object as Hex,
+    abi: worldFactoryAbi,
+    args: [initModule],
+  });
```
holic, line 133: why do we do all this again if it's returned by `getWorldFactoryContracts`?

packages/cli/src/deploy/ensureWorldFactory.ts
```diff
@@ -11,10 +11,7 @@ import { salt } from "./common";
 import { ensureContractsDeployed } from "./ensureContractsDeployed";
 import { Contract } from "./ensureContract";
 
-export async function ensureWorldFactory(
-  client: Client<Transport, Chain | undefined, Account>,
-  deployerAddress: Hex,
-): Promise<Address> {
+export function getWorldFactoryContracts(deployerAddress: Hex): readonly Contract[] {
```
holic, line 14: can we move this into its own file outside of deploy since it's used in verify now too?

packages/cli/src/verify.ts
```diff
@@ -0,0 +1,80 @@
...
+};
+
+async function verifyContract(
+  foundryProfile: string | undefined,
+  deployerAddress: Hex,
+  contract: Contract,
+  rpc: string,
+  cwd?: string,
+) {
+  if (!contract.label) {
```
holic, line 24: can we do this at the type level instead of runtime?

packages/cli/src/verify.ts
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  const rpc = await getRpcUrl(foundryProfile);
+
+  const client = createWalletClient({
+    transport: http(rpc),
+    account: privateKeyToAccount(privateKey),
+  });
+
+  const deployerAddress = await ensureDeployer(client);
```
holic, line 58: why do we have to deploy the deployer when verifying?

packages/cli/src/verify.ts
```diff
@@ -0,0 +1,80 @@
...
+    getWorldFactoryContracts(deployerAddress).map((contract) =>
+      verifyContract(foundryProfile, deployerAddress, contract, rpc, "node_modules/@latticexyz/world"),
+    ),
+  );
+
+  await Promise.all(
+    modules.map((contract) =>
+      verifyContract(foundryProfile, deployerAddress, contract, rpc, "node_modules/@latticexyz/world-modules"),
+    ),
+  );
```
holic, line 72: can these be parallelized with the above?

```
await Promise.all([
  ...getWorldFactoryContracts(...).map(...),
  modules.map(...),
]);
```

holic: and with the parallelizing, anything to be aware of as far as sourcify rate limits?

if so, we can wrap this in `p-queue` to ensure some level of concurrency

packages/cli/src/commands/verify.ts
yonadaaa: Yes, we use it to fetch the user-defined systems

packages/cli/src/deploy/ensureWorldFactory.ts
yonadaaa: Yes we should be consistent with one function-per-file!

packages/cli/src/verify.ts
yonadaaa: We could use an `as` type assertion, or rework the whole `Contract` type this is using

packages/cli/src/verify.ts
yonadaaa: We should just be getting the address, oops

packages/cli/src/commands/verify.ts
```diff
@@ -0,0 +1,84 @@
...
+};
+
+const commandModule: CommandModule<Options, Options> = {
+  command: "verify",
+
+  describe: "Verify contracts",
+
+  builder(yargs) {
+    return yargs.options({
+      configPath: { type: "string", desc: "Path to the config file" },
```
alvrs, line 27: is this the mud config?
```suggestion
      configPath: { type: "string", desc: "Path to the mud.config.ts file" },
```

packages/cli/src/commands/verify.ts
```diff
@@ -0,0 +1,84 @@
...
+      worldAddress: { type: "string", desc: "Deploy to an existing World at the given address" },
+      srcDir: { type: "string", desc: "Source directory. Defaults to foundry src directory." },
+    });
+  },
+
+  async handler(opts) {
+    const profile = opts.profile ?? process.env.FOUNDRY_PROFILE;
+
+    const configV2 = (await loadConfig(opts.configPath)) as WorldConfig;
+    const config = worldToV1(configV2);
```
alvrs, line 38: are there other utils in here that depend on this being the v1 config or could we just use the v2 config in here?

packages/cli/src/verify.ts
alvrs: could probably do something like `contract: Contract & Required<Pick<Contract, "label">>

packages/cli/src/verify.ts
```diff
@@ -0,0 +1,80 @@
...
+  rpc: string,
+  cwd?: string,
+) {
+  if (!contract.label) {
+    throw new MUDError("Need a label");
+  }
+
+  const system = getCreate2Address({ from: deployerAddress, bytecode: contract.bytecode, salt });
+
+  await forge(["verify-contract", system, contract.label, "--verifier", "sourcify", "--rpc-url", rpc], {
```
alvrs, line 30: can we make the verifier here a cli argument so it also works with other verifiers than sourcify?

alvrs: also i guess we need the verifier url (either in the foundry toml or cli arg)

packages/cli/src/verify.ts
alvrs: we could just export `deployer` from `ensureDeployer` no? looks like it's constant

packages/cli/src/verify.ts
```diff
@@ -0,0 +1,80 @@
...
+    ),
+  );
+
+  await Promise.all(
+    modules.map((contract) =>
+      verifyContract(foundryProfile, deployerAddress, contract, rpc, "node_modules/@latticexyz/world-modules"),
+    ),
+  );
+
+  if (worldAddress) {
```
alvrs, line 74: wonder if we should make this required 🤔 can't really think of a use case of wanting to verify everything except the world, and might be a tripwire if the world is not being verified

packages/cli/src/verify.ts
yonadaaa: Oh yes 🤔  Unless their chain does not allow for non EIP-155 signed transactions, which means the deployer is non-deterministic

packages/cli/src/verify.ts
yonadaaa: Let's just make it verified, not hard for people to fork in that very rare usecase

packages/cli/src/verify.ts
yonadaaa: With this method I do get odd behaviour from sourcify, lemme try a p-queue

holic: no worries if that doesn't work, there could be an order of operations thing by sourcify

it's the same reason we deploy certain chunks of contracts before others, to ensure the previous ones are there before the new ones because they depend on em and will fail otherwise

packages/cli/src/verify.ts
holic: prob need to add `deployerAddress`

packages/cli/src/deploy/ensureWorldFactory.ts
yonadaaa: Fixed, the object now returns the address

packages/cli/src/commands/verify.ts
yonadaaa: I put "Path to the MUD config file" as it felt less filename/implementation-specific

packages/cli/src/commands/verify.ts
yonadaaa: we need it to get the user-defined systems, decided to leave as is

packages/cli/src/commands/verify.ts
```diff
@@ -0,0 +1,100 @@
...
+    return yargs.options({
+      configPath: { type: "string", desc: "Path to the MUD config file" },
+      profile: { type: "string", desc: "The foundry profile to use" },
+      worldAddress: { type: "string", desc: "Verify an existing World at the given address" },
+      srcDir: { type: "string", desc: "Source directory. Defaults to foundry src directory." },
+      verifier: { type: "string", desc: "The verifier to use" },
+      verifierUrl: {
+        type: "string",
+        desc: "The verification provider.",
+      },
```
alvrs, line 39: should we also default these to the foundry profile? (i assume there is a way to define it in the profile?)

packages/cli/src/commands/verify.ts
```diff
@@ -0,0 +1,100 @@
...
+    });
+  },
+
+  async handler(opts) {
+    if (!opts.worldAddress) {
+      throw new MUDError("Need a world address");
+    }
```
alvrs, line 46: this is not an optional arg right? so `yargs` would already throw if it's not provided

packages/cli/src/commands/verify.ts
```diff
@@ -0,0 +1,100 @@
...
+        modules,
+        worldAddress: opts.worldAddress as Hex,
+        verifier: opts.verifier,
+        verifierUrl: opts.verifierUrl,
+      });
+    } catch (error) {
+      logError(error);
+      process.exit(1);
+    }
+    // process.exit(0);
```
alvrs, line 96: why is this commented out?

packages/cli/src/verifyContract.ts
```diff
@@ -0,0 +1,57 @@
...
+export async function verifyContractCreate2(
+  { name, rpc, verifier, verifierUrl, from, bytecode, salt }: Omit<VerificationOptions, "address"> & Create2Options,
+  options?: ForgeOptions,
+) {
+  const address = getCreate2Address({ from, bytecode, salt });
+
+  return verifyContract({ address, name, rpc, verifier, verifierUrl }, options);
+}
+
+export async function verifyContractCreate2DefaultSalt(
```
alvrs, line 45: why do we need a separate function for this? can we just default the `salt` parameter in `verifyContractCreate2` if it's not provided?

packages/cli/src/verifyContract.ts
```diff
@@ -0,0 +1,57 @@
...
+    args.push("--verifier", verifier);
+  }
+  if (verifierUrl) {
+    args.push("--verifier-url", verifierUrl);
+  }
+  await forge(args, options);
+}
+
+export async function verifyContractCreate2(
+  { name, rpc, verifier, verifierUrl, from, bytecode, salt }: Omit<VerificationOptions, "address"> & Create2Options,
```
alvrs, line 37: What if we just make the options this:

```ts
type VerifyContractOptions = {
  foundryProfile?: string;
  verifier?: string;
  verifierUrl?: string;
} & ({ address: Hex; } | Create2Options);

```

Then we would only need one `verifyContract` function instead of a different path for `verifyContractCreate2`.

packages/cli/src/commands/verify.ts
yonadaaa: Good shout, made it required!

packages/cli/src/verifyContract.ts
yonadaaa: Oh that's wayyy better. 

my brain lives in Solidity-land where you create overloads for each set of arguments

packages/cli/src/commands/verify.ts
yonadaaa: I wondered if we needed it but the process exits fine without it

packages/cli/src/verifyContract.ts
```diff
@@ -0,0 +1,43 @@
...
+  if (verifier) {
+    args.push("--verifier", verifier);
+  }
+  if (verifierUrl) {
+    args.push("--verifier-url", verifierUrl);
+  }
+  await forge(args, options);
+}
+
+export async function verifyContractCreate2(
```
holic, line 36: to match `getCreate2Address`
```suggestion
export async function verifyCreate2Contract(
```

yonadaaa: This function is gone!

packages/cli/src/commands/verify.ts
yonadaaa: I don't believe you can put these in the Foundry profile? I tried this and I `verify-contract` didn't pick it up:

```toml
[profile.holesky]
eth_rpc_url = "https://ethereum-holesky.publicnode.com"
verifier="blockscout"
```

alvrs (dismissed):
lgtm! have you tested this with blockscout on garnet?

yonadaaa:
> lgtm! have you tested this with blockscout on garnet?

No, let me test that first!

packages/cli/src/commands/verify.ts
holic: we'll revisit this in a follow up to allow specifying a chain like `--chain mainnet` and look these up via explorers in chain config

packages/cli/src/deploy/ensureDeployer.ts
```diff
@@ -3,7 +3,7 @@ import { getBalance, getBytecode, sendRawTransaction, sendTransaction, waitForTr
 import deployment from "./create2/deployment.json";
 import { debug } from "./debug";
 
-const deployer = `0x${deployment.address}` as const;
+export const deployer = `0x${deployment.address}` as const;
```
holic, line 6: this was intentionally not exported because the deployer is configurable by the person deploying

we should make sure to handle that!

yonadaaa: good catch!!

packages/cli/src/deploy/getWorldProxyFactoryContracts.ts
```diff
@@ -0,0 +1,96 @@
...
+import registrationSystemBuild from "@latticexyz/world/out/RegistrationSystem.sol/RegistrationSystem.json" assert { type: "json" };
+import initModuleBuild from "@latticexyz/world/out/InitModule.sol/InitModule.json" assert { type: "json" };
+import initModuleAbi from "@latticexyz/world/out/InitModule.sol/InitModule.abi.json" assert { type: "json" };
+import worldProxyFactoryBuild from "@latticexyz/world/out/WorldProxyFactory.sol/WorldProxyFactory.json" assert { type: "json" };
+import worldProxyFactoryAbi from "@latticexyz/world/out/WorldProxyFactory.sol/WorldProxyFactory.abi.json" assert { type: "json" };
+import { Hex, getCreate2Address, encodeDeployData, size } from "viem";
+import { salt } from "./common";
+import { Contract } from "./ensureContract";
+
+export function getWorldProxyFactoryContracts(deployerAddress: Hex): Record<string, Contract & { address: Hex }> {
```
holic, line 13: would it make sense to have this call `getWorldFactoryContracts`? or pull out the shared behavior into something like `getWorldDependencies`?

packages/cli/src/verify.ts
```diff
@@ -0,0 +1,103 @@
...
+        {
+          profile: foundryProfile,
+          cwd: "node_modules/@latticexyz/world",
+        },
+      ).catch((error) => {
+        console.error(`Error verifying World contract:`, error);
+      }),
+  ];
+
+  tasks.forEach((task) => verifyQueue.add(task));
```
holic, line 102: since the queue handles when these get fired, this might be easier to read as

```ts
systems.map(({ name, bytecode }) => verifyQueue.add(() => verifyContract(
```

packages/cli/src/deploy/getWorldProxyFactoryContracts.ts
holic: also any chance we could pull out this refactor into a separate PR (along with `getDeployer`)?

yonadaaa: Lets do that! it essential works for now

yonadaaa: done!

packages/cli/src/commands/verify.ts
```diff
@@ -0,0 +1,125 @@
...
+  worldAddress: string;
+  configPath?: string;
+  profile?: string;
+  rpc?: string;
+  rpcBatch?: boolean;
+  srcDir?: string;
+  verifier?: string;
+  verifierUrl?: string;
+  useProxy?: boolean;
+};
```
holic, line 25: did you decide not to flip this around and derive this from the options?

yonadaaa: Sorry, derive what from which options?

holic:
```ts
const verifyOptions = {
  deployerAddress: ...
  worldAddress: ...
  ...
};

type VerifyOptions = InferredOptionTypes<typeof verifyOptions>;
```

yonadaaa: oh the types! yes I tried totally restructuring things but it made the process exit early. `yargs` is weird. 

I will try just change the types

packages/cli/src/verify.ts
```diff
@@ -0,0 +1,129 @@
...
+  systems: { name: string; bytecode: Hex }[];
+  modules: { name: string; bytecode: Hex }[];
+  worldAddress: Hex;
+  /**
+   * Address of determinstic deployment proxy: https://github.com/Arachnid/deterministic-deployment-proxy
+   * By default, we look for a deployment at 0x4e59b44847b379578588920ca78fbf26c0b4956c.
+   * If it is not deployed or the target chain does not support legacy transactions, the user must set the deployer manually.
+   */
+  deployerAddress?: Hex;
+  useProxy?: boolean;
```
holic, line 25: chatted IRL and would like to detect proxy by calling
```
getStorageAt({
  address: ...
  slot: '0x360894a13ba1a3210667c828492db98dca3e2076cc3735a920a3ca505d382bbc'
```

packages/cli/src/commands/verify.ts
```diff
@@ -0,0 +1,132 @@
...
+        bytecode: contractData.bytecode,
+      };
+    });
+
+    // If the proxy implementation storage slot is set on the World, the World was deployed as a proxy.
+    const proxyImplementation = await publicClient.getStorageAt({
+      address: opts.worldAddress as Address,
+      slot: ERC1967_IMPLEMENTATION_SLOT,
+    });
+    const useProxy = proxyImplementation !== zeroHash;
```
holic, line 115: can we move this logic into `verify`? then we don't need the extra `verify` option

packages/cli/src/verify.ts
holic, line 25: still need to remove this CLI arg

holic (approved):

PR #2668, opened 2024-04-16
feat(store-sync): add status and block number to return type of waitForTransaction

alvrs:


packages/store-sync/src/createStoreSync.ts
```diff
@@ -271,25 +272,30 @@ export async function createStoreSync<config extends StoreConfig = StoreConfig>(
...
+        }
 
         try {
           const lastBlock = blocks[0];
           debug("fetching tx receipt for block", lastBlock.blockNumber);
-          const receipt = await publicClient.getTransactionReceipt({ hash: tx });
-          return lastBlock.blockNumber >= receipt.blockNumber;
+          const { status, blockNumber } = await publicClient.getTransactionReceipt({ hash: tx });
+          if (lastBlock.blockNumber >= blockNumber) {
+            return { status, blockNumber };
```
yonadaaa, line 286: Why not return everything from `getTransactionReceipt`? Then this is like a wrapper for that function

yonadaaa (approved):

packages/store-sync/src/createStoreSync.ts
```diff
@@ -271,25 +272,30 @@ export async function createStoreSync<config extends StoreConfig = StoreConfig>(
       // We use `mergeMap` instead of `concatMap` here to send the fetch request immediately when a new block range appears,
       // instead of sending the next request only when the previous one completed.
       mergeMap(async (blocks) => {
-        const txs = blocks.flatMap((block) => block.logs.map((op) => op.transactionHash).filter(isDefined));
-        if (txs.includes(tx)) return true;
+        for (const block of blocks) {
+          const txs = block.logs.map((op) => op.transactionHash);
+          // If the transaction caused a log, it must have succeeded
+          if (txs.includes(tx)) return { blockNumber: block.blockNumber, status: "success" as const };
```
holic, line 278:
```suggestion
          if (txs.includes(tx)) {
            return { blockNumber: block.blockNumber, status: "success" as const };
          }
```

packages/store-sync/src/createStoreSync.ts
alvrs: because we don't have access to most of this information in the early return case above and it feels wrong to only include this information in some cases based on a race condition

packages/store-sync/src/createStoreSync.ts
```diff
@@ -271,25 +272,32 @@ export async function createStoreSync<config extends StoreConfig = StoreConfig>(
...
+          }
+        }
 
         try {
           const lastBlock = blocks[0];
           debug("fetching tx receipt for block", lastBlock.blockNumber);
-          const receipt = await publicClient.getTransactionReceipt({ hash: tx });
-          return lastBlock.blockNumber >= receipt.blockNumber;
+          const { status, blockNumber, transactionHash } = await publicClient.getTransactionReceipt({ hash: tx });
+          if (lastBlock.blockNumber >= blockNumber) {
```
holic, line 287: looking at this now, I wonder if this is maybe the cause of infinite spinning sometimes

if we get a receipt, we only _sometimes_ return it?

I think this might be here to make sure the syncing is caught up to where the block for this receipt came through, but just want to check/confirm that we would refetch this transaction receipt while waiting, or if this would cause it to sort of "fall out" of the loop and ~wait indefinitely

alvrs: We would refetch it when processing the next block. Agree we should add caching to fetching tx receipts though.

holic (approved):

packages/store-sync/src/createStoreSync.ts
alvrs: https://github.com/latticexyz/mud/issues/2670

PR #2683, opened 2024-04-17
fix(cli): fix deployer warning

holic:
fixes this warning:

```
  mud:cli:deploy found CREATE2 deployer at 0x4e59b44847b379578588920ca78fbf26c0b4956c +0ms

  ⚠️ Bytecode for deployer at 0x4e59b44847b379578588920ca78fbf26c0b4956c did not match the expected CREATE2 bytecode. You may have unexpected results.
```

(I didn't realize that the bytecode in the deployer repo was the creation code, not the onchain bytecode)


packages/cli/src/deploy/ensureDeployer.ts
```diff
@@ -1,16 +1,15 @@
...
 
 const deployer = `0x${deployment.address}` as const;
-const deployerBytecode = `0x${deployment.bytecode}` as const;
 
 export async function ensureDeployer(client: Client<Transport, Chain | undefined, Account>): Promise<Address> {
   const bytecode = await getBytecode(client, { address: deployer });
   if (bytecode) {
     debug("found CREATE2 deployer at", deployer);
-    if (bytecode !== deployerBytecode) {
+    if (bytecode !== sliceHex(`0x${deployment.creationCode}`, 14)) {
```
alvrs, line 12: what's the 14?

holic, line 12: another approach here could be
```suggestion
    if (deployment.creationCode.endsWith(bytecode.slice(2))) {
```
but that feels less accurate to me

holic: image

holic, line 12:
```suggestion
    // check if deployed bytecode is the same as the expected bytecode (minus 14-bytes creation code prefix)
    if (bytecode !== sliceHex(`0x${deployment.creationCode}`, 14)) {
```

PR #2686, opened 2024-04-18
fix(world-modules): properly concat baseURI and tokenURI in ERC721 module

Kooshaba:
- previously `tokenURI` was not properly returning `concat(baseURI, tokenURI)`
- this uses a random `uintToString` function i found, lmk if you want to use OpenZeppelin or something
- launch blocker for Sky Strife


alvrs:
> his uses a random uintToString function i found, lmk if you want to use OpenZeppelin or something

Would definitely prefer the OZ implementation if possible, looks like we can just use this https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/Strings.sol#L24 (would just copy over the Math and String lib from there and refer back to the source in the header, so we don't have to add more remappings in consuming projects)

packages/world-modules/src/modules/erc721-puppet/ERC721System.sol
```diff
@@ -64,7 +64,7 @@ contract ERC721System is IERC721Mintable, System, PuppetMaster {
 
     string memory baseURI = _baseURI();
     string memory _tokenURI = TokenURI.get(_tokenUriTableId(_namespace()), tokenId);
-    _tokenURI = bytes(_tokenURI).length > 0 ? _tokenURI : string(abi.encodePacked(tokenId));
+    _tokenURI = bytes(_tokenURI).length > 0 ? _tokenURI : uintToString(tokenId);
```
yonadaaa, line 67: Agreeing with @alvrs, we can lift the [`Strings`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/Strings.sol) library that OZ uses [here](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/ERC721.sol#L92) in their ERC721 implementation.

In MUD we straight up copy a lot of files: https://github.com/latticexyz/mud/blob/main/packages/world/src/ERC165Checker.sol

yonadaaa (dismissed):

holic (approved):

holic (approved):

PR #2688, opened 2024-04-18
feat(common): kms custom account

yonadaaa:
following https://github.com/rumblefishdev/eth-signer-kms

tests need local KMS to run https://github.com/nsmithuk/local-kms


packages/cli/package.json
```diff
@@ -30,8 +30,8 @@
     "test:ci": "pnpm run test"
   },
   "dependencies": {
-    "@ethersproject/abi": "^5.7.0",
-    "@ethersproject/providers": "^5.7.2",
+    "@aws-sdk/client-kms": "*",
+    "@ethersproject/bytes": "^5.5.4",
```
holic, line 34: why 😭

packages/cli/package.json
```diff
@@ -30,8 +30,8 @@
     "test:ci": "pnpm run test"
   },
   "dependencies": {
-    "@ethersproject/abi": "^5.7.0",
-    "@ethersproject/providers": "^5.7.2",
+    "@aws-sdk/client-kms": "*",
```
holic, line 33: should make this an optional peer dep

packages/common/package.json
```diff
@@ -57,9 +57,12 @@
     "test:ci": "pnpm run test"
   },
   "dependencies": {
+    "@aws-sdk/client-kms": "*",
```
holic, line 60: we should make this an asn1 an optional peer dep and document that

packages/common/package.json
```diff
@@ -57,9 +57,12 @@
...
     "@latticexyz/schema-type": "workspace:*",
     "@solidity-parser/parser": "^0.16.0",
+    "asn1.js": "^5.4.1",
     "debug": "^4.3.4",
+    "ethers": "^5.7.2",
```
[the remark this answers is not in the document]
yonadaaa: Narrowed it down to `@ethersproject/signing-key`

packages/common/src/account/sign.ts
```diff
@@ -0,0 +1,103 @@
+// @ts-expect-error types
```
holic, line 1: what is this for?

packages/common/src/account/sign.ts
holic, line 1: this looks like kms specific stuff - should it go under kms file? or should we have a kms dir here?

packages/common/src/createKmsAccount.ts
```diff
@@ -0,0 +1,54 @@
...
+import { toAccount } from "viem/accounts";
+import { getEthAddressFromKMS } from "./account/kms";
+import { sign } from "./account/sign";
+
+/**
+ * @description Creates an Account from a KMS key and instance.
+ *
+ * @returns A Local Account.
+ */
+export async function createKmsAccount(keyId: string, kmsInstance: KMSClient): Promise<LocalAccount> {
```
holic, line 12: would be cool to upstream this and he corresponding utils to viem!

holic:
```suggestion
type CreateKmsAccountOptions = {
  keyId: string;
  client?: KMSClient;
};
export async function createKmsAccount({ keyId, client = new KMSClient() }: CreateKmsAccountOptions): Promise<LocalAccount> {
```

holic: also wonder if we should extend `LocalAccount` with a custom `source` like `aws-kms`

packages/common/package.json
```diff
@@ -72,5 +72,17 @@
     "@types/node": "^18.15.11",
     "tsup": "^6.7.0",
     "vitest": "0.34.6"
+  },
+  "peerDependencies": {
+    "@aws-sdk/client-kms": "^3.556.0",
+    "asn1.js": "^5.4.1"
```
holic, line 78: I _think_ we can be more flexible with the version ranges here

packages/common/src/account/sign.ts
yonadaaa: `asn1.js` doesn't have type definitions. I'm never sure the best way to handle this - should i add a declaration file?

packages/common/src/account/sign.ts
yonadaaa: Yes it's pretty KMS specific, lemme reorganise

packages/common/package.json
yonadaaa: Why would we want to be flexible?

packages/common/src/createKmsAccount.ts
yonadaaa: done! 

what's a `source`? I couldn't find anything on viem

packages/common/package.json
holic: if someone wanted to use `common` for the kms account but there was a bug in a particular version of either of the packages above, I think pnpm would keep you from installing other versions because they wouldn't satisfy the above semver

holic: for peer deps its common to use a wide semver (like `3.x` instead of `^3.556.0`) or just use `*`, but with the expectation that calls to our lib may break if there were breaking changes in the dep version you use

packages/common/src/account/sign.ts
holic: can we restrict this to specifically that package? or just ignore the lines that package uses? just worried about this being an overly wide "ignore these errors"

holic: I wouldn't block on adding type declarations for this lib though!

packages/common/src/createKmsAccount.ts
holic: it's an argument in account object and `LocalAccount` type

holic: I think we can do

```ts
type KMSAccount = LocalAccount<'aws-kms'>;

toAccount({
  source: 'aws-kms',
  ...
});
```

yonadaaa: Slightly different format but yeah we can do it

https://github.com/wevm/viem/blob/cfe6dee1e259f93d054ea77b18d1f4bf19a0592c/src/accounts/hdKeyToAccount.ts#L30-L34

packages/common/package.json
yonadaaa: Done!

packages/common/package.json
```diff
@@ -72,5 +72,17 @@
     "@types/node": "^18.15.11",
     "tsup": "^6.7.0",
     "vitest": "0.34.6"
+  },
+  "peerDependencies": {
+    "@aws-sdk/client-kms": "^3.x",
+    "asn1.js": "^5.x"
```
holic, line 78: surprised this semver is valid, but might be more clear as:
```suggestion
    "@aws-sdk/client-kms": "3.x",
    "asn1.js": "5.x"
```

.changeset/perfect-actors-flow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/common": patch
+---
+
+Added `createKmsAccount`, a Viem Custom Account that signs transactions using AWS KMS.
```
holic, line 5:
```suggestion
Added `createKmsAccount`, a [viem custom account](https://viem.sh/docs/accounts/custom#custom-account) that signs transactions using AWS KMS.
```

.changeset/perfect-actors-flow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/common": patch
+---
+
+Added `createKmsAccount`, a [viem custom account](https://viem.sh/docs/accounts/custom#custom-account) that signs transactions using AWS KMS.
```
holic, line 5:
````suggestion
Added `createKmsAccount`, a [viem custom account](https://viem.sh/docs/accounts/custom#custom-account) that signs transactions using AWS KMS.

To use it, you must first install `@aws-sdk/client-kms@3.x` and `asn1.js@5.x` dependencies into your project. Then create a KMS account with:

```ts
const account = createKmsAccount({ keyId: ... });
```

By default, a `KMSClient` will be created, but you can also pass one in via the `client` option. The default KMS client will use [your environment's AWS SDK configuration](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/configuring-the-jssdk.html).
````

holic: might be worth adding these details into a README file of `account/kms` dir

holic (dismissed):

PR #2695, opened 2024-04-19
feat(world): worldProxy emits hello events when upgraded

yonadaaa:
continued from https://github.com/latticexyz/mud/pull/2632
This means we can detect when the MUD protocol version has changed

### DIFF
--- a/packages/world/src/WorldProxy.sol
+++ b/packages/world/src/WorldProxy.sol
@@ -32,9 +32,6 @@ contract WorldProxy is Proxy {
     _setImplementation(implementation);
 
     StoreCore.initialize();
-    emit IStoreEvents.HelloStore(STORE_VERSION);
-
-    emit IWorldEvents.HelloWorld(WORLD_VERSION);
   }
 
   /**
@@ -44,6 +41,8 @@ contract WorldProxy is Proxy {
     StorageSlot.getAddressSlot(IMPLEMENTATION_SLOT).value = newImplementation;
 
     emit IERC1967.Upgraded(newImplementation);
+    emit IStoreEvents.HelloStore(STORE_VERSION);
+    emit IWorldEvents.HelloWorld(WORLD_VERSION);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


packages/world/src/WorldProxy.sol
```diff
@@ -44,6 +41,8 @@ contract WorldProxy is Proxy {
     StorageSlot.getAddressSlot(IMPLEMENTATION_SLOT).value = newImplementation;
 
     emit IERC1967.Upgraded(newImplementation);
+    emit IStoreEvents.HelloStore(STORE_VERSION);
+    emit IWorldEvents.HelloWorld(WORLD_VERSION);
```
holic, line 45: These need to come from the implementation, not the hardcoded constant

packages/world/src/WorldProxy.sol
```diff
@@ -44,6 +40,10 @@ contract WorldProxy is Proxy {
     StorageSlot.getAddressSlot(IMPLEMENTATION_SLOT).value = newImplementation;
 
     emit IERC1967.Upgraded(newImplementation);
+
+    IBaseWorld world = IBaseWorld(newImplementation);
+    emit IStoreEvents.HelloStore(world.storeVersion());
+    emit IWorldEvents.HelloWorld(world.worldVersion());
```
holic, line 46: although this isn't a place we need to really gas optimize, I kinda like what you had before 
```suggestion
    emit IStoreEvents.HelloStore(IBaseWorld(world).storeVersion());
    emit IWorldEvents.HelloWorld(IBaseWorld(world).worldVersion());
```

yonadaaa: Sure, I'm neither here nor there. Cool to try them both

packages/world/src/WorldProxy.sol
```diff
@@ -44,6 +40,8 @@ contract WorldProxy is Proxy {
     StorageSlot.getAddressSlot(IMPLEMENTATION_SLOT).value = newImplementation;
 
     emit IERC1967.Upgraded(newImplementation);
```
holic, line 42: is it okay that this is emitted even on a new deploy?

holic (approved):

packages/world/src/WorldProxy.sol
yonadaaa: Yes that's part of the spec here https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/proxy/ERC1967/ERC1967Proxy.sol#L26-L28

PR #2697, opened 2024-04-19
latency improvements for submitting user operations

alvrs:


packages/account-kit/src/steps/finalizing/FinalizingStep.tsx
```diff
@@ -32,11 +33,18 @@ export function FinalizingStep() {
       if (!appAccountClient) throw new Error("App account client not ready.");
       if (!registerDelegationSignature) throw new Error("No delegation signature.");
 
+      // TODO: should this use `callWithSignature`?
```
holic, line 36: It does a few lines below?

packages/common/package.json
```diff
@@ -63,6 +63,7 @@
     "execa": "^7.0.0",
     "p-queue": "^7.4.1",
     "p-retry": "^5.1.2",
+    "permissionless": "^0.1.17",
```
holic, line 66: I don’t think this is used?

packages/account-kit/src/steps/finalizing/FinalizingStep.tsx
alvrs: yeah i meant the `callWithSignature` util we have (that i believe we use for `registerSpender`)

packages/common/package.json
alvrs: oh yeah you're right, removed all changes to `common` that had to do with smart accounts again

packages/account-kit/src/steps/finalizing/FinalizingStep.tsx
holic: ohhh I see! that function is built to request the signature so wasn't immediately available for use here could be adapted

PR #2700, opened 2024-04-21
feat(world-modules): string systemId in `callWithSignature` typehash

yonadaaa:
Making signatures more legible with a `string` systemId
image

### DIFF
--- a/packages/world-modules/src/modules/callwithsignature/getSignedMessageHash.sol
+++ b/packages/world-modules/src/modules/callwithsignature/getSignedMessageHash.sol
@@ -2,7 +2,7 @@
 pragma solidity >=0.8.24;
 
 import { ResourceId } from "@latticexyz/store/src/ResourceId.sol";
-import { WorldResourceIdInstance } from "@latticexyz/world/src/WorldResourceId.sol";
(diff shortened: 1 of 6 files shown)


packages/world-modules/src/modules/callwithsignature/getSignedMessageHash.sol
```diff
@@ -34,7 +37,9 @@ function getSignedMessageHash(
       abi.encodePacked(
         "\x19\x01",
         domainSeperator,
-        keccak256(abi.encode(CALL_TYPEHASH, signer, systemId, keccak256(callData), nonce))
+        keccak256(
+          abi.encode(CALL_TYPEHASH, signer, keccak256(abi.encode(systemId.toString())), keccak256(callData), nonce)
```
holic, line 41: I think we want to do a straight cast here, not `toString`, as `toString` is more meant as the human-readable version for logging/debugging (e.g. `sy:<root>:Register`)

yonadaaa: Oh I didn't think about that! thanks

holic (dismissed):

holic:
changeset?

Issue #1755, opened 2023-10-12
consider a base PostDeploy script to inherit from


holic:
mostly so we have a place to set up boilerplate like setting the store address

```solidity
contract PostDeploy is Script {
  function run(address worldAddress) external {
    StoreSwitch.setStoreAddress(worldAddress);
```

but also could mean in the future we add support for multiple PostDeploy scripts by looking at the scripts inheritance chain and, if it inherits from our base PostDeploy, it's executed, regardless of the name (no special PostDeploy naming needed anymore)

yonadaaa:
Following how `MudTest` extends Foundry's `Test`, you could have `MudScript` extending `Script`.

holic:
> Following how `MudTest` extends Foundry's `Test`, you could have `MudScript` extending `Script`.

Definitely! And we could move away from a generic `run(address)` to something like `postDeploy(address)` function, leaving us room to add other "hooks" like `postUpgrade(address)`

holic:
with this structure, we'd be able to split big PostDeploy scripts into multiple files, which could help with issues we're seeing in Sky Strife where kicking off 300+ txs in a single script causes failures due to overlapping nonces

if we went this route, another thing that would be great is to be able to specify a list of "dependencies" for managing deploy order

```solidity
contract UploadMaps is MudScript {
  afterDeployWorld(address worldAddress) {
    ....
  }
}
```

```solidity
contract CreateDebugMatch is MudScript {
  bytes4[] public constant dependsOn = [UploadMaps.afterDeployWorld.selector];

  afterDeployWorld(address worldAddress) {
    ....
  }
}
```

PR #2703, opened 2024-04-22
feat(cli): add user-specified PostDeploy forge options

### LINKED ISSUE #2631: allow manually setting `forge script` params when running deploy with a PostDeploy
- forge script does not seem to fetch gas price from the RPC automatically
- right now i have to manually fetch it and pass it with `--with-gas-price` param

dk1a:
fixes #2631
Similar to `test`'s `forgeOptions`, this allows passing arbitrary user options to `forge script`
In my case this is needed for `--slow`, without which the script breaks because it's too big


packages/cli/src/commands/dev-contracts.ts
```diff
@@ -16,6 +16,7 @@ const devOptions = {
   rpc: deployOptions.rpc,
   configPath: deployOptions.configPath,
   alwaysRunPostDeploy: deployOptions.alwaysRunPostDeploy,
+  postDeployForgeOptions: deployOptions.postDeployForgeOptions,
```
holic, line 19: what about just `forgeOptions` to keep it the same as the other commands?

holic: if not, kinda wondering about `forgeScriptOptions` since I am kinda anticipating more than just post deploy in the future: https://github.com/latticexyz/mud/issues/1755

dk1a: > if not, kinda wondering about `forgeScriptOptions` since I am kinda anticipating more than just post deploy in the future: #1755

Sure, seems like a better name, updated

Issue #2594, opened 2024-04-02
Add support for deploying via KMS to MUD deployer


alvrs:
looks like we need a custom viem client to support this: https://github.com/wevm/viem/discussions/668

PR #2704, opened 2024-04-22
feat(cli): deploy with kms

yonadaaa:
Fixes https://github.com/latticexyz/mud/issues/2594

The AWS configuration and credentials can be set with environment variables: https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html

Specifically, KMS requires `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, `AWS_REGION` and `AWS_ENDPOINT_URL`.


packages/cli/package.json
```diff
@@ -80,5 +80,8 @@
     "tsup": "^6.7.0",
     "tsx": "^3.12.6",
     "vitest": "0.34.6"
+  },
+  "peerDependencies": {
+    "@aws-sdk/client-kms": "^3.556.0"
```
holic, line 85: I think this needs to be an actual dependency. I deleted my earlier comment after realizing this was in the CLI package, not a package end-users would depend on.

packages/cli/src/runDeploy.ts
```diff
@@ -41,6 +43,10 @@ export const deployOptions = {
     type: "string",
     desc: "The deployment salt to use. Defaults to a random salt.",
   },
+  keyId: {
```
holic, line 46: can we make this arg more specific like `awsKmsKeyId`?

also can we support this as an env var as well?

```ts
const keyId = opts.awsKmsKeyId ?? process.env.AWS_KMS_KEY_ID;
const account = keyId ? await createKmsAccount(...) : privateKeyToAccount(privateKey);
```

packages/cli/package.json
holic: I think we also need to install asn1 here

holic (approved):
cleaaannn

packages/cli/src/runDeploy.ts
```diff
@@ -41,6 +43,10 @@ export const deployOptions = {
     type: "string",
     desc: "The deployment salt to use. Defaults to a random salt.",
   },
+  awsKmsKeyId: {
+    type: "string",
+    desc: "The AWS KMS keyId. If set, the World is deployed using a KMS signer instead of local private key.",
```
holic, line 48: tiny thing since they call it "key ID" in the UI
https://docs.aws.amazon.com/kms/latest/developerguide/find-cmk-id-arn.html
```suggestion
    desc: "The AWS KMS key ID. If set, the World is deployed using a KMS signer instead of local private key.",
```

.changeset/early-teachers-greet.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/cli": patch
+---
+
+Added a `awsKmsKeyId` flag to `mud deploy` that deploys the world using an AWS KMS signer.
```
holic, line 5:
```suggestion
Added a `--awsKmsKeyId` flag to `mud deploy` that deploys the world using an AWS KMS signer.
```

holic (approved):
💯

PR #2706, opened 2024-04-22
fix(store-indexer): report correct readiness for stale block tags

alvrs:


.changeset/tricky-walls-yawn.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/store-indexer": patch
+---
+
+The indexer now checks whether it is already caught up on initial boot to report an accurate readiness state while there are no new blocks coming in.
```
holic, line 5:
```suggestion
The indexer will now initialize with the correct readiness state when the chain is not producing new blocks for the provided block tag.
```

holic (dismissed):
should wait for CI but otherwise lgtm!

PR #2709, opened 2024-04-23
fix(cli): add retry to getLogs when getting resource ID's

yonadaaa:
Attempting to fetch invalid logs from an RPC can error with a [400 Bad Request](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400) and `"block is out of range"`. In this case, Viem's `getLogs` does not [retry the query](https://github.com/wevm/viem/blob/main/src/utils/buildRequest.ts#L199-L217). 

However, when using a request router like [proxyd](https://github.com/ethereum-optimism/optimism/tree/develop/proxyd), it is possible that some RPC's are slightly behind and do not have the newer logs. In this case we do actually want to retry the query until the RPC has caught up. 

This PR adds a `pretry` to the `getLogs` call in the `getResourceIds` function which has consistently failed in deployment.

### DIFF
--- a/.changeset/five-pots-switch.md
+++ b/.changeset/five-pots-switch.md
@@ -1,0 +1,5 @@
+---
+"@latticexyz/cli": patch
+---
+
+Fixed `""block is out of range"` errors in the deployer by adding retry logic to the `getLogs` call in the `getResourceIds` function. Previously, fetching the logs could fail if the RPC was out of sync.
--- a/.changeset/five-pots-switch.md
+++ b/.changeset/five-pots-switch.md
@@ -2,4 +2,4 @@
 "@latticexyz/cli": patch
 ---
 
-Fixed `""block is out of range"` errors in the deployer by adding retry logic to the `getLogs` call in the `getResourceIds` function. Previously, fetching the logs could fail if the RPC was out of sync.
+Deploying now retries on "block is out of range" errors, for cases where the RPC is load balanced and out of sync.
--- a/packages/cli/src/deploy/getResourceIds.ts
+++ b/packages/cli/src/deploy/getResourceIds.ts
@@ -3,6 +3,8 @@ import { getLogs } from "viem/actions";
 import { storeSpliceStaticDataEvent } from "@latticexyz/store";
 import { WorldDeploy, storeTables } from "./common";
 import { debug } from "./debug";
+import pRetry from "p-retry";
+import { wait } from "@latticexyz/common/utils";
 
 export async function getResourceIds({
   client,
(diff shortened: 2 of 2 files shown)


packages/cli/src/deploy/getResourceIds.ts
```diff
@@ -15,15 +17,25 @@ export async function getResourceIds({
...
+        toBlock: worldDeploy.stateBlock,
+        event: parseAbiItem(storeSpliceStaticDataEvent),
+        args: { tableId: storeTables.store_ResourceIds.tableId },
+      }),
+    {
+      retries: 3,
+      onFailedAttempt: async (error) => {
+        const delay = error.attemptNumber * 500;
+        debug(`failed to get logs, retrying in ${delay}ms...`);
+        await wait(delay);
```
holic, line 35: no need for this here, I believe `pRetry` has its own back off internally
(I suspect this was copied from other spots where we also don't need to do this)

packages/cli/src/deploy/getResourceIds.ts
```diff
@@ -15,15 +17,25 @@ export async function getResourceIds({
...
+        strict: true,
+        address: worldDeploy.address,
+        fromBlock: worldDeploy.deployBlock,
+        toBlock: worldDeploy.stateBlock,
+        event: parseAbiItem(storeSpliceStaticDataEvent),
+        args: { tableId: storeTables.store_ResourceIds.tableId },
+      }),
+    {
+      retries: 3,
+      onFailedAttempt: async (error) => {
```
holic, line 32: we could check the error to see if it contains "block is out of range" and only retry those? but I am happy just retrying all here since its a relatively safe operation

packages/cli/src/deploy/getResourceIds.ts
yonadaaa: Oh, yes just copied other examples :D

packages/cli/src/deploy/getResourceIds.ts
yonadaaa: Done, I also like that it makes the retry logic more self-explanatory

packages/cli/src/deploy/getResourceIds.ts
```diff
@@ -15,15 +16,30 @@ export async function getResourceIds({
...
+        const shouldRetry =
+          error instanceof HttpRequestError &&
+          error.status === 400 &&
+          error.message.includes('"block is out of range"');
```
holic, line 35: do we need the quotes here?

.changeset/five-pots-switch.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/cli": patch
+---
+
+Fixed `""block is out of range"` errors in the deployer by adding retry logic to the `getLogs` call in the `getResourceIds` function. Previously, fetching the logs could fail if the RPC was out of sync.
```
holic, line 5:
```suggestion
Deploying now retries on "block is out of range" errors, for cases where the RPC is load balanced and out of sync.
```

packages/cli/src/deploy/getResourceIds.ts
holic, line 35: do we need the quotes here?
```suggestion
          error.message.includes('block is out of range');
```

holic (approved):

packages/cli/src/deploy/getResourceIds.ts
yonadaaa: oh nope! this seems less brittle also

PR #2718, opened 2024-04-24
refactor(cli): getWorldContracts

yonadaaa:
In https://github.com/latticexyz/mud/pull/2662 (contract verification) it is useful to have a list of the names and bytecode for the standard contracts deployed by MUD (`InitModule`, `World` etc.)

As pointed out in https://github.com/latticexyz/mud/pull/2662#discussion_r1577697220 we can split this into a seperate PR and rework the approach too.

### DIFF
--- a/packages/cli/src/deploy/ensureWorldFactory.ts
+++ b/packages/cli/src/deploy/ensureWorldFactory.ts
@@ -8,16 +8,27 @@ export async function ensureWorldFactory(
   deployerAddress: Hex,
   withWorldProxy?: boolean,
 ): Promise<Address> {
-  const worldFactoryContracts = withWorldProxy
-    ? getWorldProxyFactoryContracts(deployerAddress)
-    : getWorldFactoryContracts(deployerAddress);
+  if (!withWorldProxy) {
+    const contracts = getWorldFactoryContracts(deployerAddress);
 
-  // WorldFactory constructor doesn't call InitModule, only sets its address, so we can do these in parallel since the address is deterministic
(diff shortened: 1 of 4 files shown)


packages/cli/src/deploy/ensureWorldFactory.ts
```diff
@@ -1,111 +1,34 @@
...
+
+    // WorldFactory constructor doesn't call InitModule, only sets its address, so we can do these in parallel since the address is deterministic
+    await ensureContractsDeployed({
+      client,
+      deployerAddress,
+      contracts: Object.values(contracts),
+    });
+
+    return contracts.WorldProxyFactory.address;
+  }
```
holic, line 33: I like to return early in these cases

```suggestion
  if (withWorldProxy) {
    const contracts = getWorldProxyFactoryContracts(deployerAddress);
    // WorldFactory constructor doesn't call InitModule, only sets its address, so we can do these in parallel since the address is deterministic
    await ensureContractsDeployed({
      client,
      deployerAddress,
      contracts: Object.values(contracts),
    });
    return contracts.WorldProxyFactory.address;
  }
  
  const contracts = getWorldFactoryContracts(deployerAddress);
  // WorldFactory constructor doesn't call InitModule, only sets its address, so we can do these in parallel since the address is deterministic
  await ensureContractsDeployed({
    client,
    deployerAddress,
    contracts: Object.values(contracts),
  });
  return contracts.WorldFactory.address;
```

holic (approved):

PR #2721, opened 2024-04-24
fix(common): kms correctly serializes transactions

yonadaaa:
While testing the KMS deployer in https://github.com/latticexyz/mud/pull/2704 I realised that it's transactions were rejected by the RPC because they were not serialised properly. This PR fixes the KMS account by mirroring the [Viem `signTransaction` logic](https://github.com/wevm/viem/blob/main/src/accounts/utils/signTransaction.ts#L53-L71).

### DIFF
--- a/packages/common/src/account/kms/signWithKms.ts
+++ b/packages/common/src/account/kms/signWithKms.ts
@@ -1,4 +1,4 @@
-import { Hex, isAddressEqual, signatureToHex, toHex } from "viem";
+import { Hex, Signature, isAddressEqual, signatureToHex, toHex } from "viem";
 import { recoverAddress } from "viem/utils";
 import { KMSClient, SignCommandInput } from "@aws-sdk/client-kms";
 import { sign } from "./commands/sign";
@@ -65,7 +65,7 @@ type SignParameters = {
   address: Hex;
 };
 
-type SignReturnType = Hex;
+type SignReturnType = Signature;
 
 /**
  * @description Signs a hash with a given KMS key.
@@ -78,10 +78,10 @@ export async function signWithKms({ hash, address, keyId, client }: SignParamete
   const { r, s } = await getRS({ keyId, hash, client });
   const recovery = await getRecovery(hash, r, s, address);
 
-  return signatureToHex({
+  return {
     r,
     s,
     v: recovery ? 28n : 27n,
     yParity: recovery,
-  });
+  };
 }
--- a/packages/common/src/createKmsAccount.ts
+++ b/packages/common/src/createKmsAccount.ts
@@ -1,5 +1,5 @@
 import { KMSClient } from "@aws-sdk/client-kms";
-import { LocalAccount, hashMessage, hashTypedData, keccak256, serializeTransaction } from "viem";
+import { LocalAccount, hashMessage, hashTypedData, keccak256, serializeTransaction, signatureToHex } from "viem";
 import { toAccount } from "viem/accounts";
 import { signWithKms } from "./account/kms/signWithKms";
 import { getAddressFromKMS } from "./account/kms/getAddressFromKms";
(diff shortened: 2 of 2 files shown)


holic (dismissed):

.changeset/slimy-emus-remember.md
holic, line 1: good, I don't like slimy emus

packages/common/src/account/kms/createKmsAccount.ts
```diff
@@ -25,38 +25,46 @@ export async function createKmsAccount({
...
+      const signableTransaction = (() => {
+        // For EIP-4844 Transactions, we want to sign the transaction payload body (tx_payload_body) without the sidecars (ie. without the network wrapper).
+        // See: https://github.com/ethereum/EIPs/blob/e00f4daa66bd56e2dbd5f1d36d09fd613811a48b/EIPS/eip-4844.md#networking
+        if (transaction.type === "eip4844")
+          return {
+            ...transaction,
+            sidecars: false,
+          };
+        return transaction;
+      })();
```
holic, line 48: ooc why this and not
```suggestion
      // For EIP-4844 Transactions, we want to sign the transaction payload body (tx_payload_body) without the sidecars (ie. without the network wrapper).
      // See: https://github.com/ethereum/EIPs/blob/e00f4daa66bd56e2dbd5f1d36d09fd613811a48b/EIPS/eip-4844.md#networking
      const signableTransaction = transaction.type === "eip4844" ? { ...transaction, sidecars: false } : transaction;
```

packages/common/src/account/kms/createKmsAccount.ts
```diff
@@ -25,38 +25,46 @@ export async function createKmsAccount({
...
       const signature = await signWithKms({
         client,
         keyId,
-        hash,
+        hash: keccak256(serializer(signableTransaction)),
         address,
       });
 
-      return signature;
+      return serializer(transaction, signature);
```
holic, line 57: just to confirm, we _don't_ want to use `signableTransaction` here?

packages/common/src/account/kms/createKmsAccount.ts
yonadaaa: Copied verbatim from the Viem function! It could be more encapsulated imo: https://github.com/wevm/viem/blob/main/src/accounts/utils/signTransaction.ts#L53-L62

packages/common/src/account/kms/createKmsAccount.ts
yonadaaa: Nope, again this is [from viem](https://github.com/wevm/viem/blob/main/src/accounts/utils/signTransaction.ts#L68-L71). I think it's because the data you sign should (or shouldn't idk) include some data that is in the full transaction data.

packages/common/src/account/kms/createKmsAccount.ts
holic: ahh got it!

maybe worth a note above about keeping this logic aligned with viem internals?

packages/common/src/account/kms/kmsKeyToAddress.ts
```diff
@@ -0,0 +1,77 @@
+import { KMSClient } from "@aws-sdk/client-kms";
+import { LocalAccount, hashMessage, hashTypedData, keccak256, serializeTransaction, signatureToHex } from "viem";
+import { toAccount } from "viem/accounts";
+import { signWithKms } from "./signWithKms";
+import { getAddressFromKms } from "./getAddressFromKms";
+
+export type kmsKeyToAccountOptions = {
```
holic, line 7:
```suggestion
export type KmsKeyToAccountOptions = {
```

packages/common/src/account/kms/kmsKeyToAddress.ts
```diff
@@ -0,0 +1,77 @@
...
+  keyId: string;
+  client?: KMSClient;
+};
+
+export type KMSAccount = LocalAccount<"aws-kms">;
```
holic, line 12:
```suggestion
export type KmsAccount = LocalAccount<"aws-kms">;
```

packages/common/src/account/kms/kmsKeyToAddress.test.ts
```diff
@@ -102,4 +107,35 @@ describe("createKmsAccount", () => {
...
+
+    // Fund the KMS account from the admin
+    let balance = await getBalance(adminClient, { address: account.address });
+    expect(balance).toEqual(0n);
+
+    let tx = await adminClient.sendTransaction({ to: account.address, value: parseEther("1") });
+    await waitForTransaction(tx);
+
+    balance = await getBalance(adminClient, { address: account.address });
+    expect(balance).toEqual(1000000000000000000n);
```
holic, line 132: we're using a mix of tree shakable and not-tree shakable actions, would be nice to consolidate (ideally on tree shakable)

and if we're doing all tree shakable, might make sense to just use `createClient` not `createWalletClient`

packages/common/src/account/kms/kmsKeyToAddress.test.ts
```diff
@@ -102,4 +107,35 @@ describe("createKmsAccount", () => {
 
     expect(valid).toBeTruthy();
   });
+
+  it("can execute transactions", async () => {
+    const adminClient = createWalletClient({
+      chain: foundry,
+      transport: http(anvilRpcUrl),
+      account: privateKeyToAccount("0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80"),
```
holic, line 115: wanna make a note here about this being a default anvil user?

can also use `testClient` to set balance of ~any account

yonadaaa: `testClient` would be more intuitive, the funding stuff is a means to an end

holic (approved):

PR #2725, opened 2024-04-25
fix(common): use feeRef for sendTransaction calls

Kooshaba:
- if we don't use the fee ref here, we don't get the "site recommended" fee to show up in Metamask
- also doesn't seem to hurt, just makes submission a bit faster
- i believe this is what was causing nonce gap issues during the last playtest. players could improperly set their fee and get a tx stuck in the mempool indefinitely


alvrs:
Followup: #2728

Issue #2710, opened 2024-04-23
registerTable should check for existence of table ID for both table types


holic:
basically I want a unique constraint on (namespace, name) not (table type, namespace, name)

otherwise it's possible to register both an offchain and onchain table with the same namespace+name, making indexing with decoded tables more annoying (because type now needs to be part of the table name, or we drop one of the two tables)

yonadaaa:
This is on the Store level, so it's more a constraint on `type`, and a bytes30 `name`.

PR #2736, opened 2024-04-25
fix(store): enforce unique table names across types

yonadaaa:
fixes https://github.com/latticexyz/mud/issues/2710


packages/store/src/StoreCore.sol
```diff
@@ -213,12 +217,17 @@ library StoreCore {
     if (ResourceIds._getExists(tableId)) {
       revert IStoreErrors.Store_TableAlreadyExists(tableId, string(abi.encodePacked(tableId)));
     }
```
holic, line 219: Rather than adding a new table, I think we can just extend this check to look for `ResourceId` with the same name but different type.

yonadaaa: How else would you find a `ResourceId` with the same name? There is no global array of all Resource IDs we can traverse

yonadaaa: Like, just given a `bytes30` name, how do you find any/all corresponding Resource ID's unless there is a table that specifically indexes that

yonadaaa: I see, we can change the type of Table ID to check if there is either an offchain or regular table instead

packages/store/src/StoreCore.sol
```diff
@@ -214,6 +214,21 @@ library StoreCore {
...
+      if (ResourceIds._getExists(tableIdOffchain)) {
+        revert IStoreErrors.Store_TableAlreadyExists(tableId, string(abi.encodePacked(tableId)));
+      }
+    } else if (tableId.getType() == RESOURCE_OFFCHAIN_TABLE) {
+      ResourceId tableIdOnchain = ResourceIdLib.encode(RESOURCE_TABLE, tableId.getNameStore());
+
+      if (ResourceIds._getExists(tableIdOnchain)) {
+        revert IStoreErrors.Store_TableAlreadyExists(tableId, string(abi.encodePacked(tableId)));
+      }
+    }
```
holic, line 230: how much worse on gas would this be if we did something like

```solidity
{
  ResourceId onchainTableId = ResourceIdLib.encode(RESOURCE_TABLE, tableId.getNameStore());
  ResourceId offchainTableId = ResourceIdLib.encode(RESOURCE_OFFCHAIN_TABLE, tableId.getNameStore());
  if (ResourceIds._getExists(onchainTableId) || ResourceIds._getExists(offchainTableId)) {
    revert IStoreErrors.Store_TableAlreadyExists(tableId, string(abi.encodePacked(tableId)));
  }
}
```

packages/store/src/ResourceId.sol
```diff
@@ -45,4 +45,13 @@ library ResourceIdInstance {
   function getType(ResourceId resourceId) internal pure returns (bytes2) {
     return bytes2(ResourceId.unwrap(resourceId));
   }
+
+  /**
+   * @notice Get the name from a resource ID.
+   * @param resourceId The resource ID.
+   * @return A 30-byte name.
+   */
+  function getNameStore(ResourceId resourceId) internal pure returns (bytes30) {
```
holic, line 54: what about `getResourceName`?

packages/store/src/StoreCore.sol
yonadaaa: It's actually cheaper! I guess because we do two `_getExists` checks regardless.

Much clearer too.

packages/store/src/ResourceId.sol
yonadaaa: That's kinda better but the World name is also a "resource name" 🤔

.changeset/strong-lobsters-shake.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/store": patch
+---
+
+Added a check to `registerTable` that prevents registering both an offchain and onchain table with the same namespace and name, making indexing with decoded tables easier.
```
holic, line 5:
```suggestion
Added a check to `registerTable` that prevents registering both an offchain and onchain table with the same name, making indexing with decoded tables easier.
```

.changeset/strong-lobsters-shake.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/store": patch
```
holic, line 2: need to bump store/world protocol versions

yonadaaa: done

holic (dismissed):
looks like need to regenerate docs but otherwise lgtm!

holic:
waiting for @alvrs sign off too, since this is a protocol bump

alvrs (approved):

holic (approved):

PR #2738, opened 2024-04-25
feat(common): add garnet chain

holic:


packages/common/src/chains/garnet.ts
```diff
@@ -0,0 +1,32 @@
...
+      name: "Blockscout",
+      url: "https://explorer.garnetchain.com",
+    },
+  },
+  contracts: {
+    portal: {
+      [sourceId]: {
+        address: "0x57ee40586fbE286AfC75E67cb69511A6D9aF5909",
+      },
+    },
```
holic, line 30: do we have more contract addresses to expose?

[base](https://github.com/wevm/viem/blob/fdb6caabb2cf8c31144c7e427f38b9f0bd33e69f/src/chains/definitions/base.ts#L23-L46) and [base sepolia](https://github.com/wevm/viem/blob/fdb6caabb2cf8c31144c7e427f38b9f0bd33e69f/src/chains/definitions/baseSepolia.ts#L24-L47) for example have a bunch

ludns: We can expose all the ones on Base too.

yonadaaa: What about the [MUD deployer](https://github.com/latticexyz/mud/blob/main/packages/cli/src/deploy/create2/deployment.json#L6)?

packages/common/src/chains/latticeTestnet.ts
holic, line 1: we could alternatively just add `@deprecated` to this and wait for the chain to go offline before we remove the config

this is technically "breaking" in terms of semver but easy to workaround since its a static config and easy to copy+paste this back into your project if you need it, so I am fine breaking this within a patch change

holic: decided to restore this config for now since 1) templates depend on it as a testnet to deploy to for free and 2) kicks the can on the breaking change

it's now marked as `@deprecated` and we can remove in a future version

packages/common/src/chains/garnet.ts
holic: > What about the [MUD deployer](https://github.com/latticexyz/mud/blob/main/packages/cli/src/deploy/create2/deployment.json#L6)?

since CREATE2 deployment itself is deterministic, I think folks just hardcode the address

there are no viem chains I could find with this address:
https://github.com/search?q=repo%3Awevm%2Fviem%200x4e59b44847b379578588920ca78fbf26c0b4956c&type=code
https://github.com/search?q=repo%3Awevm%2Fviem+create2&type=code

alvrs: should we add the OP stack contracts: https://github.com/wevm/viem/blob/fdb6caabb2cf8c31144c7e427f38b9f0bd33e69f/src/op-stack/contracts.ts (base has them too)

holic: we extend from this chain object

PR #2741, opened 2024-04-25
fix(store-indexer): transform env var to undefined

alvrs:
Allow an empty `STORE_ADDRESS=` env variable to be passed in to the store indexer


holic (approved):
lgtm
- should test that this actually works
- could split the changeset into its own thing since its a distinct change from a thing that already existed, but I don't feel that strongly

PR #2747, opened 2024-04-26
fix(store-indexer): refine parsed type

alvrs:
The fix in #2746 changed the output type of `STORE_ADDRESS` from `Hex | undefined` to `string | undefined`. This changes it back to `Hex | undefined`.


alvrs:
shamefully merging without approval to unblock the indexer deployments

PR #2748, opened 2024-04-26
fix(store,world,world-modules): publish `solidity-files-cache.json` on npm

yonadaaa:
Publishing this file with NPM enables `forge verify-contract` to find MUD contract files when they are installed as a dependency.

### DIFF
--- a/packages/store/.npmignore
+++ b/packages/store/.npmignore
@@ -4,6 +4,7 @@
 !out/**/*.abi.json
 !out/**/*.abi.json.d.ts
 !cache/solidity-files-cache.json
+!foundry.toml
 !src/**
 !ts/**
 !mud.config.ts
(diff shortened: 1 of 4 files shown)


packages/store/.npmignore
```diff
@@ -9,3 +9,4 @@
 !package.json
 !README.md
 !dist/**
+!cache/solidity-files-cache.json
```
holic, line 12: small thing but can we pop this up near `out` so all the forge ones are together and maybe leave a comment about why this is here?

holic (dismissed):
very small nit just so we don't forget why these things are here but otherwise lgtm!

also should add a changeset so this triggers a release

yonadaaa:
closing in favour of https://github.com/latticexyz/mud/pull/2750

PR #2749, opened 2024-04-26
feat(common): add redstone chain config

holic:
continued from #2738

### DIFF
--- a/packages/common/src/chains/redstone.ts
+++ b/packages/common/src/chains/redstone.ts
@@ -18,7 +18,7 @@ export const redstone = {
   blockExplorers: {
     default: {
       name: "Blockscout",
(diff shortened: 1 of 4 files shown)


packages/common/src/chains/redstone.ts
```diff
@@ -0,0 +1,33 @@
...
+  rpcUrls: {
+    default: {
+      http: ["https://rpc.redstonechain.com"],
+      webSocket: ["wss://rpc.redstonechain.com"],
+    },
+  },
+  blockExplorers: {
+    default: {
+      name: "Blockscout",
+      url: "https://explorer.redstonechain.com",
```
ludns, line 21: we'll do explorer.redstone.xyz

PR #2750, opened 2024-04-26
fix(cli): fix verify with sourcify for dependencies

yonadaaa:
`mud verify` attempts to verify external contracts (`InitModule`, `World`, etc.) by changing the working directory to the relevant dependency (eg. `cd node_module@latticexyz/world`) and running `forge verify-contract`. 

However these dependencies on NPM are incomplete, causing issues when attempting to compile the source files:

1) Without `solidity-files-cache.json`, Forge cannot locate the contract by name alone (eg. `World`), and requires the full path. We fix this by publishing the cache on NPM.
2) The sub-dependencies of each dependency are not installed, so for example, if `world.sol` imports `schema-type`, `world.sol` cannot be compiled. We fix this by running `npm install` in each dependency during verification.

### DIFF
--- a/packages/cli/src/verify.ts
+++ b/packages/cli/src/verify.ts
@@ -74,11 +74,17 @@ export async function verify({
     ),
   );
 
-  {
-    const cwd = "node_modules/@latticexyz/world";
-
+  // If the verifier is Sourcify, attempt to verify MUD core contracts
+  if (verifier === "sourcify") {
+    // Install subdependencies so contracts can compile
     await execa("npm", ["install"], {
-      cwd,
+      cwd: "node_modules/@latticexyz/store",
+    });
+    await execa("npm", ["install"], {
+      cwd: "node_modules/@latticexyz/world",
+    });
+    await execa("npm", ["install"], {
+      cwd: "node_modules/@latticexyz/world-modules",
     });
 
     Object.entries(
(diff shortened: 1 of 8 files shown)


packages/cli/src/verify.ts
```diff
@@ -73,94 +74,114 @@ export async function verify({
...
-          cwd: "node_modules/@latticexyz/world-modules",
-        },
-      ).catch((error) => {
-        console.error(`Error verifying module contract ${name}:`, error);
-      }),
-    ),
-  );
+    await execa("npm", ["install"], {
+      cwd,
+    });
```
holic, line 82: what about structuring this file like:

```
if (verifier === 'sourcify') {
  // npm install world
  // npm install world-modules
  // verify all the deps
}
```

packages/cli/src/verify.ts
```diff
@@ -73,94 +74,114 @@ export async function verify({
...
+            address: getCreate2Address({
+              from: deployerAddress,
+              bytecode: bytecode,
+              salt,
+            }),
+          },
+          {
+            profile: foundryProfile,
+            cwd,
+          },
```
holic, line 103: thinking foundry profile makes less sense here because users can't control it (i.e. our deps may not have a "holesky" profile that you have locally) and now wondering if we should even offer the foundry profile option and instead just require the CLI flags for the things we need from the foundry config

packages/cli/src/verify.ts
```diff
@@ -54,31 +53,42 @@ export async function verify({
...
-  Object.entries(
-    usesProxy ? getWorldProxyFactoryContracts(deployerAddress) : getWorldFactoryContracts(deployerAddress),
-  ).map(([name, { bytecode }]) =>
-    verifyQueue.add(() =>
-      verifyContract(
-        {
+  // If the verifier is Sourcify, attempt to verify MUD core contracts
+  // There are path issues with verifying Blockscout and Etherscan
+  if (verifier === "sourcify") {
+    // Install subdependencies so contracts can compile
```
holic, line 75: it's less about compiling and more about the dependent files existing right?

holic (dismissed):

packages/cli/src/verify.ts
yonadaaa: well, the files have to exist in node_modules to be compiled

holic (approved):

PR #2757, opened 2024-04-28
refactor(world-modules): remove unnecessary abi.encodePacked

holic:


yonadaaa (commented):
Surprised this works! If it decreases gas, it must change the behaviour but somehow the output is the same. Cool 🤷

holic:
I'm not convinced we should do this if this deviates from the spec, but noticed this when trying to debug some issues I thought were related to signing

PR #2760, opened 2024-04-29
refactor(common,cli): kms deployer gets keyId from environment

yonadaaa:
We can declare the KMS Key ID `AWS_KMS_KEY_ID` instead of passing it into to `mud deploy`. This means the flag to enable KMS signing can just be a boolean.

### DIFF
--- a/packages/cli/src/runDeploy.ts
+++ b/packages/cli/src/runDeploy.ts
@@ -43,9 +43,9 @@ export const deployOptions = {
     type: "string",
     desc: "The deployment salt to use. Defaults to a random salt.",
   },
-  awsKmsKeyId: {
-    type: "string",
-    desc: "Optional AWS KMS key ID. If set, the World is deployed using a KMS signer instead of local private key.",
+  aws: {
+    type: "boolean",
+    desc: "Deploy the World wtih a KMS signer instead of local private key.",
   },
 } as const satisfies Record<string, Options>;
 
@@ -95,8 +95,7 @@ in your contracts directory to use the default anvil private key.`,
 
   const resolvedConfig = resolveConfig({ config, forgeSourceDir: srcDir, forgeOutDir: outDir });
 
-  const keyId = opts.awsKmsKeyId ?? process.env.AWS_KMS_KEY_ID;
-  const account = keyId ? await kmsKeyToAccount({ keyId }) : privateKeyToAccount(privateKey);
+  const account = opts.aws ? await kmsKeyToAccount() : privateKeyToAccount(privateKey);
 
   const client = createWalletClient({
     transport: http(rpc, {
--- a/packages/common/src/account/kms/kmsKeyToAccount.ts
+++ b/packages/common/src/account/kms/kmsKeyToAccount.ts
@@ -1,11 +1,11 @@
-import { KMSClient } from "@aws-sdk/client-kms";
+import { KMSClient, SignCommandInput } from "@aws-sdk/client-kms";
 import { LocalAccount, hashMessage, hashTypedData, keccak256, serializeTransaction, signatureToHex } from "viem";
 import { toAccount } from "viem/accounts";
 import { signWithKms } from "./signWithKms";
 import { getAddressFromKms } from "./getAddressFromKms";
 
 export type KmsKeyToAccountOptions = {
-  keyId: string;
+  keyId?: SignCommandInput["KeyId"];
   client?: KMSClient;
 };
 


packages/cli/src/runDeploy.ts
```diff
@@ -43,9 +43,9 @@ export const deployOptions = {
     type: "string",
     desc: "The deployment salt to use. Defaults to a random salt.",
   },
-  awsKmsKeyId: {
-    type: "string",
-    desc: "Optional AWS KMS key ID. If set, the World is deployed using a KMS signer instead of local private key.",
+  aws: {
```
holic, line 46: since we call everything "kms" I wonder if this flag should be `--kms` or `--awsKms`

packages/cli/src/runDeploy.ts
```diff
@@ -43,9 +43,9 @@ export const deployOptions = {
...
+    type: "boolean",
+    desc: "Deploy the World with a KMS signer instead of local private key.",
```
holic, line 48:
```suggestion
    desc: "Deploy the World with an AWS KMS signer instead of local private key.",
```

packages/cli/src/runDeploy.ts
```diff
@@ -95,8 +95,7 @@ in your contracts directory to use the default anvil private key.`,
 
   const resolvedConfig = resolveConfig({ config, forgeSourceDir: srcDir, forgeOutDir: outDir });
 
-  const keyId = opts.awsKmsKeyId ?? process.env.AWS_KMS_KEY_ID;
-  const account = keyId ? await kmsKeyToAccount({ keyId }) : privateKeyToAccount(privateKey);
+  const account = opts.aws ? await kmsKeyToAccount({}) : privateKeyToAccount(privateKey);
```
holic, line 98: we could also default the options object to `{}` so we don't need to pass an empty object
```suggestion
  const account = opts.aws ? await kmsKeyToAccount() : privateKeyToAccount(privateKey);
```

packages/cli/src/runDeploy.ts
yonadaaa: agreed, I also like the specificity of KMS

packages/common/src/account/kms/kmsKeyToAccount.ts
```diff
@@ -19,9 +19,9 @@ export type KmsAccount = LocalAccount<"aws-kms"> & {
  * @returns A Local Account.
  */
 export async function kmsKeyToAccount({
-  keyId,
+  keyId = process.env.AWS_KMS_KEY_ID || "",
```
yonadaaa, line 22: Defaulting in the account was inspired by this Foundry library https://github.com/foundry-rs/foundry/blob/master/crates/wallets/src/wallet.rs#L95-L97

yonadaaa: I did this because `keyId` is optional for the SDK but our `getKeyId` function returns a `string`. Gonna try something else

packages/common/src/account/kms/kmsKeyToAccount.ts
```diff
@@ -19,11 +20,15 @@ export type KmsAccount = LocalAccount<"aws-kms"> & {
  * @returns A Local Account.
  */
 export async function kmsKeyToAccount({
-  keyId,
+  keyId = process.env.AWS_KMS_KEY_ID,
```
holic, line 23: I don't think this will work with how bundlers work - they generally use env vars as string replacement during builds so this may end up as an empty string before it even gets to user code.

packages/common/src/account/kms/kmsKeyToAccount.ts
```diff
@@ -19,11 +20,15 @@ export type KmsAccount = LocalAccount<"aws-kms"> & {
...
   client = new KMSClient(),
-}: KmsKeyToAccountOptions): Promise<KmsAccount> {
+}: KmsKeyToAccountOptions = {}): Promise<KmsAccount> {
   const address = await getAddressFromKms({ keyId, client });
 
+  if (!keyId) {
+    throw new MUDError("Missing AWS KMS Key ID");
+  }
```
holic, line 30: I think this check and error should be done inside CLI not here, and leave `keyId` as a required prop

packages/cli/src/runDeploy.ts
```diff
@@ -87,9 +87,8 @@ export async function runDeploy(opts: DeployOptions): Promise<WorldDeploy> {
   const resolvedConfig = resolveConfig({ config, forgeSourceDir: srcDir, forgeOutDir: outDir });
 
   const account = await (async () => {
-    if (opts.awsKmsKeyId) {
-      const keyId = opts.awsKmsKeyId ?? process.env.AWS_KMS_KEY_ID;
-      return await kmsKeyToAccount({ keyId });
+    if (opts.kms) {
+      return await kmsKeyToAccount();
```
holic, line 91: we should do the check here like we're doing for PRIVATE_KEY

holic (dismissed):
just a couple small things around messaging but otherwise lgtm!

packages/common/src/account/kms/getAddressFromKms.ts
```diff
@@ -25,7 +25,7 @@ export async function getAddressFromKms({
   keyId,
   client,
 }: {
-  keyId: SignCommandInput["KeyId"];
+  keyId: GetPublicKeyCommandInput["KeyId"];
```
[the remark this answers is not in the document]
yonadaaa: It's `string | undefined` (same as `SignCommandInput` actually). I don't know why they make it optional! https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-aws-sdk-client-kms/Interface/GetPublicKeyCommandInput/

.changeset/quick-lions-rest.md
```diff
@@ -0,0 +1,6 @@
+---
+"@latticexyz/common": patch
```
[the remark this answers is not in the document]
yonadaaa: leftover from a previous approach

packages/common/src/account/kms/getAddressFromKms.ts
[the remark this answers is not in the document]
holic: oh, should we just make it a regular string and require it?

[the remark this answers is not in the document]
yonadaaa: We can, i figured we can pipe it through the same way they do. I copied that approach from the ethers signer https://github.com/rumblefishdev/eth-signer-kms/blob/master/src/kms.ts#L16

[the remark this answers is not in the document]
holic: I thought we had it as `string` before, not sure when it changed

holic (approved):

PR #2762, opened 2024-04-29
refactor(cli): run postdeploy with private_key flag

yonadaaa:
We now pass the deployer private key to the PostDeploy script with `forge script --private-key`.

Loading the deployer `PRIVATE_KEY` directly in the script is inflexible. For example, it breaks with AWS KMS because there is no local private key in that setup.


yonadaaa:
Closing in favour of https://github.com/latticexyz/mud/compare/yonadaaa/kms-aws-flag?expand=1

PR #2763, opened 2024-04-29
feat(store-indexer): add metric for distance from block tag to follow

alvrs:


packages/store-indexer/bin/sqlite-indexer.ts
```diff
@@ -73,6 +73,12 @@ async function getLatestStoredBlockNumber(): Promise<bigint | undefined> {
   return currentChainState?.lastUpdatedBlockNumber ?? undefined;
 }
 
+async function getDistanceFromFollowBlock(): Promise<bigint | undefined> {
+  const latestStoredBlockNumber = (await getLatestStoredBlockNumber()) ?? -1n;
+  const latestFollowBlockNumber = (await publicClient.getBlock({ blockTag: env.FOLLOW_BLOCK_TAG })).number;
```
holic, line 78: should we parallelize these to maybe have a better chance of these aligning?

packages/store-indexer/bin/sqlite-indexer.ts
```diff
@@ -73,6 +73,12 @@ async function getLatestStoredBlockNumber(): Promise<bigint | undefined> {
   return currentChainState?.lastUpdatedBlockNumber ?? undefined;
 }
 
+async function getDistanceFromFollowBlock(): Promise<bigint | undefined> {
```
holic, line 76: looks like we're always returning a bigint now - do we need `undefined` here?

packages/store-indexer/src/koa-middleware/metrics.ts
```diff
@@ -61,6 +63,16 @@ export function metrics({
     blockTagGauge.set(blockTagToValue[followBlockTag]);
   }
 
+  if (getDistanceFromFollowBlock != null) {
+    new promClient.Gauge({
+      name: "distance_from_follow_block",
+      help: "Block distance from the block tag this the indexer is following",
+      async collect(): Promise<void> {
+        this.set(Number(await getDistanceFromFollowBlock()));
```
holic, line 71: does `this.set` only support number? vs e.g. `BigInt.toString()`

alvrs: I think number is better here so we can do things like `>` for the alerts (and very unlikely the numbers ever get into a range where they don't fit into a number anymore)

holic: oh yes, I just assumed that what "gauge" metric type is for, not the value we pass into the gauge

packages/store-indexer/bin/postgres-indexer.ts
```diff
@@ -66,6 +66,14 @@ async function getLatestStoredBlockNumber(): Promise<bigint | undefined> {
   }
 }
 
+async function getDistanceFromFollowBlock(): Promise<bigint> {
+  const [latestStoredBlockNumber, latestFollowBlockNumber] = await Promise.all([
+    (await getLatestStoredBlockNumber()) ?? -1n,
+    (await publicClient.getBlock({ blockTag: env.FOLLOW_BLOCK_TAG })).number,
+  ]);
```
holic, line 73: this doesn't do anything because the `await` is in line with array, so you're calling `Promise.all` on already-awaited values
```suggestion
  const [latestStoredBlockNumber, latestFollowBlockNumber] = await Promise.all([
    getLatestStoredBlockNumber().then((blockNumber) => blockNumber ?? -1n),
    publicClient.getBlock({ blockTag: env.FOLLOW_BLOCK_TAG })).then((block) => block.number),
  ]);
```

packages/store-indexer/bin/postgres-indexer.ts
```diff
@@ -66,6 +66,14 @@ async function getLatestStoredBlockNumber(): Promise<bigint | undefined> {
...
+  return latestFollowBlockNumber - latestStoredBlockNumber;
```
holic, line 74: this doesn't do anything because the `await` is in line with array, so you're calling `Promise.all` on already-awaited values
```suggestion
  const [latestStoredBlockNumber, latestFollowBlockNumber] = await Promise.all([
    getLatestStoredBlockNumber().then((blockNumber) => blockNumber ?? -1n),
    publicClient.getBlock({ blockTag: env.FOLLOW_BLOCK_TAG })).then((block) => block.number),
  ]);
  return latestFollowBlockNumber - latestStoredBlockNumber;
```

holic: alternatively
```suggestion
  const [latestStoredBlockNumber, latestFollowBlock] = await Promise.all([
    getLatestStoredBlockNumber(),
    publicClient.getBlock({ blockTag: env.FOLLOW_BLOCK_TAG }),
  ]);
  return (latestFollowBlockNumber ?? -1n) - latestStoredBlock.number;
```

alvrs: haha damn i even remembered an article about this common mistake while writing this

holic (approved):

holic (approved):

PR #2765, opened 2024-04-29
fix(cli): do not require `PRIVATE_KEY` if using KMS

yonadaaa:


packages/cli/src/runDeploy.ts
```diff
@@ -85,7 +85,7 @@ export async function runDeploy(opts: DeployOptions): Promise<WorldDeploy> {
   }
 
   const privateKey = process.env.PRIVATE_KEY as Hex;
-  if (!privateKey) {
+  if (!opts.awsKmsKeyId && !privateKey) {
```
holic, line 88: can we move this PK block down under `resolvedConfig` so its a bit more localized with the KMS stuff?

and could potentially wrap it in a self invoking function like

```ts
const account = keyId ? await kmsKeyToAccount({ keyId }) : (() => {
  const privateKey = process.env.PRIVATE_KEY as Hex;
  if (!privateKey) {
    throw new MUDError(
      `Missing PRIVATE_KEY environment variable.
Run 'echo "PRIVATE_KEY=0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80" > .env'
in your contracts directory to use the default anvil private key.`,
    );
  }
  return privateKeyToAccount(privateKey);
})();
```

or maybe pull that out into something like `envPrivateKeyToAccount()`

yonadaaa: wrapped it all in a self-invoking function! I like that the two environment variables are only used there

holic: and one we might be able to ditch in https://github.com/latticexyz/mud/pull/2760 ?

holic (approved):

PR #2766, opened 2024-04-29
fix(cli): run postdeploy with aws flag when kms is enabled

yonadaaa:


.changeset/hip-chairs-fly.md
```diff
@@ -0,0 +1,14 @@
+---
+"@latticexyz/cli": patch
+---
+
+Fixed `mud deploy` to use the `forge script --aws` flag when executing `PostDeploy` with a KMS signer.
+
+Note that you may need to update your `PostDeploy` script, with `vm.startBroadcast` receiving no arguments instead of reading a private key from the environment:
```
holic, line 7:
```suggestion
Note that you may need to update your `PostDeploy.s.sol` script, with `vm.startBroadcast` receiving no arguments instead of reading a private key from the environment:
```

holic (dismissed):

PR #2768, opened 2024-04-30
ci: add timeout to e2e tests

### LINKED ISSUE #2761: add CI timeouts
just spotted e2e tests on main running for >2h, which sounds wrong

should add a timeout so this doesn't eat resources and block other tests from running

roaminro:
This PR add a timeout of 25 minutes to the e2e tests.
After looking at the history of the e2e tests runs, I saw that it usually takes 8-12 minutes to complete these tests so I arbitrarly chose 25 minutes which is approximately 2x of the max runtime observed.

Let me know if you feel that 25 minutes is still too high.

Closes #2761

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -14,6 +14,7 @@ jobs:
   main:
     name: Run end-to-end tests
     runs-on: depot-ubuntu-22.04-16
+    timeout-minutes: 25
     services:
       postgres:
         image: postgres:12.1-alpine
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e.yml
```diff
@@ -14,6 +14,7 @@ jobs:
   main:
     name: Run end-to-end tests
     runs-on: depot-ubuntu-22.04-16
+    timeout-minutes: 25
```
holic, line 17: can we move this down to the "Run tests" step? that's where it was stalling before:
image

maybe also worth lowering this to ~15 mins as it seems the average time on main is ~5 mins for this step

holic:
thank you! just a small suggestion to time out the exact thing that was stalling

roaminro:
Thanks for the review, just pushed the changes ;)

holic (approved):
thanks so much!

PR #2769, opened 2024-04-30
fix(cli): postDeploy uses correct kms flag

yonadaaa:
Crept in by not merging main into https://github.com/latticexyz/mud/pull/2766


holic (approved):
did e2e tests not catch this?

yonadaaa:
> did e2e tests not catch this?

They did, but I think I merged the conflicting PR's at the same time before tests kicked off.

PR #2771, opened 2024-04-30
feat(common): add indexer URL to chain configs

yonadaaa:


holic (approved):

e2e/packages/client-vanilla/src/mud/setupNetwork.ts
```diff
@@ -38,7 +38,7 @@ export async function setupNetwork() {
     address: networkConfig.worldAddress as Hex,
     publicClient,
     startBlock: BigInt(networkConfig.initialBlockNumber),
-    indexerUrl: networkConfig.indexerUrl ?? undefined,
+    indexerUrl: networkConfig.indexerUrl,
```
holic, line 41: I think this was intentionally here to make types happy because `params.get` can return null

e2e/packages/client-vanilla/src/mud/getNetworkConfig.ts
```diff
@@ -28,7 +28,7 @@ export async function getNetworkConfig() {
     faucetServiceUrl: params.get("faucet") ?? chain.faucetUrl,
     worldAddress,
     initialBlockNumber,
-    indexerUrl: params.get("indexerUrl"),
+    indexerUrl: params.get("indexerUrl") ?? chain.indexerUrl,
```
holic, line 31: since the chain is part of `createStoreSync` (possibly on the `publicClient`), we could check this in there and default it, rather than passing it through in every template

packages/store-sync/src/createStoreSync.ts
```diff
@@ -96,7 +97,7 @@ export async function createStoreSync<config extends StoreConfig = StoreConfig>(
       filters,
       initialState,
       initialBlockLogs,
-      indexerUrl,
+      indexerUrl: indexerUrl ?? (publicClient.chain as MUDChain).indexerUrl,
```
yonadaaa, line 100: Do we want to pass the chain more explicitly to avoid this cast?

holic: no, I don't want to have a potential deviation between publicClient's chain and a chain passed in

this cast is fine but you could be super defensive with something like
```
"indexerUrl" in publicClient.chain && typeof publicClient.chain.indexerUrl === 'string' ? publicClient.chain.indexerUrl : undefined
```

packages/common/src/chains/redstone.ts
```diff
@@ -21,6 +21,7 @@ export const redstone = {
       url: "https://explorer.redstone.xyz",
     },
   },
+  indexerUrl: "indexer.mud.redstonechain.com",
```
holic, line 24:
```suggestion
  indexerUrl: "https://indexer.mud.redstonechain.com",
```

packages/common/src/chains/garnet.ts
```diff
@@ -22,6 +22,7 @@ export const garnet = {
       url: "https://explorer.garnetchain.com",
     },
   },
+  indexerUrl: "indexer.mud.garnetchain.com",
```
holic, line 25:
```suggestion
  indexerUrl: "https://indexer.mud.garnetchain.com",
```

packages/store-sync/src/createStoreSync.ts
holic: we prob also still want some way to turn off the indexer if needed

maybe we also allow `indexerUrl: false` here?

.changeset/rotten-rules-switch.md
```diff
@@ -0,0 +1,7 @@
+---
+"@latticexyz/common": patch
+---
+
+Added an optional `indexerUrl` property to `MUDChain` which is used as the default indexer URL in `createStoreSync`.
+
+Also added indexer URL's to the Redstone and Garnet chain configs.
```
holic, line 7: can we add a changelog for store-sync that talks about `createStoreSync` defaulting to `chain.indexerUrl` (or `false` to intentionally unset/not default)

and then this one can just focus on adding `indexerUrl` to `MUDChain` and filling it in for redstone, garnet chains

holic (approved):

PR #2775, opened 2024-04-30
feat(cli): blockscout is default verifier

yonadaaa:


packages/cli/src/commands/verify.ts
```diff
@@ -25,7 +25,7 @@ const verifyOptions = {
     desc: "Enable batch processing of RPC requests in viem client (defaults to batch size of 100 and wait of 1s)",
   },
   srcDir: { type: "string", desc: "Source directory. Defaults to foundry src directory." },
-  verifier: { type: "string", desc: "The verifier to use. Default to sourcify", default: "sourcify" },
+  verifier: { type: "string", desc: "The verifier to use. Default to sourcify", default: "blockscout" },
```
holic, line 28:
```suggestion
  verifier: { type: "string", desc: "The verifier to use. Defaults to blockscout", default: "blockscout" },
```

holic (approved):
looked into this briefly and doing conditional verifying based on chain is a little more involved (because we have no chain reference, it's inferred via rpc), so doing the simple/easy thing to unblock

PR #2779, opened 2024-05-01
Feat(world): support batchCall in callFrom action

dhvanipa:
Support `batchCall`

### DIFF
--- a/packages/world/ts/actions/callFrom.ts
+++ b/packages/world/ts/actions/callFrom.ts
@@ -63,9 +63,30 @@ export function callFrom<TChain extends Chain, TAccount extends Account>(
         writeArgs.address !== params.worldAddress ||
         writeArgs.functionName === "call" ||
         writeArgs.functionName === "callFrom" ||
+        writeArgs.functionName === "batchCallFrom" ||
         writeArgs.functionName === "callWithSignature"
       ) {
         return getAction(client, writeContract, "writeContract")(writeArgs);
+      }
+
+      if (writeArgs.functionName === "batchCall") {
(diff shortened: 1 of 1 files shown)


packages/world/ts/actions/callFrom.ts
```diff
@@ -63,11 +63,32 @@ export function callFrom<TChain extends Chain, TAccount extends Account>(
...
       }
 
+      if (writeArgs.functionName === "batchCall") {
+        if (!writeArgs.args || !Array.isArray(writeArgs.args) || writeArgs.args.length === 0) {
+          throw new Error("batchCall args should be an array with at least one element");
+        }
+
+        const systemCallFromData = writeArgs.args[0].map((systemCallData: Hex[]) => {
+          return [params.delegatorAddress, ...systemCallData];
+        });
```
holic, line 79: I am not confident in the shape of this data. I think this works when it comes from our `encodeSystemCalls` but doesn't seem to exactly correspond to the shape of `WriteContractParameters<typeof IBaseWorldAbi, "batchCall">`

```ts
(property) args: readonly [readonly {
    systemId: `0x${string}`;
    callData: `0x${string}`;
}[]] | ([readonly {
    systemId: `0x${string}`;
    callData: `0x${string}`;
}[]] & readonly [readonly {
    systemId: `0x${string}`;
    callData: `0x${string}`;
}[]])
```

But if I inspect the broader `WriteContractParameters<typeof IBaseWorldAbi>` and do an if statement, the expected `args` type is a bit wider and includes the array of hex values.

Not really sure how best to approach this, but might mean we need more tests here.

holic:
closing in favor of https://github.com/latticexyz/mud/pull/2796

PR #2782, opened 2024-05-01
ci: separate publish workflow to ensure order

holic:
moves our npm publish step to a single workflow that calls out to the previous workflows, but with `needs` to ensure order of operations

noticed that we had both "snapshot" and "release" CI steps running simultaneously, where "snapshot" would sometimes run first (due to concurrency), which would publish a new npm version for a release but not properly tag+release on github

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,32 @@
+name: Publish to npm
+
+on:
+  push:
+    branches:
+      - "main"
+    paths-ignore:
+      - "docs/**"
+
+# Ensure only one workflow/job is publishing to npm at a time
+concurrency: publish-to-npm
+
+jobs:
+  release:
+    uses: "./.github/workflows/release.yml"
+    secrets: inherit
+
+  snapshot:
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -16,7 +12,6 @@ env:
 jobs:
   release:
     name: Release
-    if: github.repository == 'latticexyz/mud'
```
alvrs, line 19: let's add this to the new top level one

alvrs (approved):

.github/workflows/publish.yml
```diff
@@ -0,0 +1,36 @@
...
+
+jobs:
+  release:
+    if: github.repository == 'latticexyz/mud'
+    uses: "./.github/workflows/release.yml"
+    secrets: inherit
+
+  snapshot:
+    if: github.repository == 'latticexyz/mud'
+    needs: [release]
```
holic, line 21: not sure if the conditional is required here and below if `needs` is not met by earlier conditional, but added it here anyway

PR #2783, opened 2024-05-01
ci: refactor docker build

holic:
attempting to refactor the docker build step so that they run in order on the same worker, rather than separate workers all performing the same work (building a big docker image)

### DIFF
--- a/.github/actions/docker-build-push/action.yml
+++ b/.github/actions/docker-build-push/action.yml
@@ -1,0 +1,32 @@
+name: Docker build and push
+description: Common setup steps used by our workflows
+inputs:
+  image:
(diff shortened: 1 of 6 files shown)


alvrs (approved):

kylegalbraith:
@holic I was taking a look at your Depot usage and wanted to suggest that you have 5,000 accelerated Docker image build minutes with us. You could cut these Docker image builds over to a Depot project or two and get significantly faster Docker image builds to go with the GHA runners you are already using. Here are some docs on getting going with that, but feel free to ping @jacobwgillespie or me if we can help out: https://depot.dev/docs/integrations/github-actions

PR #2786, opened 2024-05-02
chore: upgrade to ejs 3.1.10

yonadaaa:


holic (dismissed):

packages/store/package.json
```diff
@@ -76,7 +76,7 @@
     "@types/mocha": "^9.1.1",
     "@types/node": "^18.15.11",
     "ds-test": "https://github.com/dapphub/ds-test.git#e282159d5170298eb2455a6c05280ab5a73a4ef0",
-    "ejs": "^3.1.8",
+    "ejs": "^3.1.10",
```
holic, line 79: I don't think this is used in store, world, and world-modules? wonder if we can safely remove

.changeset/heavy-toes-attend.md
```diff
@@ -0,0 +1,8 @@
+---
+"@latticexyz/world-modules": patch
+"@latticexyz/store": patch
+"@latticexyz/world": patch
+"@latticexyz/cli": patch
+---
+
+Upgraded the EJS package to 3.1.10, and removed it from `store`, `world`, and `cli`.
```
holic, line 8: this might read better in the changelog if we add the "upgrade" changeset to the corresponding packages and "removed" from the others

I try to think of changesets in the context of a single package and how it'll look as if I was reading just that package's changelog (aside from the unique way we roll them up to a single changelog)

yonadaaa: I like this! They are kind of unrelated changes too.

.changeset/dirty-ears-happen.md
```diff
@@ -0,0 +1,7 @@
+---
+"@latticexyz/world-modules": patch
+"@latticexyz/store": patch
+"@latticexyz/cli": patch
+---
+
+Removed the `ejs` package from `store`, `world-modules`, and `cli`.
```
holic, line 7:
```suggestion
Removed the unused `ejs` dependency.
```

.changeset/heavy-toes-attend.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/world": patch
+---
+
+Upgraded the `ejs` package to 3.1.10.
```
holic, line 5:
```suggestion
Upgraded the `ejs` dependency to 3.1.10.
```

holic (approved):

PR #2796, opened 2024-05-07
feat(world): support batchCall in callFrom action

### LINKED ISSUE #1728: make batchCall easier to work with in the client
there's several places in #1611 that I avoided wiring up `batchCall` because it was trickier to use

it would be nice if we had a typed helper for `batchCall` that would take in a list of abi-typed calls (like `writeContract`) and execute them through `batchCall`

### LINKED ISSUE #1761: encodeSystemCalls(From) should use batchCall(From) interface
`encodeSystemCalls` and `encodeSystemCallsFrom` are both meant to be used for `batchCall` and `batchCallFrom`, but the batch call arguments expect an array of object (i.e. struct), not tuples

we should probably rename them to `encodeBatchCall` and `encodeBatchCallFrom` as well, for clarity (and maybe `encodeSystemCall` -> `encodeCall` and `encodeSystemCallFrom` -> `encodeCallFrom`)

holic:
closes #2779 (couldn't push to their branch so re-opening here)
closes #1761
closes #1728


holic:
@dhvanipa It looks like you're trying to do `batchCall(callFrom)` via `batchCallFrom`. I wonder if we could bypass `batchCallFrom` and use `callFrom(batchCall)` instead?

packages/world/ts/actions/callFrom.ts
```diff
@@ -55,11 +56,28 @@ export function callFrom(
...
+        const batchCallArgs = writeArgs as unknown as WriteContractParameters<worldCallAbi, "batchCall">;
+        const [systemCalls] = batchCallArgs.args;
+        if (!systemCalls.length) {
+          throw new Error("`batchCall` should have at least one system call.");
+        }
+
+        return _writeContract({
+          ...batchCallArgs,
+          functionName: "batchCallFrom",
+          args: [systemCalls.map((systemCall) => ({ from: params.delegatorAddress, ...systemCall }))],
```
dhvanipa, line 77: This expansion of `systemCall` fails if the way the user passes in these args is using `encodeSystemCalls`/`encodeSystemCall`.

This is because these functions are meant for `call` instead of `batchCall`. Both signatures have the same data except `batchCall` takes in a struct while call does not.

dhvanipa: updating call to take in a struct would require upgrading the world so not sure if thats a good idea

other options:
- add `encodeBatchSystemCall` 
- encodeSystemCall gives you back a SystemCall object, then to convert to use with .call you just Object.values(systemCall)

dhvanipa: Ended up creating a wrapper in our client around `encodeSystemCalls` and `encodeSystemCall` for batch calls. Sharing here for posterity:

```
export type EncodedSystemCallData = {
  systemId: Hex;
  callData: Hex;
};

export type SystemCallData = {
  systemId: Hex;
  functionName: string;
  args: any[];
};

export function encodeSystemCallForBatchCall(abi: Abi, systemCall: SystemCallData): EncodedSystemCallData {
  const call = encodeSystemCall({
    abi,
    systemId: systemCall.systemId,
    functionName: systemCall.functionName,
    args: systemCall.args,
  });
  return {
    systemId: call[0],
    callData: call[1],
  };
}

export function encodeSystemCallsForBatchCall(abi: Abi, systemCalls: SystemCallData[]): EncodedSystemCallData[] {
  const batchCalls: EncodedSystemCallData[] = [];
  for (const call of encodeSystemCalls(abi, systemCalls)) {
    batchCalls.push({
      systemId: call[0],
      callData: call[1],
    });
  }
  return batchCalls;
}
```

holic: Turns out I had the wrong return types in `encodeSystemCallFrom` and `encodeSystemCallsFrom`. I fixed those up and now I think these utils should work with the `callFrom` action changes here.

PR #2803, opened 2024-05-08
feat(cli): deploy with external modules

holic:
- [ ] module input/output config tests
- [x] update verify with new approach
- [x] deprecate `name`, move to a compat layer
- [ ] (for later) replace table ID resolving with something that has stronger type inference+autocomplete


e2e/packages/contracts/mud.config.ts
```diff
@@ -60,9 +60,10 @@ export default defineWorld({
   },
   modules: [
     {
-      name: "Unstable_CallWithSignatureModule",
+      name: "CallWithSignatureModule",
```
holic, line 63: `name` isn't really tied to anything when using `artifactPath`, except the label that gets printed in log messages

not sure if we should make the types reflect this better or leaving a field here for a "nice name" is nice?

alvrs (approved):
needs a changeset but looking good otherwise!

PR #2807, opened 2024-05-09
feat(store,world): usable enum values from config

### LINKED ISSUE #1957: config should return enums as a usable TS object
We allow defining enums in the MUD config like 

```ts
export default mudConfig({
  enums: {
    Direction: ["North", "East", "South", "West"],
  },
});
```

which outputs Solidity

```
enum Direction {
  North,
  East,
  South,
  West
}
```

but we currently don't have a way to reference an enum in TS for calling contracts, so user land is forced to define this.

It would be nice if we could do something like

```ts
import config from "./mud.config";

await worldContract.write.move([config.enums.Direction.North]);
```
where `config.enums.Direction.North` is strongly typed and has a value of `0` to match the contract enum value

holic:
transforms enum input arrays into an object so they can be used in TS

closes #1957

### DIFF
--- a/.changeset/sweet-lemons-eat.md
+++ b/.changeset/sweet-lemons-eat.md
@@ -19,7 +19,7 @@ config.enumValues.TerrainType.Grass;
 //                              ^? (property) Grass: 1
 ```
 
-This should allow for easier referencing of enum values (i.e. `uint8` equivalent) in contract calls.
+This allows for easier referencing of enum values (i.e. `uint8` equivalent) in contract calls.
 
 ```ts
 writeContract({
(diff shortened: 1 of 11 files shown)


alvrs:
i very much like the new feature but is there any way we could turn this into a non-breaking change? maybe as a new property on the output? there might be teams out there using the current structure

.changeset/sweet-lemons-eat.md
```diff
@@ -0,0 +1,30 @@
...
+//                              ^? (property) Grass: 1
+```
+
+This should allow for easier referencing of enum elements in contract calls.
+
+```ts
+writeContract({
+  // …
+  functionName: "setTerrainType",
+  args: [config.mappedEnums.TerrainType.Grass],
```
holic, line 28:
```suggestion
  args: [config.enumValues.TerrainType.Grass],
```

🤔

alvrs (dismissed):

PR #2802, opened 2024-05-08
increasing node memory to be able to build the store-sync package

joergkiesewetter:
fixes #2801

PR #2810, opened 2024-05-10
build: up node memory during docker build

joergkiesewetter:
applying the same change as on #2802, but on main branch

(cherry picked from commit 31e0a4b14e914e08f878785b3a040cc528e4146d)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,6 +42,7 @@ WORKDIR /app
 # this resolves some chicken-and-egg problems with using workspace bins before they're created (install -> build -> install)
 RUN pnpm recursive run prepare
 RUN --mount=type=cache,id=pnpm,target=/pnpm/store pnpm install --frozen-lockfile
+ENV NODE_OPTIONS=--max-old-space-size=4096
 RUN pnpm run -r build
 
 FROM mud AS store-indexer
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -42,6 +42,7 @@ WORKDIR /app
 # this resolves some chicken-and-egg problems with using workspace bins before they're created (install -> build -> install)
 RUN pnpm recursive run prepare
 RUN --mount=type=cache,id=pnpm,target=/pnpm/store pnpm install --frozen-lockfile
+ENV NODE_OPTIONS=--max-old-space-size=4096
 RUN pnpm run -r build
```
holic, line 46: do you know `ENV` sticks around in the docker container? as in running the indexer would use 4gb memory too?

holic: https://docs.docker.com/reference/dockerfile/#env
> The ENV instruction sets the environment variable <key> to the value <value>. This value will be in the environment for all subsequent instructions in the build stage and can be [replaced inline](https://docs.docker.com/reference/dockerfile/#environment-replacement) in many as well.

sounds like just during build?

holic: but also seeing
> If an environment variable is only needed during build, and not in the final image, consider setting a value for a single command instead:

so I wonder if we use this instead:
```
RUN NODE_OPTIONS=--max-old-space-size=4096 pnpm run -r build
```

joergkiesewetter:
ok, changed it. For me, the build works like this, as well. I will also do a PR to change it in the "historical_indexer"-branch

holic (approved):
thanks!

PR #2816, opened 2024-05-13
feat(template/vanilla): use a namespace 🏎 

qbzzt:


qbzzt:
I don't know how to fix the test, but I released a snapshot and it works.

holic:
> I don't know how to fix the test, but I released a snapshot and it works.

need to run codegen!

qbzzt:
> > I don't know how to fix the test, but I released a snapshot and it works.
> 
> need to run codegen! `pnpm run build` from `packages/contracts`

Thank you! I worked here, and I'm now doing the same on the other PRs.

holic:
any chance we can do all the templates in one PR?

templates/vanilla/packages/contracts/mud.config.ts
```diff
@@ -1,6 +1,7 @@
 import { defineWorld } from "@latticexyz/world";
 
 export default defineWorld({
+  namespace: "counter",
```
alvrs, line 4: thoughts about using `home` as default namespace in templates? cc @holic

qbzzt:
How about `main`?

qbzzt:
Closed in favor of a single PR with all the templates.

PR #2817, opened 2024-05-13
feat(templates/react-ecs): use a namespace 🏎 

qbzzt:


qbzzt:
Closed in favor of a single PR with all the templates.

PR #2828, opened 2024-05-15
build: clean up TS build pipeline

ssalbdivad:
This PR addresses a big underlying problem with downstream type-perf due to `.ts` files being referenced from `package.json`.

To allow types to resolve to `.ts` files while in dev without affecting build output, I've set up a repo-wide set of tsconfig paths aligned to each package name, combined with any subpath exports defined in its package.json.

Changes include:

- Adding base tsconfig.json at the repo root defining paths and common settings
- Extend the base tsconfig.json in each package and dedupe other settings
- Remove unnecessary nested .gitignore files
- Replace .npmignore files with `files` in package.json

I'd appreciate some sanity checks on local dev working as expected etc. before this is merged.

This PR is already quite large in scope, and the type improvements it offers are quite significant, so rather than letting it continue to grow, I have a few suggestions for follow up work:

- Ensure `tsx` is used throughout development and Vitest adds the `vite-tsconfig-paths` plugin so that imports also resolve to `.ts` during dev.
- Implement a build script that creates a temporary tsconfig with `compilerOptions/paths` set to `{}` for building so that the build will fail if the tsconfig paths defined for a given package aren't also resolvable in build output. This could then replace `tsup` and all the corresponding configs could be removed, which would be great for deduplication and simplicity.
- Take an additional pass at the project-specific tsconfig settings, removing as many overrides as possible to standardize behavior across the repo. In some cases these may be needed, but largely the overrides may just be an artifact of the way the tsconfigs were initially defined.
- Migrate to  module NodeNext, i.e. .js suffixes on all local imports. Kind of an eye sore, but the settings has lots of other advantages. I think you mentioned wanting to get rid of barrel files- this would be a good change accompanying that. That is unless you want to use Bundler with .ts imports or similar and then have some tool handle converting the build output for you, but if you want to use tsc which I think is ideal for a repo like this, .js endings are probably your best bet.
- In `packages/cli/src/commands/set-version.ts` on line 85, there was an `npmResult` that was being implicitly resolved as any due to a resolution issue. That should have an explicit type, but in absence of one existing, I've explicitly typed it as `any` for now and added a TODO referencing this PR.
- A few packages like `CLI` which probably shouldn't rely on DOM types do indirectly through utils, e.g.:

```
packages/cli test:ci: ../common/src/utils/waitForIdle.ts(3,16): error TS2304: Cannot find name 'requestIdleCallback'.
```
 
I've added `DOM` to their `tsconfig.json` for now to unblock the build, but eventually utils should probably either be split into a separate package for DOM APIs or have a separate entry point.

- I had to make a few suboptimal changes to get the `.abi.json` files to resolve to their corresponding .d.ts with a default `moduleResolution` setting of `"bundler"` (by default with this setting and `resolveJsonModule`, the imports seem to resolve to the non-narrowed `.json` files rather than the `.d.ts` with narrowed types:

`packages/world/tsconfig.json`:     "resolveJsonModule": false,
`packages/cli/tsconfig.json`:       "moduleResolution": "Node",

 
Happy to schedule a call to go over all this in more detail- I know it's a lot!


packages/config/package.json
```diff
@@ -16,25 +16,9 @@
...
-        "./src/deprecated/register/index.ts"
-      ],
-      "node": [
-        "./src/deprecated/node/index.ts"
-      ]
-    }
-  },
+  "files": [
+    "dist"
+  ],
```
holic, line 21: just noting my own learnings here:
we don't need to include everything from `.npmignore` here because by default, npm picks up `package.json` as well as `README` (with any extension): https://docs.npmjs.com/cli/v10/configuring-npm/package-json#files

packages/cli/package.json
```diff
@@ -69,7 +71,7 @@
     "@types/debug": "^4.1.7",
     "@types/ejs": "^3.1.1",
     "@types/glob": "^7.2.0",
-    "@types/node": "^18.15.11",
+    "@types/node": " 20.12.12",
```
holic, line 74: why the leading space? and why the version bump to 20 if we're still using 18?

packages/common/package.json
```diff
@@ -20,37 +20,10 @@
...
-      ],
-      "utils": [
-        "./src/utils/index.ts"
-      ],
-      "kms": [
-        "./src/exports/kms.ts"
-      ]
-    }
-  },
+  "main": "./dist/index.js",
```
holic, line 23: why did `main` get added here and not in others?

packages/react/package.json
```diff
@@ -12,7 +12,9 @@
   "exports": {
     ".": "./dist/index.js"
   },
-  "types": "src/index.ts",
+  "files": [
+    "dist"
```
holic, line 16: I noticed CHANGELOG was previously included via npmignore but not here, curious if you think it's worth including? cc @alvrs

packages/schema-type/package.json
```diff
@@ -14,19 +14,12 @@
...
-      "deprecated": [
-        "./src/typescript/exports/deprecated.ts"
-      ]
-    }
-  },
+  "files": [
+    "dist",
+    "out",
+    "foundry.toml",
+    "cache"
```
holic, line 21: This is a carry over from npmignore's line:
```
!cache/solidity-files-cache.json
```

But I wonder if we actually need this? cc @alvrs 

```suggestion
```

holic, line 21: This is a carry over from npmignore's line:
```
!cache/solidity-files-cache.json
```

But I wonder if we actually need this? cc @alvrs 

```suggestion
    "foundry.toml"
```

packages/schema-type/tsconfig.json
```diff
@@ -1,16 +1,8 @@
...
-    "declaration": true,
-    "sourceMap": true,
-    "isolatedModules": true,
-    "esModuleInterop": true,
-    "forceConsistentCasingInFileNames": true,
-    "skipLibCheck": true,
-    "strict": true
-  }
+    "outDir": "dist",
+    "types": []
```
holic, line 5: any reason to explicitly define this as an empty array?

packages/store/package.json
```diff
@@ -19,28 +19,12 @@
...
-      "register": [
-        "./ts/register/index.ts"
-      ]
-    }
-  },
+  "files": [
+    "dist",
+    "out",
+    "foundry.toml",
+    "cache"
```
holic, line 26: same as above, need to check if we actually need this here

packages/world/package.json
```diff
@@ -18,25 +18,13 @@
...
-        "./ts/node/index.ts"
-      ]
-    }
-  },
+  "files": [
+    "dist",
+    "out",
+    "foundry.toml",
+    "cache",
+    "test/MudTest.t.sol"
```
holic, line 26: @alvrs I wonder if we should move this into src

packages/world/ts/worldEvents.test.ts
```diff
@@ -1,7 +1,7 @@
 import { describe, expect, it } from "vitest";
 import IWorldFactoryAbi from "../out/IWorldFactory.sol/IWorldFactory.abi.json";
 import IBaseWorldAbi from "../out/IBaseWorld.sol/IBaseWorld.abi.json";
-import { helloWorldEvent, worldDeployedEvent } from "./worldEvents";
+import { helloWorldEvent, worldDeployedEvent } from "./worldEvents.js";
```
holic, line 4: I realize we wanna move towards this but is the `.js` extension here needed?

packages/react/package.json
alvrs: no strong opinion here

packages/schema-type/package.json
alvrs: i don't remember why it was there, seems not very useful to include the files cache in the published packages 🤔

packages/world/package.json
alvrs: Changing the import path would be a breaking change, would prefer to avoid unless there is a strong reason to do so (but we could add a reexport from src if you feel strongly about importing it from there)

packages/react/package.json
ssalbdivad: CHANGELOG, package.json, README.md and LICENSE.md are automatically included, regardless of whether they are explicitly listed in files.

packages/world/ts/worldEvents.test.ts
ssalbdivad: I can get rid of this, it was likely just an accident I was testing out some different settings to get json resolution working

packages/cli/package.json
ssalbdivad: I was having some trouble getting `fetch` to resolve, so tried upgrading node types. That seemed to work, but likely only because it forced TS to reload a cache or something. Reverting this.

packages/common/package.json
ssalbdivad: moduleResolution "Node" requires this be specified here as IIRC it doesn't respect exports.

I believe only this package was being imported in a problematic way, but to be on the safe side it might be best as a followup to add this to the other packages as well for the primary entry point.

packages/schema-type/tsconfig.json
ssalbdivad: It would prevent @types from being resolved, but I'm not sure why I added it here. I'll remove it.

packages/react/package.json
holic: going to go with no changelog for now to keep package smaller, can revisit in the future

holic (approved):

PR #2830, opened 2024-05-15
build: use correct error message 🏎

qbzzt:
It gave me an error message saying I need to run `gas-report`, even though that's not used in the templates. So I fixed it.

### DIFF
--- a/.github/workflows/templates.yml
+++ b/.github/workflows/templates.yml
@@ -43,7 +43,7 @@ jobs:
         working-directory: ./templates/${{ matrix.template }}
         run: pnpm run build
 
-      - name: Outdated files, run `pnpm gas-report` and commit them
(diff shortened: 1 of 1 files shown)


.github/workflows/templates.yml
```diff
@@ -43,7 +43,7 @@ jobs:
         working-directory: ./templates/${{ matrix.template }}
         run: pnpm run build
 
-      - name: Outdated files, run `pnpm gas-report` and commit them
+      - name: Out of date files, run `pnpm build` in `pacakges/contracts`
```
alvrs, line 46:
```suggestion
      - name: Outdated files, run `pnpm build` and commit them
```

alvrs: for consistency with the other places where we use this message

alvrs (approved):

Issue #2826, opened 2024-05-14
fix hanging test


holic:
(from https://github.com/latticexyz/mud/pull/2823#issuecomment-2111072372)

```
packages/store-sync test:ci: close timed out after 500ms
packages/store-sync test:ci: Failed to terminate worker while running /home/runner/work/mud/mud/packages/store-sync/src/recs/recsStorage.test.ts.
packages/store-sync test:ci: Tests closed successfully but something prevents Vite server from exiting
packages/store-sync test:ci: You can try to identify the cause by enabling "hanging-process" reporter. See https://vitest.dev/config/#reporters
```

maybe related:
https://github.com/vitest-dev/vitest/issues/3077
https://github.com/nodejs/undici/issues/2026#issuecomment-2000017934

sounds like maybe bumping to node >=18.20.1 could fix the above

in the meantime, we can add some CI timeouts like in #2768 to ensure this doesn't run longer than it should

holic:
node bumped and timeout added in https://github.com/latticexyz/mud/pull/2836

PR #2831, opened 2024-05-15
build: bump to node 18.20.2, pnpm 9.1.1

holic:
addresses part of #2826 and maybe fixes vite process hanging due to anvil.js

to catch up your local environment
```sh
# install the node version in .nvmrc
nvm install
# use the node version in .nvmrc
nvm use
# install the pnpm version we use
npm install -g pnpm@9.1.1
```


.github/actions/setup-prerequisites/action.yml
```diff
@@ -5,8 +5,6 @@ runs:
   steps:
     - name: Setup pnpm
       uses: pnpm/action-setup@v2
-      with:
-        version: 8.x
```
holic, line 9: this will default to the `packageManager` version in `package.json`

Dockerfile
```diff
@@ -31,7 +31,7 @@ RUN curl -L https://foundry.paradigm.xyz/ | bash && \
 # pnpm
 ENV PNPM_HOME="/pnpm"
 ENV PATH="${PATH}:${PNPM_HOME}"
-RUN npm install pnpm@8 --global && pnpm --version
+RUN npm install pnpm@9.1.1 --global && pnpm --version
```
holic, line 34: ideally this would install from the `packageManager` version so we don't have to redefine here but unclear how to do that

alvrs (dismissed):

PR #2832, opened 2024-05-15
fix(cli): fixed module artifactPath imports 

holic:
I should have removed this in https://github.com/latticexyz/mud/pull/2803 because we had already removed artifact imports there so we weren't using the dep anymore

this clears the path for using `mud` CLI inside world-modules, which means we can split up the big, gnarly mud config there


packages/cli/src/deploy/common.ts
```diff
@@ -22,7 +21,7 @@ export const worldTables = configToTables(worldToV1(worldConfig));
 
 export const worldDeployEvents = [helloStoreEvent, helloWorldEvent] as const;
 
-export const worldAbi = [...IBaseWorldAbi, ...IModuleAbi] as const;
+export const worldAbi = IBaseWorldAbi;
```
holic, line 24: the `IBaseWorld` ABI now includes the errors we wanted from `IModule` ABI

alvrs (approved):

alvrs (approved):

PR #2840, opened 2024-05-16
feat(store,world): add option to codegen tables into namespace dirs

holic:


packages/store/ts/config/v2/table.ts
```diff
@@ -126,7 +126,7 @@ export function resolveTableCodegen<input extends TableInput>(input: input): res
     storeArgument: get(options, "storeArgument") ?? TABLE_CODEGEN_DEFAULTS.storeArgument,
     // dataStruct is true if there are at least 2 value fields
     dataStruct: get(options, "dataStruct") ?? Object.keys(input.schema).length - input.key.length > 1,
-  } satisfies TableCodegen as resolveTableCodegen<input>;
+  } satisfies TableCodegen as never;
```
yonadaaa, line 129: 🫨

packages/store/mud.config.ts
```diff
@@ -4,7 +4,7 @@ export default defineStore({
   codegen: {
     storeImportPath: "../../",
   },
-  namespace: "store" as const,
```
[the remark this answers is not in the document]
holic: yep! this was leftover from previous iterations

packages/store/ts/config/v2/table.ts
holic: this just aligns with the rest of our `as never` usage, context here: https://discord.com/channels/865335009915961364/1199772289495990404/1230849460842532915

yonadaaa: very interesting, thank you!

packages/store/ts/scripts/generate-test-tables.ts
```diff
@@ -1,14 +1,15 @@
...
+
+const configPath = fileURLToPath(import.meta.url);
 
 const config = defineStore({
+  sourceDirectory: "../../test",
   codegen: {
     storeImportPath: "../../../src/",
-    outputDirectory: "../test/codegen",
   },
-  namespace: "store",
```
yonadaaa, line 11: Why was this removed? 

And shouldn't we have an example of a namespaced table?

yonadaaa (approved):
works for me!

alvrs (approved):

PR #2844, opened 2024-05-20
refactor(world): store names and types of function arguments and return values onchain

yonadaaa:


yonadaaa:
closing in favour https://github.com/latticexyz/mud/pull/2849 because this is quite complex and we don't really need named params

Issue #2845, opened 2024-05-21
Generate system interfaces from onchain function signatures


yonadaaa:
We can fetch the function definitions for a given system ID from the `FunctionSignatures` table and use them to codegen the system interface. Users can interact with this interface or compile it to generate the system ABI.

This can start as a helper function then we can make it a CLI command.

PR #2849, opened 2024-05-21
feat(cli): generate system ABI's from World address

yonadaaa:
Adds a new MUD `generate-abi` command that generates an ABI for a given World by reading the onchain system registrations

fixes https://github.com/latticexyz/mud/issues/2845


packages/cli/src/commands/index.ts
```diff
@@ -14,6 +14,7 @@ import test from "./test";
 import trace from "./trace";
 import devContracts from "./dev-contracts";
 import verify from "./verify";
+import abigen from "./abigen";
```
holic, line 17: not a huge fan of continuing the tablegen, worldgen, etc. naming pattern

I could see us eventually combining all of these features under a `codegen` command, maybe with sub commands like `codegen tables`), but in the meantime, something like "codegen abi" or "generate abi" for the CLI command and "world to abi" for the helper methods?

yonadaaa: hmm I like this line of thinking, lemme marinate on it

It's interesting that `abigen` and `worldgen` have different inputs but basically the same output

yonadaaa: The command is now `generate-abi` and the helper is just `systemToAbi`

packages/cli/src/commands/generate-abi.ts
```diff
@@ -0,0 +1,66 @@
...
+  const worldAbi = [];
+
+  // render system ABI's
+  for (const system of systems) {
+    const fullOutputPath = `${resourceToLabel(system)}.json`;
+    const abi = systemToAbi(system);
+    await fs.writeFile(fullOutputPath, JSON.stringify(abi));
+
+    // prepare IWorld ABI
+    worldAbi.push(...abi);
```
holic, line 58: is this accurate?

systems will have a function name like `function increment()` but world may have it as `function app__increment()`

yonadaaa: wowww good point! i think we will have to prepare the world ABI seperately

packages/cli/src/commands/generate-abi.ts
```diff
@@ -0,0 +1,66 @@
...
+    return yargs.options(generateAbiOptions);
+  },
+
+  async handler(args) {
+    await generateAbiHandler(args);
+    process.exit(0);
+  },
+};
+
+export async function generateAbiHandler(opts: Options) {
```
holic, line 36: thoughts on putting this into a separate dir like `packages/cli/src/init` or something to prepare for more of these types of files?

packages/cli/src/commands/generate-abi.ts
```diff
@@ -0,0 +1,66 @@
+import type { CommandModule, InferredOptionTypes } from "yargs";
+import { Hex, createWalletClient, http } from "viem";
+import { getSystems } from "../deploy/getSystems";
+import { getWorldDeploy } from "../deploy/getWorldDeploy";
```
holic, line 4: we should consider "lifting" these out of `deploy` dir if we're sharing them across different commands/utils

yonadaaa: Yup, and I think having separate directories for each command makes less sense with these new tools. We'll basically be reusing everything except for the `ensureX` functions that actually deploy the contracts

packages/cli/src/commands/generate-abi.ts
yonadaaa: What do you mean by "combining" the CLI commands? Like just putting some of this logic in a helper function? I imagine if we have one command that combines the logic of other commands, they all still live in the `commands` directory but share helpers

PR #2850, opened 2024-05-21
fix(cli,world): resolve table by just name

holic:
`resolveTableId` (or specifically the context passed into `resolveWithContext`) changed slightly in #2803 where we started using namespaced table names, which broke some previous usages of just table names without a namespace. This adds back in support for just table names, with an error thrown if we can't resolve the table due to collisions (with an actionable error message to quickly resolve).

We had set up a new `resolveTableId` with the new config but never used it/exported it, so this also fixes that and removes the long-deprecated one as well as the now-unused v1 modules config.


packages/world/ts/exports/internal.ts
```diff
@@ -11,3 +11,5 @@ export * from "../encodeSystemCallsFrom";
 export * from "../actions/callFrom";
 
 export * from "../callWithSignatureTypes";
+
+export { resolveTableId, resolveWithContext } from "../config/v2/dynamicResolution";
```
holic, line 15: I am intentionally exporting `resolveTableId` in the internal bundle for now as I want to reconsider the API before we make it concrete.

packages/world/ts/config/v2/dynamicResolution.ts
```diff
@@ -1,10 +1,8 @@
-export enum DynamicResolutionType {
-  TABLE_ID,
-  SYSTEM_ADDRESS,
-}
+import { World } from "./output";
 
 export type DynamicResolution = {
-  type: DynamicResolutionType;
+  // TODO: add systemAddress support
+  type: "tableId";
```
holic, line 5: this is an internal thing and changing this to strings instead of enums leads to better errors in the `resolveWithContext` function

packages/world/ts/config/v2/dynamicResolution.ts
```diff
@@ -38,20 +36,29 @@ export function isValueWithType(value: unknown): value is ValueWithType {
...
+        ([tableName, table]) => tableName === input.input || table.name === input.input,
+      );
 
-    if (input.type === DynamicResolutionType.TABLE_ID) {
-      const tableId = context.tableIds?.[input.input];
-      resolved = tableId && { value: tableId, type: "bytes32" };
-    }
+      if (tableEntries.length > 1) {
+        throw new Error(
+          `Found more than one table with name "${input.input}". Try using one of the following table names instead: ${tableEntries.map(([tableName]) => tableName).join(", ")}`,
```
alvrs, line 51: nice!

alvrs (approved):
Needs a changeset but aside from that lgtm

PR #2851, opened 2024-05-21
docs(config): add module definitions

qbzzt:


docs/pages/config.mdx
```diff
@@ -48,6 +56,23 @@ export default defineWorld({
...
+      args: [
+        {
+          type: "bytes32",
+          value: resourceToHex({
+            type: "table",
+            namespace: "app",
+            name: "Tasks",
+          }),
+        },
+      ],
```
holic, line 73: can we make the default recommendation use `resolveTableId` since it's less complex? and save this for the more advanced usage?
```suggestion
      args: [resolveTableId("Tasks")],
```

docs/pages/config.mdx
```diff
@@ -17,6 +17,14 @@ The is an example of a `World` config:
 ```tsx
 import { defineWorld } from "@latticexyz/world";
 
+// Used for the module parameters, not required otherwise.
+// If you use this, make sure to run:
+//
+// pnpm install @latticexyz/common
+//
+// In the packages/contracts directory.
+import { resourceToHex } from "@latticexyz/common";
```
ludns, line 26: this is importing the wrong thing right? seems like you use `resolveTableId`, not `resourceToHex`

qbzzt: Yes, I should have fixed it when I changed it below. Thank you for catching this - fixed.

docs/pages/config.mdx
```diff
@@ -125,3 +130,78 @@ The global configuration keys are all optional.
...
+
+       ```sh
+       cd packages/contracts
+       pnpm install @latticexyz/config
+       ```
+
+    1. In `mud.config.ts`, import `resolveTableId`.
+
+       ```typescript
+       import { resolveTableId } from "@latticexyz/config/register";
```
holic, line 163:
```suggestion
       import { resolveTableId } from "@latticexyz/world/internal";
```

docs/pages/config.mdx
```diff
@@ -17,6 +17,14 @@ The is an example of a `World` config:
 ```tsx
 import { defineWorld } from "@latticexyz/world";
 
+// Used for the module parameters, not required otherwise.
+// If you use this, make sure to run:
+//
+// pnpm install @latticexyz/common
+//
+// In the packages/contracts directory.
+import { resolveTableId } from "@latticexyz/config/register";
```
holic, line 26: like my previous comment, this should be imported from world (see https://mud.dev/changelog for details)

docs/pages/config.mdx
```diff
@@ -125,3 +123,64 @@ The global configuration keys are all optional.
...
+    </details>
+
+    <details>
+
+    <summary>Use `resourceToHex` to specify the full resource information.</summary>
+
+    1. In `mud.config.ts`, import `resourceToHex`.
+
+       ```typescript
+       import { resourceToHex } from "@latticexyz/world/internal";
```
holic, line 168: this still lives in common

and I'm not sure we need to document this as I'd rather folks not lean on this particular approach (gives us less flexibility to change the underlying implementation)

qbzzt: OK, I'll remove that.

docs/pages/config.mdx
```diff
@@ -12,18 +12,19 @@ The config is used to define:
   If you want greater control over your systems (to change their public access or their name), you can use the `systems` object in the config.
 - The [modules](/world/modules) that will be installed in the `World`.
 
-The is an example of a `World` config:
+This is an example of a `World` config:
 
 ```tsx
 import { defineWorld } from "@latticexyz/world";
+import { resolveTableId } from "@latticexyz/world/internal";
```
alvrs, line 19: We should not document things that live in `/internal`, the idea with those exports is that they're only used internally. This makes me wonder though whether we should make this util an external export since it's needed for this use case @holic?

docs/pages/config.mdx
```diff
@@ -90,19 +99,8 @@ The global configuration keys are all optional.
   The keys are the names of the enumerations.
   The values are arrays of the strings for the values the enumeration can take.
 
-- **`tables`**: a record of tables. The keys in the record are table names.
```
alvrs, line 93: why was this removed?

qbzzt: Because in the current version we have two separate `tables` lines. [See here for the other one](https://github.com/latticexyz/mud/blob/main/docs/pages/config.mdx?plain=1#L61)

docs/pages/config.mdx
holic: I intentionally made it internal because I am not happy with this API and want to replace it without it being considered a breaking change

qbzzt:
The config file documentation changed so much this is no longer relevant as is.

Issue #2847, opened 2024-05-21
Generate worlds.json from contract address


yonadaaa:
Given a world address, we can get the deployment block number by fetching `HelloWorld` events from the RPC (the earliest one in case the world is a proxy).

PR #2853, opened 2024-05-22
feat(cli): generate worlds.json given a world address

yonadaaa:
does https://github.com/latticexyz/mud/issues/2847

This allows projects that are extending an existing World to define what address they are interacting with

### DIFF
--- a/packages/cli/src/commands/index.ts
+++ b/packages/cli/src/commands/index.ts
@@ -14,6 +14,7 @@ import test from "./test";
 import trace from "./trace";
 import devContracts from "./dev-contracts";
 import verify from "./verify";
+import generateWorlds from "./generateWorlds";
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any -- Each command has different options
 export const commands: CommandModule<any, any>[] = [
@@ -30,4 +31,5 @@ export const commands: CommandModule<any, any>[] = [
   devContracts,
   abiTs,
   verify,
+  generateWorlds,
 ];
--- a/packages/cli/src/runDeploy.ts
+++ b/packages/cli/src/runDeploy.ts
@@ -18,6 +18,8 @@ import { build } from "./build";
 import { kmsKeyToAccount } from "@latticexyz/common/kms";
 import { configToModules } from "./deploy/configToModules";
 
+export const localChains = [1337, 31337];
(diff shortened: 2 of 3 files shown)


packages/cli/src/runDeploy.ts
```diff
@@ -18,6 +18,8 @@ import { build } from "./build";
 import { kmsKeyToAccount } from "@latticexyz/common/kms";
 import { configToModules } from "./deploy/configToModules";
 
+export const localChains = [1337, 31337];
```
yonadaaa, line 21: Given all of these new tools that lift logic from `deploy`, we need to restructure as suggested in https://github.com/latticexyz/mud/pull/2849#discussion_r1609704642

packages/cli/src/commands/init.ts
```diff
@@ -0,0 +1,49 @@
+import type { CommandModule, InferredOptionTypes } from "yargs";
+import { loadConfig } from "@latticexyz/config/node";
+import { World as WorldConfig } from "@latticexyz/world";
+import { getRpcUrl } from "@latticexyz/common/foundry";
+import { Hex, createWalletClient, http } from "viem";
+import { getWorldDeploy } from "../deploy/getWorldDeploy";
+import { writeDeploymentResult } from "../utils/writeDeploymentResult";
+
+const verifyOptions = {
+  configPath: { type: "string", desc: "Path to the MUD config file" },
```
holic, line 10: what's the expected behavior if you already have a config?

I thought the goal of `mud init` was to generate a mud config given a world address?

packages/cli/src/commands/index.ts
```diff
@@ -30,4 +31,5 @@ export const commands: CommandModule<any, any>[] = [
   devContracts,
   abiTs,
   verify,
+  init,
```
holic, line 34: prob need a changeset for this (and in a future PR, can edit/expand the changeset to include new behavior)

packages/cli/src/commands/init.ts
yonadaaa: Good point, the config will ultimately be generated as part of this!

holic:
going with a different approach for now (see https://github.com/latticexyz/mud/pull/3171)

PR #2860, opened 2024-05-22
docs(guides / extending): update for the `app` namespace 🚗 

qbzzt:


docs/pages/guides/extending-a-world/index.mdx
```diff
@@ -631,13 +629,14 @@ We will fix these problems now.
    } = await setup();
 
    // Components expose a stream that triggers when the component is updated.
-   components.Counter && components.Counter.update$.subscribe((update) => {
-     const [nextValue, prevValue] = update.value;
-     console.log("Counter updated", update, { nextValue, prevValue });
-     document.getElementById("counter")!.innerHTML = String(nextValue?.value ?? "unset"$
-   });
+   components.Counter &&
+     components.Counter.update$.subscribe((update) => {
```
holic, line 633: why is the conditional needed? can this be simplified to
```suggestion
   components.Counter.update$.subscribe((update) => {
```

holic: ahh, in that case, would this work?
```suggestion
   components.Counter?.update$.subscribe((update) => {
```

holic (approved):

PR #2870, opened 2024-05-24
feat(world): system namespace based on file name

yonadaaa:


packages/common/src/labelToResource.ts
```diff
@@ -0,0 +1,20 @@
+import { ROOT_NAMESPACE } from "./resourceToLabel";
+
+export function labelToResource(label: string): {
```
holic, line 3: fwiw I don't think we should encourage this behavior of going from a string label to a resource

we should probably just keep this as an internal util close to where we need it (i.e. worldgen)

packages/common/src/resourceToLabel.ts
```diff
@@ -1,9 +1,9 @@
-const rootNamespace = "";
+export const ROOT_NAMESPACE = "";
```
holic, line 1: we should probably lift this out into its own file and use it more throughout the TS codebase for clarity

packages/common/src/labelToResource.ts
yonadaaa: Okay that makes sense! I think we need it in a few places lemme see

packages/common/src/resourceToLabel.ts
yonadaaa: Absolutelyy, I use this a lot

packages/world/ts/config/resolveWorldConfig.ts
```diff
@@ -17,7 +17,7 @@ export function resolveWorldConfig(
 ) {
   // Include contract names ending in "System", but not the base "System" contract, and not Interfaces
   const defaultSystemNames =
-    existingContracts?.filter((name) => name.endsWith("System") && name !== "System" && !name.match(/^I[A-Z]/)) ?? [];
+    existingContracts?.filter((name) => name.match(/^(?!I[A-Z]).*System$/) && !name.match(/^.+__I.+System$/)) ?? [];
```
holic, line 20: I think `(^|__)` is what you're looking for? "start of string or prefixed by __"

```suggestion
    existingContracts?.filter((name) => !/(^|___)I.+System$/.test(name)) ?? [];
```

packages/cli/src/deploy/resolveConfig.ts
```diff
@@ -44,8 +44,10 @@ export function resolveConfig<config extends ConfigInput>({
     .map(toFunctionSignature);
 
   const systems = Object.entries(resolvedConfig.systems).map(([systemName, system]): System => {
-    const namespace = config.namespace;
-    const name = system.name;
+    const resource = labelToResource(systemName);
```
holic, line 47: why are we parsing this when `system` already has `system.namespace` and `system.name`?

yonadaaa: This `System` type doesn't have a separate namespace and name? https://github.com/latticexyz/mud/blob/main/packages/world/ts/config/resolveWorldConfig.ts#L81

holic: ahh that's an oversight and we should fix that!

yonadaaa: Parsing the resource in `resolveSystemConfig` now!

examples/minimal/packages/contracts/script/PostDeploy.s.sol
```diff
@@ -6,10 +6,11 @@ import { console } from "forge-std/console.sol";
 import { StoreSwitch } from "@latticexyz/store/src/StoreSwitch.sol";
 import { ResourceId, WorldResourceIdLib, WorldResourceIdInstance } from "@latticexyz/world/src/WorldResourceId.sol";
 import { RESOURCE_SYSTEM } from "@latticexyz/world/src/worldResourceTypes.sol";
+import { Systems } from "@latticexyz/world/src/codegen/tables/Systems.sol";
 
 import { MessageTable } from "../src/codegen/index.sol";
 import { IWorld } from "../src/codegen/world/IWorld.sol";
-import { ChatNamespacedSystem } from "../src/systems/ChatNamespacedSystem.sol";
+import { namespace__ChatNamespacedSystem } from "../src/systems/namespace__ChatNamespacedSystem.sol";
```
holic, line 13: we could probably just name this `namespace__ChatSystem` since it's clear from the name that it is namespaced

packages/world/ts/config/resolveWorldConfig.ts
```diff
@@ -57,8 +59,18 @@ export function resolveWorldConfig(
  * Default value for accessListAddresses is []
  * Default value for accessListSystems is []
  */
-export function resolveSystemConfig(systemName: string, config?: SystemConfig, existingContracts?: string[]) {
-  const name = config?.name ?? systemName;
+export function resolveSystemConfig(
+  systemName: string,
+  configNamespace: string,
+  config?: SystemConfig,
+  existingContracts?: string[],
```
holic, line 66: can we convert this to an object? it gets really hard to reason about past 2 args

packages/world/ts/node/render-solidity/worldgen.ts
```diff
@@ -42,10 +43,14 @@ export async function worldgen(
         };
       }
     });
-    const systemInterfaceName = `I${system.basename}`;
+
+    const namespaceIsSet = system.basename.includes("__");
+    const { namespace, name } = resolvedConfig.systems[system.basename];
+
+    const systemInterfaceName = namespaceIsSet ? `${namespace}__I${name}` : `I${name}`;
```
holic, line 50: what about

```suggestion
    const systemInterfaceName = namespace === ROOT_NAMESPACE ? `I${name}` : `${namespace}__I${name}`;
```

.changeset/small-snakes-guess.md
```diff
@@ -0,0 +1,7 @@
+---
+"@latticexyz/common": patch
+"@latticexyz/world": patch
+"@latticexyz/cli": patch
+---
+
+Added support for namespaced systems.
```
holic, line 7: can you describe how this should be used?

packages/cli/src/deploy/resolveConfig.ts
```diff
@@ -44,7 +44,7 @@ export function resolveConfig<config extends ConfigInput>({
     .map(toFunctionSignature);
 
   const systems = Object.entries(resolvedConfig.systems).map(([systemName, system]): System => {
-    const namespace = config.namespace;
+    const namespace = system.namespace;
```
holic, line 47: does the deployer properly consider system namespaces when preparing to deploy this? registering namespace, etc

packages/world/ts/config/resolveWorldConfig.ts
```diff
@@ -12,12 +12,14 @@ export type ResolvedWorldConfig = ReturnType<typeof resolveWorldConfig>;
...
 export function resolveWorldConfig(
-  config: Pick<StoreConfig & WorldConfig, "systems" | "excludeSystems">,
+  config: Pick<StoreConfig & WorldConfig, "namespace" | "systems" | "excludeSystems">,
   existingContracts?: string[],
 ) {
   // Include contract names ending in "System", but not the base "System" contract, and not Interfaces
   const defaultSystemNames =
-    existingContracts?.filter((name) => name.endsWith("System") && name !== "System" && !name.match(/^I[A-Z]/)) ?? [];
+    existingContracts?.filter(
+      (name) => name.endsWith("System") && name !== "System" && !name.match(/^I[A-Z]/) && !name.match(/__I/),
```
holic, line 21: no change needed here, but just FYI, `.match` returns data where as `.test` just returns a boolean, so I tend to use it more often for conditionals like this

```ts
/__I/.test(name)
```

holic: not sure if you find this more or less readable but could combine the regex test and also ensure the `__I` variant matches the `[A-Z]` constraint
```suggestion
      (name) => name.endsWith("System") && name !== "System" && !/(^|__)I[A-Z]/.test(name),
```

examples/minimal/packages/contracts/script/PostDeploy.s.sol
yonadaaa: True, that's better now

packages/cli/src/deploy/resolveConfig.ts
yonadaaa: Yes - but good point. In the deployer we [get every system ID](https://github.com/latticexyz/mud/blob/main/packages/cli/src/deploy/deploy.ts#L93) in the project then [register their namespace](https://github.com/latticexyz/mud/blob/main/packages/cli/src/deploy/ensureNamespaceOwner.ts#L17) if needed.

.changeset/small-snakes-guess.md
yonadaaa: yes this is a WIP, just thinking about how best to put it!

.changeset/small-snakes-guess.md
yonadaaa, line 1: Wonder if we want to go into more detail here. There are a lot of nuances around root systems, the `namespace` vs `namespaces` key etc.

holic: I think this is great for now! Can always add more nuance in the `namespaces` config PR

packages/world/ts/node/render-solidity/worldgen.ts
yonadaaa: I agree, if the interface is user-facing we can change it to make things more consistent. 

It might be confusing though that the interface would now be named differently to your system. If you have `namespace: "app"`, and declare `SomeSystem.sol`, the interface would be `app_ISomeSystem.sol`

packages/world/ts/config/resolveWorldConfig.ts
yonadaaa: done!

packages/world/ts/node/render-solidity/worldgen.ts
yonadaaa: By this logic, why don't we change the default behaviour of https://github.com/latticexyz/mud/pull/2880 to always use `namespaceDirectories`?

holic: > It might be confusing though that the interface would now be named differently to your system. If you have `namespace: "app"`, and declare `SomeSystem.sol`, the interface would be `app_ISomeSystem.sol`

to clarify, it would be `codegen/world/app__ISomeSystem.sol` (note that we're in a world codegen directory), which I think makes more sense than `codegen/world/ISomeSystem.sol` for a system defined in `app` namespace

holic: > By this logic, why don't we change the default behaviour of #2880 to always use `namespaceDirectories`?

because these ones _are_ imported/used in user land and generating them depends on the config alone

I don't think the generated world interfaces were ever meant to be imported in Solidity, they just separate/build up the `IWorld` interface to have an ABI that we can build and import into TS

yonadaaa: Thought about this more and totally agree. We don't expect users to import the interface files so changing what they are called shouldn't break any existing code. Will change the behaviour!!

.changeset/small-snakes-guess.md
```diff
@@ -0,0 +1,15 @@
+---
+"@latticexyz/common": patch
```
holic, line 2:
```suggestion
```
I don't think this changeset makes much sense looking at this package alone

If we want a changeset for common, should add one that is just something like "Added export for root namespace string" with an example of how to use it

holic (dismissed):
just a small change to changeset but otherwise lgtm!

.changeset/small-snakes-guess.md
yonadaaa: 100% agree

PR #2873, opened 2024-05-24
feat(common): export base tsconfig

ssalbdivad:
This migrates `tsconfig.base.json` to `@latticexyz/common` so it can be referenced in templates, and updated overtime without creating breaking changes.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,7 +1,4 @@
 {
-  "extends": [
-    // Note that the base config gets copied over to each template workspace too
-    "./tsconfig.base.json",
-    "./tsconfig.paths.json"
-  ]
+  "extends": ["./packages/common/tsconfig.base.json", "./tsconfig.paths.json"],
+  "exclude": ["**/dist", "**/node_modules", "**/docs", "**/e2e"]
 }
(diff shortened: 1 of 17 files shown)


tsconfig.json
```diff
@@ -1,7 +1,4 @@
 {
-  "extends": [
-    // Note that the base config gets copied over to each template workspace too
-    "./tsconfig.base.json",
-    "./tsconfig.paths.json"
-  ]
+  "extends": ["./packages/common/tsconfig.base.json", "./tsconfig.paths.json"],
+  "exclude": ["**/dist", "**/node_modules", "**/docs", "**/e2e"]
```
holic, line 3: can you share a bit about these exclusions? specifically `docs` and `e2e` and wondering about the impacts (e2e tsconfigs extend this one and unclear how that impacts exclude here)

ssalbdivad: This was migrated directly from the previous `tsconfig.base.json`.

The intent I suppose is to avoid prematurely pulling those projects in in-editor, but we could take a closer look in a future commit if you'd like.

holic: That makes sense for most other packages, but if I were working in e2e dir and that tsconfig inherits from this base, where e2e is ignored, does that mean we don't get typechecking in e2e?

holic: > I could add that to this PR if you'd like but I do want to reiterate that this was not a change I made in this PR.

Totally, I think it was added as part of https://github.com/latticexyz/mud/pull/2828 ? I just missed the potential implications in that PR.

ssalbdivad: Cool, just let me know whether you'd like to address that here or tackle it via a separate PR.

holic (approved):

PR #2876, opened 2024-05-28
feat(cli): get systems from contract names instead of filenames

yonadaaa:
Instead of enforcing one system per file, we compile the project and check the output for contracts that end in `System`.


yonadaaa:
Closing as not priority, interesting idea though

PR #2877, opened 2024-05-28
fix(world-modules): register total supply table in erc20 module

yonadaaa:


.changeset/light-tables-drive.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/world-modules": patch
+---
+
+Fixed `ERC20Module` to register the `TotalSupply` table when creating a new token.
```
holic, line 5: can we put some instructions for existing worlds/tokens to apply this fix as well?

packages/world-modules/src/modules/erc20-puppet/ERC20Module.sol
```diff
@@ -80,6 +81,7 @@ contract ERC20ModuleRegistrationLibrary {
     // Register the tables
     Allowances.register(_allowancesTableId(namespace));
     Balances.register(_balancesTableId(namespace));
+    TotalSupply.register(_totalSupplyTableId(namespace));
```
holic, line 84: are there any strategies we have available to keep this sort of thing from happening again? ideally we don't want to add runtime checks (for extra gas in hot path) and tests have a similar "manual maintenance" problem

yonadaaa: This is non-trivial because there is no "list" of all the tables that are registered by a module - the table registration is just inlined in some library. 

We sort of have to manually maintain this, but we can make it a bit more sticky by "bundling" it in with other tests. Like we can adopt a pattern of "whenever we test reading a table value, we check it exists first"

yonadaaa: Although in this case, we never test `TotalSupply.get()`, we only ever read it via `totalSupply()` 🤔

.changeset/light-tables-drive.md
yonadaaa: Added an example script!

packages/world-modules/src/modules/erc20-puppet/ERC20Module.sol
holic: are there other kinds of tests we can do? if each module had its own mud config, we could potentially have a test that reads the config and checks that each table has been registered during deploy?

.changeset/light-tables-drive.md
```diff
@@ -3,3 +3,30 @@
...
+
+import { Script } from "forge-std/Script.sol";
+import { ResourceId } from "@latticexyz/store/src/ResourceId.sol";
+import { StoreSwitch } from "@latticexyz/store/src/StoreSwitch.sol";
+
+import { TotalSupply } from "@latticexyz/world-modules/src/modules/erc20-puppet/tables/TotalSupply.sol";
+import { _totalSupplyTableId } from "@latticexyz/world-modules/src/modules/erc20-puppet/utils.sol";
+
+contract RegisterTotalSupply is Script {
+  function run(address worldAddress, ResourceId namespace) external {
```
holic, line 21: how do I call this from forge? how do I get the correct namespace resource ID?

packages/world-modules/src/modules/erc20-puppet/ERC20Module.sol
yonadaaa: We could have separate MUD configs for each config, although even that wouldn't cover every case. For example `KeysWithValue` uses `tableIdArgument` so you'd have to figure out the target table ID.

holic: just to clarify, we don't have to solve this in this PR but just wanna think about how we can prevent these sorts of things falling through the cracks in the future

alvrs: Maybe an end to end test that throws if a store event is retrieved but no corresponding table registration event? (Would only have captured this case if we used the ERC20 module in an end-to-end test though...)

holic: ohh! I think there's a way to track storage reads with forge-std, so we could have a general test to detect storage reads and ensure that the associated table is registered before reading

holic: added a follow up here: https://github.com/latticexyz/mud/issues/2883

.changeset/light-tables-drive.md
```diff
@@ -0,0 +1,38 @@
+---
+"@latticexyz/world-modules": patch
+---
+
+Fixed `ERC20Module` to register the `TotalSupply` table when creating a new token.
+
+You can register the table on existing Worlds with a Forge script:
```
holic, line 7:
```suggestion
If you've deployed a world with the `ERC20Module`, we recommend patching your world to register this table so that indexers can properly decode its record. You can do so with a simple Forge script:
```

holic (approved):

PR #2878, opened 2024-05-30
feat: restore typesVersions

holic:
closes https://github.com/latticexyz/mud/issues/2871

for backwards compatibility with `moduleResolution: "node"`
we had removed this in https://github.com/latticexyz/mud/pull/2828 but it broke projects when upgrading MUD verisons

(intentionally didn't include a changeset because this can be considered part of the changes in https://github.com/latticexyz/mud/pull/2828 and https://github.com/latticexyz/mud/pull/2873)

### DIFF
--- a/packages/gas-report/package.json
+++ b/packages/gas-report/package.json
@@ -12,14 +12,20 @@
   "exports": {
     ".": "./dist/index.js"
   },
+  "typesVersions": {
+    "*": {
+      "index": [
+        "./dist/index.d.ts"
+      ]
+    }
+  },
   "bin": {
     "gas-report": "./dist/gas-report.js"
   },
   "files": [
     "dist",
     "out",
-    "src",
-    "foundry.toml"
+    "src"
   ],
   "scripts": {
     "build": "tsup",
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "release:check": "changeset status --verbose --since=origin/main",
     "release:publish": "pnpm install && pnpm build && changeset publish",
     "release:version": "changeset version && pnpm install --lockfile-only && pnpm run changelog:generate",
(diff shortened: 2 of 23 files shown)


holic:
confirmed this works in templates by
- creating a project using mud latest (doesn't yet have these TS changes)
- upgrading mud to tag `main`
- checking that there are type errors
- upgrading mud to tag `restore-typeversions`
- checking that there are no type errors and the project can be built
```sh
pnpm create mud@latest test-project-types
cd test-project types
pnpm mud set-version --tag main && pnpm install
pnpm mud set-version --tag restore-typeversions && pnpm install
```

packages/gas-report/package.json
```diff
@@ -12,14 +12,20 @@
...
+    }
+  },
   "bin": {
     "gas-report": "./dist/gas-report.js"
   },
   "files": [
     "dist",
     "out",
-    "src",
-    "foundry.toml"
```
ssalbdivad, line 22: I think you re-added this in a previous PR right? There are several other packages where this was removed as well, so just wanted to make sure that was intended.

holic: yeah I had added one `foundry.toml` in https://github.com/latticexyz/mud/pull/2867 but can't recall if this was motivated by an error that appeared or if I just spotted it was missing and inconsistent with the rest.

but I don't think these are actually needed (they're similar to tsconfig.json)

holic: to be certain, I just made a fresh template with this version/PR (published to npm via snapshot action) and confirmed that a local anvil deploy rolled out fine! so I think the resulting app has all it needs inside the `out` dir and doesn't need `foundry.toml` in the npm package

ssalbdivad: Cool! I know I wanted to remove it originally, just wanted to make sure there wasn't some weird reason it needed to be kept.

alvrs (approved):

PR #2880, opened 2024-05-30
feat(world): enable config namespaces key

yonadaaa:


packages/world/ts/config/v2/namespaces.ts
```diff
@@ -23,7 +24,7 @@ export type validateNamespaces<namespaces, scope extends Scope = AbiTypeScope> =
   [namespace in keyof namespaces]: {
     [key in keyof namespaces[namespace]]: key extends "tables"
       ? validateTables<namespaces[namespace][key], scope>
-      : namespaces[namespace][key];
+      : ErrorMessage<`Expected full tables config.`>;
```
holic, line 27: This error isn't accurate. Maybe something like

```suggestion
      : ErrorMessage<`\`${key & string}\` is not a valid namespace config option.`>
```

holic: @ssalbdivad Do you happen to have a TS util to do something along the lines of `(keyof NamespacesInput).join(", ")`?

packages/world/ts/config/v2/world.ts
```diff
@@ -75,22 +74,40 @@ export function resolveWorld<const world extends WorldInput>(world: world): reso
...
   const namespaces = world.namespaces ?? {};
 
+  if (world.namespaces) {
+    if (!world.codegen) {
+      world.codegen = {};
+    }
+    if (world.codegen.namespaceDirectories === undefined) {
+      world.codegen = { ...world.codegen, namespaceDirectories: true };
+    }
+  }
```
holic, line 84: we shouldn't override the input object

packages/world/ts/config/v2/namespaces.ts
ssalbdivad: @holic Yep, there is a type in `@arktype` util called `stringifyUnion`!

examples/minimal/packages/contracts/mud.config.ts
```diff
@@ -37,6 +37,32 @@ export default defineWorld({
       key: ["owner", "item", "itemVariant"],
     },
   },
+  namespaces: {
+    royale: {
```
holic, line 41: this "minimal example" is deviating from being a minimal example

is there another place we can put this?

packages/world/ts/config/v2/world.ts
```diff
@@ -75,29 +74,42 @@ export function resolveWorld<const world extends WorldInput>(world: world): reso
   const scope = extendedScope(world);
   const namespaces = world.namespaces ?? {};
 
+  const resolvedStore = resolveStore(world);
+
+  const codegen = mergeIfUndefined(resolvedStore.codegen, resolveCodegen(world.codegen));
+  codegen.namespaceDirectories = world.namespaces !== undefined;
+  codegen.indexFilename = world.namespaces === undefined ? codegen.indexFilename : false;
```
holic, line 81: using both resolved store codegen and resolved world codegen is a bit weird since the resolvers will attempt to never leave something `undefined`, making it hard to set some defaults here

I think this is closer to accurate, where we set up some store codegen defaults if they're not already set:

```suggestion
  const resolvedStore = resolveStore({
    ...world,
    codegen: mergeIfUndefined(world.codegen ?? {}, {
      namespaceDirectories: world.namespaces != null,
      indexFilename: world.namespaces != null ? false : undefined,
    }),
  });
```

examples/minimal/packages/contracts/mud.config.ts
yonadaaa: Agreed - I'll make another example or maybe put it in some test

holic: we have `test/mock-game-contracts`, could add another `test/multiple-namespaces`

.changeset/violet-flowers-poke.md
```diff
@@ -0,0 +1,45 @@
+---
+"@latticexyz/store": patch
+"@latticexyz/world": patch
+---
+
+Enabled the `namespaces` key in the config. This enables declaring resources in different namespaces in a single MUD config.
+
+For example:
```
holic, line 8:
```suggestion
You can now configure resources across multiple namespaces in a single MUD config using the `namespaces` key.

Note that once you start using `namespaces` config option, you will need to move your existing tables, systems, etc. under the corresponding namespace within `namespaces`.
```

examples/multiple-namespaces/mud.config.ts
```diff
@@ -1,23 +1,53 @@
...
+        Score: {
+          schema: {
+            player: "address",
+            score: "uint256",
+          },
+          key: ["player"],
+        },
+      },
+    },
+    gameFork: {
```
holic, line 34: it's not immediately obvious to me what this example is supposed to communicate

I wonder if there's another example we can use here that is more self explanatory

packages/world/ts/config/v2/world.ts
```diff
@@ -29,8 +29,9 @@ export type validateWorld<world> = {
       : key extends "enums"
         ? narrow<world[key]>
         : key extends "namespaces"
-          ? // ? validateNamespaces<world[key], extendedScope<world>>
-            ErrorMessage<`Namespaces config will be enabled soon.`>
+          ? world extends { namespace?: unknown; tables?: unknown; systems?: unknown }
```
holic, line 32: would like to revisit this pattern soon and refactor some things so is easier to maintain and doesn't become a place for things to fall through cracks as we expand on what can be in a namespace

thinking something like

```ts
// in store
type NamespaceInput = { tables?: TablesInput };
type SingleNamespaceInput = { namespace?: string } & NamespaceInput;
type StoreInput = SingleNamespaceInput & { codegen?: CodegenInput, deploy?: DeployInput };

// in world
type NamespaceInput = StoreNamespaceInput & { systems?: SystemsInput };
type SingleNamespaceInput = { namespace?: string } & NamespaceInput;
type WorldInput = SingleNamespaceInput & { namespaces?: NamespacesInput, codegen?: CodegenInput, deploy?: DeployInput };
```

then this validation can be something like
```ts
world extends SingleNamespaceInput ? ErrorMessage<...> : validateNamespaces<...>
```

examples/multiple-namespaces/mud.config.ts
holic: oh whoops, I was confusing this with the changeset! I think this is fine as an example app, just wanted to make sure the example config in the changeset was intuitive/self-explanatory

.changeset/violet-flowers-poke.md
```diff
@@ -0,0 +1,45 @@
...
+          schema: {
+            player: "address",
+            x: "int32",
+            y: "int32",
+          },
+          key: ["player"],
+        },
+      },
+    },
+    somePlugin: {
```
holic, line 32: good to give folks a concrete example rather than "some" abstract one
```suggestion
    scoreboardPlugin: {
```

yonadaaa: that's the name i was looking for 🙇

packages/world/ts/config/v2/world.ts
```diff
@@ -75,22 +76,40 @@ export function resolveWorld<const world extends WorldInput>(world: world): reso
...
   const namespaces = world.namespaces ?? {};
 
+  const resolvedStore = resolveStore({
+    ...world,
+    codegen: mergeIfUndefined(world.codegen ?? {}, {
+      // put codegen into namespaced directories if we're using `namespaces` key
+      namespaceDirectories: world.namespaces != null,
+      // don't generate `index.sol` if we're using `namespaces` key, to avoid naming conflicts for table names across namespaces
+      // this is a deprecated option anyway, so this default guides folks towards not using it
+      indexFilename: world.namespaces != null ? false : undefined,
```
holic, line 86: this comment mentions that this is deprecated but I don't think we actually deprecated this yet, can we?

examples/multiple-namespaces/mud.config.ts
holic: > The hacker namespace demonstrates access control by attempting to write to a table in game

I think this might not be clear and the name itself could be misleading

Maybe we can use tokens as a better example? where we might have many of the same tables across namespaces

packages/world/ts/config/v2/input.ts
```diff
@@ -103,7 +107,16 @@ export type WorldInput = evaluate<
...
 
-export type NamespaceInput = Pick<StoreInput, "tables">;
+export type NamespaceInput = Pick<StoreInput, "tables"> & {
+  /**
+   * Contracts named *System will be deployed by default
+   * as public systems at `namespace/ContractName`, unless overridden
+   *
+   * The key is the system name (capitalized).
+   * The value is a SystemConfig object.
+   */
```
holic, line 117: this comment could use an update

packages/world/ts/config/v2/world.ts
yonadaaa: How do we mark something as deprecated?

Link to original discussion https://github.com/latticexyz/mud/pull/2880#discussion_r1622475589

holic:
```ts
// @deprecated
```
or
```ts
/**
 * @deprecated
 */
```

.changeset/violet-flowers-poke.md
```diff
@@ -0,0 +1,45 @@
+---
+"@latticexyz/store": patch
```
holic, line 2:
```suggestion
```
there's no real user-facing store changes, just conditionally defaulting `indexFilename` to `false` if `namespaces` is used, so would probably opt to exclude this changeset from this description

I could see leaving this here if store had `namespaces` support too, but we haven't migrated that yet

.changeset/violet-flowers-poke.md
yonadaaa: This namespace name is actually too long, it causes the project to fail in in unexpected ways

holic: we should fix that and/or add it to validation

packages/store/ts/config/v2/defaults.ts
```diff
@@ -5,8 +5,9 @@ export const CODEGEN_DEFAULTS = {
   userTypesFilename: "common.sol",
   outputDirectory: "codegen",
   // TODO: default to true if using top-level `namespaces` key (once its migrated to store)
-  namespaceDirectories: false,
-  indexFilename: "index.sol",
+  namespaceDirectories: false as boolean,
+  // @deprecated import table libraries from their individual files instead */
```
holic, line 9: this got added to defaults, not the input type

packages/store/ts/config/v2/output.ts
```diff
@@ -59,7 +59,8 @@ export type Codegen = {
    */
   // TODO: move `namespaces` key handling into store so we can conditionally turn this on/off
   readonly namespaceDirectories: boolean;
-  readonly indexFilename: string;
+  // @deprecated import table libraries from their individual files instead */
```
holic, line 62: should check that this strikesthrough `indexFilename` usage, I can't recall if tsdoc tags needed to be in a `/* */` style comment or can be in any comment

(if we leave the `//` style comment, should remove the `*/` at the end)

packages/store/ts/config/v2/defaults.ts
yonadaaa: Good spot, actually `indexFilename` isn't even on the input type

packages/store/ts/config/v2/output.ts
yonadaaa: Fixed now, it does need to be `/* */`!

.changeset/violet-flowers-poke.md
yonadaaa: Working on it in https://github.com/latticexyz/mud/pull/2897

examples/multiple-namespaces/mud.config.ts
yonadaaa: Just removed the `hacker__` example, we don't need to demonstrate access control here

packages/world/ts/config/v2/world.ts
holic: working on this here: https://github.com/latticexyz/mud/pull/2898

packages/world/ts/config/v2/input.ts
```diff
@@ -90,9 +90,13 @@ export type WorldInput = evaluate<
      * The value is a SystemConfig object.
      */
     systems?: SystemsInput;
-    /** System names to exclude from automatic deployment */
+    /**
+     * System names to exclude from automatic deployment
+     * @deprecated
+     */
+    // TODO: move this to systems.System.deploy.disabled
```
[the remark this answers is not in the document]
holic: working on this in https://github.com/latticexyz/mud/pull/2898

packages/world/ts/config/v2/input.ts
```diff
@@ -87,12 +87,16 @@ export type WorldInput = evaluate<
...
      *
      * The key is the system name (capitalized).
-     * The value is a SystemConfig object.
+     * The value is a SystemInput object.
      */
     systems?: SystemsInput;
-    /** System names to exclude from automatic deployment */
+    /**
+     * System names to exclude from automatic deployment
+     * @deprecated
```
holic, line 95:
```suggestion
```
let's not deprecate this yet - I am not sure the exact approach we're taking since most things in `SystemInput` are deploy related and unclear if its worth moving them all under a `deploy` key

packages/world/ts/config/v2/world.ts
```diff
@@ -28,8 +29,9 @@ export type validateWorld<world> = {
       : key extends "enums"
         ? narrow<world[key]>
         : key extends "namespaces"
-          ? // ? validateNamespaces<world[key], extendedScope<world>>
-            ErrorMessage<`Namespaces config will be enabled soon.`>
+          ? world extends { namespace?: unknown; tables?: unknown; systems?: unknown }
```
holic, line 32:
```suggestion
          // TODO: revisit this pattern (https://github.com/latticexyz/mud/pull/2898)
          ? world extends { namespace?: unknown; tables?: unknown; systems?: unknown }
```

packages/world/ts/config/v2/world.ts
```diff
@@ -28,8 +29,9 @@ export type validateWorld<world> = {
...
+            ? ErrorMessage<`\`Can only use \`namespaces\` with \`namespace\`, \`tables\`, or \`systems\` keys.`>
```
holic, line 33:
```suggestion
            ? ErrorMessage<"Can only use `namespaces` with `namespace`, `tables`, or `systems` keys.">
```

packages/world/ts/config/v2/world.ts
```diff
@@ -39,7 +41,10 @@ export function validateWorld(world: unknown): asserts world is WorldInput {
   const scope = extendedScope(world);
   validateStore(world);
 
-  if (hasOwnKey(world, "namespaces")) {
+  if (hasOwnKey(world, "namespaces") && isObject(world.namespaces)) {
+    if (hasOwnKey(world, "namespace")) {
+      throw new Error(`\`Can only use \`namespaces\` with \`namespace\`, \`tables\`, or \`systems\` keys.`);
```
holic, line 46:
```suggestion
      throw new Error("Cannot use `namespaces` with `namespace`, `tables`, or `systems` keys.");
```

packages/world/ts/config/v2/world.ts
```diff
@@ -39,7 +41,10 @@ export function validateWorld(world: unknown): asserts world is WorldInput {
   const scope = extendedScope(world);
   validateStore(world);
 
-  if (hasOwnKey(world, "namespaces")) {
+  if (hasOwnKey(world, "namespaces") && isObject(world.namespaces)) {
+    if (hasOwnKey(world, "namespace")) {
```
holic, line 45: we should check for `tables` and `systems` keys here too, otherwise runtime doesn't have the same behavior as types

packages/world/ts/config/v2/worldWithShorthands.ts
```diff
@@ -33,7 +34,9 @@ export type validateWorldWithShorthands<world> = {
   [key in keyof world]: key extends "tables"
     ? validateTablesWithShorthands<world[key], extendedScope<world>>
     : key extends "namespaces"
-      ? validateNamespacesWithShorthands<world[key], extendedScope<world>>
+      ? world extends { namespace?: unknown; tables?: unknown; systems?: unknown }
+        ? ErrorMessage<`\`Can only use \`namespaces\` with \`namespace\`, \`tables\`, or \`systems\` keys.`>
```
holic, line 38:
```suggestion
        ? ErrorMessage<"Cannot use `namespaces` with `namespace`, `tables`, or `systems` keys.">
```
also wondering if there's a way to not duplicate this behavior here

packages/world/ts/config/v2/worldWithShorthands.ts
```diff
@@ -65,12 +68,14 @@ export function resolveWorldWithShorthands<world extends WorldWithShorthandsInpu
...
     return isTableShorthandInput(table) ? resolveTableShorthand(table, scope) : table;
   });
-  const namespaces = mapObject(world.namespaces ?? {}, (namespace) => ({
-    ...namespace,
-    tables: mapObject(namespace.tables ?? {}, (table) => {
-      return isTableShorthandInput(table) ? resolveTableShorthand(table, scope) : table;
-    }),
-  }));
+  const namespaces = world.namespaces
+    ? mapObject(world.namespaces, (namespace) => ({
```
holic, line 72: can we do this same conditional for `tables` to avoid setting it?

packages/world/ts/config/v2/input.ts
```diff
@@ -87,12 +87,16 @@ export type WorldInput = evaluate<
...
+     */
+    // TODO: move this to systems.System.deploy.disabled
```
holic, line 97:
```suggestion
```

packages/world/ts/config/v2/worldWithShorthands.ts
```diff
@@ -33,7 +34,9 @@ export type validateWorldWithShorthands<world> = {
   [key in keyof world]: key extends "tables"
     ? validateTablesWithShorthands<world[key], extendedScope<world>>
     : key extends "namespaces"
-      ? validateNamespacesWithShorthands<world[key], extendedScope<world>>
+      ? world extends { namespace?: unknown; tables?: unknown; systems?: unknown }
```
holic, line 37:
```suggestion
      // TODO: revisit this pattern (https://github.com/latticexyz/mud/pull/2898)
      ? world extends { namespace?: unknown; tables?: unknown; systems?: unknown }
```

packages/world/ts/config/v2/world.ts
holic: I think we also need this check in `validateWorldWithShorthands`

packages/world/ts/config/v2/world.test.ts
```diff
@@ -177,27 +182,28 @@ describe("defineWorld", () => {
...
-    attest<"AnotherOne">(config.tables.AnotherOne__Example.namespace).equals("AnotherOne");
-    attest(config.tables.AnotherOne__Example.tableId).equals(
-      resourceToHex({ type: "table", name: "Example", namespace: "AnotherOne" }),
-    );
+      });
+    })
+      .throws("Can only use `namespaces` with `namespace`, `tables`, or `systems` keys.")
+      .type.errors(
+        // eslint-disable-next-line max-len
+        `Type '{ AnotherOne: { tables: { Example: { schema: { id: string; name: string; age: string; }; key: string[]; }; }; }; }' is not assignable to type '"\`Can only use \`namespaces\` with \`namespace\`, \`tables\`, or \`systems\` keys.`,
```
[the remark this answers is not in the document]
yonadaaa: It will match if we only include part. Why would you not as specific as possible though?

packages/world/ts/config/v2/worldWithShorthands.ts
yonadaaa: Yes, and we need to or else the later `hasOwnKey(world, "table")` check will always pass!

packages/world/ts/config/v2/world.test.ts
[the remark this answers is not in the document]
holic: we're only testing if the error we expect is there

if we change the shape of the object its using, then we have to update the error message too, which is a little annoying

holic:
replaced by https://github.com/latticexyz/mud/pull/2968

PR #2888, opened 2024-06-05
refactor(store): remove V1 `codegenIndexFilename`

yonadaaa:
This usages of this property can be replaced by the V2 `config.codegen.indexFilename`.

Pulled out of https://github.com/latticexyz/mud/pull/2880


holic (approved):
nice nice!

PR #2891, opened 2024-06-05
chore: add example for multiple namespaces

yonadaaa:
pulled from https://github.com/latticexyz/mud/pull/2880


examples/multiple-namespaces/packages/contracts/package.json
holic, line 1: since this is only the contracts package, we probably don't need the workspace - this whole example dir could just be the contracts dir

yonadaaa: agreed!

examples/multiple-namespaces/pnpm-workspace.yaml
holic, line 1: don't need this if we're not using a workspace anymore

examples/multiple-namespaces/package.json
holic, line 1: might be nice to add to this list (I see we haven't been doing this very well)
https://github.com/latticexyz/mud/blob/main/.github/workflows/examples.yml#L24

holic (approved):

PR #2893, opened 2024-06-05
fix: use legacy tx for postdeploy on garnet/redstone

transmissions11:
Manually patches postdeploy issue with Garnet/Redstone (https://github.com/foundry-rs/foundry/issues/7806) by using `--legacy` for `forgeScriptOptions` if chainid matches.


holic:
closed for now as https://github.com/foundry-rs/foundry/issues/7806 looks close to fixed and we have `--forgeScriptOptions="--legacy"` as a workaround for now

PR #2894, opened 2024-06-05
refactor(world): add validateNamespace helper

yonadaaa:


packages/world/ts/config/v2/namespaces.ts
```diff
@@ -19,12 +20,16 @@ export type namespacedTableKeys<world> = world extends { namespaces: infer names
...
+      : ErrorMessage<`\`${key & string}\` is not a valid Namespace config option.`>;
+};
+
 export type validateNamespaces<namespaces, scope extends Scope = AbiTypeScope> = {
-  [namespace in keyof namespaces]: {
-    [key in keyof namespaces[namespace]]: key extends "tables"
-      ? validateTables<namespaces[namespace][key], scope>
-      : namespaces[namespace][key];
-  };
+  [namespace in keyof namespaces]: validateNamespace<namespace, scope>;
```
holic, line 32:
```suggestion
  [namespace in keyof namespaces]: validateNamespace<namespaces[namespace], scope>;
```

packages/world/ts/config/v2/namespaces.ts
```diff
@@ -19,12 +20,16 @@ export type namespacedTableKeys<world> = world extends { namespaces: infer names
     }[keyof namespaces]
   : never;
 
+export type validateNamespace<namespace, scope extends Scope = AbiTypeScope> = {
+  readonly [key in keyof namespace]: key extends "tables"
+    ? validateTables<namespace[key], scope>
+    : key extends keyof NamespaceInput
+      ? conform<namespace[key], NamespaceInput[key]>
+      : ErrorMessage<`\`${key & string}\` is not a valid Namespace config option.`>;
```
holic, line 28:
```suggestion
      : ErrorMessage<`\`${key & string}\` is not a valid namespace config option.`>;
```

packages/world/ts/config/v2/worldWithShorthands.ts
```diff
@@ -54,7 +54,7 @@ export type validateNamespacesWithShorthands<namespaces, scope extends Scope = A
   [namespace in keyof namespaces]: {
     [key in keyof namespaces[namespace]]: key extends "tables"
       ? validateTablesWithShorthands<namespaces[namespace][key], scope>
-      : validateNamespaces<namespaces[namespace], scope>[key];
+      : validateNamespace<namespaces[namespace], scope>[key];
```
holic, line 57: might be nice to add some regression tests to make sure this doesn't slip through the cracks again!

packages/world/ts/config/v2/namespaces.ts
```diff
@@ -19,12 +20,16 @@ export type namespacedTableKeys<world> = world extends { namespaces: infer names
     }[keyof namespaces]
   : never;
 
+export type validateNamespace<namespace, scope extends Scope = AbiTypeScope> = {
```
holic, line 23: we ought to have the corresponding `validateNamespace` runtime function as well

yonadaaa: yess that would be neater

packages/world/ts/config/v2/worldWithShorthands.ts
yonadaaa: Agreed!

yonadaaa: Agreed!

Would we just test that each namespace in a shorthand config is valid?

holic: I think a type+runtime test for a valid+invalid namespace should be good

yonadaaa: Only thing is we can't test until `namespaces` is enabled in https://github.com/latticexyz/mud/pull/2880

holic: why? there's already tests for namespaces in `worldWithShorthands.test.ts`

yonadaaa: Doesn't that mean that the `namespaces` key is enabled on `defineWorldWithShorthands`? That feels incorrect

holic, line 57: need to remember to add this to `validateWorld` too

holic (approved):

packages/world/ts/config/v2/worldWithShorthands.ts
yonadaaa: will do that in [yonadaaa/fix-tables-without-namespace](https://github.com/latticexyz/mud/pull/2880)

PR #2895, opened 2024-06-06
refactor: clean up json imports

holic:


holic:
outdated and doesn't really solve the issue, have some other promising ideas coming in #3029

PR #2897, opened 2024-06-06
fix(world): truncate system names and namespaces

yonadaaa:


packages/world/ts/config/resolveWorldConfig.ts
```diff
@@ -77,8 +78,8 @@ export function resolveSystemConfig({
   // If the namespace is not set in the system name, default to the config namespace
   const parts = systemName.split("__");
   const namespaceIsSet = parts.length === 2;
-  const namespace = namespaceIsSet ? parts[0] : configNamespace;
-  const name = namespaceIsSet ? parts[1] : systemName;
+  const namespace = (namespaceIsSet ? parts[0] : configNamespace).slice(0, NAMESPACE_MAX_LENGTH);
```
yonadaaa, line 81: Would `config.namespace` already be truncated here?

PR #2899, opened 2024-06-06
docs(guides/modules): how to write a module

qbzzt:


alvrs (commented):
This looks correct, but I wonder if there is a simpler module we could use as an example? Maybe it's because i'm tired, but I had difficulties wrapping my head around what's going on in this module even though I already understand MUD modules, so I'm concerned someone reading this who isn't familiar with modules would be scared by the complexity of _this_ module and think modules in general are very complex (when they are really just a simple deployed installation script).

qbzzt:
@alvrs , changed to a much simpler example.

alvrs (approved):

PR #2901, opened 2024-06-13
Fix(store-sync): Handle unsynced RPC nodes

dhvanipa:
**Context:** https://indexsupply.com/shovel/docs/#unsynchronized-ethereum-nodes
   - `3rd party Ethereum API providers may load balance RPC requests across a set of unsynchronized nodes. In rare cases, the eth_getLogs request will be routed to a node that doesn’t have the latest block.`
   - This issue has been noticed multiple times on Redstone

**Solution:** Batch fetch the latest block number with the logs, and ensure that the latest block number is past the `toBlock`. Otherwise, wait for the RPC to catch up to the latest block.
   - This fix relies on the client making a batch call to the RPC, which needs to be enabled by clients in their [Viem HTTP transport](https://viem.sh/docs/clients/transports/http.html#batch-json-rpc).

For the sync stack, the `fromBlock` is decided from a local variable and the `toBlock` is decided by a [subscription](https://viem.sh/docs/actions/public/watchBlocks) to the latest block number. So we need an additional filter from the latest block number subscription to not emit when the `toBlock` is behind the `fromBlock`, which happens with unsynchronized RPCs.
  - There is currently silent failure when this happens as when `toBlock` is less than `fromBlock`, no error is thrown, and no logs are fetched either, as [this while loop](https://github.com/latticexyz/mud/blob/main/packages/block-logs-stream/src/fetchLogs.ts#L89) will resolve false. So in this PR, we add a filter to not emit incorrect `toBlock`'s and also a safety error check to throw if this case occurs.
      - Because the batch call actually seems like only a fix for HTTP, I think this is the real issue that's causing incorrect logs with the sync-stack (since the default is to use WS). This is in-line with what we're seeing on the logs of the Biomes indexer pod, where it doesn't show there are 0 logs, but that the storageAdapter is never called.

### DIFF
--- a/packages/block-logs-stream/src/fetchLogs.ts
+++ b/packages/block-logs-stream/src/fetchLogs.ts
@@ -90,7 +90,18 @@ export async function* fetchLogs<TAbiEvents extends readonly AbiEvent[]>({
     try {
       const toBlock = fromBlock + blockRange;
       debug("getting logs", { fromBlock, toBlock });
-      const logs = await publicClient.getLogs({ ...getLogsOpts, fromBlock, toBlock, strict: true });
+
+      const [latestBlockNumber, logs] = await Promise.all([
+        publicClient.getBlockNumber({ cacheTime: 0 }),
+        publicClient.getLogs({ ...getLogsOpts, fromBlock, toBlock, strict: true }),
+      ]);
+      if (latestBlockNumber < toBlock) {
+        const blockTimeInSeconds = 2;
+        const seconds = Number(toBlock - latestBlockNumber) * blockTimeInSeconds;
+        debug(`latest block number ${latestBlockNumber} is less than toBlock ${toBlock}, retrying in ${seconds}s`);
+        await wait(1000 * seconds);
+        continue;
+      }
       yield { fromBlock, toBlock, logs };
       fromBlock = toBlock + 1n;
       blockRange = bigIntMin(maxBlockRange, getLogsOpts.toBlock - fromBlock);
--- a/packages/store-sync/src/createStoreSync.ts
+++ b/packages/store-sync/src/createStoreSync.ts
@@ -187,18 +187,21 @@ export async function createStoreSync<config extends StoreConfig = StoreConfig>(
     tap((startBlock) => debug("starting sync from block", startBlock)),
   );
 
+  let startBlock: bigint | null = null;
+  let endBlock: bigint | null = null;
+  let lastBlockNumberProcessed: bigint | null = null;
+
   const latestBlock$ = createBlockStream({ publicClient, blockTag: followBlockTag }).pipe(shareReplay(1));
   const latestBlockNumber$ = latestBlock$.pipe(
     map((block) => block.number),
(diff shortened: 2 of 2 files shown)


packages/store-sync/src/createStoreSync.ts
```diff
@@ -187,35 +187,43 @@ export async function createStoreSync<config extends StoreConfig = StoreConfig>(
...
+  let lastBlockNumberProcessed: bigint | null = null;
+
   const latestBlock$ = createBlockStream({ publicClient, blockTag: followBlockTag }).pipe(shareReplay(1));
   const latestBlockNumber$ = latestBlock$.pipe(
     map((block) => block.number),
     tap((blockNumber) => {
       debug("on block number", blockNumber, "for", followBlockTag, "block tag");
     }),
+    filter((blockNumber) => {
+      return lastBlockNumberProcessed == null || blockNumber > lastBlockNumberProcessed;
```
dhvanipa, line 201: unsure if we want to add a debug log here

packages/block-logs-stream/src/fetchLogs.ts
```diff
@@ -90,7 +90,18 @@ export async function* fetchLogs<TAbiEvents extends readonly AbiEvent[]>({
     try {
       const toBlock = fromBlock + blockRange;
       debug("getting logs", { fromBlock, toBlock });
-      const logs = await publicClient.getLogs({ ...getLogsOpts, fromBlock, toBlock, strict: true });
+
+      const [latestBlockNumber, logs] = await Promise.all([
```
dhvanipa, line 94: This only works if `batch` is enabled in the HTTP transport. This is decided by clients though, so maybe this is put into docs?

packages/block-logs-stream/src/fetchLogs.ts
```diff
@@ -90,7 +90,18 @@ export async function* fetchLogs<TAbiEvents extends readonly AbiEvent[]>({
...
+        publicClient.getBlockNumber({ cacheTime: 0 }),
+        publicClient.getLogs({ ...getLogsOpts, fromBlock, toBlock, strict: true }),
+      ]);
+      if (latestBlockNumber < toBlock) {
+        const blockTimeInSeconds = 2;
```
dhvanipa, line 99: Not sure what the best way to handle this is, ie should it be a deploy config? I dont think there's a way to detect from the chosen chain what the block time is?

packages/block-logs-stream/src/fetchLogs.ts
holic: I wonder if we should have an option to enable the "validate unsynchronized nodes" behavior, where we require `batch` to be enabled or, ideally, do the batch call here directly rather than relying on some client configuration.

Since this approach isn't useful unless the call is atomic (going to the same RPC) and I suspect not all RPC providers support batching, seems like we should put this behind a flag to avoid doubling RPC calls for syncing data from RPC. Could make this the default approach with a way to opt out of this new behavior.

PR #2913, opened 2024-06-26
docs(guide/hello-world/browser-wallet): initial version

qbzzt:


docs/pages/guides/hello-world/browser-wallet.mdx
```diff
@@ -0,0 +1,224 @@
...
+});
+```
+
+Request the [address list](https://viem.sh/docs/accounts/jsonRpc) from the browser extension (confusingly, also called "wallet").
+This request prompts the extension to ask for user permission to connect to the application.
+
+```ts
+const extensionWalletClient = createWalletClient({
+  ...clientOptions,
+  transport: custom(window.ethereum!),
```
holic, line 181: I don't think this is a pattern we should support. A complete browser wallet solution will come with Account Kit (name TBD).

qbzzt: I think we should explain all the available options (Account Kit when it's available) and let developers decide what's best for them. 

@alvrs , can you tie break and decide?

docs/pages/guides/hello-world/add-chain-client.mdx
alvrs, line 1: Can we split out this guide to a separate PR? I think we can merge this one first and the wallet one later.

docs/pages/guides/hello-world/add-chain-client.mdx
```diff
@@ -0,0 +1,93 @@
...
+    default:
+      console.log("No URL type I can use for the chain");
+  }
+
+  const clientOptions = {
+    chain: networkConfig.chain,
+    transport: transportObserver(fallback(fallbacks)),
+    pollingInterval: 1000,
+  } as const satisfies ClientConfig;
+  ```
```
alvrs, line 64: I would argue this is not something we have to explain how to make dynamic here. We can simplify to mentioning that depending on the chain you want to use you might have to adjust your `transport` client option for viem. Anything beyond that falls in the user's responsibility to customize / make dynamic as they wish.

docs/pages/guides/hello-world/browser-wallet.mdx
alvrs: We're gonna prioritize shipping a usable (and documentable) version of the account kit libraries very soon so I would agree with @holic here that we shouldn't merge this one yet so we don't send folks off a path they have to walk back later. (Note that the account kit libraries don't require you to use all of account kit, but also just make setting up an external wallet and delegations to a local signer much easier.)

If we extract this change to its own PR you can still send the WIP to folks asking about external wallets in the discord, with the disclaimer that a much better solution is in the works and ready soon.

docs/pages/guides/hello-world/add-chain-client.mdx
qbzzt: Done. This is the now just the chain PR. Browser wallet coming in a separate PR.

docs/pages/guides/hello-world/add-chain-client.mdx
qbzzt: Removed, replaced with a much simpler note.

docs/pages/guides/hello-world/browser-wallet.mdx
qbzzt: Removed here, added as https://github.com/latticexyz/mud/pull/2924

alvrs (approved):

PR #2917, opened 2024-07-02
feat(common): throw instead of truncating namespace

### LINKED ISSUE #2905: throw instead of truncating namespaces
example:
https://github.com/latticexyz/mud/blob/e0c9eaaf2990ea031ee3ba98b8560fcfec82889d/packages/common/src/resourceToHex.ts#L20
(there might me more)

namespaces are tied to access control, so it's probably not safe to automatically truncate these, thus changing access control policies, and instead we should just require a <=14 char namespace and throw otherwise

holic:
fixes #2905

### DIFF
--- a/.changeset/tender-badgers-ring.md
+++ b/.changeset/tender-badgers-ring.md
@@ -2,4 +2,4 @@
 "@latticexyz/common": patch
 ---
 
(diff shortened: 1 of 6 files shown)


alvrs (approved):

qbzzt:
Great, thank you!

PR #2922, opened 2024-07-03
chore: bump glob

holic:
encountered some issues with older versions when working on https://github.com/latticexyz/mud/pull/2880

had to add `.sort()` everywhere because new versions of glob no longer automatically sort output and instead lean on natural ordering from fs (which is unfortunately non-deterministic)


holic:
> ooc what where the issues you encountered with the older version?

```
Error: Dynamic require of "fs" is not supported
```

PR #2923, opened 2024-07-03
feat: initial system manifest

holic:


packages/store/ts/config/v2/input.ts
```diff
@@ -43,6 +43,12 @@ export type StoreInput = {
    * Defaults to `src` to match `foundry.toml`'s default. If you change this from the default, you may also need to configure foundry with the same source directory.
    */
   readonly sourceDirectory?: string;
+  /**
+   * Directory of MUD metadata generated during build (e.g. systems manifest) and deploy steps (e.g. system and module addresses). This directory is relative to the MUD config.
+   *
+   * Defaults to `.mud`.
+   */
+  readonly metadataDirectory?: string;
```
holic, line 51: I was by default making this configurable, but now reconsidering since it makes it much more complicated with static imports.

packages/world/ts/config/v2/worldWithShorthands.ts
```diff
@@ -62,17 +62,32 @@ export function resolveWorldWithShorthands<world extends WorldWithShorthandsInpu
...
+  const fullConfig = {
+    ...world,
+    ...(world.tables
+      ? {
+          tables: mapObject(world.tables, (table) =>
+            isTableShorthandInput(table) ? resolveTableShorthand(table, scope) : table,
+          ),
+        }
+      : null),
+    ...(world.namespaces
```
holic, line 75: The previous `resolveWorldWithShorthands` code was always setting the `namespaces` key, which kept the downstream `resolveWorld` function from determining if we're using namespaces or not.

holic:
see https://github.com/latticexyz/mud/pull/3050

PR #2926, opened 2024-07-04
refactor: build step cleanup

holic:
pulled out of #2923

- changed `configPath` to `rootDir` everywhere, because we almost always want the project root dir and want to avoid using `path.dirname` in all the downstream tooling
- moved store/world build steps into a single build script so they can reuse vars etc. and map more closely to the `mud build` command
- renamed `all-codegen` to `all-build` because that's actually what it's doing


packages/world/ts/node/findSolidityFiles.ts
```diff
@@ -0,0 +1,14 @@
+import path from "node:path";
+import { glob } from "glob";
+import { World } from "../config/v2/output";
+
+export async function findSolidityFiles({ rootDir, config }: { rootDir: string; config: World }) {
```
holic, line 5: may need to exclude `.t.sol` and `.s.sol`?

holic: ah likely because we're all good about putting source in `src` but technically possible to have test files, etc. live in this dir as well (and I personally like colocating tests with the files they're testing)

holic:
this doesn't change anything about the CLI commands, just the arg name/intent of the internal functions, and shouldn't impact users unless someone imported `tablegen` from `@latticexyz/store/codegen` or `worldgen` from `@latticexyz/world/node` (which feels less intentional)

PR #2928, opened 2024-07-05
refactor(store,world): rename TS keys of resource labels

holic:
just a small internal naming change for key names used in TS mappings to align with the terminology we've started using

### DIFF
--- a/packages/world/ts/config/v2/input.ts
+++ b/packages/world/ts/config/v2/input.ts
@@ -103,7 +103,7 @@ export type WorldInput = evaluate<
 
 export type NamespacesInput = { [label: string]: NamespaceInput };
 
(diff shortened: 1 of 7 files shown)


packages/world/ts/config/v2/input.ts
```diff
@@ -101,9 +101,9 @@ export type WorldInput = evaluate<
   }
 >;
 
-export type NamespacesInput = { [key: string]: NamespaceInput };
+export type NamespacesInput = { [label: string]: NamespaceInput };
 
-export type NamespaceInput = Pick<StoreInput, "tables">;
+export type NamespaceInput = Pick<StoreInput, "namespace" | "tables">;
```
holic, line 106:
```suggestion
export type NamespaceInput = Pick<StoreInput, "tables">;
```

PR #2929, opened 2024-07-05
WIP namespaces

holic:
TODO
- [x] restore flattened, namespaced to `config.tables` output
  - [ ] figure out how to deprecate keys in a mapped type
- [ ] output namespaces for single-namespace config
- [x] enforce/validate one mode (single namespace or multi namespace)
- [ ] remove strong types for truncated strings (namespace for namespaces, name for tables)


packages/store/ts/config/v2/store.test.ts
```diff
@@ -460,9 +468,9 @@ describe("defineStore", () => {
     });
 
     attest<"namespace">(config.namespace).equals("namespace");
-    attest<"namespace">(config.tables.namespace__Example.namespace).equals("namespace");
-    attest<"Example">(config.tables.namespace__Example.name).equals("Example");
-    attest(config.tables.namespace__Example.tableId).equals(
+    attest<"namespace">(config.tables.Example.namespace).equals("namespace");
+    attest<"Example">(config.tables.Example.name).equals("Example");
+    attest(config.tables.Example.tableId).equals(
```
holic, line 473: We should probably also include the namespaced variant of these for backwards compatibility, but would be considered deprecated usage.

holic:
replaced by https://github.com/latticexyz/mud/pull/2968

Issue #2934, opened 2024-07-11
enable type checking in template tests


holic:
recently some spotted some TS issues that we didn't catch before

holic:
not actually an issue, just misunderstood how some things were working

PR #2823, opened 2024-05-14
feat(create-mud): use app namespace

qbzzt:
Replacing #2816 , #2817 , #2818 , #2819 , #2821

PR #2935, opened 2024-07-11
fix(create-mud): update component references

holic:
These broke in https://github.com/latticexyz/mud/pull/2823 but weren't caught by TS (see #2934)


holic:
not actually broken in main, since the config v2 -> v1 compat step reverts the table names and removes the namespace prefix

PR #2937, opened 2024-07-11
chore: remove unused store-sync/query-cache

holic:
goes with #2539 and #2936

we're not using the query API yet and I don't wanna burn time refactoring these as we update the config, so will remove for now


holic:
gonna yolo merge since nothing depends on these files, just cleaning up

PR #2939, opened 2024-07-12
fix(world/worldgen): remove \r to support windows 🚗 

qbzzt:
Need shown in https://discord.com/channels/865335009915961364/1261009544004501574


holic:
I wanna think about the consequences of this more before we do this.

I would be happier if we throw if we detect a `\r` somewhere and suggest users save files with Unix line endings, as there's probably a lot more than just system source files that would be impacted by this.

holic:
closing this out for now as it feels safer and more correct to suggest folks use unix line endings

PR #2941, opened 2024-07-12
refactor(cli): move off of old config

holic:


packages/cli/src/commands/worldgen.ts
```diff
@@ -1,17 +1,12 @@
...
-import { getSrcDirectory } from "@latticexyz/common/foundry";
-import path from "path";
-import { rmSync } from "fs";
-import { getExistingContracts } from "../utils/getExistingContracts";
 
 type Options = {
   configPath?: string;
   clean?: boolean;
-  srcDir?: string;
-  config?: WorldConfig;
```
holic, line 14: I don't think this option ever worked - it wasn't part of `builder` method and isn't possible to pass in an object to CLI anyway.

alvrs (approved):

PR #2943, opened 2024-07-15
fix(store-sync): add NoInfer to syncToRecs options

alvrs:
No strong types are needed for this part of the config and @ssalbdivad spotted that the part of the config is a major typescript performance bottleneck.

### DIFF
--- a/.changeset/four-peas-swim.md
+++ b/.changeset/four-peas-swim.md
@@ -1,0 +1,5 @@
+---
+"@latticexyz/store-sync": patch
+---
+
+TypeScript performance improvements.
--- a/.changeset/four-peas-swim.md
+++ b/.changeset/four-peas-swim.md
@@ -2,4 +2,4 @@
 "@latticexyz/store-sync": patch
 ---
 
-TypeScript performance improvements.
+Added `NoInfer` to a part of `SyncToRecsOptions` to improve TypeScript performance.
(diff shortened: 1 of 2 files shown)


holic (approved):
Not blocking but do we have or should we have some type tests to make sure this doesn't affect strongly typed components?

ssalbdivad:
@alvrs To be clear `NoInfer` isn't necessarily less safe it's about prioritizing where TS looks when inferring a variable type. My guess is this type is not written as intended in the first place as a structure where the `config` object both exists as a key and all of its keys exist at the top level of the object is probably not realistic.

It would definitely be valuable to have more tests so we know what might be relying on behavior like this though.

alvrs:
> a structure where the config object both exists as a key and all of its keys exist at the top level of the object is probably not realistic

`SyncOptions` includes `config` (`StoreConfig`) as a subkey! Here we're adding all general `SyncOptions` to `SyncToRecsOptions`. The reason we're doing `Omit<..., "config"> & { config }` is because `config` is an optional key in `SyncOptions`, but a required key in `SyncToRecsOptions` (maybe there is a more elegant way to perform this modification?)

https://github.com/latticexyz/mud/blob/706986b7ababbec819b02ca40a02b14594a2f656/packages/store-sync/src/common.ts#L66-L70

PR #2944, opened 2024-07-15
WIP: state lib benchmarks

alvrs:


packages/zustand-query/src/store.test.ts
```diff
@@ -0,0 +1,155 @@
...
+import { Table } from "@latticexyz/config";
+import { createStore as createZustandStore } from "zustand/vanilla";
+import { StoreApi } from "zustand";
+import { mutative } from "zustand-mutative";
+import { dynamicAbiTypeToDefaultValue, staticAbiTypeToDefaultValue } from "@latticexyz/schema-type/internal";
+
+/**
+ * A Record is one row of the table. It includes both the key and the value.
+ */
+type Record = { [field: string]: number | string | bigint | number[] | string[] | bigint[] };
```
holic, line 11: also `boolean` and `boolean[]`

and beware that `Record` conflicts with the TS native `Record`

I think I've used `Row` or `TableRecord` in other areas

packages/store-sync/src/dozer/fetchRecordsDozerSql.ts
```diff
@@ -0,0 +1,69 @@
...
+    })
+  ).json();
+
+  if (isDozerResponseFail(response)) {
+    console.warn(`Dozer response: ${response.msg}\n\nTry reproducing via cURL:
+    curl ${dozerUrl} \\
+    --compressed \\
+    -H 'Accept-Encoding: gzip' \\
+    -H 'Content-Type: application/json' \\
+    -d '[${queries.map((query) => `{"address": "${storeAddress}", "query": "${query.sql.replaceAll('"', '\\"')}"}`).join(",")}]'`);
```
holic, line 56: what about `JSON.stringify` here?
```suggestion
    -d '${JSON.stringify(queries.map((query) => ({ address: storeAddress, query: query.sql })))}.replaceAll("'", "\\'")'`);
```

packages/store-sync/src/zustand-query/createStorageAdapter.ts
```diff
@@ -0,0 +1,147 @@
...
+        });
+      }
+    }
+
+    // Convert logs in (partial) table updates
+    for (const log of logs) {
+      const { namespace, name } = hexToResource(log.args.tableId);
+      // TODO: use label once available
+      const tableConfig = store.getState().config[namespace][name];
+      const boundTable = store.getState().actions.getTable({ namespace, label: name });
```
holic, line 56: do we need to use bound tables here if we have the table config? would it be better to use the regular/non-bound API?

packages/zustand-query/src/recs/index.ts
holic, line 1: nice!

packages/zustand-query/src/createStore/getTable.ts
```diff
@@ -0,0 +1,59 @@
...
+  subscribe: (args: BoundSubscribeArgs) => SubscribeResult;
+};
+
+export type GetTableArgs = TableLabel;
+
+export type GetTableResult = BoundTable;
+
+export const getTable =
+  (context: Context): ((args: GetTableArgs) => GetTableResult) =>
+  (tableLabel) => {
```
holic, line 41: this is hard to read

I wonder of it would be easier as functions?

```
export function getTable(context: Context): GetTableResult {
  return function bindTable(...) {
    return ...
  }
}
```

packages/zustand-query/src/createStore/getTables.ts
```diff
@@ -0,0 +1,26 @@
...
+  [namespace: string]: {
+    [table: string]: BoundTable;
+  };
+};
+
+export type GetTablesResult = BoundTables;
+
+export const getTables =
+  (context: Context): (() => GetTablesResult) =>
+  (): BoundTables => {
```
holic, line 14: same as above

packages/zustand-query/src/createStore/registerTable.ts
```diff
@@ -0,0 +1,32 @@
...
+import { Context } from "./common";
+import { getTable, BoundTable } from "./getTable";
+
+export type RegisterTableArgs = TableInput;
+
+export type RegisterTableResult = BoundTable;
+
+export const registerTable =
+  (context: Context): ((args: RegisterTableArgs) => RegisterTableResult) =>
+  (tableInput) => {
```
holic, line 11: same as above

packages/store/ts/config/v2/table.ts
```diff
@@ -159,13 +162,15 @@ export function resolveTable<input extends TableInput, scope extends Scope = Abi
   const label = input.label;
   const type = input.type ?? TABLE_DEFAULTS.type;
   const namespace = input.namespace ?? TABLE_DEFAULTS.namespace;
+  const namespaceLabel = input.namespace ?? TABLE_DEFAULTS.namespace;
```
holic, line 165:
```suggestion
  const namespaceLabel = input.namespaceLabel ?? TABLE_DEFAULTS.namespace;
```

holic: we might want to require `namespaceLabel` in this context, otherwise its kind of unclear if the namespace label should inherit the input namespace or the default namespace

PR #2951, opened 2024-07-16
docs(introduction): add video to the introduction

qbzzt:


docs/pages/introduction.mdx
```diff
@@ -1,5 +1,16 @@
...
+<iframe
+  width="560"
+  height="315"
+  src="https://www.youtube.com/embed/Unfj6yN9wmA?si=0bfRVGFb2V_CDJ-z"
+  title="YouTube video player"
+  frameborder="0"
+  allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture; web-share"
+  referrerpolicy="strict-origin-when-cross-origin"
+  allowfullscreen
+></iframe>
```
holic, line 12: Can we make this look better? I feel like this disrupts the flow of the page

image

alvrs:
wonder if we should rather turn this into small video section (with thumbnails of each video) with a couple useful introductory videos. Two more i'd include:
- https://www.youtube.com/watch?v=UvIWmzscWp8
- https://www.youtube.com/watch?v=pla6wfWN7tE

ludns:
Some of these videos are out of date (eg: config format is different). Is that a big deal?

qbzzt:
> Some of these videos are out of date (eg: config format is different). Is that a big deal?

I don't think so because we state the date they were taken, and it's clear in a fast-moving industry that they might be inaccurate.

holic:
I feel pretty strongly we shouldn't clutter the Quickstart guide with these - they're not actually quick start videos.

And I agree that them being out of date doesn't help either. They would be more misleading and confusing than helpful in this context.

Can we make a separate videos section? Maybe as part of "Guides"?

PR #2955, opened 2024-07-17
feat(world): system labels

holic:


packages/world/ts/config/v2/system.ts
```diff
@@ -0,0 +1,82 @@
...
+>;
+
+export type validateSystem<input, options extends ValidateSystemOptions = {}> = {
+  [key in keyof input | requiredSystemKey<options["inNamespace"]>]: key extends keyof SystemInput
+    ? key extends "label" | "namespace"
+      ? options["inNamespace"] extends true
+        ? ErrorMessage<"Overrides of `label` and `namespace` are not allowed for systems in this context.">
+        : key extends keyof input
+          ? narrow<input[key]>
+          : never // TODO: should this be SystemInput[key] ?
```
alvrs, line 22: i think `never` is good, this clause should never be reached in practice (since `input` will only have keys that appear in `input`)

alvrs (dismissed):
a new typescript wizard is born 🙇

packages/world/ts/config/v2/system.ts
holic, line 22:
```suggestion
          : never
```

PR #2968, opened 2024-07-23
feat(store,world): enable namespaces input

holic:
- [x] multiple namespace tests
- [x] validate that each namespace is only specified once (since the same namespace can be used with diff labels)
- [x] output table libs to namespaced dirs
- [x] support deploying systems from namespace dirs
- [x] enable systems input in namespaces


packages/store/ts/config/v2/output.ts
```diff
@@ -62,15 +62,13 @@ export type Codegen = {
...
   readonly outputDirectory: string;
   /**
-   * Whether or not to organize codegen output (table libraries, etc.) into directories by namespace.
+   * Tables index filename.
    *
-   * For example, a `Counter` table in the `app` namespace will have codegen at `codegen/app/tables/Counter.sol`.
+   * Defaults to `"index.sol"` when in single-namespace mode, and `false` for multi-namespace mode.
    *
-   * Defaults to `true` when using top-level `namespaces` key, `false` otherwise.
+   * @deprecated We recommend importing directly from table libraries rather than from the index for better compile times and deterministic deploys.
```
alvrs, line 69: ooc do we know that it impacts compile times?

alvrs (approved):
🔥🔥🔥

packages/world/ts/config/v2/namespaces.ts
```diff
@@ -1,42 +1,11 @@
...
-export function validateNamespace<scope extends Scope = AbiTypeScope>(
-  namespace: unknown,
-  scope: scope,
-): asserts namespace is NamespaceInput {
-  if (hasOwnKey(namespace, "tables")) {
-    validateTables(namespace.tables, scope);
-  }
-}
+// Copied from store/ts/config/v2/namespaces.ts but using world namespace validate/resolve methods
+// TODO: figure out how to dedupe these?
```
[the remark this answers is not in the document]
alvrs: don't really have a good strategy for this either :/

packages/store/ts/config/v2/output.ts
holic: I haven't measured it but it's very likely given 1) more files need to be compiled, even if not used and 2) reports from @0xhank and others on more files = worse compile times (which may be fixable upstream in forge/foundry/solc)

The main motivation to deprecate this here is that any change to tables impacts bytecode of all things import `index.sol`, even if they aren't importing the table that changed.

PR #2970, opened 2024-07-23
chore(docs): tell vscode that .mdx files are markdown 🚗 

qbzzt:


.vscode/settings.json
```diff
@@ -1,3 +1,5 @@
 {
-  "typescript.tsdk": "node_modules/typescript/lib"
```
holic, line 2: can we restore this please?

.vscode/settings.json
```diff
@@ -1,3 +1,5 @@
...
+  "files.associations": {
+    "*.mdx": "markdown"
```
holic, line 3: mdx is not technically markdown

have you tried the MDX plugin?

image

qbzzt: Thank you, plugin fixed it.

PR #2971, opened 2024-07-24
docs(services/indexer): expand on how to use, promote our indexer when possible

qbzzt:
Also:

1. Specify when you need to run the front end
2. Update the `World` address
3. Reorganize some content

Closing: #2997


holic (commented):
Thoughts on splitting this out into a few different pages instead of one massive page with all the possible options? I feel like it's a bit hard to traverse in its current state and hard to answer "what should I do?" without reading all the technical detail/steps of every option.

qbzzt:
> Thoughts on splitting this out into a few different pages instead of one massive page with all the possible options? I feel like it's a bit hard to traverse in its current state and hard to answer "what should I do?" without reading all the technical detail/steps of every option.

Working on that now.

qbzzt:
> Thoughts on splitting this out into a few different pages instead of one massive page with all the possible options? I feel like it's a bit hard to traverse in its current state and hard to answer "what should I do?" without reading all the technical detail/steps of every option.

Consider it done.

alvrs (approved):

PR #2972, opened 2024-07-25
refactor(store): update tablegen with namespaces output

holic:
reminders to self when refactoring worldgen:
- [ ] update `worldImportPath` like we did here for `storeImportPath`
- [ ] replace usages of `usedInPath` and then get rid of `solidityRelativeImportPath` and related code


packages/cli/scripts/generate-test-tables.ts
```diff
@@ -16,10 +15,10 @@ const config = defineStore({
     Enum2: ["E1"],
   },
   userTypes: {
-    TestTypeAddress: { filePath: "./contracts/src/types.sol", type: "address" },
-    TestTypeInt64: { filePath: "./contracts/src/types.sol", type: "int64" },
-    TestTypeBool: { filePath: "./contracts/src/types.sol", type: "bool" },
-    TestTypeUint128: { filePath: "./contracts/src/types.sol", type: "uint128" },
+    TestTypeAddress: { filePath: "../contracts/src/types.sol", type: "address" },
```
holic, line 18: previously, codegen expected this path to be relative to `process.cwd()` (almost always the project root/same dir as the `mud.config.ts`)

once I changed the paths to be relative to the project root/`mud.config.ts`, this broke, because for this particular script, the config is in a slightly different place

packages/cli/scripts/generate-test-tables.ts
```diff
@@ -16,10 +15,10 @@ const config = defineStore({
...
+    TestTypeInt64: { filePath: "../contracts/src/types.sol", type: "int64" },
+    "TestTypeLibrary.TestTypeBool": { filePath: "../contracts/src/types.sol", type: "bool" },
+    "TestTypeLibrary.TestTypeUint128": { filePath: "../contracts/src/types.sol", type: "uint128" },
```
holic, line 21: previously, codegen would parse the file at `filePath` to determine where in the file was the specified user type and then propagate the fully qualified name everywhere

this felt too "automagic" to me and added a lot of complexity to codegen that I was able to remove by not automatically inferring the code path for the user type (with the nice side effect that we no longer need remappings for tablegen)

we can still support this (imo rare) case by specifying the fully qualified type name like above

(maybe in the future could improve this with a similar "label" approach where the label is used in config references like schemas, and allow specifying a fully qualified code path override of `internalType` that gets used in codegen)

alvrs (approved):

packages/store/mud.config.ts
```diff
@@ -2,7 +2,7 @@ import { defineStore } from "./ts/config/v2/store";
 
 export default defineStore({
   codegen: {
-    storeImportPath: "../../",
+    storeImportPath: "./src",
```
holic, line 5: fyi this is an internal config option

PR #2980, opened 2024-07-29
fix: changeset minor version

### LINKED ISSUE #2977: fix changesets minor version bump
for some reason, changesets thinks the next version is 3.0.0 rather than 2.1.0, even though we only have `minor` bumps in changesets

holic:
closes #2977 

looks like changeset has some weird logic for resolving peer dependencies: https://github.com/changesets/changesets/issues/1279

we can work around this by:
- enabling experimental peer dep flag in changeset
- specifying a wider version range for workspace peer deps

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -10,5 +10,8 @@
   "snapshot": {
     "useCalculatedVersion": true,
     "prereleaseTemplate": "{tag}-{commit}"
+  },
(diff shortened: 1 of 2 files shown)


holic:
I would be happy to make a repro for the changeset repo but I am not convinced it'll ever get addressed given the git history/activity, so going to punt on this effort for now

PR #2987, opened 2024-07-29
docs(solidity to MUD dictionary): first version

qbzzt:
Explain how to deal with state variables.


docs/pages/_meta.js
```diff
@@ -29,6 +29,7 @@ export default {
   },
   guides: "Guides",
   templates: "Templates",
+  dictionary: "Solidity to MUD Dictionary",
```
alvrs, line 32: What do you think of calling this "MUD for Solidity programmers"? (leaning on naming of books like [Rust for Embedded C Programmers](https://opentitan.org/book/doc/rust_for_c_devs.html))

docs/pages/dictionary.mdx
```diff
@@ -0,0 +1,262 @@
+import { Callout } from "nextra/components";
+
+# Solidity to MUD Dictionary
+
+To properly use MUD you need to do some things differently than the way you do them normally in Solidity.
+
+## State variables
+
+A `System` should be stateless, for several reasons:
```
alvrs, line 9:
```suggestion
A `System` contract should not store any internal state, but only interact with `World` state via `Tables`.
```
Making it more explicit that systems can work with state, just not internal state

docs/pages/dictionary.mdx
```diff
@@ -0,0 +1,262 @@
...
+
+- By concentrating all the state information in a single contract (the `World`) we make it easier to apply [access control](/world/namespaces-access-control), [execute hooks when data changes](/store/store-hooks), etc.
+- It is a lot easier to [upgrade functionality](/cli/deploy#upgrading-a-world) when you do not need to deal with storage because the contract does not have a state.
+- The same `System` can be used by multiple `World`s.
+
+The functionality of [state variables](https://docs.soliditylang.org/en/latest/internals/layout_in_storage.html) is implemented using [tables](/store/tables).
```
alvrs, line 15:
```suggestion
Instead of using [state variables](https://docs.soliditylang.org/en/latest/internals/layout_in_storage.html) you can achieve the same functionality with [tables](/store/tables).
```

docs/pages/dictionary.mdx
```diff
@@ -0,0 +1,262 @@
...
+});
+```
+
+</details>
+
+<Callout type="info" emoji="📝">
+  MUD key fields have to be fixed-length.
+  This means that some mappings, for example `mapping(string => uint) balances;`, cannot be directly translated to a MUD table.
+
+An easy workaround is to use the hash of the string as the key instead of the string itself.
```
alvrs, line 155:
```suggestion
An easy workaround is to use the hash of the string as the key instead of the string itself (which is exactly what happens under the hood of vanilla Solidity with the keys of a mapping).
```

docs/pages/dictionary.mdx
```diff
@@ -0,0 +1,262 @@
...
+   ```
+
+   </details>
+
+1. An array is a mapping between an unsigned integer and the value type of the array.
+   We can use a table with `index` as the key. That way, we can get around the limitation that arrays of variable-length types such as `string` or `bytes` are not supported.
+
+    <details>
+
+   {" "}
```
alvrs, line 198:
```suggestion
```

docs/pages/dictionary.mdx
```diff
@@ -0,0 +1,262 @@
...
+       },
+     },
+   });
+   ```
+
+    </details>
+
+{
+
+// ## Events
```
alvrs, line 229: Why is this part commented out?

docs/pages/_meta.js
qbzzt: Done, good idea.

docs/pages/dictionary.mdx
qbzzt: Because I didn't manage to get it working on React with Zustand yet. So I decided it would be better to publish what I have and wait until we have stash to explain events.

docs/pages/dictionary.mdx
qbzzt: No. It doesn't change anything real, but I can't pass prettier without it.

PR #2988, opened 2024-07-29
test: improve type benches with baseline + .ts files, avoiding build

ssalbdivad:
While investigating the type performance issues that led to https://github.com/wevm/viem/pull/2557, I made some improvements to the local type benchmarks.

The new "baseline expression" allows them to be run accurately directly from `.ts` files and gives a much better linear representation of the cost of each bench (note that the scaling between 1 table and 5 tables is close to linear).

Also updated `@arktype` dependencies to the latest version and new `@ark` scope.

closes #2988

### DIFF
--- a/packages/config/package.json
+++ b/packages/config/package.json
@@ -40,7 +40,7 @@
     "test:ci": "pnpm run test"
   },
   "dependencies": {
-    "@arktype/util": "0.0.40",
+    "@ark/util": "catalog:",
     "@latticexyz/common": "workspace:*",
     "@latticexyz/schema-type": "workspace:*",
     "esbuild": "^0.17.15",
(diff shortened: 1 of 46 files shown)


packages/config/package.json
```diff
@@ -40,7 +40,7 @@
     "test:ci": "pnpm run test"
   },
   "dependencies": {
-    "@arktype/util": "0.0.40",
+    "@ark/util": "catalog:",
```
holic, line 43: TIL https://pnpm.io/catalogs

ssalbdivad: It's a great feature! I figured I'd use it since I was updating `@ark/util` anyways, but you could migrate to this for all packages you share as non-dev dependencies across multiple packages.

Having multiple package versions can cause lots of problems, so the centralization is super valuable.

holic (approved):

PR #2996, opened 2024-07-31
feat(store-sync): add util to fetch snapshot from indexer with SQL API

alvrs:
Adding utils and docs to interact with experimental SQL API


packages/store-sync/package.json
```diff
@@ -41,9 +45,6 @@
       ],
       "trpc-indexer": [
         "./dist/trpc-indexer/index.d.ts"
-      ],
-      "zustand": [
-        "./dist/zustand/index.d.ts"
```
alvrs, line 46: todo: readd

packages/protocol-parser/src/common.ts
```diff
@@ -51,3 +52,5 @@ export type ValueArgs = {
   encodedLengths: Hex;
   dynamicData: Hex;
 };
+
+export type PartialTable = Pick<Table, "schema" | "key">;
```
holic, line 56: should probably give this a better name if we're going to use it outside the context of a single module

packages/protocol-parser/src/decodeDozerField.ts
```diff
@@ -0,0 +1,24 @@
+import { AbiType } from "@latticexyz/config";
+import {
+  ArrayAbiType,
+  SchemaAbiTypeToPrimitiveType,
+  arrayToStaticAbiType,
+  schemaAbiTypeToDefaultValue,
+} from "@latticexyz/schema-type/internal";
+
+export function decodeDozerField<abiType extends AbiType>(
```
holic, line 9: chatted IRL, should move these into `store-sync/dozer` since they're not store/world protocol things

packages/protocol-parser/src/getKeySchema.ts
```diff
@@ -1,11 +1,13 @@
-import { Schema, Table } from "@latticexyz/config";
+import { KeySchema, StaticAbiType, Table } from "@latticexyz/config";
```
holic, line 1: need to bring over `KeySchema` into this PR

but I don't think it belongs in the config package, because config has no notion of key schema/value schema, so I think should prob bring that over here instead

packages/store-sync/src/dozer/decodeDozerField.ts
```diff
@@ -0,0 +1,24 @@
+import { AbiType } from "@latticexyz/config";
+import {
+  ArrayAbiType,
+  SchemaAbiTypeToPrimitiveType,
+  arrayToStaticAbiType,
+  schemaAbiTypeToDefaultValue,
+} from "@latticexyz/schema-type/internal";
+
+export function decodeDozerField<abiType extends AbiType>(
```
holic, line 9: what about `decodeField` or `decodeColumnValue`? (we're in the dozer export, so dozer in the name seems duplicative)

packages/store-sync/src/dozer/fetchRecordsSql.ts
```diff
@@ -0,0 +1,69 @@
...
+
+type FetchRecordsSqlResult =
+  | {
+      blockHeight: bigint;
+      result: {
+        table: Table;
+        records: DecodeDozerRecordsResult;
+      }[];
+    }
+  | undefined;
```
holic, line 33: not blocking but fwiw I found it quite nice to model this sort of thing with arktype, then can parse/validate/return strongly typed response from a JSON API request

packages/store-sync/src/dozer/fetchRecordsSql.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  if (isDozerResponseFail(response)) {
+    console.warn(`Dozer response: ${response.msg}\n\nTry reproducing via cURL:
+    curl ${dozerUrl} \\
+    --compressed \\
+    -H 'Accept-Encoding: gzip' \\
+    -H 'Content-Type: application/json' \\
+    -d '[${queries.map((query) => `{"address": "${storeAddress}", "query": "${query.sql.replaceAll('"', '\\"')}"}`).join(",")}]'`);
+    return;
+  }
```
holic, line 58:
```suggestion
  const json = JSON.stringify(queries.map((query) => ({ address: storeAddress, query: query.sql })));
  const response: DozerResponse = fetch(dozerUrl, {
    method: "POST",
    headers: {
      "Content-Type": "application/json",
    },
    body: json,
  }).then((res) => res.json();

  if (isDozerResponseFail(response)) {
    console.warn(`Dozer response: ${response.msg}\n\nTry reproducing via cURL:
    curl ${dozerUrl} \\
    --compressed \\
    -H 'Accept-Encoding: gzip' \\
    -H 'Content-Type: application/json' \\
    -d '${json.replaceAll("'", "\\'")}'`);
    return;
  }
```

packages/store-sync/src/dozer/fetchRecordsSql.ts
```diff
@@ -0,0 +1,69 @@
...
+  | {
+      blockHeight: bigint;
+      result: {
+        table: Table;
+        records: DecodeDozerRecordsResult;
+      }[];
+    }
+  | undefined;
+
+export async function fetchRecordsSql({
```
holic, line 35: what about just `fetchRecords`? or `query`? it's dozer so I think SQL can be implied

packages/store-sync/src/dozer/fetchRecordsSql.ts
```diff
@@ -0,0 +1,69 @@
...
+  ).json();
+
+  if (isDozerResponseFail(response)) {
+    console.warn(`Dozer response: ${response.msg}\n\nTry reproducing via cURL:
+    curl ${dozerUrl} \\
+    --compressed \\
+    -H 'Accept-Encoding: gzip' \\
+    -H 'Content-Type: application/json' \\
+    -d '[${queries.map((query) => `{"address": "${storeAddress}", "query": "${query.sql.replaceAll('"', '\\"')}"}`).join(",")}]'`);
+    return;
```
holic, line 57: I find it a little weird that a function to fetch a thing returns undefined on a failure rather than throwing or something.

I would usually reserve this behavior for the place where "failure is optional" e.g. `getSnapshot` or `syncToStash`

packages/store-sync/src/dozer/getSnapshot.ts
```diff
@@ -0,0 +1,88 @@
...
+  // Each individual request is expected to be executed against the same db state so it
+  // can't be parallelized.
+  const results = (await Promise.all([fetchLogs(), ...fetchSql()])).filter(isDefined);
+
+  // The block number passed in the overall result will be the min of all queries and the logs.
+  const initialBlockLogs = {
+    blockNumber: results.length > 0 ? bigIntMin(...results.map((result) => result.blockNumber)) : startBlock,
+    logs: results.flatMap((result) => result.logs),
+  };
+
```
holic, line 86: ooc what happens if `fetchLogs` fails but `fetchSql` doesn't? or vice versa? or some of `fetchSql` does? do we end up in a weird state?

packages/store-sync/src/dozer/getSnapshot.ts
```diff
@@ -0,0 +1,88 @@
...
+      chainId,
+      address: storeAddress,
+      filters: logsFilters,
+      indexerUrl: dozerUrl,
+    });
+  };
+
+  const fetchSql = (): Promise<StorageAdapterBlock | undefined>[] => {
+    return sqlFilters.map(async (filter) => {
+      const result = await fetchRecordsSql({ dozerUrl, storeAddress, queries: [filter] });
```
holic, line 64: more of a stylistic thing but I wonder if this would be more readable

```ts
  const fetchRecords = async (query: TableQuery): Promise<StorageAdapterBlock | undefined> => {
    const result = await fetchRecordsSql({ dozerUrl, storeAddress, queries: [filter] });
```

```ts
const results = (await Promise.all([fetchLogs(), ...sqlFilters.map(fetchRecords)])).filter(isDefined);
```

packages/store-sync/src/dozer/selectFrom.ts
```diff
@@ -0,0 +1,19 @@
+import { Table } from "@latticexyz/config";
+import { TableQuery } from "./common";
+
+// For autocompletion but still allowing all SQL strings
+export type Where<table extends Table> = `"${keyof table["schema"] & string}"` | (string & {});
```
holic, line 5: arktype has `autocomplete<...>` for this!

packages/store-sync/src/dozer/getSnapshot.ts
```diff
@@ -0,0 +1,88 @@
...
+import { StorageAdapterBlock, SyncFilter as LegacyLogFilter } from "../common";
+import { fetchRecordsSql } from "./fetchRecordsSql";
+import { recordToLog } from "../recordToLog";
+import { getSnapshot as getSnapshotLogs } from "../getSnapshot";
+import { bigIntMin, isDefined } from "@latticexyz/common/utils";
+
+export type GetSnapshotArgs = {
+  dozerUrl: string;
+  storeAddress: Hex;
+  filters?: SyncFilter[];
```
holic, line 12: would it better or worse to split this into two args: `filters` and `queries`?

packages/store-sync/src/dozer/fetchRecordsSql.ts
alvrs: no i think you're right, can't remember why i put it in here. moved it into `getSnapshot` now.

packages/store-sync/src/dozer/getSnapshot.ts
alvrs: I kinda like the fact that `table` is shared by both so you can start simple by providing just the table (if you wanna fetch the entire table) and only add sql to the ones where it's needed later

packages/store-sync/src/dozer/getSnapshot.ts
alvrs: yeah i think before that was the case, now all of `getSnapshot` fails if one of the individual requests fails

holic (approved):

packages/protocol-parser/src/getKeySchema.ts
```diff
@@ -1,11 +1,22 @@
-import { Schema, Table } from "@latticexyz/config";
+import { StaticAbiType, Table } from "@latticexyz/config";
 
 type PartialTable = Pick<Table, "schema" | "key">;
 
+type KeySchema = {
```
holic, line 5: we should prob export this?

packages/store-sync/src/dozer/decodeRecords.ts
```diff
@@ -0,0 +1,41 @@
...
+};
+
+/**
+ * Trim the header row from the query result
+ */
+function trimHeader(result: DozerQueryResult): DozerQueryRecord[] {
+  return result.slice(1);
+}
+
+export type DecodeDozerRecordsResult<schema extends Schema = Schema> = getSchemaPrimitives<schema>[];
```
holic, line 23: should we update all the type names without "Dozer"?

alvrs: ahh yes missed that one, good catch

holic (approved):

Issue #2995, opened 2024-07-31
pre 2.1 config input clean up


holic:
- [x] remove `tableId` from config table input, this is derived from type + namespace + name
- [x] remove `registerFunctionSelectors` from system input as its unused
- [ ] deprecate `excludeSystems`
- [x] mark `worldInterfaceName` as internal?
- [ ] rename `worldgenDirectory` as `worldInterfacesDirectory`?

holic:
done in https://github.com/latticexyz/mud/pull/2998
closing as the rest can be revisited in https://github.com/latticexyz/mud/issues/2979

PR #2998, opened 2024-08-01
feat(store,world): clean up a few config inputs

holic:
see #2995

found these when working on docs:
- `tableId` input is not used and is instead derived from `type`, `namespace`, and `name` inputs
- `registerFunctionSelectors` was not in use and will be replaced with `deploy.disabled` in the future
  -  unclear why you'd want to deploy a system but not register its selectors?
- marked `worldInterfaceName` as internal as many things in docs, etc. depend on the name being `IWorld` and we only made this configurable so we could override it ourselves for `IBaseWorld`

there's a few more in #2995 to clean up but will save those for when I have good alternatives implemented before deprecating


ludns:
You might want to have a system that can only be called via `world.call` and not register its selectors. 
It's probably going to be quite common actually for big worlds where function selectors will clash quite early because of the birthday paradox. The deployer will err otherwise. We could make it gracefully fail but i think most people will just turn off function selectors on big worlds.
I think it's important to let people deploy systems and not register the selectors.

packages/store/ts/config/v2/input.ts
```diff
@@ -39,7 +38,6 @@ export type TableInput = {
   readonly name?: string;
   readonly schema: SchemaInput;
   readonly key: readonly string[];
-  readonly tableId?: Hex;
```
alvrs, line 42: i thought the idea with allowing a `tableId` input is that it could technically be overridden to an id that is not representable in UTF8? I guess that would break a lot of other assumptions, is that why you're removing it here?

alvrs (dismissed):
lgtm, pending question about `tableId`

holic:
> You might want to have a system that can only be called via `world.call` and not register its selectors. It's probably going to be quite common actually for big worlds where function selectors will clash quite early because of the birthday paradox. The deployer will err otherwise. We could make it gracefully fail but i think most people will just turn off function selectors on big worlds. I think it's important to let people deploy systems and not register the selectors.



> You might want to have a system that can only be called via `world.call` and not register its selectors.

Ohh good call, totally forgot about this use case. System would still be callable via system ID and system function selector, but just not from the world. I have a use case in mind for this already!

packages/store/ts/config/v2/input.ts
holic: That was the intent but now that there's a namespace and name override and they're contextual, it's a little scary to allow overriding the `tableId` directly. I think it leads to hard to predict/understand behavior.

I'm not aware of any strings JS can't represent/store accurately. There's various encoding options for these too.

We could still add this option again in the future when we have a strong use case, but we'll wanna be sure to validate that the `tableId` matches the table type or namespace it's used within.

PR #3001, opened 2024-08-02
feat: explorer package

karooolis:


packages/explorer/public/next.svg
holic, line 1: can we get rid of these?

packages/explorer/src/app/(home)/DataExplorer.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { useSearchParams } from "next/navigation";
+import { useState } from "react";
+import { useQuery } from "@tanstack/react-query";
+import { SQLEditor } from "./SQLEditor";
+import { TableSelector } from "./TableSelector";
+import { TablesViewer } from "./TablesViewer";
+
+export function DataExplorer() {
+  const searchParams = useSearchParams();
+  const [query, setQuery] = useState<string | undefined>();
```
holic, line 12: would be nice if this was part of the URL as well

holic, line 12: would be nice if this was part of the URL as well

packages/explorer/public/next.svg
karooolis: good catch, removed

templates/react-ecs/mprocs.yaml
```diff
@@ -8,3 +8,16 @@ procs:
...
     cwd: packages/contracts
     shell: anvil --base-fee 0 --block-time 2
+  indexer:
+    shell: rimraf $SQLITE_FILENAME && pnpm sqlite-indexer
+    env:
+      RPC_HTTP_URL: "http://127.0.0.1:8545"
+      FOLLOW_BLOCK_TAG: "latest"
+      SQLITE_FILENAME: "indexer.db"
+  explorer:
+    cwd: /Users/karolis/Code/Lattice.xyz/experiments/mud-2/packages/explorer
```
holic, line 18: prob wnat something like this?
```suggestion
    cwd: ../../packages/explorer
```

karooolis: ah damn, this should not have been added now, removing now. But generally yes, I was using it for testing some stuff

templates/react-ecs/mprocs.yaml
```diff
@@ -8,3 +8,16 @@ procs:
...
+    shell: pnpm dev
```
holic: actually maybe instead of changing `cwd`,  what we want is this to be
```suggestion
    shell: cd ../../packages/explorer && pnpm dev
```

then we can use a relative path to indexer DB

karooolis: I'll try this and consult with you when updating templates, should have not added this now, my bad

packages/explorer/tsconfig.json
```diff
@@ -0,0 +1,26 @@
...
+    "isolatedModules": true,
+    "jsx": "preserve",
+    "incremental": true,
+    "plugins": [
+      {
+        "name": "next"
+      }
+    ],
+    "paths": {
+      "@/*": ["./src/*"]
```
holic, line 21: this isn't documented anywhere but I tend to not use import path aliases as it makes refactoring with IDE really bad, where moving a file causes all sorts of weird and unpredictable behavior

I tend to prefer relative imports instead

packages/explorer/tsconfig.json
holic, line 1: can we inherit from the root tsconfig like we do in all other packages?

packages/explorer/tailwind.config.ts
```diff
@@ -0,0 +1,85 @@
...
+    extend: {
+      fontFamily: {
+        sans: ["var(--font-jetbrains-mono)"],
+        mono: ["var(--font-jetbrains-mono)"],
+      },
+
+      colors: {
+        border: "hsl(var(--border))",
+        input: "hsl(var(--input))",
+        ring: "hsl(var(--ring))",
```
holic, line 29: where do these vars come from?

packages/explorer/src/store.ts
```diff
@@ -0,0 +1,59 @@
...
+  balances: Balances;
+  fetchBalances: () => void;
+};
+
+export const useStore = create<Store>((set) => ({
+  account: ACCOUNTS[0],
+  setAccount: (account) => set({ account }),
+
+  balances: {},
+  fetchBalances: async () => {
```
holic, line 27: ooc why this instead of a component for an account balance, using wagmi hook, then iterating through accounts -> balance in react tree?

packages/explorer/tailwind.config.ts
karooolis: From `globals.css`, it's the way `shadcn/ui` is set up by default.

packages/explorer/src/components/ui/input.tsx
holic, line 1: our current pattern is naming the file exactly as the component that is exported, so this would be `Input.tsx` rather than `input.tsx`

(a rename like this is a little tricky with git and macos given that macos filenames are case insensitive)

packages/explorer/src/components/ui/input.tsx
```diff
@@ -0,0 +1,25 @@
+import * as React from "react";
+import { cn } from "@/lib/utils";
+
+export interface InputProps
+  extends React.InputHTMLAttributes<HTMLInputElement> {}
```
holic, line 5: curious about this pattern vs
```
export type Props =  React.InputHTMLAttributes<HTMLInputElement>
```

packages/explorer/src/components/ui/input.tsx
karooolis: Yeah, dealt with filenames changes on macos, a bit annoying. I'll change to match the component name. The current lowercase components is how `shadcn/ui` sets them up but there is nothing stopping to change that.

packages/explorer/package.json
```diff
@@ -0,0 +1,71 @@
...
+    "@radix-ui/react-toast": "^1.2.1",
+    "@radix-ui/themes": "^3.0.5",
+    "@tanstack/react-query": "^5.51.3",
+    "@tanstack/react-table": "^8.19.3",
+    "@uidotdev/usehooks": "^2.4.1",
+    "@wagmi/core": "^2.12.1",
+    "better-sqlite3": "^8.6.0",
+    "class-variance-authority": "^0.7.0",
+    "clsx": "^2.1.1",
+    "lodash": "^4.17.21",
```
holic, line 37: I have PTSD from lodash and bundle sizes. Looks like we only use `camelCase`, so wondering if there's a lightweight version of this that we can include as a function and use throughout?

packages/explorer/src/components/ui/input.tsx
karooolis: Good point, I'm not aware of why it's done this way (default shadcn/ui exported component), I'll simplify it with your suggestion.

packages/explorer/package.json
karooolis: Ah good point, not worth importing just for one function, left it during prototyping. I'll implement our own function for this.

packages/explorer/src/store.ts
karooolis: Wagmi can only fetch one balance at a time, no? Couldn't find how to fetch several balances all at once. The nice micro-optimization with this one is that all balances are fetched in one POST request.

packages/explorer/src/app/api/table/route.ts
```diff
@@ -0,0 +1,32 @@
...
+  const { searchParams } = new URL(req.url);
+  const tableId = searchParams.get("tableId") as Hex;
+
+  if (!tableId) {
+    return Response.json({ error: "tableId is required" }, { status: 400 });
+  }
+
+  const db = getDatabase();
+  const table = db
+    ?.prepare(`SELECT * FROM __mudStoreTables WHERE id='${tableId}'`)
```
holic, line 28: I think we're negating the use of prepared statements and also prone to SQL injection (because `tableId` is user-provided)
```suggestion
    ?.prepare("SELECT * FROM __mudStoreTables WHERE id=?").bind(tableId)
```

packages/explorer/src/app/(home)/SQLEditor.tsx
```diff
@@ -0,0 +1,44 @@
...
+          style={{ flex: "1" }}
+          placeholder="SQL query…"
+          value={deferredQuery}
+          onChange={(evt) => setDeferredQuery(evt.target.value)}
+        >
+          <TextField.Slot></TextField.Slot>
+        </TextField.Root>
+
+        <Button type="submit">Execute query</Button>
+      </Flex> */}
```
holic, line 41: should this be deleted if not used?

karooolis: Yeah, should not be used for now, removed

packages/explorer/src/store.ts
holic: you can enable batch RPC calls in your viem/wagmi client

then do something like (pseudocode)

```tsx
<>
  <h1>Accounts</h1>
  {accounts.map((account) => (
    <div key={account}>{account}: <AccountBalance account={account} /></div>
  )}
</>
```

```tsx
function AccountBalance({ account }) {
  const balance = useBalance({ account });
  return <>{formatEther(balance)}</>;
```

packages/explorer/src/components/ui/input.tsx
karooolis: Updated to match all filenames to component names.

packages/explorer/tsconfig.json
karooolis: I personally prefer aliases but not fixated on this, and good point about refactoring. Changed to relative paths

packages/explorer/src/components/ui/input.tsx
karooolis: so looking at other shadcn components it seems like it's just a pattern that makes it easy to extend the type, such as in `Button` component:

```typescript
export interface ButtonProps
  extends React.ButtonHTMLAttributes<HTMLButtonElement>,
    VariantProps<typeof buttonVariants> {
  asChild?: boolean;
}
```

However, I don't see why it's needed given there are no additional props we need to add so changed to which is consistent to how types are declared in other components:

```typescript
const Input = React.forwardRef<
  HTMLInputElement,
  React.InputHTMLAttributes<HTMLInputElement>
>(({ className, type, ...props }, ref) => {
```

packages/explorer/src/app/api/table/route.ts
karooolis: I am unable to avoid using prepare statement with `better-sqlite3` it seems but should have parameterized SQL, now done. I was unable to parameterize `FROM` so instead checking if table name is valid i.e.

```typescript
  try {
    if (!table || !doesTableExist(table)) {
      return Response.json({ error: "table does not exist" }, { status: 400 });
    }

    const db = getDatabase();
    const query = `SELECT * FROM "${table}" LIMIT 30`;
    const rows = db?.prepare(query).all() as RowsResponse;
```

packages/explorer/src/app/api/rows/route.ts
```diff
@@ -0,0 +1,33 @@
...
+type RowsResponse = Row[] | undefined;
+
+function convertKeysToCamelCase(rows: RowsResponse): Row[] {
+  if (rows == undefined) {
+    return [];
+  }
+
+  return rows.map((row) => {
+    return Object.keys(row).reduce((result: { [key: string]: string }, key) => {
+      const camelKey = _.camelCase(key);
```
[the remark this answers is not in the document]
karooolis: It's because when value schema is defined in camelCase e.g.

{"createdAt":"uint256","completedAt":"uint256","description":"string"}}

and when I fetch table schema, the keys are snake cased.

CleanShot 2024-08-09 at 12 35 35@2x

Then, when updating the content in `EditableTableCell` I am using column's name to determine which field is being updated which couldn't work due to differing keys (one snake cased, one camel cased). I simplified the approach by only camel casing the column key when updating the value so I can determine the correct key (https://github.com/latticexyz/mud/pull/3001/files#diff-824dfa3431242e4dabe23c890ddf66a1c1058a22e9345b840a848938f3c72272R41) rather than camel casing all columns like before.

packages/explorer/tsconfig.json
karooolis: I tried it several ways but then I'm getting the below error, do you have any insights perhaps?

CleanShot 2024-08-09 at 13 12 00@2x

packages/explorer/package.json
```diff
@@ -0,0 +1,69 @@
...
+    "@types/react": "^18",
+    "@types/react-dom": "^18",
+    "@types/react-highlight-words": "^0.20.0",
+    "eslint": "^8",
+    "eslint-config-next": "14.2.3",
+    "postcss": "^8",
+    "prettier": "3.2.5",
+    "prettier-plugin-tailwindcss": "^0.6.5",
+    "tailwindcss": "^3.4.1",
+    "typescript": "^5"
```
holic, line 67: you don't need to define this here as its already in the workspace root

packages/explorer/package.json
```diff
@@ -0,0 +1,69 @@
...
+    "zustand": "^4.3.7"
+  },
+  "devDependencies": {
+    "@trivago/prettier-plugin-sort-imports": "^4.3.0",
+    "@types/better-sqlite3": "^7.6.4",
+    "@types/node": "^20",
+    "@types/react": "^18",
+    "@types/react-dom": "^18",
+    "@types/react-highlight-words": "^0.20.0",
+    "eslint": "^8",
```
holic, line 61: same with this, this is already in the workspace root

packages/explorer/package.json
```diff
@@ -0,0 +1,69 @@
...
+    "eslint-config-next": "14.2.3",
+    "postcss": "^8",
+    "prettier": "3.2.5",
```
holic, line 64: and this

packages/explorer/tsconfig.json
holic: guessing its a TS version mismatch because your TS version in package.json is a different semver range than the one we use elsewhere

you should be able to remove the TS dev dep from this package because its part of the workspace root already

packages/explorer/src/store.ts
karooolis: Was a bit hesitant at first but after refactoring I think it's much nicer now

packages/explorer/package.json
karooolis: removed, couldn't spot any more

packages/explorer/package.json
karooolis: removed, couldn't spot any more

.changeset/swift-rabbits-appear.md
```diff
@@ -0,0 +1,34 @@
+---
+"@latticexyz/explorer": patch
+---
+
+### Initial Release of Explorer
+
+Explorer is a GUI tool designed for visually exploring and manipulating the state of worlds.
```
holic, line 7: We've generally avoided the use of headlines in changesets because the changelog makes use of headlines and the changeset including them mucks up formatting a bit.

Can we frame this in the context of a changeset?

> Initial release of the `@latticexyz/explorer` package. Explorer is a standalone tool designed to explore and manage worlds. This initial release includes support for only local worlds, with support for any world in the near future.

.changeset/swift-rabbits-appear.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Getting Started
```
holic, line 9: With the above framing and next sentence below, we can just remove this headline I think

.changeset/swift-rabbits-appear.md
```diff
@@ -0,0 +1,34 @@
...
+
+To set up and start using the Explorer, follow these steps:
+
+1. **Configure Paths to Indexer's Database**
```
holic, line 13:
```suggestion
1. **Configure paths to SQLite indexer database**
```

.changeset/swift-rabbits-appear.md
```diff
@@ -0,0 +1,34 @@
...
+
+   Set up either a relative or an absolute path to the indexer's database using environment variables. You may use `@latticexyz/store-indexer` for indexing your world's data.
+
+   - `INDEXER_DB_PATH_ABSOLUTE` - Set the absolute path to the database.
+   - `INDEXER_DB_PATH` - Set the relative path to the database.
+
+2. **Install Dependencies**
```
holic, line 20:
```suggestion
2. **Install dependencies**
```

.changeset/swift-rabbits-appear.md
```diff
@@ -0,0 +1,34 @@
...
+
+   Run the following command to install the necessary dependencies:
+
+   ```sh
+   pnpm install
+   ```
+
+3. **Start the Development Server**
```
holic, line 28:
```suggestion
3. **Start the development server**
```

.changeset/swift-rabbits-appear.md
```diff
@@ -0,0 +1,34 @@
...
+
+   Use the following command to start the development server:
+
+   ```sh
+   pnpm dev
+   ```
```
holic, line 34: Instead of the install and dev steps, can we build/export a binary with tsup for this service and run it like `npx @latticexyz/explorer`?

examples/explorer-template/.eslintrc
holic, line 1: how about just `examples/local-explorer/...` for now?

examples are meant to showcase how to do something, templates are meant to be cloned by our CLI tool, and I don't wanna mix em up

examples/explorer-template/packages/client/package.json
```diff
@@ -0,0 +1,37 @@
...
+  },
+  "dependencies": {
+    "@latticexyz/common": "2.1.0",
+    "@latticexyz/dev-tools": "2.1.0",
+    "@latticexyz/react": "2.1.0",
+    "@latticexyz/recs": "2.1.0",
+    "@latticexyz/schema-type": "2.1.0",
+    "@latticexyz/store-sync": "2.1.0",
+    "@latticexyz/utils": "2.1.0",
+    "@latticexyz/world": "2.1.0",
```
holic, line 21: we're going to have version issues if we have some MUD packages coming from npm and some linked locally

we should change all these to `link:../../packages/*`

examples/explorer-template/packages/client/src/App.tsx
```diff
@@ -0,0 +1,29 @@
...
+    components: { Counter },
+    systemCalls: { increment },
+  } = useMUD();
+
+  const counter = useComponentValue(Counter, singletonEntity);
+
+  return (
+    <>
+      <div>
+        Counter: <span>{counter?.value ?? "??"}</span>
```
holic, line 16: thoughts on using the plain react template rather than the ECS one? so we get a task list instead of a counter

examples/explorer-template/packages/contracts/package.json
```diff
@@ -0,0 +1,35 @@
...
+    "prettier": "prettier --write 'src/**/*.sol'",
+    "solhint": "solhint --config ./.solhint.json 'src/**/*.sol' --fix",
+    "test": "tsc --noEmit && mud test"
+  },
+  "dependencies": {
+    "@latticexyz/cli": "2.1.0",
+    "@latticexyz/schema-type": "2.1.0",
+    "@latticexyz/store": "2.1.0",
+    "@latticexyz/world": "2.1.0",
+    "@latticexyz/world-modules": "2.1.0"
```
holic, line 23: same as above

examples/explorer-template/packages/contracts/package.json
```diff
@@ -0,0 +1,35 @@
...
+  },
+  "devDependencies": {
+    "@types/node": "^18.15.11",
+    "ds-test": "https://github.com/dapphub/ds-test.git#e282159d5170298eb2455a6c05280ab5a73a4ef0",
+    "forge-std": "https://github.com/foundry-rs/forge-std.git#74cfb77e308dd188d2f58864aaf44963ae6b88b1",
+    "prettier": "3.2.5",
+    "prettier-plugin-solidity": "1.3.1",
+    "solhint": "^3.3.7",
+    "solhint-config-mud": "2.1.0",
+    "solhint-plugin-mud": "2.1.0"
```
holic, line 33: same as above

package.json
```diff
@@ -8,7 +8,7 @@
     "url": "https://github.com/latticexyz/mud.git"
   },
   "scripts": {
-    "all-build": "for dir in packages/store packages/world packages/world-modules packages/cli test/mock-game-contracts e2e/packages/contracts examples/*/packages/contracts examples/multiple-namespaces templates/*/packages/contracts; do (cd \"$dir\" && pwd && pnpm build); done",
+    "all-build": "for dir in packages/store packages/world packages/explorer packages/world-modules packages/cli test/mock-game-contracts e2e/packages/contracts examples/*/packages/contracts examples/multiple-namespaces templates/*/packages/contracts; do (cd \"$dir\" && pwd && pnpm build); done",
```
holic, line 11: this task was originally meant to be able to generate all the codegen that gets checked into git (rather than build all packages, which we have `pnpm build` for)

does explorer have any usages of `mud build` where we'd wanna check in those files? if not, can remove this

packages/explorer/README.md
```diff
@@ -0,0 +1,30 @@
+### Explorer
+
+Explorer is a GUI tool designed for visually exploring and manipulating the state of worlds.
+
+## Getting Started
```
holic, line 5: once we have an exported binary, we should separate the tasks of using explorer (add package to project, run binary with env var) from developing explorer (installing package, running mprocs dev server)

packages/explorer/package.json
```diff
@@ -0,0 +1,67 @@
...
+    "@radix-ui/react-select": "^2.1.1",
+    "@radix-ui/react-separator": "^1.1.0",
+    "@radix-ui/react-slot": "^1.1.0",
+    "@radix-ui/react-toast": "^1.2.1",
+    "@radix-ui/themes": "^3.0.5",
+    "@tanstack/react-query": "^5.51.3",
+    "@tanstack/react-table": "^8.19.3",
+    "@uidotdev/usehooks": "^2.4.1",
+    "@wagmi/core": "^2.12.1",
+    "better-sqlite3": "^8.6.0",
```
holic, line 34: haven't checked but we should make sure these common dep versions are aligned with the rest of our packages (wagmi/viem used in templates, sqlite in indexer, react in templates and react package, etc.)

packages/explorer/src/app/(home)/EditableTableCell.tsx
```diff
@@ -0,0 +1,144 @@
...
+  const worldAddress = useWorldAddress();
+
+  const [loading, setLoading] = useState(false);
+  const [prevValue, setPrevValue] = useState<unknown>(defaultValue);
+  const [value, setValue] = useState<unknown>(defaultValue);
+
+  const tableId = config?.table_id;
+  const fieldType = config?.value_schema[camelCase(name)] as SchemaAbiType;
+
+  const handleSubmit = async (newValue: unknown) => {
```
holic, line 46: not blocking, but you might consider using a react-query mutation for this use case, which includes pending states, etc.

packages/explorer/src/app/(home)/DataExplorer.tsx
holic, line 1: what's the advantage of putting these in `(home)`?

packages/explorer/src/app/(home)/SQLEditor.tsx
holic: looks like this is still here?

packages/explorer/src/app/(home)/TableSelector.tsx
```diff
@@ -0,0 +1,53 @@
...
+          <SelectValue placeholder="Select a table ..." />
+        </SelectTrigger>
+        <SelectContent>
+          {options?.map((option) => {
+            return (
+              <SelectItem key={option} value={option} className="font-mono">
+                {NON_EDITABLE_TABLES.includes(option) && (
+                  <Lock className="mr-2 inline-block opacity-70" size={14} />
+                )}
+                {option.replace(`${worldAddress}__`, "")}
```
holic, line 45: is there another way to derive the table name rather than string replacement?

packages/explorer/src/components/ui/Button.tsx
```diff
@@ -0,0 +1,56 @@
...
+      size: "default",
+    },
+  },
+);
+
+export interface ButtonProps
+  extends React.ButtonHTMLAttributes<HTMLButtonElement>,
+    VariantProps<typeof buttonVariants> {
+  asChild?: boolean;
+}
```
holic, line 40:
```suggestion
export type ButtonProps = React.ButtonHTMLAttributes<HTMLButtonElement> & VariantProps<typeof buttonVariants> & { asChild?: boolean; }
```

packages/explorer/src/components/AccountSelect.tsx
```diff
@@ -0,0 +1,59 @@
...
+  address,
+  name,
+}: {
+  address: Address;
+  name: string;
+}) {
+  const balance = useBalance({
+    address,
+    query: {
+      refetchInterval: CONFIG.BALANCES_REFETCH_INTERVAL,
```
holic, line 24: does this need to be configurable?

packages/explorer/src/app/interact/FunctionField.tsx
```diff
@@ -0,0 +1,194 @@
...
+          {(abi.stateMutability === "view" || abi.stateMutability === "pure") &&
+            "Read"}
+          {(abi.stateMutability === "payable" ||
+            abi.stateMutability === "nonpayable") &&
+            "Write"}
+        </Button>
+
+        {result && (
+          <pre className="text-md rounded border p-3 text-sm">{result}</pre>
+        )}
```
holic, line 188: I personally have found this pattern to be too much of a foot-gun and now use the following in all my React ternaries:
```suggestion
        {result ? (
          <pre className="text-md rounded border p-3 text-sm">{result}</pre>
        ) : null}
```

the reason is if `result` is ever `0` (zero), it will be rendered as `0` in the UI rather than nothing

although this may not be a concern now or in this context, in some cases we may refactor a variable used in rendering with the possibility of being zero and then suddenly have weird rendering bugs in UI that could be hard to spot or track down

packages/explorer/src/app/interact/Form.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+              return (
+                <li key={idx}>
+                  <a
+                    href={`#${(abi as AbiFunction).name}`}
+                    className={cn(
+                      "whitespace-nowrap text-sm hover:text-orange-500 hover:underline",
+                      {
+                        "text-orange-500": hash === (abi as AbiFunction).name,
+                      },
```
holic, line 66: you can also do
```suggestion
                      abi.name === hash ? "text-orange-500" : null,
```

packages/explorer/src/app/interact/Form.tsx
```diff
@@ -0,0 +1,103 @@
...
+          >
+            {filteredFunctions.map((abi, idx) => {
+              if ((abi as AbiFunction).type !== "function") {
+                return null;
+              }
+
+              return (
+                <li key={idx}>
+                  <a
+                    href={`#${(abi as AbiFunction).name}`}
```
holic, line 61: why does this type need to be casted if we already have it casted above in the array definition?

packages/explorer/src/app/interact/Form.tsx
```diff
@@ -0,0 +1,103 @@
...
+  };
+};
+
+export function Form({ data }: Props) {
+  const [hash] = useHashState();
+  const [filterValue, setFilterValue] = useState("");
+  const deferredFilterValue = useDeferredValue(filterValue);
+  const abiFunctions = data.abi.filter(
+    (abi) => (abi as AbiFunction).type === "function",
+  ) as AbiFunction[];
```
holic, line 23: rather than casting, you could also do narrowing with

```ts
const abiFunctions = data.abi.filter((item): item is AbiFunction => item.type === "function");
```

packages/explorer/src/app/api/world/utils/deduplicateAbi.ts
```diff
@@ -0,0 +1,29 @@
+import { Abi, AbiItem } from "viem";
+
+function hasNamedInputs(entry: AbiItem) {
+  if (!("inputs" in entry) || !entry.inputs || !entry.inputs.length) {
+    return false;
+  }
+  return entry.inputs.some((input) => input.name);
+}
+
+export function deduplicateAbi(abi: Abi) {
```
holic, line 10: what's this help with?

holic (commented):
mostly small stylistic things and a few questions about the intent of various functions, but otherwise looking good!

main thing I'd like to see is exporting a binary we can use to run this with something like `npx @latticexyz/explorer`

packages/explorer/src/app/interact/FunctionField.tsx
karooolis: Understood, probably result is not going be 0 in this case but down the line it could change so yeah, makes sense, updated.

.changeset/swift-rabbits-appear.md
karooolis: Agreed, removed

.changeset/swift-rabbits-appear.md
karooolis: Yeah, like your description, changed and removed the title.

packages/explorer/src/app/interact/Form.tsx
karooolis: Good catch, that's a leftover

packages/explorer/src/app/interact/Form.tsx
karooolis: oh this is much better with type predicate

packages/explorer/src/app/(home)/DataExplorer.tsx
karooolis: Just enables grouping files better. Otherwise, home page files would be at the same level as `api`, `interact` page, etc. and I'm not too fond of that.

packages/explorer/src/app/(home)/utils/getFieldIdx.ts
```diff
@@ -0,0 +1,19 @@
+import {
+  isDynamicAbiType,
+  isStaticAbiType,
+} from "@latticexyz/schema-type/internal";
+
+export function getFieldIdx(
```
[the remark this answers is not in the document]
karooolis: noted!

packages/explorer/src/app/api/world/utils/deduplicateAbi.ts
karooolis: Discussed in https://github.com/latticexyz/mud/pull/3025

packages/explorer/src/app/_providers.tsx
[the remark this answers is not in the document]
karooolis: Adjusted, need to find `eslint` to prevent this

packages/explorer/src/app/(home)/TablesViewer.tsx
```diff
@@ -0,0 +1,295 @@
...
+  const [columnFilters, setColumnFilters] = useState<ColumnFiltersState>([]);
+  const [columnVisibility, setColumnVisibility] = useState<VisibilityState>({});
+  const [rowSelection, setRowSelection] = useState({});
+  const [globalFilter, setGlobalFilter] = useState("");
+  const [showAllColumns, setShowAllColumns] = useState(false);
+
+  const { data: schema } = useQuery({
+    queryKey: ["schema", { table: selectedTable }],
+    queryFn: async () => {
+      const response = await fetch(`/api/schema?table=${selectedTable}`);
```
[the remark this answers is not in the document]
karooolis: Oh nice tip, adjusted

examples/explorer-template/.eslintrc
karooolis: Agreed, should have not used `template` as part of the name, changed to `local-explorer`

examples/explorer-template/packages/client/src/App.tsx
karooolis: Yeah, task list is a more interactive example so makes sense to use it as well, changed to it

packages/explorer/src/app/layout.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+const jetbrains = JetBrains_Mono({
+  subsets: ["latin"],
+  variable: "--font-jetbrains-mono",
+});
+
+export const metadata: Metadata = {
+  title: "Worlds Explorer",
+  description: "Worlds Explorer",
+};
```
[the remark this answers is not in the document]
karooolis: Good catch, adjusted

packages/explorer/src/hooks/useHash.ts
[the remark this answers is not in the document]
karooolis: Updated filename to `useHashState`

packages/explorer/src/hooks/useHash.ts
```diff
@@ -0,0 +1,51 @@
+"use client";
+
+import { useParams, useRouter } from "next/navigation";
+import queryString from "query-string";
+import { useEffect, useMemo, useState } from "react";
+
+// source: https://github.com/vercel/next.js/discussions/49465
+export const useHashState = () => {
```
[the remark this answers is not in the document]
karooolis: To get hash from the URL so as to highlight properly the selected function in `Interact` page, just a minor Ui detail.

packages/explorer/src/app/(home)/abi.ts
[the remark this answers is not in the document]
karooolis: only need `setField` ABI, leftover from quick prototyping, my bad

packages/explorer/src/app/(home)/abi.ts
```diff
@@ -0,0 +1,63 @@
+export const abi = [
```
holic, line 1: ahh is this how you worked around the strongly typed json imports issue?

packages/explorer/src/components/AccountSelect.tsx
karooolis: Generally, I like to use config for such values in case it repeats across several similar situations. Then I can configure all of them from one place, and stay consistent. But after having a second though, it's not necessary here, removed.

packages/explorer/src/app/(home)/abi.ts
karooolis: changed to IBaseWorldABI from world package

package.json
karooolis: Ah I see, regular `build` should be good, removed to original

packages/explorer/src/app/(home)/EditableTableCell.tsx
karooolis: refactored, although will still have to do some improvements here

packages/explorer/src/app/(home)/TableSelector.tsx
karooolis: Alternatively, I could also just slice to show the last two parts of the name, namespace and table name e.g.

`0x8d8b6b8414e1e3dcfd4168561b9be6bd3bf6ec4b__world__balances` -> `world__balances`

I'm not too fond of string replacement either but table names are stored in such format by the indexer, and can't think of how else to do this?

packages/explorer/src/app/interact/page.tsx
```diff
@@ -0,0 +1,26 @@
+import { headers } from "next/headers";
+import { Form } from "./Form";
+
+async function getABI() {
+  const headersList = headers();
+  const protocol = headersList.get("x-forwarded-proto");
+  const host = headersList.get("host");
+
+  const res = await fetch(
+    `${protocol}://${host}/api/world?address=${process.env.NEXT_PUBLIC_WORLD_ADDRESS}`,
```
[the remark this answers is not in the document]
karooolis: The thing in node is that it doesn't have browser context, and thus the full URI has to be provided, including the protocol.

packages/explorer/package.json
karooolis: Found and updated these disrepancies: `@types/node`, `@types/react`, `@types/react-dom`. Also removed `prettier` as it was in root `package.json` already.

packages/explorer/README.md
karooolis: Updated changeset to include sections on installing explorer, example setup, and contributing/development guide. Lmk what you think

.changeset/swift-rabbits-appear.md
```diff
@@ -0,0 +1,71 @@
+---
+"@latticexyz/explorer": patch
+---
+
+Initial release of the @latticexyz/explorer package. Explorer is a standalone tool designed to explore and manage worlds. This initial release supports local worlds, with plans to extend support to any world in the future.
+
+## Using explorer
```
holic, line 7: I think everything below this line should be in the README instead of the changeset. We can point to the README or docs from the changeset.

examples/local-explorer/explorer-watcher.mjs
```diff
@@ -0,0 +1,92 @@
...
+const EXPLORER_DIR = path.resolve(__dirname, "..", "..", "packages", "explorer");
+const WORLDS_JSON_PATH = path.resolve(__dirname, "packages", "contracts", "worlds.json");
+
+const MODE = process.env.MODE || "development";
+const CHAIN_ID = parseInt(process.env.CHAIN_ID || 31337);
+let explorerProcess = null;
+
+async function startExplorer() {
+  process.env.INIT_PWD = process.cwd();
+  process.env.NEXT_PUBLIC_CHAIN_ID = CHAIN_ID;
```
holic, line 19: mutating the env feels dangerous, especially when we can just pass this into `spawn` via `env` arg

examples/local-explorer/explorer-watcher.mjs
holic, line 1: Is there a way to expose this from explorer itself rather than this being part of user land code?

maybe you pass in the path to the `worlds.json` to the binary and it figures out how to resolve it from the cwd and also watch it

alternatively, we could leave out the watcher for the first version and revisit later, so that we can land this sooner than later and not block on the watcher (which feels like a nice-to-have rather than a requirement?)

examples/local-explorer/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "mud-template-react",
```
holic, line 2:
```suggestion
  "name": "local-explorer",
```

packages/explorer/src/app/worlds/[worldAddress]/interact/Form.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { AbiFunction } from "viem";
+import { useDeferredValue, useState } from "react";
+import { Input } from "../../../../components/ui/Input";
+import { useHashState } from "../../../../hooks/useHashState";
+import { cn } from "../../../../lib/utils";
+import { FunctionField } from "./FunctionField";
+
+type Props = {
+  data: {
+    abi: AbiFunction[];
```
Comment (holic) [inline packages/explorer/src/app/worlds/[worldAddress]/interact/Form.tsx:13]: ooc why is this nested inside `data`?

packages/explorer/src/app/worlds/[worldAddress]/interact/page.tsx
```diff
@@ -0,0 +1,24 @@
+import { headers } from "next/headers";
+import { Hex } from "viem";
+import { Form } from "./Form";
+
+async function getABI(worldAddress: Hex) {
+  const headersList = headers();
+  const protocol = headersList.get("x-forwarded-proto");
+  const host = headersList.get("host");
+
+  const res = await fetch(`${protocol}://${host}/api/world?address=${worldAddress}`);
```
Comment (holic) [inline packages/explorer/src/app/worlds/[worldAddress]/interact/page.tsx:10]:
```suggestion
  const res = await fetch(`${protocol}://${host}/api/world?${new URLSearchParams({ address: worldAddress })}`);
```

packages/explorer/src/app/worlds/[worldAddress]/utils/getFieldIndex.ts
```diff
@@ -0,0 +1,13 @@
+import { isDynamicAbiType, isStaticAbiType } from "@latticexyz/schema-type/internal";
+
+export function getFieldIndex(valueSchema: Record<string, string>, fieldName: string) {
```
Comment (holic) [inline packages/explorer/src/app/worlds/[worldAddress]/utils/getFieldIndex.ts:3]: might be worth adding this into protocol-parser

packages/explorer/src/components/ui/Button.tsx
```diff
@@ -0,0 +1,44 @@
+import { type VariantProps, cva } from "class-variance-authority";
+import * as React from "react";
+import { Slot } from "@radix-ui/react-slot";
+import { cn } from "../../lib/utils";
+
+const buttonVariants = cva(
+  // eslint-disable-next-line max-len
+  "inline-flex items-center justify-center whitespace-nowrap rounded-md text-sm font-medium ring-offset-background transition-colors focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring focus-visible:ring-offset-2 disabled:pointer-events-none disabled:opacity-50",
```
holic, line 8: no change expected here, but just wanted to mention a thing I've been doing lately to make this a bit more readable is splitting these up into multiple strings, where each string groups properties by use (e.g. layout things vs text formatting vs state modifiers):

```suggestion
  "inline-flex items-center justify-center rounded-md",
  "whitespace-nowrap text-sm font-medium",
  "ring-offset-background transition-colors",
  "focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring focus-visible:ring-offset-2",
  "disabled:pointer-events-none disabled:opacity-50",
```

packages/explorer/src/consts.ts
```diff
@@ -0,0 +1,28 @@
+import { Hex } from "viem";
+import { privateKeyToAccount } from "viem/accounts";
+
+export const NON_EDITABLE_TABLES = ["__chainState", "__mudStoreTables"];
```
holic, line 4: we could prob import this from store-sync/sqlite so we don't have to hardcode here

packages/explorer/src/consts.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export const PRIVATE_KEYS: Hex[] = [
```
holic, line 6: can we clarify that this is local/anvil private keys?

packages/world/ts/deduplicateAbi.ts
```diff
@@ -0,0 +1,24 @@
...
+
+function hasNamedInputs(entry: AbiItem) {
+  if (!("inputs" in entry) || !entry.inputs || !entry.inputs.length) {
+    return false;
+  }
+  return entry.inputs.some((input) => input.name);
+}
+
+// during deduplication, favor ABI items with named inputs
+export function deduplicateAbi(abi: Abi) {
```
holic, line 11: do we need this anymore?

holic (commented):
nice! just a few more small things

packages/explorer/src/components/ui/Button.tsx
karooolis: Yeah, that's nicer, agreed. It would make for a nice prettier plugin

packages/world/ts/deduplicateAbi.ts
karooolis: Argh absolutely not

packages/explorer/src/app/worlds/[worldAddress]/interact/Form.tsx
Comment (karooolis) [reply packages/explorer/src/app/worlds/[worldAddress]/interact/Form.tsx]: Generally, I try to pass only the least required props and when it comes to backend response, prefer to destructure it in the component where it's used. But tbh, probably applying the principle of "least-privilege" makes sense even here. Will apply it moving forward

packages/explorer/src/app/worlds/[worldAddress]/utils/getFieldIndex.ts
Comment (karooolis) [reply packages/explorer/src/app/worlds/[worldAddress]/utils/getFieldIndex.ts:3]: totally, created a PR https://github.com/latticexyz/mud/pull/3046

packages/explorer/src/consts.ts
karooolis: I've imported the names from `store-sync`, although it's pretty horrible due to how types are created in `drizzle-orm`. Will try to check again if I can come up with anything better

.changeset/swift-rabbits-appear.md
karooolis: totally, should have thought about that, adjusted the README.md with a link to it from the changeset

packages/explorer/src/consts.ts
```diff
@@ -0,0 +1,38 @@
...
+
+type Table = {
+  [key: symbol]: string;
+};
+
+const TableName = Symbol.for("drizzle:Name");
+export const NON_EDITABLE_TABLES = [
+  (chainState as unknown as Table)[TableName],
+  (mudStoreTables as unknown as Table)[TableName],
+];
```
holic, line 13: I think you can use `getTableName` from drizzle to do this

```ts
import { getTableName } from "drizzle-orm";
const internalTableNames = [chainState, mudStoreTables].map(getTableName);
```

but if we're not using drizzle here, it probably doesn't make much sense to add it just for this, so maybe we export these table names from store-sync/sqlite?

examples/local-explorer/explorer-watcher.mjs
karooolis: ah missed it, very good point, now passing environment variables as `spawn` args,

examples/local-explorer/explorer-watcher.mjs
karooolis: To give some context, there are 2 reasons why I've created the explorer. 

First, I was initially trying to separate the `explorer` package from how it should be set up as part of a project. So my thinking was that all the "glue" parts should not live inside the package itself. However, after your comment and having considered it again, it seems like it would indeed be better to resolve `worlds.json` in the package itself as long as the path is known.

The second reason for `explorer-watcher.mjs` existence is to simplify switching between `development` and `production` environments. It's not trivial to set up a development environment for the `explorer` so wanted to make it easily available by toggling a flag inside `mprocs.yaml`. Having said that, perhaps `local-explorer` example could always run in `dev` environment, while all the templates will be set up with `prod`?

All in all, I think it would still be nice to keep the `explorer-watcher.mjs` as part of the `local-explorer` example but agreed that the `worlds.json` import, and world address resolving, can be moved to the package itself. So it would be left to only serve the purpose of toggling between `prod` and `dev`. Would like to hear your thoughts on that.

packages/explorer/src/consts.ts
karooolis: Yeah, tried to do that but as you said, not great to import the whole package for one function. And then it still didn't work because of "barrel files imports" issue due to some incompatibility between Next.js and the package itself. So sure, exporting from `store-sync/sqlite` makes most sense to me, will create a PR.

examples/local-explorer/explorer-watcher.mjs
holic: my thought is that the local-explorer example is meant to demonstrate how we'd eventually make use of the explorer in e.g. templates, and including this watcher script in every template/MUD project is a little sad when we can probably pull all this into the package itself

something like `explorer --watch` and a separate arg for `--worldsFile` (but give this a sane default)

karooolis: Definitely, wouldn't want to ask people to include the whole file in their own projects. This is just for the purpose of an example. Similarly to how `client` in `templates` resolves `worlds.json`, for example, although all that setup would not be necessary for a deployed project.

Having said that, I agree with the premise that we should encapsulate as much as possible inside the package itself. Will rearrange it in such a way where `explorer` can be run with various flags, so `explorer-watcher.mjs` in some form might still be kept in the package but will experiment further on the best way to do it. Initially, the idea was that it could be run with environment variables but flags is definitely better, good suggestion.

holic: I'm happy to leave this here as an example if the intent is _not_ to bring this over to templates in the future, then we can iterate towards this being part of explorer itself that templates can lean on with CLI flags instead of scripts

karooolis: I'm going to look a bit more into this now before deciding on how to proceed with templates down the road but we're on the same page regarding that no extra files should be required for standard use

karooolis: I moved the script into the package itself. It also accepts CLI args as suggested, and it feels much cleaner now tbf. Will update README as well to reflect the changes on how the explorer should be run. 

The accepted CLI args include:
- `worldAddress`
- `worldsConfigPath` - if it's configured, the script aims to resolve world address from here
- `chainId`
- `port`
- `env` (production | development)

examples/local-explorer/LICENSE
holic, line 1: delete?

packages/explorer/explorer-runner.mjs
holic, line 1: can we move this into a `bin` dir?

packages/explorer/package.json
```diff
@@ -0,0 +1,64 @@
+{
+  "name": "@latticexyz/explorer",
+  "version": "2.1.0",
+  "private": true,
+  "description": "Explorer is a tool for visually exploring and manipulating the state of worlds",
+  "bin": {
+    "explorer": "./explorer-runner.mjs"
```
holic, line 7: do we need to build this?

packages/explorer/explorer-runner.mjs
holic: also do we need to add
```
#!/usr/bin/env node
```
or similar at the top so its treated as a node thing not a bash script?

packages/explorer/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+   Alternatively, if you have a worlds configuration file:
+
+   ```sh
+   npx @latticexyz/explorer --worldsConfigPath <PATH_TO_WORLDS_CONFIG>
+   ```
+
+   Note: You can use `@latticexyz/store-indexer` for indexing your world's data.
+
+## CLI Arguments
```
holic, line 31:
```suggestion
## CLI arguments
```
sentence case 🙏

packages/explorer/explorer-runner.mjs
```diff
@@ -0,0 +1,110 @@
...
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+
+const argv = minimist(process.argv.slice(2));
+const port = argv.port || 13690;
+const chainId = argv.chainId || 31337;
+const env = argv.env || "production";
+const indexerDbPath = argv.indexerDbPath || "indexer.db";
+const worldsConfigPath = argv.worldsConfigPath || null;
```
holic, line 17: can we call this `worldsFile` to match the config?

packages/explorer/explorer-runner.mjs
```diff
@@ -0,0 +1,110 @@
...
+import path from "path";
+import process from "process";
+import { fileURLToPath } from "url";
+import { spawn } from "child_process";
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+
+const argv = minimist(process.argv.slice(2));
+const port = argv.port || 13690;
```
holic, line 13:
```suggestion
const port = argv.port || process.env.PORT || 13690;
```

packages/explorer/explorer-runner.mjs
```diff
@@ -0,0 +1,110 @@
...
+    args = ["dev"];
+  }
+
+  explorerProcess = spawn(command, args, {
+    cwd: __dirname,
+    stdio: "inherit",
+    env: {
+      ...process.env,
+      PORT: port,
+      INIT_PWD: process.cwd(),
```
holic, line 39: "cwd" is "current working directory" and "pwd" is "print working directory", so this env arg is kinda confusing

doesn't look like we're overriding this so maybe just ditch it for now?

packages/explorer/explorer-runner.mjs
```diff
@@ -0,0 +1,110 @@
...
+import { fileURLToPath } from "url";
+import { spawn } from "child_process";
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+
+const argv = minimist(process.argv.slice(2));
+const port = argv.port || 13690;
+const chainId = argv.chainId || 31337;
+const env = argv.env || "production";
```
holic, line 15:
```suggestion
const env = argv.env || process.NODE_ENV || "production";
```

packages/explorer/explorer-runner.mjs
```diff
@@ -0,0 +1,110 @@
...
+const indexerDbPath = argv.indexerDbPath || "indexer.db";
```
holic, line 16:
```suggestion
const indexerDbPath = argv.indexerDbPath || process.env.INDEXER_DB_PATH || "indexer.db";
```

packages/explorer/explorer-runner.mjs
```diff
@@ -0,0 +1,110 @@
...
+const worldsConfigPath = argv.worldsConfigPath || null;
+
+let worldAddress = argv.worldAddress || null;
```
holic, line 19:
```suggestion
let worldAddress = argv.worldAddress || process.env.WORLD_ADDRESS || null;
```

packages/explorer/explorer-runner.mjs
```diff
@@ -0,0 +1,110 @@
...
+
+  explorerProcess = spawn(command, args, {
+    cwd: __dirname,
+    stdio: "inherit",
+    env: {
+      ...process.env,
+      PORT: port,
+      INIT_PWD: process.cwd(),
+      WORLD_ADDRESS: worldAddress,
+      INDEXER_DB_PATH: indexerDbPath,
```
holic, line 41: rather than feeding in `INIT_PWD` maybe we just do a `path.join(process.cwd(), indexerDbPath)` here?

holic, line 41: if we do this, maybe we don't need `INIT_PWD`?
```suggestion
      INDEXER_DB_PATH: path.join(process.cwd(), indexerDbPath),
```

packages/explorer/explorer-runner.mjs
```diff
@@ -0,0 +1,110 @@
...
+    }
+  } catch (error) {
+    console.error("Error reading worlds.json:", error);
+    return null;
+  }
+}
+
+async function restartExplorer() {
+  if (explorerProcess) {
+    explorerProcess.kill();
```
holic, line 67: I am wondering if `worlds.json` changes the address but we don't wipe the DB, what happens?

Or if `worlds.json` changes but the address _doesn't_ change?

packages/explorer/explorer-runner.mjs
karooolis: ah crap, yeah bad naming

packages/explorer/explorer-runner.mjs
karooolis: Oh that's nice. With the current setup the script is always run in `node` but it's still a good point in case somebody tries to run it like `./explorer.mjs`, updated. And moved to `bin`.

packages/explorer/package.json
karooolis: it doesn't have to be built since it can be used without transpilers but this comment made me realize that perhaps this script should be converted to typescript, as it's done in `store-indexer` package, etc. So changed it to `explorer.ts` and linking the `explorer` command to the transpiled executable i.e. `./dist/explorer.mjs`.

packages/explorer/explorer-runner.mjs
karooolis: Hehe all good, should think through all the scenarios and consider if it's worth handling. If `worlds.json` changes, only in case the `worldAddress` has actually changed will the process get restarted. In the context of development environment, it's probably most relevant when the `worlds.json` is initially empty, and only later the address is set so the `explorer` gets set up properly without manual intervention.

As for `indexer.db`, the `explorer` shows whatever is included in the database. So for v0 I don't think it's an issue since we are focusing on local development environment where there is only 1 database to fetch from, plus as you said it all happens on local anvil chain. But in the future versions when this actually becomes "worlds" explorer, we should consider this more carefully.

packages/explorer/package.json
```diff
@@ -0,0 +1,71 @@
...
+  "name": "@latticexyz/explorer",
+  "version": "2.1.1",
+  "private": true,
+  "description": "Explorer is a tool for visually exploring and manipulating the state of worlds",
+  "bin": {
+    "explorer": "./dist/explorer.mjs"
+  },
+  "scripts": {
+    "build": "pnpm run build:explorer && pnpm run build:bin",
+    "build:bin": "tsup && pnpm link .",
```
holic, line 11: why link?

packages/explorer/package.json
```diff
@@ -0,0 +1,71 @@
+{
+  "name": "@latticexyz/explorer",
+  "version": "2.1.1",
+  "private": true,
+  "description": "Explorer is a tool for visually exploring and manipulating the state of worlds",
```
holic, line 5: are we able to use `"type": "module"` here? that would maybe let us avoid `.mjs` etc everywhere?

packages/explorer/package.json
karooolis: ah, it was useful in the beginning for testing the `explorer` command but actually not really needed, removed

packages/explorer/package.json
```diff
@@ -0,0 +1,71 @@
+{
+  "name": "@latticexyz/explorer",
+  "version": "2.1.1",
+  "private": true,
```
holic, line 4: either we should remove this or, if we're not yet ready to publish, remove the changeset

.changeset/swift-rabbits-appear.md
```diff
@@ -0,0 +1,7 @@
+---
+"@latticexyz/explorer": patch
```
holic, line 2: this should probably be `minor`

packages/explorer/package.json
karooolis: the idea is to low-key publish it already, removed `private` 😬

.changeset/swift-rabbits-appear.md
karooolis: Updated. Does it require version bump? Based on my understanding no because none of the existing packages are affected

packages/explorer/package.json
karooolis: Works with `type: module`! Updated, no more `mjs`

.changeset/swift-rabbits-appear.md
holic: we currently have all MUD packages tied together in their versioning/release cycle, so in the context of MUD it's a new (and big!) feature, so I think tagging it as a bump to v2.2 makes the most sense?

karooolis: Agreed on v2.2!

packages/explorer/tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": ["../../packages/common/tsconfig.base.json"],
   "compilerOptions": {
+    "lib": ["DOM", "ESNext"],
```
holic, line 4: this is already in the base config afaik

karooolis: nice catch! removed, couldn't spot any more redundant ones

pnpm-lock.yaml
holic, line 1: one last request:
```
git fetch origin
git checkout origin/main pnpm-lock.yaml
rm -rf **/node_modules
pnpm install
```
just to ensure all the various merging and such over time hasn't mucked too much with the lockfile

karooolis: updated, good idea

holic (approved):
let's gooooooo

PR #3002, opened 2024-08-02
WIP: typescript performance

alvrs:


packages/store-sync/src/common.ts
```diff
@@ -75,7 +75,7 @@ export type SyncOptions<config extends StoreConfig = StoreConfig> = {
    *
    * [0]: https://viem.sh/docs/clients/public.html
    */
-  publicClient: PublicClient;
+  publicClient: any; //PublicClient;
```
alvrs, line 78: this change alone reduced overall type instantiations in `setup()` by 66%, see https://github.com/latticexyz/mud/compare/b6ce4340cf02a283f03417e074978369e3b15cd2...289d1d1

PR #3007, opened 2024-08-05
chore: fetch main branch when running prerelease action

Kooshaba:
the main branch wasn't present in the action and was failing when attempting to checkout the changelog file


holic:
I think this PR came out of running the wrong action (prerelease instead of snapshot)

If we need this, I wonder if it would be better to use the `if` option to check that its only run for `main` branch?

Alternatively, we could disable this action for now to prevent misuse until we actually need it.

PR #3012, opened 2024-08-06
refactor(cli,world): move helper functions into world package, add getWorldAbi

karooolis:
`getWorldAbi` pulled out of #2849


packages/cli/src/deploy/debug.ts
```diff
@@ -1,10 +0,0 @@
-import { debug as parentDebug } from "../debug";
-
-export const debug = parentDebug.extend("deploy");
-export const error = parentDebug.extend("deploy");
```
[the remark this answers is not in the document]
karooolis: I moved it to `packages/world/ts/debug.ts`, it is needed there as well in the helper functions that got moved. Should I keep two copies then, one in `cli`, one in `world`?

packages/world/ts/getWorldDeploy.ts
holic, line 1: do we need to move this? I don't think this is used by any helpers here

packages/world/ts/logsToWorldDeploy.ts
holic, line 1: similarly, I don't think we need to move this

packages/world/ts/common.ts
```diff
@@ -0,0 +1,68 @@
...
+  readonly systemId: Hex;
+  readonly allowAll: boolean;
+  readonly allowedAddresses: readonly Hex[];
+  readonly allowedSystemIds: readonly Hex[];
+  readonly functions: readonly WorldFunction[];
+};
+
+export type DeployedSystem = Omit<System, "abi" | "prepareDeploy" | "deployedBytecodeSize" | "allowedSystemIds"> & {
+  address: Address;
+};
```
holic, line 68: I would rather not move all of these deployment-specific types over and instead just move what we need for these helpers and adjust their arguments accordingly, e.g.

```diff
 export async function getResourceIds({
   client,
-  worldDeploy,
+  worldAddress,
+  fromBlock,
+  toBlock,
 }: {
   readonly client: Client;
-  readonly worldDeploy: WorldDeploy;
+  worldAddress: Hex,
+  fromBlock: bigint,
+  toBlock: bigint,
 }): Promise<readonly Hex[]> {

File: packages/world/ts/getWorldDeploy.ts
Comment (karooolis) [reply packages/world/ts/getWorldDeploy.ts:1]: I need this because of how `getWorldAbi` is expecting the world deploy to be passed as one of the parameters https://github.com/latticexyz/mud/pull/2849/files#diff-c387c5680d2d7ed3fc8008e9769b2c37be5e4e2a0f11dda28f02521618047d1fR48-R53 . Then the world deploy param is used in `getResourceIds({ client, worldDeploy })`, `getFunctions({ client, worldDeploy })`,  `getResourceAccess({ client, worldDeploy })`, `getTableValue`, and maybe some more.

File: packages/world/ts/logsToWorldDeploy.ts
Comment (karooolis) [reply packages/world/ts/logsToWorldDeploy.ts:1]: `logsToWorldDeploy` is used in `getWorldDeploy`. But if `getWorldDeploy` is not needed, then will not move this one as well.

File: packages/world/ts/common.ts (hunk 2 above)
Comment (karooolis) [reply packages/world/ts/common.ts:68]: Makes sense, will do

File: packages/cli/src/deploy/debug.ts (hunk 1 above)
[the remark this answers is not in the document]
Comment (holic) [reply packages/cli/src/deploy/debug.ts:4]: yes, the loggers should always be local to their own source, not imported from another project, otherwise things using `debug` in CLI get logged out as if it was from the world package

File: packages/world/ts/getWorldDeploy.ts
Comment (holic) [reply packages/world/ts/getWorldDeploy.ts:1]: if we change the helper signatures to not be deployment specific and take in just what they need (i.e. world address, from/to block), we won't need this!

File: packages/cli/src/deploy/debug.ts (hunk 1 above)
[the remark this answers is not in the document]
Comment (karooolis) [reply packages/cli/src/deploy/debug.ts:4]: Got it, didn't think of that. Moved back

File: packages/world/ts/getWorldDeploy.ts
Comment (karooolis) [reply packages/world/ts/getWorldDeploy.ts:1]: Ah that's right, just removed the `worldDeploy` dependency everywhere, and updated `getSystems` params to be:

```typescript
export async function getSystems({
  client,
  worldAddress,
  stateBlock,
  deployBlock,
}: {
  readonly client: Client;
  readonly worldAddress: Address;
  readonly stateBlock: bigint;
  readonly deployBlock: bigint;
}): Promise<readonly DeployedSystem[]> {
...
```

Also, now resolving `address`, `stateBlock` and `deployBlock` inside `getWorldAbi()`. What do you think?

packages/world/ts/common.ts
karooolis: `getSystems` no longer has `WorldDeploy` as param. But it returns `DeployedSystem` as promise so still need to move it in such case. Or do you have other suggestions?

packages/world/ts/getWorldDeploy.ts
karooolis: And updated all other helper functions to take in only the required params i.e. no more `worldDeploy`.

packages/world/ts/getWorldAbi.ts
```diff
@@ -0,0 +1,39 @@
...
+
+import { getSystems } from "./getSystems";
+import { functionSignatureToAbiItem } from "./functionSignatureToAbiItem";
+
+export async function getWorldAbi({
+  client,
+  worldAddress,
+}: {
+  readonly client: Client;
+  readonly worldAddress: Address;
```
holic, line 13: I think this could also just take in `fromBlock` and `toBlock` to pass down into `getSystems`, rather than fetching the deploy block in here.

packages/world/ts/debug.ts
```diff
@@ -0,0 +1,10 @@
+import createDebug from "debug";
+
+export const debug = createDebug("mud:cli");
+export const error = createDebug("mud:cli");
```
holic, line 4:
```suggestion
export const debug = createDebug("mud:world");
export const error = createDebug("mud:world");
```

packages/world/ts/common.ts
holic: what we need for the purposes of world ABI is just the world function selectors for systems, so I don't think we even need `getSystems`? we can just call to `getFunctions`?

packages/world/ts/getWorldAbi.ts
karooolis: Good idea, accepting `fromBlock` and `toBlock` now.

packages/world/ts/common.ts
karooolis: Ah great suggestion, changed to `getFunctions` and moved `getSystems` back into `cli` package.

holic (approved):
overall good, but might consider including the entire base interface of the world alongside the world-scoped system functions

packages/world/ts/getWorldAbi.ts
```diff
@@ -0,0 +1,25 @@
...
+}): Promise<Abi> {
+  const worldFunctions = await getFunctions({
+    client,
+    worldAddress: getAddress(worldAddress),
+    fromBlock,
+    toBlock,
+  });
+  const worldAbi = worldFunctions.map((func) => functionSignatureToAbiItem(func.signature));
+
+  return worldAbi;
```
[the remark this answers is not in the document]
holic: otherwise, maybe we rename this function to `getWorldFunctionsAbi`?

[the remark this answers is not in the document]
karooolis: Good idea to add it here, have been doing it on the consuming-side i.e. world explorer. Just added.

karooolis:
> overall looks good, but might consider including the entire base interface of the world alongside the world-scoped system functions

great! Good idea, just added.

packages/world/ts/getWorldAbi.ts
```diff
@@ -0,0 +1,27 @@
+import { Client, Abi, Address, getAddress } from "viem";
+import IBaseWorldAbi from "../out/IBaseWorld.sol/IBaseWorld.abi.json";
+import { functionSignatureToAbiItem } from "./functionSignatureToAbiItem";
+import { getFunctions } from "./getFunctions";
+
+export async function getWorldAbi({
+  client,
+  worldAddress,
+  fromBlock,
+  toBlock,
```
holic, line 10: we could probably default this
```suggestion
  toBlock = 'latest',
```

holic (approved):

packages/world/ts/getWorldAbi.ts
karooolis: Agreed, can change to it after trying it out with world explorer.

holic (approved):

PR #3021, opened 2024-08-09
wip system deploys

holic:
We're exploring what it might look like for package-exportable representation of MUD resources, where the current MUD config is only a partial representation since systems are defined by what's on the filesystem. This adds an intermediate build step and JSON representation of those systems so that we can then export a complete view of what's to be deployed.

TODO
- [ ] extract and include system functions to register onchain
- [ ] translate `accessList` to a set of addresses + system IDs? (could still be done by deployer but might be nice to move here)
- [ ] update deployer to use this new JSON file


holic:
a thing I fiddled with in https://github.com/latticexyz/mud/pull/3021/commits/5c7ba7606b676cf01df9b18fb6a39aabd1cc22ad is a "pure" representation of a given system and all its dependent bytecode, where references to public libs in the bytecode are replaced by the library bytecode that the deployer would "unwind" and deploy deepest first, swap with address, then continue outward until the system is deployed

this is ~double the filesize as the current state where we now reference "deployables" and have a top-level "deployables" object with bytecode of libraries

even though it's larger, I still like the former because it's more "pure" and easier to merge without naming conflicts across projects ("deployable" source file structures and contracts may be similarly named within a namespace), but hard to each opton until we get to the step where we start combine configs/manifests for the entire world deploy

holic:
this is _almost_ a representation of the things needed to deploy but aren't represented by config output, just missing system functions to be registered (wouldn't be too hard to add)

then would need to bake this new JSON into deployer and figure out how to build a metadata package with this new format + bake in as a default during world deploy

but worried about the number of unknowns still here when we get to the config combining step that I am thinking of pausing this work for now to just unblock metadata via regular world module then revisit this later

PR #3022, opened 2024-08-09
fix(world): workaround resolve json issue

holic:
can either turn on `resolveJsonModule` but get weaker types or turn it off and get strong types from our generated `.json.d.ts`, but not both

this attempts to work around this issue by manually importing the ABI types we need when we need them 😭


holic:
this was a workaround for issues in https://github.com/latticexyz/mud/pull/3001 when extending the workspace root tsconfig, but we decided we'll just let explorer have its own tsconfig (since its never imported as a package, only used as a binary/service)

PR #3025, opened 2024-08-12
fix(world): dedupe functions in getWorldAbi result

karooolis:
Dynamically fetched ABI from a deployed world returns user-created custom functions along with default world methods such as `batchCall`. However, it is not able to fetch system methods such as `setField` that should also be included in the final ABI for completion.

To include all system methods, the dynamically fetched ABI is concatenated with statically generated World's ABI, resulting in some duplicated functions.

`deduplicateABI` removes all the duplicate functions, and in case a duplicate is encountered, favors the ABI item that has named imports (i.e. statically generated World ABI), so that the ABI contains as much information as possible. In `explorer` case, for example, it's helpful to see the function parameter names when interacting with various functions.


packages/world/ts/deduplicateAbi.ts
```diff
@@ -0,0 +1,24 @@
...
+  for (const entry of abi) {
+    const key = "name" in entry ? `${entry.type}_${entry.name}` : entry.type;
+    const existingEntry = uniqueEntries.get(key);
+
+    if (!existingEntry || (hasNamedInputs(entry) && !hasNamedInputs(existingEntry))) {
+      uniqueEntries.set(key, entry);
+    }
+  }
+
+  return Array.from(uniqueEntries.values());
```
holic, line 23: we have a `uniqueBy` helper in common that might make this easier to implement

packages/world/ts/deduplicateAbi.ts
```diff
@@ -0,0 +1,24 @@
...
+  }
+  return entry.inputs.some((input) => input.name);
+}
+
+// during deduplication, favor ABI items with named inputs
+export function deduplicateAbi(abi: Abi) {
+  const uniqueEntries = new Map();
+
+  for (const entry of abi) {
+    const key = "name" in entry ? `${entry.type}_${entry.name}` : entry.type;
```
holic, line 15: I think this is assuming each function is unique by name but Solidity supports overloads 

might make use of abitype or viem helpers to convert this to a function signature before comparing

holic:
can you fill in the PR description about why this is important to do? Unclear if the side effects of duplicate entries (and seems intended given solidity supports overloads)

packages/world/ts/deduplicateAbi.ts
```diff
@@ -0,0 +1,24 @@
+import { Abi, AbiItem } from "viem";
+
+function hasNamedInputs(entry: AbiItem) {
+  if (!("inputs" in entry) || !entry.inputs || !entry.inputs.length) {
+    return false;
+  }
+  return entry.inputs.some((input) => input.name);
```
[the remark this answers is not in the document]
karooolis: I am not sure if there is any downside to have both, named and unnamed inputs. I am favoring named inputs so I can display more useful information to users in the explorer. And I am also deduplicating same functions just because it might be confusing to have two equivalent `batchCall` shown in the UI.

As for counting, I am not sure how helpful that would be in this case because the world ABI that gets fetched dynamically does not have any named inputs while the static (imported) ABI has all of its inputs named. So it's always either 0 named inputs, or some number > 0.

packages/world/ts/deduplicateAbi.ts
karooolis: Ah that's very good point, and a massive oversight. Adjusted to take into this into account.

packages/world/ts/deduplicateAbi.ts
```diff
@@ -0,0 +1,29 @@
...
+}
+
+// favor ABI items with named inputs when deduplicating
+export function deduplicateAbi(abi: Abi) {
```
holic, line 11: can we add tests for this?

karooolis:
> can you fill in the PR description about why this is important to do? Unclear if the side effects of duplicate entries (and seems intended given solidity supports overloads)

Definitely, solidity overloads should be taken into account but ABI should contain no duplicates imo. Included problem description.

karooolis:
> thinking about another approach where, instead of deduplicating an ABI as a whole, we just construct to the world a bit differently for our use case, starting with the base ABI (since it has named params) and add the functions registered on the world that we haven't seen yet (using function selectors as the "key" for uniqueness check, since its ultimately what is called)
> 
> ```ts
> const baseFunctionSelectors = IBaseWorldAbi.filter(isAbiFunction).map(toFunctionSelector);
> const worldFunctionsAbi = worldFunctions
>   .map((func) => functionSignatureToAbiItem(`function ${func.signature}`))
>   .filter((abiItem) => !baseFunctionSelectors.includes(toFunctionSelector(abiItem)));
> 
> const worldAbi = [
>   ...IBaseWorldAbi,
>   ...worldFunctionsAbi,
> ];
> ```

I ended up liking this approach more too, somehow feels more clean. Only renamed the function to `concatBaseAbi` instead of `deduplicate`, and added a test.

packages/world/ts/concatBaseAbi.test.ts
```diff
@@ -0,0 +1,12 @@
+import { describe, expect, it } from "vitest";
+import { Abi } from "viem";
+import worldAbi from "./mocks/worldAbi";
+import { concatBaseAbi } from "./concatBaseAbi";
+import combinedAbi from "./mocks/combinedAbi";
+
+describe("World ABI", () => {
+  it("should concat base and world ABI", () => {
+    const abi = concatBaseAbi(worldAbi as Abi);
+    expect(abi).toMatchObject(combinedAbi);
```
holic, line 10: tests are meant to describe and demonstrate behavior, but that's really hard to see when these objects are huge and hidden away in files

can we do some smaller tests instead of using mocks? like

```
const abi = concatBaseAbi([{type: "function", name: "addTask", ...}]);
expect(abi).toContainEqual({ type: "function", name: "addTask", ... });
```

karooolis: agreed, I made it so verbose to make sure it works with "real-world" data, and also handles correctly events deduplication, etc. and then also knew what output to expect. But regardless, it could have been split into smaller pieces.

Anyhow, now it's clear that since the fetched world ABI only contains functions, the testing can be simplified, updated to your suggestion.

packages/world/ts/concatBaseAbi.ts
```diff
@@ -0,0 +1,16 @@
+import { Abi, AbiItem, AbiFunction, toFunctionSelector } from "viem";
+import IBaseWorldAbi from "../out/IBaseWorld.sol/IBaseWorld.abi.json";
+
+function isAbiFunction(abiItem: AbiItem): abiItem is AbiFunction {
+  return abiItem.type === "function";
+}
+
+export function concatBaseAbi(worldAbi: Abi): Abi {
```
holic, line 8: sorry to keep going back and forth on this but now that this is only handling functions, I think this logic is worth inlining into `getWorldAbi` rather than being a separate helper

means we'll likely need to mock `getFunctions` if we want to be able to test it, but I think the logic is simple enough to not necessarily need it

karooolis: All good, just removed `concatbaseAbi` and adjusted the tests, lmk what you think

packages/world/ts/getWorldAbi.test.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  return {
+    getFunctions,
+  };
+});
+
+describe("World ABI", () => {
+  it("should concat base and world ABI", async () => {
+    const abi = await getWorldAbi({
+      client: testClient,
```
holic, line 18: if `getFunctions` is mocked, do we even need to return a real client here? can this just be a very basic `createTestClient()`?

packages/world/ts/getWorldAbi.ts
```diff
@@ -20,7 +24,10 @@ export async function getWorldAbi({
     fromBlock,
     toBlock,
   });
-  const worldFunctionsAbi = worldFunctions.map((func) => functionSignatureToAbiItem(func.signature));
+  const baseFunctionSelectors = (IBaseWorldAbi as Abi).filter(isAbiFunction).map(toFunctionSelector);
```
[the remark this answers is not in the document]
karooolis: Without it, even with type predicate, typescript fails to recognize in the `.map(..)` that it's of type `AbiFunction`

packages/world/ts/getWorldAbi.ts
```diff
@@ -20,7 +24,10 @@ export async function getWorldAbi({
...
+  const worldFunctionsAbi = worldFunctions
+    .map((func) => functionSignatureToAbiItem(func.signature))
+    .filter((abiItem) => !baseFunctionSelectors.includes(toFunctionSelector(abiItem as AbiFunction)));
```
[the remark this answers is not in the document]
karooolis: Adjusted to  return `AbiFunction` but in order for that to work, have to ensure that it indeed returns function ABI item, otherwise the function throws.

packages/world/ts/getWorldAbi.test.ts
karooolis: Good point, simplified `createTestClient` as much as I could

packages/world/ts/common.ts
```diff
@@ -7,3 +7,19 @@ export type WorldFunction = {
...
+
+export const anvilRpcUrl = `http://${anvilHost}:${anvilPort}/${poolId}`;
+
+export const testClient = createTestClient({
+  mode: "anvil",
+  transport: http(anvilRpcUrl),
+  pollingInterval: 10,
+})
+  .extend(publicActions)
+  .extend(walletActions);
```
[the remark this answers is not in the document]
karooolis: Removed, copied too hastily from another package

holic (approved):

Issue #2981, opened 2024-07-29
register resource labels


holic:
we need a table for resource labels (and likely for metadata in general) so that we can later hydrate a MUD config from these labels (when available)

PR #3026, opened 2024-08-13
feat(world-module-metadata): add metadata module

holic:
part of #2981


packages/cli/src/deploy/configToModules.ts
```diff
@@ -14,8 +14,18 @@ export async function configToModules<config extends World>(
   // TODO: remove/replace `forgeOutDir`
   forgeOutDir: string,
 ): Promise<readonly Module[]> {
+  const configModules = [
+    {
+      name: "metadata",
+      artifactPath: "@latticexyz/world-modules/out/MetadataModule.sol/MetadataModule.json",
```
holic, line 20: need to import this way otherwise we create a circular dep between CLI and world-modules

downside is this would require projects to have a world-modules dep in their package that uses `mud` CLI

maybe could make this more explicit as a peer dep for CLI?

packages/world-module-metadata/src/MetadataModule.sol
```diff
@@ -0,0 +1,51 @@
...
+      RESOURCE_SYSTEM,
+      namespace.getNamespace(),
+      "MetadataSystem"
+    );
+    if (!ResourceIds.getExists(metadataSystemId)) {
+      world.registerSystem(metadataSystemId, metadataSystem, true);
+      world.registerFunctionSelector(metadataSystemId, "setResource(bytes32,bytes32,string)");
+    }
+
+    world.transferOwnership(namespace, _msgSender());
```
holic, line 49: this pattern feels a little weird to me, but curious to hear what you think:

on first install, metadata module creates `metadata` namespace, registers tables/systems, hands ownership back to original caller

on second install (idempotent, would upgrade systems or add tables), you'd first need to hand ownership of the namespace to the module, then it hands it back after install

alvrs: i think this pattern is fine! makes it possible to upgrade, which is nice

packages/world-module-metadata/test/MetadataModule.t.sol
```diff
@@ -0,0 +1,115 @@
...
+    assertEq(NamespaceOwner.get(namespace), address(this));
+
+    // Installing again will revert because metadata namespace isn't owned by the module, so the module is unable to write to it.
+    vm.expectRevert(
+      abi.encodeWithSelector(IWorldErrors.World_AccessDenied.selector, namespace.toString(), address(metadataModule))
+    );
+    world.installModule(metadataModule, new bytes(0));
+
+    // Transferring the namespace to the module and installing should be a no-op/idempotent and the module will return namespace ownership.
+    // TODO: is this a security concern? could someone frontrun by putting a tx in between the two calls below?
```
alvrs, line 43: you can always make it atomic by doing both via a batchcall. But even without i don't think there is a security concern here, the module would just own the namespace, but unless the module has functions to do something malicious with or is upgradable, another tx in between couldn't do anything else with it

packages/world-module-metadata/ts/build.ts
```diff
@@ -0,0 +1,17 @@
+import path from "node:path";
+import { fileURLToPath } from "node:url";
+import { tablegen } from "@latticexyz/store/codegen";
+import { worldgen } from "@latticexyz/world/node";
+
+/**
+ * To avoid circular dependencies, we run a very similar `build` step as `cli` package here.
+ */
+
+// TODO: move tablegen/worldgen to CLI commands from store/world we can run in package.json instead of a custom script
```
alvrs, line 10: wanna open an issue for this? seems like a "good first issue"

alvrs (approved):

packages/world-module-metadata/ts/build.ts
holic: done: https://github.com/latticexyz/mud/issues/3030

PR #3028, opened 2024-08-13
feat(cli): register resource labels

### LINKED ISSUE #2981: register resource labels

holic:
builds on #3026 
closes #2981

the changeset in #3026 captures this change


alvrs (dismissed):

packages/cli/src/deploy/configToModules.ts
```diff
@@ -27,6 +27,8 @@ export async function configToModules<config extends World>(
   forgeOutDir: string,
 ): Promise<readonly Module[]> {
   const defaultModules: Module[] = [
+    // TODO: replace metadata install here with custom logic inside `ensureModules` or an`ensureDefaultModules` to check
+    //       if metadata namespace exists, if we own it, and if so transfer ownership to the module before reinstalling
```
holic, line 31:
```suggestion
    // TODO: replace metadata install here with custom logic inside `ensureModules` or an `ensureDefaultModules` to check
    //       if metadata namespace exists, if we own it, and if so transfer ownership to the module before reinstalling
    //       (https://github.com/latticexyz/mud/issues/3035)
```

PR #3033, opened 2024-08-15
docs(state-sync/dozer): first version

qbzzt:


docs/pages/state-query/dozer.mdx
```diff
@@ -0,0 +1,573 @@
...
+If there is a dozer instance serving a blockchain, as there is for [Redstone](https://redstone.xyz/) and [Garnet](https://garnetchain.com/docs/what-is-redstone), you can use it to run queries on the data of any `World` on that blockchain.
+
+The query language is a subset of [the SQL `SELECT` command](<https://en.wikipedia.org/wiki/Select_(SQL)>).
+
+## Example `World`
+
+On Garnet there is a `World` at address [`0x95F5d049B014114E2fEeB5d8d994358Ce4FFd06e`](https://explorer.garnetchain.com/address/0x95F5d049B014114E2fEeB5d8d994358Ce4FFd06e) that runs a slightly modified version of [the React template](https://github.com/latticexyz/mud/tree/main/templates/react).
+You can see the data schema for the `World` [in the block explorer](https://explorer.garnetchain.com/address/0x95F5d049B014114E2fEeB5d8d994358Ce4FFd06e?tab=mud).
+
+### Create a client to access the `World`
```
alvrs, line 15: should we move this section below the curl section given it's not needed for the curl section?

docs/pages/state-query/dozer.mdx
```diff
@@ -0,0 +1,573 @@
...
+   ```sh copy
+   pnpm create mud@latest tasks --template react
+   cd tasks
+   pnpm dev
+   ```
+
+1. [Browse to the application](http://localhost:3000/?chainId=17069&worldAddress=0x95f5d049b014114e2feeb5d8d994358ce4ffd06e).
+   The URL specifies the `chainId` and `worldAddress` for the `World`.
+
+1. In Mud DevTools see your account address and [fund it on Garnet](https://garnetchain.com/faucet).
```
alvrs, line 31:
```suggestion
1. In MUD DevTools see your account address and [fund it on Garnet](https://garnetchain.com/faucet).
```
always `MUD` in regular text

docs/pages/state-query/dozer.mdx
```diff
@@ -0,0 +1,573 @@
...
+         },
+         key: ["id"],
+       },
+     },
+   });
+   ```
+
+   </CollapseCode>
+
+The Garnet dozer instance is at `https://dozer.mud.garnetchain.com`.
```
alvrs, line 68: Would put the urls for both redstone and garnet at the very top of the doc

docs/pages/state-query/dozer.mdx
```diff
@@ -0,0 +1,573 @@
...
+
+## cUrl queries
+
+You can run dozer queries by communicating directly with the server's API.
+
+### Simple query
+
+This query looks for some fields from a single table.
+
+1. Create a file, `query.json`, with this content.
```
alvrs, line 78: to simplify even more and have a "just copy and run this" experience, should we skip creating a file and parsing it with jq, and instead just put the query inline in the curl command?

docs/pages/state-query/dozer.mdx
```diff
@@ -0,0 +1,573 @@
...
+  ["taskcreator"],
+  ["0x428b1853e5ec29d35c84a218ec5170efc7621b58"],
+  ["0x735b2f2c662ebedffa94027a7196f0559f7f18a4"],
+  ["0x8225d72f2c39f3729d7f3fc03c6aa8731eaeef48"]
+]
+```
+
+</details>
+
+## Using dozer from a MUD client
```
alvrs, line 356: for the typescript section i think it would be useful to have the part about "how to modify a MUD template to hydrate from dozer", but additionally before that i would put a simpler example with just a single typescript file:

```ts
const config = defineStore({ ... });
const result = fetchRecordsSql({ 
  dozerUrl: ...,
  storeAddress: ...,
  queries: [
    selectFrom(...)
  ]
 });
```

docs/pages/state-query/dozer.mdx
qbzzt: IMHO, no. It's easier to edit a file than it is to edit the query on the command line.

docs/pages/state-query/dozer.mdx
qbzzt: I wasn't sure, because you can also use it to see what the curl results should be like. But once the explorer is out we can use that, so no need to worry about it.

docs/pages/state-query/dozer.mdx
qbzzt: OK. And after this is merged let's put them in the docs for the networks too (https://redstone.xyz/docs/network-info and the garnet equivalent)

Issue #1853, opened 2023-11-01
Combine `AccessControlLib` in `world-module` and `AccessControl` in `world`


alvrs:
> Though I understand why it's in `world-modules`, would it maybe be better to have a single `AccessControl` lib with `hasAccess`/`_hasAccess` etc, to have all the logic in 1 place and let it mirror the table methods?

_Originally posted by @dk1a in https://github.com/latticexyz/mud/pull/1793#discussion_r1378611057_

PR #3034, opened 2024-08-15
refactor(world): make AccessControl lib usable outside of world package

holic:
fixes https://github.com/latticexyz/mud/issues/1853


packages/world/gas-report.json
```diff
@@ -3,43 +3,43 @@
     "file": "test/AccessControl.t.sol",
     "test": "testAccessControl",
     "name": "AccessControl: hasAccess (cold)",
-    "gasUsed": 6521
+    "gasUsed": 9111
```
alvrs, line 6: I assume this increased because the tests still use the non-`_`-prefixed method, should we add variants of each of these for the `_` methods to compare and still have a gas report for the internally used methods?

alvrs (approved):

packages/world/gas-report.json
holic: I didn't think it was necessary to have gas reports for all the internal `_` method variants (since they're internal) and the gas use is captured by the tests using these internal methods (which didn't update in this PR since there was no gas change)

PR #3040, opened 2024-08-15
feat(stash): add stash, the new MUD client state library

alvrs:


packages/stash/src/decorators/default.ts
holic, line 1: `defaultActions.ts` to match what it exports?

packages/stash/src/common.ts
```diff
@@ -0,0 +1,167 @@
...
+
+export type StoreUpdatesSubscriber<config extends StoreConfig = StoreConfig> = (updates: StoreUpdates<config>) => void;
+
+export type StoreSubscribers<config extends StoreConfig = StoreConfig> = Set<StoreUpdatesSubscriber<config>>;
+
+export type Stash<config extends StoreConfig = StoreConfig> = {
+  /**
+   * Get a readonly reference to the current state
+   */
+  get: () => State<config>;
```
holic, line 156:
```suggestion
  readonly get: () => State<config>;
```

packages/stash/src/common.ts
```diff
@@ -0,0 +1,167 @@
...
+  /**
+   * Internal references for interacting with the state.
+   * @internal
+   * @deprecated Do not use this internal reference externally.
+   */
+  _: {
+    tableSubscribers: TableSubscribers;
+    storeSubscribers: StoreSubscribers;
+    state: MutableState<config>;
+  };
```
holic, line 166:
```suggestion
  readonly _: {
    readonly tableSubscribers: TableSubscribers;
    readonly storeSubscribers: StoreSubscribers;
    readonly state: MutableState<config>;
  };
```

packages/stash/src/createStash.ts
```diff
@@ -0,0 +1,55 @@
...
+  const stash = {
+    get: () => state,
+    _: {
+      state,
+      tableSubscribers,
+      storeSubscribers,
+    },
+  } satisfies Stash;
+
+  return extend({ stash, actions: defaultActions(stash) }) as never;
```
holic, line 54: neat

packages/stash/src/decorators/default.ts
```diff
@@ -0,0 +1,86 @@
...
+import { RegisterTableArgs, RegisterTableResult, registerTable } from "../actions/registerTable";
+import { RunQueryArgs, RunQueryOptions, RunQueryResult, runQuery } from "../actions/runQuery";
+import { SetRecordArgs, SetRecordResult, setRecord } from "../actions/setRecord";
+import { SetRecordsArgs, SetRecordsResult, setRecords } from "../actions/setRecords";
+import { SubscribeQueryArgs, SubscribeQueryResult, subscribeQuery } from "../actions/subscribeQuery";
+import { SubscribeStoreArgs, SubscribeStoreResult, subscribeStore } from "../actions/subscribeStore";
+import { SubscribeTableArgs, SubscribeTableResult, subscribeTable } from "../actions/subscribeTable";
+import { Table } from "@latticexyz/config";
+
+export type StashBoundDecodeKeyArgs<table extends Table = Table> = Omit<DecodeKeyArgs<table>, "stash">;
```
holic, line 20: `StashBound` -> `Stashed`?


jk jk jk don't do it

packages/stash/src/actions/deleteRecord.ts
```diff
@@ -0,0 +1,37 @@
...
+  table: table;
+  key: Key<table>;
+};
+
+export type DeleteRecordResult = void;
+
+export function deleteRecord<table extends Table>({ stash, table, key }: DeleteRecordArgs<table>): DeleteRecordResult {
+  const { namespaceLabel, label } = table;
+
+  if (stash.get().config[namespaceLabel] == null) {
```
holic, line 17: ooc any advantage of using `stash.get()` over `stash._.get`?

packages/stash/src/actions/extend.ts
```diff
@@ -0,0 +1,15 @@
+import { Stash } from "../common";
+
+export type ExtendArgs<stash extends Stash, actions> = {
+  stash: stash;
+  actions: actions;
+};
+
+export type ExtendResult<stash extends Stash, actions> = stash & actions;
```
holic, line 8: I guess the expectation/assumption here is there are no overlapping keys?

packages/stash/src/actions/getConfig.ts
```diff
@@ -0,0 +1,14 @@
...
+import { Stash } from "../common";
+
+export type GetConfigArgs = {
+  stash: Stash;
+  table: { label: string; namespaceLabel?: string };
+};
+
+export type GetConfigResult<table extends Table = Table> = table;
+
+export function getConfig({ stash, table }: GetConfigArgs): GetConfigResult<Table> {
```
holic, line 11: is this still necessary if we're using full table objects now?

packages/stash/src/queryFragments.ts
```diff
@@ -0,0 +1,106 @@
...
+  const getInitialKeys = () => ({});
+  return { table, match, getInitialKeys };
+}
+
+/**
+ * Matches all records that match the provided partial record.
+ * This works for both value and key, since both are part of the record.
+ * RECS equivalent (only for value match): HasValue(Component, value)
+ */
+export function MatchRecord<table extends Table>(
```
holic, line 73: I am torn on naming of these, where we don't use "table" in `In` but use "record" for `MatchRecord`

wondering if this could be just `Match` and `NotMatch`

packages/stash/src/queryFragments.ts
```diff
@@ -0,0 +1,106 @@
...
+
+/**
+ * Matches all records that don't exist in the table.
+ * RECS equivalent: Not(Component)
+ */
+export function NotIn<table extends Table>(table: table): QueryFragment<table> {
+  const match = (stash: Stash, encodedKey: string) => !(encodedKey in getRecords({ stash, table }));
+  const getInitialKeys = () => ({});
+  return { table, match, getInitialKeys };
+}
```
holic, line 66: do we need separate `Not*` fragments? could we compose?

```ts
export function Not<table extends Table>(fragment: QueryFragment<table>): QueryFragment<table> {
  const match = (...args) => !fragment.match(...args);
  return { ...fragment, match };
}
```

packages/stash/src/actions/deleteRecord.ts
alvrs: yeah exactly, trying to only use `._.state` when necessary, which is only when needing to write

packages/stash/src/actions/getConfig.ts
alvrs: The idea with this one is being able to get the config from a table in situations where it's dynamically synced from a store registration log. Will rename to `getTableConfig`

packages/stash/src/queryFragments.ts
alvrs: Could also be `Equals` 🤔

packages/stash/src/actions/extend.ts
holic: you would?

https://www.typescriptlang.org/play/?#code/AQFwngDgpsCiAeIoDsAmAlKBnArgGxAB4sQBDLACwBphSBjEASwHtksA+YAXmBPIuAAyWgxZsA3AChJoSDCiIUqKKm5xFaTLgKEA3sADmUEAGFWAM0YGAXMAAUASm6cAbs0aqAvjX1HTFq1tHZ2BkHABbACMoACdgT3ZpWWhDYzNkSwM1BSQ0FQBtAHI-dMzCgF1JAHoq4Dq6gD0AfkkgA

looks like you get a union for each action

I wonder if we wanna use `merge` from arktype instead?

packages/stash/src/queryFragments.ts
holic: I like `Matches` to indicate "partial match", I would think of "Equals" would mean "matches all keys/values for this object"

packages/stash/src/queryFragments.ts
alvrs: added `Not`, in the future we might want to add specific negative fragments if it becomes necessary for performance optimization reasons

packages/stash/src/actions/extend.ts
holic: > That means it is now technically possible to override existing functions.

this was already happening at runtime, the types just better reflect it now

packages/stash/README.md
```diff
@@ -0,0 +1,79 @@
+# Stash
+
+Stash is a client state library optimized for the MUD data model.
+It uses the MUD store config to define local tables, which support writing, reading and subscribing to table updates.
+It comes with a query engine optimized for "ECS-style" queries (similar to `@latticexyz/recs`) but with native support for composite keys.
```
holic, line 5: is there a docs page we can link to to clarify "ECS-style queries"?

packages/stash/README.md
```diff
@@ -0,0 +1,79 @@
...
+      key: ["player"],
+    },
+  },
+);
+
+// Initialize stash
+const stash = createStash(config);
+
+// Write to a table
+const Position = config.tables;
```
holic, line 39:
```suggestion
const { Position } = config.tables;
```

packages/stash/package.json
```diff
@@ -0,0 +1,59 @@
+{
+  "name": "@latticexyz/stash",
```
holic, line 2:
```suggestion
  "private": true,
  "name": "@latticexyz/stash",
```
do we wanna make this private for now so we don't publish at the same time as explorer?

packages/stash/README.md
alvrs: added a link to the ECS explainer

holic (approved):

Issue #3042, opened 2024-08-16
Revert details for `error Reason(param-type)` don't make it from the `System` to the client


qbzzt:
I have this Solidity code:

```solidity
error Fail();
error ParamFailString(string);

contract TasksSystem is System {

  function errorRevert() public {
    revert Fail();
  }

  function stringParamRevert() public {
    revert ParamFailString("Parameter fail");
  }
```

When I call `app__errorRevert`, I catch this error:

```
{
  "details": "execution reverted: custom error 552670ff:",
  "metaMessages": [
    "Estimate Gas Arguments:",
    "  from:                  0x735B2F2c662eBEDFFA94027A7196F0559f7f18a4\n  to:                    0x8d8b6b8414e1e3dcfd4168561b9be6bd3bf6ec4b\n  data:                  0xce1cd432\n  maxFeePerGas:          0 gwei\n  maxPriorityFeePerGas:  0 gwei\n  nonce:                 0"
  ],
  "shortMessage": "Execution reverted with reason: custom error 552670ff:.",
  "name": "EstimateGasExecutionError",
  "version": "viem@2.9.20",
  "message": "Execution reverted with reason: custom error 552670ff:.\n\nEstimate Gas Arguments:\n  from:                  0x735B2F2c662eBEDFFA94027A7196F0559f7f18a4\n  to:                    0x8d8b6b8414e1e3dcfd4168561b9be6bd3bf6ec4b\n  data:                  0xce1cd432\n  maxFeePerGas:          0 gwei\n  maxPriorityFeePerGas:  0 gwei\n  nonce:                 0\n\nDetails: execution reverted: custom error 552670ff:\nVersion: viem@2.9.20",
  "cause": {
    "details": "execution reverted: custom error 552670ff:",
    "shortMessage": "Execution reverted with reason: custom error 552670ff:.",
    "name": "ExecutionRevertedError",
    "version": "viem@2.9.20",
    "message": "Execution reverted with reason: custom error 552670ff:.\n\nDetails: execution reverted: custom error 552670ff:\nVersion: viem@2.9.20",
    "cause": {
      "details": "execution reverted: custom error 552670ff:",
      "metaMessages": [
        "URL: http://127.0.0.1:8545",
[...]

PR #3043, opened 2024-08-16
refactor(common): simplify writeContract/sendTransaction actions

holic:
workaround until https://github.com/wevm/viem/pull/2624 lands
fixes https://github.com/latticexyz/mud/issues/3042

before:
image

after:
image

(note "module already installed" message, which came from being able to catch the decoded error)


holic:
this was landed in viem: https://github.com/wevm/viem/pull/2624
bumping viem to pull this in here: https://github.com/latticexyz/mud/pull/3038

holic:
actually, this is still useful because we call `prepareTransactionRequest` here, which is where estimation errors are thrown and this bypasses viem's error handling in `writeContract`

we won't need this once we refactor this to use the viem nonce manager and can just wrap `writeContract` in a queue without the additional `prepare` step, but will save that for a future PR

holic:
ended up refactoring our approach here, since we can use our nonce manager in the same way the viem one would be used and it cleans up a lot + improves error messages

PR #3046, opened 2024-08-19
feat(protocol-parser): add `getFieldIndex` to internal exports

karooolis:


packages/protocol-parser/src/getFieldIndex.ts
```diff
@@ -0,0 +1,14 @@
+import { isDynamicAbiType, isStaticAbiType } from "@latticexyz/schema-type/internal";
+
+export function getFieldIndex(valueSchema: Record<string, string>, fieldName: string): number {
```
holic, line 3: can we make use of `ValueSchema` type instead of just `Record`?

would also be nice to use a generic so we can provide hinting for `fieldName` as well

```suggestion
export function getFieldIndex<valueSchema extends ValueSchema>(valueSchema: valueSchema, fieldName: keyof valueSchema): number {
```

karooolis: Definitely. Only had to cast `fieldName` into string since `fieldName` can be a `string | number | symbol`. My understanding is that we could make `ValueSchema` keys types stricter to disable numeric and symbol keys, so something like below would not be allowed:

```typescript
const valueSchema = {
 "1": "value1",
  2: "value2"
}
```

holic: `ValueSchema` is already defined with `string` keys so you could just do `fieldName: keyof valueSchema & string`

karooolis: not sure why it couldn't resolve automagically but added the `string` intersection, thanks

holic (approved):

PR #3047, opened 2024-08-19
chore(store-sync): export internal table names

karooolis:


packages/store-sync/src/sqlite/internalTables.ts
```diff
@@ -2,15 +2,18 @@ import { blob, integer, sqliteTable, text } from "drizzle-orm/sqlite-core";
 import { address, asHex, json } from "./columnTypes";
 import { PartialTable } from "./common";
 
-export const chainState = sqliteTable("__chainState", {
+export const CHAIN_STATE_TABLE_NAME = "__chainState";
+export const MUD_STORE_TABLES_NAME = "__mudStoreTables";
```
holic, line 6: we don't really do uppercase constants like this in MUD

I think we can probably leave the table names where they are (so they're close to their definition) and then do something like

```
export const internalTables = [chainState, mudStoreTables];
export const internalTableNames = internalTables.map(getTableName);
```

karooolis: noted, and updated

holic (approved):

PR #3050, opened 2024-08-20
feat(cli,world): register system ABI onchain

holic:
- [x] fix ABI race condition in world (world uses build script to generate ABIs but that same build script needs ABIs now)
  - [x] maybe easiest to fix my moving to parsing AST instead of ABI
- [x] register ABIs in metadata table during deploy
- [x] add `.mud` to `.gitignore` in templates, etc.
- [x] add `namespaceLabel` to config's system output


holic:
okay so after two attempts, I understand this a bit better

we currently have a constraint/order of operations
- parse solidity AST to codegen system interfaces + world interface
- import/use those codegen interfaces
- compile
- register ABI from compile artifact

I tried to simplify by just
- compile
- register ABI from compile artifact
- (eventually codegen interface from ABI)

but we lean on just ABI for a number of reasons:
- source won't compile if it references codegen interfaces that don't exist yet (chicken and egg problem)
- ABI is a subset of interface and is missing things like param modifiers (`string memory` vs `string calldata`, `address` vs `address payable`)
 
so then I went the other way
- parse solidity AST to codegen ABI
- register ABI
- (eventually codegen interface from ABI)

but that is also fraught because just parsing solidity skips parts of the compiler we need, like resolving imports with remappings, resolving underlying types of enums/user types, converting structs to tuples

so we can't actually create pure/true representation of the ABI and reimplementing all that in the "parse solidity" step seems like a bad idea

all this to figure out where the right "entry point" is for generating the local system manifest and I think it's in `mud build` after `forge build`

to recap:
- we need to keep our order-of-operations (parse source, codegen interfaces, compile, extract ABI)
- generate system manifest from ABI, after `forge build` but before `abi-ts` (in the future we may want these JSON manifests to be strongly typed)

there's still the problem of the ABI not including enough data for a 1:1 interface (e.g. `memory` vs `calldata`) but maybe we can just make some assumptions for now and fill that in later when we have some time to combine the ABI with AST (both available in forge output, which is nice, but still a pain to unwind)

alvrs (approved):
🔥🔥🔥

PR #3057, opened 2024-08-22
feat(world): add namespaceLabel to system config

holic:
goes with #3039 and closes some gaps we had there

### DIFF
--- a/packages/config/src/common.ts
+++ b/packages/config/src/common.ts
@@ -24,13 +24,39 @@ export type Schema = {
 };
 
 export type Table = {
-  readonly label: string;
-  readonly type: satisfy<ResourceType, "table" | "offchainTable">;
-  readonly namespace: string;
+  /**
+   * Human-readable label for this table's namespace. Used for namespace config keys and directory names.
+   */
   readonly namespaceLabel: string;
+  /**
+   * Human-readable label for this table. Used as config keys, library names, and filenames.
+   * Labels are not length constrained like resource names, but special characters should be avoided to be compatible with the filesystem, Solidity compiler, etc.
+   */
+  readonly label: string;
+  /**
+   * Table type used in table's resource ID and determines how storage and events are used by this table.
+   */
+  readonly type: satisfy<ResourceType, "table" | "offchainTable">;
+  /**
+   * Table namespace used in table's resource ID and determines access control.
+   */
+  readonly namespace: string;
+  /**
+   * Table name used in system's resource ID.
+   */
   readonly name: string;
+  /**
+   * Table's resource ID.
+   */
   readonly tableId: Hex;
+  /**
+   * Schema definition for this table's records.
+   */
   readonly schema: Schema;
+  /**
+   * Primary key for records of this table. An array of zero or more schema field names.
+   * Using an empty array acts like a singleton, where only one record can exist for this table.
+   */
   readonly key: readonly string[];
 };
 
--- a/packages/config/src/common.ts
+++ b/packages/config/src/common.ts
@@ -25,14 +25,14 @@ export type Schema = {
 
 export type Table = {
   /**
-   * Human-readable label for this table's namespace. Used for namespace config keys and directory names.
-   */
-  readonly namespaceLabel: string;
-  /**
    * Human-readable label for this table. Used as config keys, library names, and filenames.
    * Labels are not length constrained like resource names, but special characters should be avoided to be compatible with the filesystem, Solidity compiler, etc.
    */
   readonly label: string;
+  /**
+   * Human-readable label for this table's namespace. Used for namespace config keys and directory names.
+   */
+  readonly namespaceLabel: string;
   /**
    * Table type used in table's resource ID and determines how storage and events are used by this table.
    */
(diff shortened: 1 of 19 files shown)


packages/config/src/common.ts
```diff
@@ -24,13 +24,39 @@ export type Schema = {
...
+   * Table namespace used in table's resource ID and determines access control.
+   */
   readonly namespace: string;
-  readonly namespaceLabel: string;
+  /**
+   * Table name used in table's resource ID.
+   */
   readonly name: string;
+  /**
+   * Table's resource ID.
```
karooolis, line 49: Like the property descriptions. Just curious is there a reason why it uses single-line comments in `packages/cli/src/deploy/common.ts` but then multi-line comments here, even for single line comments?

packages/config/src/common.ts
```diff
@@ -24,13 +24,39 @@ export type Schema = {
 };
 
 export type Table = {
+  /**
+   * Human-readable label for this table. Used as config keys, library names, and filenames.
+   * Labels are not length constrained like resource names, but special characters should be avoided to be compatible with the filesystem, Solidity compiler, etc.
```
karooolis, line 29: Wonder if these special characters should be specified here? Curious if there are any other constraints like can a label start with number, etc?

packages/store/ts/config/v2/defaults.ts
```diff
@@ -24,9 +24,9 @@ export const TABLE_DEPLOY_DEFAULTS = {
 export type TABLE_DEPLOY_DEFAULTS = typeof TABLE_DEPLOY_DEFAULTS;
 
 export const TABLE_DEFAULTS = {
-  namespace: "",
+  namespaceLabel: "",
   type: "table",
```
karooolis, line 28: just wonder why only these two properties are used as `TABLE_DEFAULTS`? Why is `namespace` not part of, and all the others in `TableInput`?

karooolis: I see now, `namespace` defaults to `namespaceLabel`.

packages/store/ts/config/v2/table.ts
```diff
@@ -171,8 +176,10 @@ export function resolveTable<input extends TableInput, scope extends Scope = Abi
   input: input,
   scope: scope = AbiTypeScope as unknown as scope,
 ): resolveTable<input, scope> {
-  const namespaceLabel = input.namespaceLabel ?? TABLE_DEFAULTS.namespace;
+  const namespaceLabel = input.namespaceLabel ?? TABLE_DEFAULTS.namespaceLabel;
+  // validate ensures this is length constrained
   const namespace = input.namespace ?? namespaceLabel;
```
karooolis, line 181: I find it slightly odd that `namespace` is assigned with `namespaceLabel` rather than vice-versa. This means that `namespaceLabel` is treated as the ID rather than the `namespace`. I'd expect it to be in reverse but I suspect it's due to some constraints I'm not aware of yet.

packages/store/ts/config/v2/input.ts
```diff
@@ -18,11 +18,17 @@ export type TableDeployInput = Partial<TableDeploy>;
...
-   * Human-readable table label. Used as config keys, table library names, and filenames.
+   * Human-readable label for this table. Used as config keys, library names, and filenames.
    * Labels are not length constrained like resource names, but special characters should be avoided to be compatible with the filesystem, Solidity compiler, etc.
    */
   readonly label: string;
   /**
+   * Human-readable label for this table's namespace. Used for namespace config keys and directory names.
+   * Defaults to the nearest namespace in the config or root namespace if not set.
+   */
+  readonly namespaceLabel?: string;
```
karooolis, line 29: probably too big of a change now and not worth the effort but wonder if defining namespace as its own type would make sense e.g:

type NamespaceInput = {
  id: string;
  label?: string;
}

type Table Input = {
  namespace?: NamespaceInput;
  // ...
}

karooolis (approved):
Some nits and questions but looks good to me!

packages/config/src/common.ts
holic: that file is internal, so not as important to have good docs and those are mostly internal notes

the tsdoc here will get included as part of your IDE tooltips and maybe eventually generated into documentation

packages/config/src/common.ts
holic: I don't know the range of special characters in various OS's so I'll leave it to user discretion

if a label has a non-fs compatible character, they'll know quickly because tablegen/worldgen will fail

packages/store/ts/config/v2/table.ts
holic: `namespaceLabel` is the "entry point" and the strongly typed thing, which propagates to the namespace if it meets namespace criteria (<=14 chars) of the resource ID, otherwise throws in the validate function and requires either adjusting the label or specifying an explicit `namespace`

packages/store/ts/config/v2/input.ts
holic: this would be a breaking change unfortunately

PR #3061, opened 2024-08-23
feat: apply logs from receipt

holic:


holic:
replaced by https://github.com/latticexyz/mud/pull/3215

PR #3062, opened 2024-08-26
feat(explorer): show transactions

karooolis:
CleanShot 2024-10-01 at 17 43 18@2x


.changeset/blue-starfishes-fry.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/explorer": patch
+---
+
+World Explorer now includes a "Transactions" tab for reviewing completed transactions. It provides detailed information about emitted events, logs, and additional transaction data.
```
holic, line 5:
```suggestion
Observe tab is now populated by transactions flowing through the world, in addition to local transactions when using the `observer` transport wrapper.
```

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/observe/BlockExplorerLink.tsx
```diff
@@ -0,0 +1,18 @@
+import { ExternalLinkIcon } from "lucide-react";
+import { Hex } from "viem";
+import { useChain } from "../../../../hooks/useChain";
+import { explorerForChainId } from "../../../../utils/explorerForChainId";
+
+export function BlockExplorerLink({ hash, children }: { hash?: Hex; children: React.ReactNode }) {
+  const { id: chainId } = useChain();
+  const explorerUrl = explorerForChainId({ hash, chainId });
```
Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/observe/BlockExplorerLink.tsx:8]: what about
```suggestion
  const url = blockExplorerTransactionUrl({ chainId, hash });
```

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/observe/BlockExplorerLink.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+  if (!explorerUrl) return children;
+  return (
+    <a href={`${explorerUrl}/tx/${hash}`} target="_blank" rel="noopener noreferrer" className="flex hover:underline">
+      <div className="flex items-center">
```
Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/observe/BlockExplorerLink.tsx:13]:
```suggestion
    <a href={`${explorerUrl}/tx/${hash}`} target="_blank" rel="noopener noreferrer" className="flex items-center hover:underline">
```

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/observe/Confirmations.tsx
```diff
@@ -0,0 +1,28 @@
...
+    },
+  });
+
+  if (!confirmations) return <Skeleton className="h-4 w-[50px]" />;
+  return (
+    <span className="flex items-center text-xs font-extrabold text-green-600">
+      <span
+        className="mr-2 inline-block h-2 w-2 animate-pulse rounded-full"
+        style={{
+          background: "rgb(64, 182, 107)",
```
Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/observe/Confirmations.tsx:22]: would be great to use tailwind colors or lift this into tailwind config

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/observe/TimeAgoCell.tsx
```diff
@@ -0,0 +1,20 @@
+import { useEffect, useState } from "react";
+import { timeAgo } from "../../../../utils/timeAgo";
+
+export function TimeAgoCell({ timestamp }: { timestamp: bigint }) {
```
Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/observe/TimeAgoCell.tsx:4]: this doesn't seem particularly cell-oriented so could prob name it:
```suggestion
export function TimeAgo({ timestamp }: { timestamp: bigint }) {
```
and I'd tend to leave out the color styling of this and let it inherit the color of the parent

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/observe/TransactionsTableContainer.tsx
```diff
@@ -0,0 +1,149 @@
...
+        mergedMap.set(transaction.hash, { ...transaction, write: existing.write });
+      } else {
+        mergedMap.set(transaction.hash, { ...transaction });
+      }
+    }
+
+    return Array.from(mergedMap.values()).reverse();
+  }, [transactions, observerWrites]);
+
+  async function handleTransaction(hash: Hex, timestamp: bigint) {
```
Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/observe/TransactionsTableContainer.tsx:75]: hmm, feels weird that we're not writing to the transaction list until after all the things below resolve (getting tx, getting receipt, simulating result

wouldn't we want to render it right away, then enrich it with data as it comes in?

could we do this all inline with individual components rather than in a big handler like this? or does that impact the resulting order?

packages/explorer/src/app/(explorer)/layout.tsx
```diff
@@ -31,7 +31,7 @@ export default function RootLayout({
       <body className={`${inter.variable} ${jetbrains.variable} dark`}>
         <Theme>
           <div
-            className="container"
+            className="container pb-8"
             style={{
               fontFamily: "var(--font-jetbrains-mono)",
```
holic, line 36: can we lift this into tailwind and use `font-jetbrains-mono`?

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/observe/TransactionsTableContainer.tsx
Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/observe/TransactionsTableContainer.tsx:75]: good point, did it like that initially for simplicity reasons but definitely better to update state as soon as it's available.

As for individual components, I explored that approach but it didn't quite work due to how tanstack table works. It expects all the data to be passed to the `useReactTable` beforehand, and then provides utilities for rendering rows and cells. We could do without the tanstack table ofc but wanted to use it in case we need filtering, sorting and any other capabilities. Also, as you noted, there would be other considerations when it comes to sorting that would be need handling, mainly seems like handling incoming txs from observer would need more thought where tx hash is not immediately available.

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/observe/TimeAgoCell.tsx
Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/observe/TimeAgoCell.tsx]: yeah makes sense, renamed to `TimeAgo`. As for styling, in order to do that it would require to be wrapped with a `div` element for that column just for this style. I understand the rationale to keep shared styling at parent-level as much as possible but in this case, each column has different styling when it comes to colors, and don't have a common parent in that regard.

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/observe/TransactionsTableContainer.tsx
Comment (holic) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/observe/TransactionsTableContainer.tsx]: OOC do we need two components? seems like we could do one component and one hook? since most of the logic in `TransactionTableContainer` is now hook-like and we only wrap it in a component to render `TransactionTableView`

packages/explorer/src/components/LatestBlock.tsx
```diff
@@ -7,18 +7,16 @@ export function LatestBlock() {
...
     <div className="inline-block w-[50px]">
       {block ? (
         <div className="flex items-center justify-end text-xs font-extrabold text-green-600">
-          <span
-            className="mr-2 inline-block h-[8px] w-[8px] animate-pulse rounded-full"
-            style={{
-              background: "rgb(64, 182, 107)",
-            }}
-          ></span>
+          <span className="mr-2 inline-block h-[8px] w-[8px] animate-pulse rounded-full bg-success"></span>
```
holic, line 19: as a general rule and where possible, I try not to add spacing to individual elements to move siblings, but rather change this at the parent

since we already have `flex` on the parent, we can just add `gap-2` above to do the same kind of spacing here

holic (dismissed):
no action needed on the spacing comment, just a note as we continue to build out and refine pages

prob worth moving the `TransactionTableContainer` logic into a hook then use it in `TransactionsTableView` (which could prob get back to being `TransactionsTable`)

otherwise lgtm!

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/observe/TransactionsTableContainer.tsx
```diff
@@ -0,0 +1,197 @@
...
+          receipt = await getTransactionReceipt(wagmiConfig, { hash });
+          if (receipt) break;
+        } catch (error) {
+          console.error(`Failed to fetch receipt, attempt ${attempt + 1}:`, error);
+        }
+
+        if (attempt < maxRetries - 1) {
+          await new Promise((resolve) => setTimeout(resolve, retryInterval));
+        }
+      }
```
Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/observe/TransactionsTableContainer.tsx:125]: this is basically what `waitForTransactionReceipt` does, but with caching/batching - can we use that here?

Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/observe/TransactionsTableContainer.tsx]: oh so nice!!! great suggestion, changed

packages/explorer/src/components/LatestBlock.tsx
karooolis: Yeah good point and generally try to follow the same logic, should be more mindful about it. Added gap property.

PR #3063, opened 2024-08-27
refactor(cli): register labels as tags

holic:
small refactor pulled out of #3050 in prep for registering tags for systems


packages/cli/package.json
```diff
@@ -39,6 +39,7 @@
     "test:ci": "pnpm run test"
   },
   "dependencies": {
+    "@ark/util": "catalog:",
```
alvrs, line 42: what does `catalog:` do?

alvrs (approved):

packages/cli/package.json
holic: it's a better version of aligning versions of a given package: https://pnpm.io/catalogs
https://github.com/latticexyz/mud/blob/main/pnpm-workspace.yaml#L5-L11

PR #3065, opened 2024-08-27
feat: init explorer package

karooolis:


karooolis:
publishing this locally to get around circular dependency issue when publishing package + templates both at once.

PR #3066, opened 2024-08-27
fix(create-mud): update changeset package name + description

karooolis:


.changeset/real-waves-bathe.md
```diff
@@ -2,4 +2,4 @@
 "@latticexyz/create-mud": patch
```
holic, line 2:
```suggestion
"create-mud": patch
```

holic (approved):

.changeset/real-waves-bathe.md
karooolis: dangers of copy-paste

PR #3067, opened 2024-08-27
feat: add changeset validation workflow

karooolis:


.github/workflows/changeset.yml
```diff
@@ -0,0 +1,21 @@
...
+    runs-on: depot-ubuntu-22.04-16
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Setup
+        uses: ./.github/actions/setup
+
+      - name: Changeset
+        run: pnpm changeset
```
holic, line 21:
```suggestion
        run: pnpm changeset status
```

.github/workflows/changeset.yml
```diff
@@ -0,0 +1,21 @@
+name: Changeset
+
+on:
+  push:
+    branches:
+      - "main"
```
karooolis, line 6: Initially, thought to run workflow only if `.changeset/**` changes but then realized that `package.json` of packages may change too, causing changeset validation to fail. So now including all paths, although could still narrow down to `.changeset/**` and `packages/**/package.json`?

.github/workflows/changeset.yml
```diff
@@ -0,0 +1,18 @@
+name: Changeset
+
+on:
+  pull_request:
```
karooolis, line 4: Initially, thought to run workflow only if .changeset/** changes but then realized that package.json of packages may change too, causing changeset validation to fail. So now including all paths, although could still narrow down to .changeset/** and packages/**/package.json?

holic: sounds reasonable to run as-needed (just for changeset + package.json changes) although may also need to consider the `pnpm-workspace.yaml` and ~any `package.json` (e.g. `**/package.json`)

but also running it for every commit is fine too, should be lightweight enough to not matter much

karooolis: Perhaps then best to not restrict any paths since it is indeed lightweight, and don't want to miss anything accidentally.

.github/workflows/changeset.yml
```diff
@@ -0,0 +1,20 @@
...
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Setup
+        uses: ./.github/actions/setup
+
+      - name: Changeset
+        run: pnpm changeset status --since origin/main
```
karooolis, line 20: Have to do it like this due to base branch set as `main`. Relevant issue: https://github.com/changesets/changesets/issues/1055

holic (approved):

PR #3073, opened 2024-08-27
fix(common): use pending block tag in tx queue

holic:
we refactored these actions in https://github.com/latticexyz/mud/pull/3043, but missed that `estimateGas` is not called with `pending` block tag by default (but was what we used before)


packages/common/src/sendTransaction.ts
```diff
@@ -70,13 +73,14 @@ export async function sendTransaction<
...
+          const params = {
+            blockTag,
             ...request,
             nonce,
             ...feeRef.fees,
-          };
+          } as const satisfies SendTransactionParameters<chain, account, chainOverride, request>;
           debug("sending tx to", request.to, "with nonce", nonce);
-          return await getAction(client, viem_sendTransaction, "sendTransaction")(params);
+          return await getAction(client, viem_sendTransaction, "sendTransaction")(params as never);
```
holic, line 83: viem types don't like `blockTag` here, but hoping to get this fixed upstream

packages/common/src/sendTransaction.ts
```diff
@@ -38,10 +39,11 @@ export type SendTransactionExtraOptions<chain extends Chain | undefined> = {
 export async function sendTransaction<
   chain extends Chain | undefined,
   account extends Account | undefined,
-  chainOverride extends Chain | undefined,
+  const request extends SendTransactionRequest<chain, chainOverride>,
+  chainOverride extends Chain | undefined = undefined,
 >(
   client: Client<Transport, chain, account>,
-  request: SendTransactionParameters<chain, account, chainOverride>,
+  request: SendTransactionParameters<chain, account, chainOverride, request>,
```
holic, line 46: these changes align with viem's type signature (forgot to do this last version bump)

PR #3076, opened 2024-08-28
feat(explorer): fetch partial abi on load

karooolis:


karooolis:
Closed in favor of #3077 and #3079

PR #3077, opened 2024-08-28
chore(explorer): setup zustand with context

karooolis:
Previous implementation was possibly prone to SSR-related issues. Updated zustand setup according to the official guide: https://zustand.docs.pmnd.rs/guides/nextjs


holic (approved):
oof so much boilerplate! but glad there's a solution, I am pretty sure I've come across this too

"app store" makes me think of iOS/MacOS but I don't have a good alternative to suggest 🤷

karooolis:
> oof so much boilerplate! but glad there's a solution, I am pretty sure I've come across this too
> 
> "app store" makes me think of iOS/MacOS but I don't have a good alternative to suggest 🤷

Same thoughts about "app store" :/ considered using "app data store" but too long. Might come up with a better name eventually.

PR #3074, opened 2024-08-27
docs(explorer): add world explorer docs page

karooolis:

PR #3088, opened 2024-08-29
docs(explorer): update README

karooolis:


packages/explorer/README.md
```diff
@@ -23,7 +23,7 @@ World Explorer is a GUI tool designed for visually exploring and manipulating th
    Alternatively, if you have a worlds configuration file:
 
    ```sh
-   npx @latticexyz/explorer --worldsConfigPath <PATH_TO_WORLDS_CONFIG>
+   npx @latticexyz/explorer --worldsFile <PATH_TO_WORLDS_CONFIG>
    ```
```
holic, line 27: since this defaults to the location we put the file in, I think we can remove this arg and let folks figure out how to customize the arg

and maybe instead of "if you have a worlds config file" something like "if you've deployed a world from your project"

karooolis: Closing this PR in favor of @qbzzt https://github.com/latticexyz/mud/pull/3074/files , let's take the discussion there, it includes an updated README.

PR #3094, opened 2024-08-29
feat(modules): standalone erc20 with external store

TamaraRingas:


packages/world-module-erc20-external-store/src/Store.sol
holic, line 1: I don't think you need this copied here?

TamaraRingas: I thought it might be good to have it there because it would have to be deployed for the token contracts to connect to? But otherwise I can move it to `/test`?

packages/world-module-erc20-external-store/test/ERC20.t.sol
```diff
@@ -0,0 +1,204 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.24;
+
+import { Test } from "forge-std/Test.sol";
+import { StoreSwitch } from "@latticexyz/store/src/StoreSwitch.sol";
+import { GasReporter } from "@latticexyz/gas-report/src/GasReporter.sol";
+import { RESOURCE_TABLE } from "@latticexyz/store/src/storeResourceTypes.sol";
+import { ResourceId, ResourceIdLib } from "@latticexyz/store/src/ResourceId.sol";
+
+import { Store } from "../src/Store.sol";
```
holic, line 10: there's a `StoreMock` you could use from the store package that does this (but we'd need to make sure its exported in the npm package, like `MudTest` in world package)

TamaraRingas: The StoreMock was missing several functions that are needed when using Store without a World, like `getStaticField()` and `getDynamicField()`

holic: doesn't look like we're using those methods anymore?

either way, would be good to add those to the mock

PR #3097, opened 2024-08-29
feat(explorer): remove dev command from build

karooolis:
First attempt to compile out `next dev` path for `@latticexyz/explorer` build but so far didn't succeed, will continue experimenting with it.


karooolis:
Close in favor of https://github.com/latticexyz/mud/pull/3108

PR #3099, opened 2024-08-30
fix(explorer): interpret hex as string (minimist)

karooolis:


holic (approved):

holic (approved):

package.json
```diff
@@ -58,5 +58,10 @@
   "engines": {
     "node": "^18.20.1",
     "pnpm": "^9.6.0"
+  },
+  "pnpm": {
+    "patchedDependencies": {
```
holic, line 63: ooc does this have to be in the workspace root for it to work? or could this be in `packages/explorer/package.json`

package.json
```diff
@@ -58,5 +58,10 @@
...
+      "minimist@1.2.8": "patches/minimist@1.2.8.patch"
```
holic, line 64: might be nice to name this something like `patches/minimist@1.2.8-allow-hex.patch`

karooolis: updated in https://github.com/latticexyz/mud/pull/3105

package.json
karooolis: looks like it's not possible to scope it down to a package, pnpm complains if I do so :/

PR #3122, opened 2024-09-03
fix(cli): await module deployment to catch error

alvrs:


alvrs:
too late https://github.com/latticexyz/mud/commit/ef6f7c0c6afcc46e7463d18c00fa99c7cafcae65 :(

PR #3123, opened 2024-09-03
chore: include main branch when fetching during pre-release action

alvrs:
pulled out from #2996


holic:
it was pulled out here https://github.com/latticexyz/mud/pull/3007

but iirc it ended up being the wrong action to be run, he wanted snapshot not prerelease?

PR #3124, opened 2024-09-04
docs(state-query/dozer): add live query and metadata discovery 🚗 

qbzzt:


ludns:
Can we break this into table metadata and live query?
Live query on Dozer is not quite ready (we are missing a notification from Dozer when a record gets deleted).

qbzzt:
> Can we break this into table metadata and live query? Live query on Dozer is not quite ready (we are missing a notification from Dozer when a record gets deleted).

Yes, see https://github.com/latticexyz/mud/pull/3186 .

qbzzt:
Live queries don't really work (due to the lack of a delete), so no need for this yet.

PR #3129, opened 2024-09-05
chore(stash): release package to npm

alvrs:


holic:
replaced by https://github.com/latticexyz/mud/pull/3184 so we can start publishing snapshots

PR #3130, opened 2024-09-05
feat(cli): speed up dev deploy with temporary automine during deploy

alvrs:


packages/cli/src/runDeploy.ts
```diff
@@ -155,6 +159,10 @@ export async function runDeploy(opts: DeployOptions): Promise<WorldDeploy> {
       opts.kms ? true : false,
     );
   }
+
+  // Reset mining mode after deploy
+  resetMiningMode();
```
holic, line 164: should this be `await`ed?

packages/cli/src/utils/enableAutomine.ts
```diff
@@ -0,0 +1,107 @@
...
+    headers: {
+      "Content-Type": "application/json",
+    },
+    body: JSON.stringify({
+      jsonrpc: "2.0",
+      method: "anvil_getAutomine",
+      params: [],
+      id: 1,
+    }),
+  }).then((res) => res.json());
```
holic, line 43: why do we need to call the rpc manually when viem has actions for this? https://viem.sh/docs/actions/test/getAutomine

```ts
await getAction(client, getAutomine, 'getAutomine')({});
```

alvrs: agreed, much better now

holic (approved):

PR #3155, opened 2024-09-09
fix(explorer): refactor to yargs for cli

karooolis:


.changeset/itchy-trees-retire.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/explorer": patch
+---
+
+Migrated the World Explorer CLI to yargs to resolve the issue with the worldAddress option, which was being incorrectly interpreted as a number rather than a hexadecimal value.
```
holic, line 5:
```suggestion
Fixed an issue with `--worldAddress` CLI flag being incorrectly interpreted as a number rather hex.
```

packages/explorer/bin/explorer.ts
```diff
@@ -1,23 +1,63 @@
...
+      default: process.env.WORLD_ADDRESS || null,
+    },
+  })
+  .parseSync();
+
+const port = argv.port;
+const chainId = argv.chainId;
+const indexerDatabase = argv.indexerDatabase;
+const worldsFile = argv.worldsFile;
+const isDev = argv.dev;
```
holic, line 59: is it necessary to map these to their own variables?

could we either just refer directly to `argv.port` below or explode with `const { port, ... } = yargs(...)`

packages/explorer/bin/explorer.ts
```diff
@@ -1,23 +1,63 @@
...
-let worldAddress = argv.worldAddress || process.env.WORLD_ADDRESS || null;
+// Define command-line options with yargs
+const argv = yargs(process.argv.slice(2))
+  .options({
+    port: {
+      alias: "p",
+      description: "Port number for the server",
+      type: "number",
+      default: process.env.PORT || 13690,
+    },
```
holic, line 21: should we add `--host` or `--hostname` while we're here?

packages/explorer/bin/explorer.ts
karooolis: True, now that `yargs` is used here, adjusted

packages/explorer/bin/explorer.ts
```diff
@@ -1,43 +1,88 @@
...
+    hostname: {
+      alias: "H",
+      description: "Host for the server",
+      type: "string",
+    },
```
holic, line 26:
```suggestion
      type: "string",
      default: process.env.HOSTNAME || "127.0.0.1",
    },
```

packages/explorer/bin/explorer.ts
```diff
@@ -1,43 +1,88 @@
...
+  if (dev) {
     command = "pnpm";
     args = ["dev"];
   } else {
     command = "pnpm";
     args = ["start"];
   }
 
+  if (hostname) {
+    args.push("--hostname", hostname);
```
holic, line 76: could we pass this as an env var instead like PORT?

holic (approved):

PR #3156, opened 2024-09-09
feat(explorer): external wallet connect

karooolis:


.changeset/popular-socks-rush.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/explorer": patch
+---
+
+Integrated RainbowKit provider to be used for external wallet connections.
```
holic, line 5: is it possible to connect? as in do we have a connect button yet?

if not, maybe no need for a changelog because it's not usable yet

karooolis: Removed changeset as discussed

Issue #3137, opened 2024-09-08
Documentation: move best practices out of guide


ludns:
I wouldn't expect to find best practices inside "Guide".
They are quite good, and I think we should make sure people read them even if they are too cool to read the guides. Maybe we should link to them from a couple places where it makes sense.

PR #3157, opened 2024-09-09
docs(best-practices): move out of guides

qbzzt:
Closing: #3137


docs/next.config.mjs
```diff
@@ -256,6 +256,26 @@ export default withNextra({
...
         permanent: false,
       },      
+      {
+        source: "/guides/best-practices/dividing-into-systems",
+        destination: "/best-practices/dividing-into-systems",
+        permanent: false,
+      },
+      {
+        source: "/guides/best-practices/system-best-practices",
+        destination: "/best-practices/system-best-practices",
```
holic, line 266: how about just `/best-practices/systems` instead of duplicating the section title?

qbzzt: You're right.

PR #3180, opened 2024-09-13
fix(common): use latest block tag in nonce manager

alvrs:
fixes 1776
fixes 2675

### DIFF
--- a/.changeset/bright-rabbits-trade.md
+++ b/.changeset/bright-rabbits-trade.md
@@ -1,0 +1,11 @@
+---
+"@latticexyz/common": patch
+---
+
+To reset an account's nonce, the nonce manager uses the [`eth_getTransactionCount`](https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_gettransactioncount) RPC method,
+which returns the number of transactions sent from the account.
+When using the `pending` block tag, this includes transactions in the mempool that have not been included in a block yet.
+If an account submits a transaction with a nonce higher than the next valid nonce, this transaction will stay in the mempool until the nonce gap is closed and the transactions nonce is the next valid nonce.
+This means if an account has gapped transactions "stuck in the mempool", the `eth_getTransactionCount` method with `pending` block tag can't be used to get the next valid nonce
+(since it includes the number of transactions stuck in the mempool).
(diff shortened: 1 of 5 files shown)


.changeset/bright-rabbits-trade.md
```diff
@@ -0,0 +1,11 @@
...
+"@latticexyz/common": patch
+---
+
+To reset an account's nonce, the nonce manager uses the [`eth_getTransactionCount`](https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_gettransactioncount) RPC method,
+which returns the number of transactions sent from the account.
+When using the `pending` block tag, this includes transactions in the mempool that have not been included in a block yet.
+If an account submits a transaction with a nonce higher than the next valid nonce, this transaction will stay in the mempool until the nonce gap is closed and the transactions nonce is the next valid nonce.
+This means if an account has gapped transactions "stuck in the mempool", the `eth_getTransactionCount` method with `pending` block tag can't be used to get the next valid nonce
+(since it includes the number of transactions stuck in the mempool).
+Since the nonce manager only resets the nonce on reload or in case of a nonce error, using the `latest` block tag by default is the safer choice to be able to recover from nonce gaps.
```
holic, line 11:
```suggestion
Since the nonce manager only resets the nonce on reload or in case of a nonce error, using the `latest` block tag by default is the safer choice to be able to recover from nonce gaps.

Note that this change may reveal more "transaction underpriced" errors than before. These errors will now be retried automatically and should go through after the next block is mined.
```
(we need to add in support for retrying transaction underpriced)

holic (approved):

packages/common/src/sendTransaction.ts
```diff
@@ -94,6 +93,12 @@ export async function sendTransaction<
               debug("got nonce error, retrying", error.message);
               return;
             }
+
+            if (String(error).includes("transaction underpriced")) {
```
holic, line 97: - https://github.com/ethereum/go-ethereum/blob/c0b5d428a9a88ddd82605983bbd1b612dc61672d/core/txpool/errors.go#L35
- https://github.com/paradigmxyz/reth/blob/ef1d9e7a038a34563522c19f796c01773fb0ea0b/crates/transaction-pool/src/error.rs#L203

PR #3181, opened 2024-09-13
feat(explorer): active chain as dynamic param

karooolis:
todo:

- [x] Handle layout, error pages, etc.
- [x] Make sure correct chain is configured in production mode
- [x] Throw if non-valid chain is used
- [x] Keep connected between tabs


packages/explorer/src/app/icon.svg
karooolis, line 1: Unusual bug but `useParams` would return incorrect params in certain occasions due to favicon not being present i.e. https://github.com/vercel/next.js/discussions/13301 . Not sure if we want to use this as the favicon though.

holic: could also 404 that page with middleware

karooolis: sure but why not just add a favicon? Kind of like the MUD logo as favicon, also shows up in MetaMask wallet when connecting.

holic: we can still add a favicon via meta tag, etc, I just dont love `favicon.ico` specifically haha, it's oooold browser norms and annoying that they still request this by default

and unclear what the expectation is of this file in terms of appropriate format + size

karooolis: I get you, the next.js convention is nice to reduce boilerplate but making it implicit is not so great. Made it more explicit by setting favicon in layout's metadata. Still not ideal but better than before, lmk what you think.

holic: Oh yeah I think we should set a favicon in metadata!

It just doesn't need to be at the exact name/path of `/favicon.ico`. And browsers will attempt to hit that path regardless of what metadata says. That's why I usually just nuke this route in favor of metadata tags with png/svg icons

packages/explorer/public/favicon.svg
holic, line 1: there's another of these svgs somewhere with the paths fixed so they don't have lines between each pixel - can we try to find those?

packages/explorer/src/app/[chainName]/(explorer)/worlds/[worldAddress]/page.tsx
Comment (holic) [inline packages/explorer/src/app/[chainName]/(explorer)/worlds/[worldAddress]/page.tsx:1]: can we put chain name param inside the (explorer) dir?

packages/explorer/src/common.ts
```diff
@@ -1,22 +1,36 @@
...
+  [anvil.id]: "anvil",
+  [garnet.id]: "garnet",
+  [redstone.id]: "redstone",
+} as const;
+
+export const chainNameId: Record<SupportedChainNames, SupportedChainIds> = {
+  anvil: anvil.id,
+  garnet: garnet.id,
+  redstone: redstone.id,
+};
```
holic, line 18: there's nothing in these types to ensure we maintain parity with the above map

can we compute this or add type constraints to make sure they stay aligned?

packages/explorer/src/common.ts
```diff
@@ -1,22 +1,36 @@
...
+
+export const chains: Record<SupportedChainIds, Chain> = {
   [anvil.id]: anvil,
-  [redstone.id]: redstone,
   [garnet.id]: garnet,
+  [redstone.id]: redstone,
 };
```
holic, line 24: alternatively could name the chains here like

```ts
const supportedChains = { redstone, garnet, anvil };
```

then compute the look ups based on that, or do look ups like

```ts
const chain = Object.values(supportedChains).find((c) => c.id === chainId);
```

packages/explorer/src/common.ts
```diff
@@ -1,22 +1,36 @@
...
+    throw new Error(`Invalid chain id. Supported chains are: ${Object.keys(chainIdName).join(", ")}.`);
   }
-
-  return chain;
 }
 
-export function isAnvil() {
-  const chain = getChain();
-  return chain.id === anvil.id;
+export function isValidChainName(name: string | string[] | undefined): asserts name is SupportedChainNames {
```
holic, line 32:
```suggestion
export function validateChainName(name: string | string[] | undefined): asserts name is SupportedChainNames {
```

something named `isValid` makes me think this should return a boolean, not throw

packages/explorer/src/hooks/useChain.ts
```diff
@@ -0,0 +1,16 @@
...
+export function useChain() {
+  const { chainName } = useParams();
+  isValidChainName(chainName);
+
+  const chainId = chainNameId[chainName];
+  const chain = chains[chainId];
+  return chain;
+}
+
+export function useChainId() {
```
holic, line 13: Is this hook necessary?

karooolis: it's a nice helper used in `useContractMutation` and such where we need `chainId` only but not strictly necessary. Do you think it's an overkill?

packages/explorer/src/common.ts
karooolis: ah very good point, will do

packages/explorer/src/common.ts
karooolis: Initially computed with `reduce` but thought it's too much boilerplate for that. However, your comment made me realize that keeping it aligned is definitely preferred, will change. This is where `_.lodash` is nice but will make without it.

packages/explorer/src/hooks/useChain.ts
holic: yeah I find it a bit overkill when you can do 
```ts
const { id: chainId } = useChain();
```

karooolis: makes sense, took inspiration from wagmi but indeed not necessary

packages/explorer/public/favicon.svg
karooolis: Found one in https://lattice.xyz/brand, looks to have no lines

holic: image

🤔 

eh this is fine haha, it'll get shrunk way down

holic (dismissed):

packages/explorer/public/favicon.svg
holic: oh nm, I see you replaced it with the square one

PR #3184, opened 2024-09-17
feat(stash): release package to npm

alvrs:
doing this now so we can build on it, even though its still not stabilized


holic:
- [ ] `CreateStoreResult` -> `CreateStashResult` or maybe `ExtendedStash`

PR #3185, opened 2024-09-17
feat(explorer): dozer integration

karooolis:
Key changes:
- Added Dozer support.
- Created `DataExplorerSqlite` and `DataExplorerDozer` to distinguish which indexer is used under-the-hood.
- Standardized `TablesViewer` and `TableSelector` interfaces so they work for both, `DataExplorerSqlite` and `DataExplorerDozer`.
- Added a basic `SQLEditor` that is included as part of `DataExplorerDozer`.
- Removed `tables`, `rows` and other routes. Only left with `api/world` for fetching world ABI, and `api/table` to fetch a specific data table (when using sqlite indexer).
- All queries have been moved to `queries/**` helpers.
- In SQLite and Dozer case all available MUD tables are fetched from `store__tables` (previously it was fetched from all available SQL databases using its pragma).

Demo world - http://localhost:13690/garnet/worlds/0x9ca094a38027bb331656e4f1f6ed40c9982dbe70/explore

todo:
- [x] handle anvil
- [x] improve types
- [x] same component interfaces for sqlite and dozer 
- [x] remove redundant api endpoints


packages/explorer/src/common.ts
```diff
@@ -9,6 +11,9 @@ export type supportedChainId = supportedChains[supportedChainName]["id"];
 export const chainIdToName = Object.fromEntries(
   Object.entries(supportedChains).map(([chainName, chain]) => [chain.id, chainName]),
 );
+export const supportedChainsIdName = Object.fromEntries(
+  Object.entries(supportedChains).map(([name, chain]) => [chain.id, name]),
+);
```
holic, line 16: this does the same as `chainIdToName`?

karooolis: yeah, accidental leftover from merge. Will setup knip to detect unused exports

.changeset/cyan-chefs-obey.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/explorer": patch
+---
+
+Dozer integration has been added for Redstone and Garnet. When the Explorer is run on Anvil, the local indexer is used. For Garnet or Redstone, the indexed data is retrieved from a remotely hosted Dozer instance.
```
holic, line 5: I don't think end users know/care what Dozer is and we've started moving towards calling this SQL API (and could say "hosted SQL API" or "Redstone SQL API" or similar when referring to the hosted variant).

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/utils/decodeTable.ts
```diff
@@ -0,0 +1,45 @@
...
+import { hexToResource } from "@latticexyz/common";
+import { Schema, Table } from "@latticexyz/config";
+import { hexToSchema } from "@latticexyz/protocol-parser/internal";
+
+export type DeployedTable = Omit<Table, "label" | "namespaceLabel"> & {
+  valueSchema: Schema;
+  keySchema: Schema;
+};
+
+export const decodeTable = (row: Hex[]): DeployedTable => {
```
Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/utils/decodeTable.ts:11]: what's this for? sqlite specifically or dozer specifically?

a lot of this logic looks duplicative of https://github.com/latticexyz/mud/blob/main/packages/store-sync/src/logToTable.ts

no change needed here, can do this as a follow up, but thinking we could lift this out into protocol-parser or something so you can go from the onchain representation of a table (a record from `Tables` table) into a proper table object

packages/explorer/src/common.ts
```diff
@@ -1,5 +1,7 @@
 import { anvil, garnet, redstone } from "viem/chains";
 
+export const internalMUDNamespaces = ["world", "store", "metadata"];
```
holic, line 3:
```suggestion
export const internalNamespaces = ["world", "store", "metadata"];
```
(we're in MUD and namespaces are a MUD concept so prob don't need to include this in the name)

packages/explorer/src/queries/dozer/useTableDataQuery.ts
```diff
@@ -0,0 +1,53 @@
...
+};
+
+export function useTableDataQuery({ schema, query }: Props) {
+  const { worldAddress, chainName } = useParams();
+  const dozerUrl = useDozerUrl();
+
+  return useQuery({
+    queryKey: ["table", worldAddress, chainName, query],
+    queryFn: async () => {
+      const response = await fetch(dozerUrl, {
```
holic, line 23: broad question, not strictly tied to this PR: have we fiddled with server actions instead of this react-query+fetch pattern?

packages/explorer/src/app/api/table/route.ts
```diff
@@ -1,34 +1,34 @@
...
-    return Response.json({ error: "table is required" }, { status: 400 });
-  }
+function doesTableExist(tableId: string) {
+  const db = getDatabase();
+  const result = db?.prepare("SELECT name FROM sqlite_master WHERE type='table' AND name = ?").get(tableId);
+
+  return Boolean(result);
+}
+
+export async function GET(request: Request) {
```
holic, line 18: similar question as above, curious if this could be a server action instead of a separate API route to fetch from

packages/explorer/src/queries/dozer/useDeployedTablesQuery.ts
```diff
@@ -0,0 +1,50 @@
...
+};
+
+export function useDeployedTablesQuery() {
+  const { worldAddress, chainName } = useParams();
+  const dozerUrl = useDozerUrl();
+
+  return useQuery({
+    queryKey: ["deployedTables", worldAddress, chainName],
+    queryFn: async () => {
+      const response = await fetch(dozerUrl, {
```
holic, line 20: maybe could use the helper for this?
https://github.com/latticexyz/mud/blob/main/packages/store-sync/src/sql/fetchRecords.ts

packages/explorer/src/queries/sqlite-indexer/useDeployedTablesQuery.ts
```diff
@@ -0,0 +1,44 @@
...
+};
+
+export function useDeployedTablesQuery() {
+  const { worldAddress } = useParams();
+  const TABLES_QUERY_TABLE = `${worldAddress}__store__tables`;
+
+  return useQuery({
+    queryKey: ["deployedTables", TABLES_QUERY_TABLE],
+    queryFn: async () => {
+      const response = await fetch(`/api/table?tableId=${TABLES_QUERY_TABLE}`);
```
holic, line 25: what if this endpoint returned a well-formed `Table` object instead of requiring the frontend to compute this? then the frontend only has to know about `Table`

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/utils/decodeTable.ts
Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/utils/decodeTable.ts:11]: This was originally used for Dozer endpoint since it returns all the data in hex form. It's now used for SQLite as well due to a different approach of how it's being fetched i.e. the data for the table is now coming from `store__Tables`.

packages/explorer/src/queries/sqlite-indexer/useDeployedTablesQuery.ts
karooolis: yeah, this might not the right level to do it at, and could be moved up to the endpoint. However, I'm also thinking that in case the local indexer exposes the same endpoint that matches Dozer one, it would be convenient to use the same query, just with different endpoints. Looking more into this

packages/explorer/src/queries/dozer/useTableDataQuery.ts
karooolis: interesting question I don't have a definite answer to yet. Even if we were to use server actions, react-query is helpful for polling. Another point to consider is that server-actions are blocking meaning they can't run in parallel. For the current setup I don't see that as a problem but could be down the line.

I was meaning to dive deeper into server actions, specifically the way they behave and impact behavior in next.js, to understand exactly where it could be useful. Just for the sake of type-safety I'd love to incorporate it if it actually makes sense but still need to do more research. For now the good old API is a battle-tested solution in my eyes.

packages/explorer/src/queries/dozer/useDeployedTablesQuery.ts
karooolis: oh this looks really useful, checking

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/DataExplorer.tsx
```diff
@@ -1,39 +1,40 @@
...
-  });
-  const selectedTable = searchParams.get("tableId") || (tables?.length > 0 ? tables[0] : null);
-
-  if (isLoading) {
-    return <Loader className="animate-spin" />;
-  }
+  useEffect(() => {
+    if (!deployedTable || !tableId) return;
+    if (chainId === anvil.id) {
+      setQuery(`SELECT * FROM "${tableId}"`);
```
Comment (karooolis) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/DataExplorer.tsx:27]: The handling of queries is still in question. At the moment it's being passed as a parameter but will adapt once there is a clearer consensus on how to handle it.

packages/explorer/src/app/(explorer)/hooks/useIndexerApiUrl.ts
```diff
@@ -0,0 +1,19 @@
+import { anvil, garnet, redstone } from "viem/chains";
+import { useChain } from "./useChain";
+
+const dozerUrl = {
+  [garnet.id]: "https://indexer.mud.garnetchain.com/q",
+  [redstone.id]: "https://indexer.mud.redstone.xyz/q",
```
alvrs, line 6:
```suggestion
  [redstone.id]: "https://indexer.mud.redstonechain.com/q",
```

.changeset/cyan-chefs-obey.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/explorer": patch
+---
+
+Hosted SQL API integration has been added for Redstone and Garnet.
```
holic, line 5: Exploring worlds on Redstone and Garnet chains will now retrieve data from the hosted SQL indexer.

packages/explorer/src/app/(explorer)/hooks/useIndexerApiUrl.ts
```diff
@@ -0,0 +1,19 @@
+import { anvil, garnet, redstone } from "viem/chains";
+import { useChain } from "./useChain";
+
+const dozerUrl = {
+  [garnet.id]: "https://indexer.mud.garnetchain.com/q",
+  [redstone.id]: "https://indexer.mud.redstonechain.com/q",
```
holic, line 6: these are on the chain config exported from MUD
https://github.com/latticexyz/mud/blob/6c056de6090a6f4a9633b96513ca1738dc0993c1/packages/common/src/chains/redstone.ts#L7
https://github.com/latticexyz/mud/blob/6c056de6090a6f4a9633b96513ca1738dc0993c1/packages/common/src/chains/garnet.ts#L7

packages/explorer/src/app/(explorer)/hooks/useIndexerApiUrl.ts
```diff
@@ -0,0 +1,19 @@
...
+  const { id: chainId } = useChain();
+  if (chainId === anvil.id) {
+    return "/api/sqlite-indexer";
+  }
+
+  if (!(chainId in dozerUrl)) {
+    throw new Error(`Dozer URL not found for chain id ${chainId}`);
+  }
+  return dozerUrl[chainId as keyof typeof dozerUrl];
+}
```
holic, line 19: can we move this to a regular function rather than hook and have it return an object like
```
{ type: "sqlite", url: "/api/sqlite-indexer" }
```

```
{ type: "hosted", url: `${chain.indexerUrl}/q` }
```

then in other areas where we query each indexer, rather than checking the chain ID, we check the `type` to determine the format of the sql?

packages/explorer/src/app/(explorer)/queries/useDeployedTablesQuery.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  return useQuery<DozerResponse, Error, DeployedTable[]>({
+    queryKey: ["deployedTables", worldAddress, chainName],
+    queryFn: async () => {
+      const storeTablesKey = "store__Tables";
+      const tableName = chainId === anvil.id ? `${worldAddress}__${storeTablesKey}` : storeTablesKey;
+      const query =
+        chainId === anvil.id
+          ? `SELECT * FROM "${tableName}"`
+          : `SELECT ${Object.keys(mudConfig.tables[storeTablesKey].schema).join(", ")} FROM ${tableName}`;
```
holic, line 25: this is one of those areas I am worried about the chain ID getting sprinkled around
```suggestion
      const tableName = "store__Tables";
      const query =
        indexer.type === 'sqlite'
          ? `SELECT * FROM "${worldAddress}__${tableName}"`
          : `SELECT ${Object.keys(mudConfig.tables[tableName].schema).join(", ")} FROM ${tableName}`;
```

packages/explorer/src/components/ui/Table.tsx
```diff
@@ -1,5 +1,5 @@
 import * as React from "react";
-import { cn } from "../../lib/utils";
+import { cn } from "../../app/(explorer)/lib/utils";
```
holic, line 2: maybe we need to move `cn` out of this dir?

karooolis: sure, move back to the root? I moved all utils used only from within the `explorer` into here but agree that somehow doesn't look right

packages/explorer/src/app/(explorer)/hooks/useIndexerApiUrl.ts
karooolis: good point, will adjust

packages/explorer/src/app/(explorer)/hooks/useIndexerApiUrl.ts
karooolis: perfect, updated

packages/explorer/src/components/ui/Table.tsx
karooolis: moved it back to root so that common small util functions potentially useful across the whole app are stored in `src/utils.ts`.

packages/explorer/src/app/(explorer)/queries/useDeployedTablesQuery.ts
karooolis: Yeah see your point, like this improvement, updated

packages/explorer/src/components/ui/Table.tsx
holic: I generally follow the rule that when you import a thing from somewhere else deep in the app, it's probably time to lift it up into some common dir that can be used in both places.

packages/explorer/src/app/(explorer)/api/utils/decodeTable.ts
```diff
@@ -0,0 +1,48 @@
+import { Hex, decodeAbiParameters, parseAbiParameters } from "viem";
+import { hexToResource } from "@latticexyz/common";
+import { Schema, Table } from "@latticexyz/config";
+import { getSchemaPrimitives, hexToSchema } from "@latticexyz/protocol-parser/internal";
+
+export type DeployedTable = Table & {
+  valueSchema: Schema;
+  keySchema: Schema;
+};
```
holic, line 9: where do we use `keySchema` and `valueSchema` in explorer? curious if we can move away from it or if it's important to have

packages/explorer/src/app/(explorer)/api/utils/decodeTable.ts
```diff
@@ -0,0 +1,48 @@
...
+  ) satisfies Schema;
+  const valueSchema = Object.fromEntries(
+    valueAbiTypes.map((abiType, i) => [fieldNames[i], { type: abiType, internalType: abiType }]),
+  ) satisfies Schema;
+
+  return {
+    tableId: tableId as Hex,
+    name,
+    namespace,
+    label: namespace,
```
holic, line 37:
```suggestion
    label: name,
```

packages/explorer/src/app/(explorer)/api/utils/decodeTable.ts
holic: ah okay, so it just sounds like the helpers we have don't cover the use cases here to avoid having to pass around keySchema/valueSchema

we should add those to protocol-parser etc. but doesn't have to happen here

karooolis: just checked, look like it does have some useful utilities like `getKey` or `getKeyTuple`. I'll check more which ones I can use

karooolis: Alright, did some refactoring, and made use of `getValueSchema()` + `getKeySchema()`. So removed `DeployedTable` type, and just using `Table` from `@latticexyz/config`. 

A nice utility that would be nice to have and that I can't find is constructing the key field for `setField` method. Not sure though if I just can't find it.

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/EditableTableCell.tsx
```diff
@@ -4,44 +4,48 @@ import { toast } from "sonner";
...
-import { TableConfig } from "../../../../../api/table/route";
+import { cn } from "../../../../../../utils";
+import { useChain } from "../../../../hooks/useChain";
 
 type Props = {
   name: string;
   value: string;
-  keyTuple: string[];
-  config: TableConfig;
+  tableConfig: Table;
```
Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/EditableTableCell.tsx:25]: I liked the use of `table` throughout 🙈

holic (approved):
Looks good!

I personally prefer the `table` naming of `Table` objects rather than `tableConfig` as it matches the rest of MUD (minus stash)

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/EditableTableCell.tsx
Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/EditableTableCell.tsx]: moved to `table`

PR #3188, opened 2024-09-17
docs(guide/dozer): move hydration to its own guide 🚗 

qbzzt:
Moved the "adding dozer hydration guide" from the dozer page to its own page.


ludns:
Let's add the callout to that page too? The one that explains the two different types of sync

qbzzt:
> Let's add the callout to that page too? The one that explains the two different types of sync

Done

PR #3195, opened 2024-09-18
fix(explorer): various fixes

holic:
- made `observer()` options arg optional
- fixed favicon
- fixed regression in explorer vs internal layouts
- fixed redirects
- fixed issues with validating chain IDs and names
- reworked iframe rendering to reuse existing iframes for the same url


alvrs (approved):

packages/explorer/src/app/(explorer)/icon.svg
karooolis, line 1: See it has been changed back to `icon.svg`. Would be nice to use layout's metadata but can't figure out quickly how to make that work from a non-root path.

packages/explorer/src/common.ts
```diff
@@ -1,21 +1,23 @@
 import { anvil, garnet, redstone } from "viem/chains";
 
 export const supportedChains = { anvil, garnet, redstone } as const;
-export const supportedChainsById = Object.fromEntries(
-  Object.entries(supportedChains).map(([, chain]) => [chain.id, chain]),
-);
+export type supportedChains = typeof supportedChains;
```
karooolis, line 4: Wonder if it should start uppercase given it's a type, and that's been the convention so far?

karooolis (approved):

packages/explorer/src/app/(explorer)/icon.svg
holic: yeah I couldn't figure out why specifying the path in page `meta` lead to a bunch of invalid requests to the icon, even though the meta tag rendered fine and had the correct path

changing this to just `icon.svg` follows the next.js convention which serves the icon automatically: https://nextjs.org/docs/app/api-reference/file-conventions/metadata/app-icons#image-files-ico-jpg-png

packages/explorer/src/common.ts
holic: this is just a shortcut to avoid having to sprinkle `typeof` everywhere you're using it with a `const`

when the name mirrors the type, you can use the same name/reference in both types and runtime

PR #3197, opened 2024-09-18
fix(cli): improve performance of linked library resolution during deployment

alvrs:
Significantly improved the deployment performance for large projects with public libraries by implementing a more efficient algorithm to resolve public libraries during deployment.
The local deployment time on a large reference project was reduced from over 10 minutes to 4 seconds.

### DIFF
--- a/packages/cli/src/deploy/createPrepareDeploy.ts
+++ b/packages/cli/src/deploy/createPrepareDeploy.ts
@@ -1,7 +1,7 @@
 import { DeterministicContract, LibraryPlaceholder, salt } from "./common";
 import { spliceHex } from "@latticexyz/common";
 import { Hex, getCreate2Address, Address } from "viem";
-import { LibraryMap, getLibraryKey } from "./getLibraryMap";
+import { LibraryMap } from "./getLibraryMap";
 
 export function createPrepareDeploy(
   bytecodeWithPlaceholders: Hex,
@@ -10,16 +10,8 @@ export function createPrepareDeploy(
   return function prepareDeploy(deployer: Address, libraryMap: LibraryMap) {
     let bytecode = bytecodeWithPlaceholders;
     for (const placeholder of placeholders) {
-      const libraryKey = getLibraryKey(placeholder);
-      const library = libraryMap[libraryKey];
-      if (!library) {
-        throw new Error(`Could not find library for bytecode placeholder ${placeholder.path}:${placeholder.name}`);
-      }
-
-      // Store the prepared address in the library map to avoid preparing the same library twice
-      library.address ??= library.prepareDeploy(deployer, libraryMap).address;
-
-      bytecode = spliceHex(bytecode, placeholder.start, placeholder.length, library.address);
+      const address = libraryMap.getAddress({ name: placeholder.name, path: placeholder.path, deployer });
+      bytecode = spliceHex(bytecode, placeholder.start, placeholder.length, address);
     }
     return {
       bytecode,
--- a/packages/cli/src/deploy/createPrepareDeploy.ts
+++ b/packages/cli/src/deploy/createPrepareDeploy.ts
@@ -7,8 +7,17 @@ export function createPrepareDeploy(
   bytecodeWithPlaceholders: Hex,
   placeholders: readonly LibraryPlaceholder[],
 ): DeterministicContract["prepareDeploy"] {
-  return function prepareDeploy(deployer: Address, libraryMap: LibraryMap) {
+  return function prepareDeploy(deployer: Address, libraryMap?: LibraryMap) {
     let bytecode = bytecodeWithPlaceholders;
+
+    if (placeholders.length === 0) {
+      return { bytecode, address: getCreate2Address({ from: deployer, bytecode, salt }) };
+    }
+
+    if (!libraryMap) {
+      throw new Error("Libraries must be provided if there are placeholders");
+    }
+
     for (const placeholder of placeholders) {
       const address = libraryMap.getAddress({ name: placeholder.name, path: placeholder.path, deployer });
       bytecode = spliceHex(bytecode, placeholder.start, placeholder.length, address);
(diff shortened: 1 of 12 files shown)


packages/cli/src/deploy/createPrepareDeploy.ts
```diff
@@ -1,24 +1,25 @@
...
   bytecodeWithPlaceholders: Hex,
   placeholders: readonly LibraryPlaceholder[],
 ): DeterministicContract["prepareDeploy"] {
-  return function prepareDeploy(deployer: Address, libraries: readonly Library[]) {
+  return function prepareDeploy(deployer: Address, libraryMap: LibraryMap) {
     let bytecode = bytecodeWithPlaceholders;
     for (const placeholder of placeholders) {
-      const library = libraries.find((lib) => lib.path === placeholder.path && lib.name === placeholder.name);
+      const libraryKey = getLibraryKey(placeholder);
+      const library = libraryMap[libraryKey];
```
alvrs, line 14: Previously we were doing a linear search to find the matching library every time is function was called, now we're just looking up the matching library in constant time. Doesn't make a big difference with small numbers of libraries but compounds with many libraries.

packages/cli/src/deploy/createPrepareDeploy.ts
```diff
@@ -1,24 +1,25 @@
...
-        bytecode,
-        placeholder.start,
-        placeholder.length,
-        library.prepareDeploy(deployer, libraries).address,
-      );
+
+      // Store the prepared address in the library map to avoid preparing the same library twice
+      library.address ??= library.prepareDeploy(deployer, libraryMap).address;
+
+      bytecode = spliceHex(bytecode, placeholder.start, placeholder.length, library.address);
```
alvrs, line 22: Previously we were resolving each library multiple times if it was used in multiple contracts/libraries. The time complexity of this quickly explodes if there are many libraries that use other libraries. Now we're caching a library's resolved address, so each library is only resolved once.

packages/cli/src/runDeploy.ts
```diff
@@ -133,6 +133,7 @@ export async function runDeploy(opts: DeployOptions): Promise<WorldDeploy> {
         : undefined,
     }),
     account,
+    pollingInterval: 100,
```
alvrs, line 136: Reducing polling time for a short time during deployment should be safe and makes a significant impact on deployment time (with the project i was testing with this change brought the time from 16s to 4s after the previous improvements were implemented)

alvrs: (moved this to `waitForTransactions` instead to only increase the polling rate where it matters)

alvrs: Moved this change to #3198 since it's unrelated to library linking, and CI wasn't happy with it yet

packages/cli/src/deploy/getLibraryMap.ts
```diff
@@ -0,0 +1,12 @@
...
+import { Library } from "./common";
+
+export type LibraryMap = { [key: string]: Library & { address?: Hex } };
+
+export function getLibraryMap(libraries: readonly Library[]): LibraryMap {
+  return Object.fromEntries(libraries.map((library) => [getLibraryKey(library), library]));
+}
+
+export function getLibraryKey({ path, name }: { path: string; name: string }): string {
+  return `${path}:${name}`;
```
holic, line 11: should we include the deployer in the key since it changes the bytecode and thus the address?

packages/cli/src/deploy/deployWorld.ts
```diff
@@ -24,13 +24,7 @@ export async function deployWorld(
...
   });
 
-  debug("waiting for world deploy");
-  const receipt = await waitForTransactionReceipt(client, { hash: tx });
-  if (receipt.status !== "success") {
-    console.error("world deploy failed", receipt);
-    throw new Error("world deploy failed");
-  }
-
+  const [receipt] = await waitForTransactions({ client, hashes: [tx], debugLabel: "world deploy" });
```
holic, line 27: iirc I intentionally didn't replace this one with `waitForTransactions` because I wanted to throw a more specific error

packages/cli/src/verify.ts
```diff
@@ -105,7 +105,7 @@ export async function verify({
     );
 
     modules.map(({ name, prepareDeploy }) => {
-      const { address } = prepareDeploy(deployerAddress, []);
+      const { address } = prepareDeploy(deployerAddress, {});
```
holic, line 108: thoughts on defaulting the second arg in `prepareDeploy` so callers don't need to know the shape of the thing to pass in

packages/cli/src/deploy/getLibraryMap.ts
```diff
@@ -0,0 +1,12 @@
+import { Hex } from "viem";
+import { Library } from "./common";
+
+export type LibraryMap = { [key: string]: Library & { address?: Hex } };
+
+export function getLibraryMap(libraries: readonly Library[]): LibraryMap {
+  return Object.fromEntries(libraries.map((library) => [getLibraryKey(library), library]));
```
holic, line 7: not blocking but I wonder if we should just return some interface with `get` and `set` and leave the internal representation and key shape as implementation details, then you can use this almost like a regular map but with 2-3 args for the key (path + name + maybe deployer address)

holic (dismissed):
lgtm!

main question is whether or not we need to key by deployer since we cache the address and the address is based on the deployer

packages/cli/src/deploy/getLibraryMap.ts
alvrs: the deployer address is passed in later (in `prepareDeploy`) while the map is created before, but we can change `address` in the map to `{ [deployer: Address]: Address }` to cache per deployer address

packages/cli/src/deploy/deployWorld.ts
alvrs: reason i changed this was bc i changed polling settings in `waitForTransactions` and would have had to duplicate the changes here, but moved that to #3198 so could also move this change there

holic (approved):
nice! I dig the new `LibraryMap`

PR #3203, opened 2024-09-19
feat(explorer): filterable tables selector

karooolis:
CleanShot 2024-09-19 at 14 15 41


.changeset/rare-ducks-relate.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/explorer": patch
+---
+
+World Explorer now includes filterable tables selector.
```
holic, line 5:
```suggestion
Table selector of the Explore tab now has an input for searching/filtering tables by name.
```

Don't think we need to mention "World Explorer" here because the changeset is already "bound" to this package

packages/explorer/src/components/ui/Command.tsx
holic, line 1: is this a shadcn/ui thing?

karooolis: [yeah](https://ui.shadcn.com/docs/components/command), although heavily relies on [cmdk](https://github.com/pacocoursey/cmdk)

holic: yeah I had the same thought, would be cool!

PR #3209, opened 2024-09-20
chore(explorer): detect dead-code

karooolis:
Added `knip` to help detect dead-code.


holic (approved):

packages/explorer/package.json
```diff
@@ -31,6 +31,7 @@
     "dev": "tsup --watch",
     "explorer:dev": "next dev --port 13690",
     "explorer:start": "node .next/standalone/packages/explorer/server.js",
+    "knip": "knip",
```
holic, line 34: if this is just meant to be run as a one-off as-needed, could achieve the same thing with `pnpx knip` and avoid having to add the dep to the repo

PR #3215, opened 2024-09-20
feat(store-sync): apply optimistic logs

holic:


packages/store-sync/src/createStoreSync.ts
```diff
@@ -219,29 +243,49 @@ export async function createStoreSync({
...
+          concatMap(async () => {
+            if (lastBlockNumberProcessed != null) {
+              await applyOptimisticLogs(lastBlockNumberProcessed);
+            }
+          }),
+          tap(() => storageAdapterLock$.next(false)),
+          ignoreElements(),
+        ),
+      );
+    }),
```
alvrs, line 270: does the `storageAdapterLock$` logic need to be rxjs based or could this be a boolean variable? i always need extra brain cycles to think through rxjs logic

packages/store-sync/src/createStoreSync.ts
```diff
@@ -199,6 +202,27 @@ export async function createStoreSync({
   let endBlock: bigint | null = null;
   let lastBlockNumberProcessed: bigint | null = null;
 
+  // For chains that provide guaranteed receipts ahead of block mining, we can apply the logs immediately.
+  // This works because, once the block is mined, the same logs will be applied. Store events are defined in
+  // such a way that reapplying the same logs, even if the order changes, will mean that the storage adapter
```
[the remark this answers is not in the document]
holic: ahh I guess I didn't describe the nuance well here

order is important, but we can apply some/all of the logs before the block is mined and then applying the logs from the mined block is effectively a no-op for all the previously set records (assuming the node guarantees the receipt)

packages/store-sync/src/createStoreSync.ts
holic: it's an observable because we need to subscribe to it in two places (one here to wait for the lock to free up and one in `waitForTransaction`)

alvrs: Maybe it's because it's late but i've stared at this part of the code for a while and am still not fully sure what's going on 🙈 What is the `concat` of three streams doing, out of which the first and last have a `ignoreElements`? why is the last stream `of(true)`?

alvrs: couldn't we do something like
```ts
for(const block of storedBlocks) {
  storageAdapterLock.next(true);
  await applyOptimisticLogs(block.blockNumber);
  storageAdapterLock.next(false);
}
```
or 
```ts
storedBlock$.pipe(concatMap(block => {
  storageAdapterLock$.next(true);
  await applyOptimisticLogs(lastBlockNumberProcessed);
  storageAdapterLock$.next(false);
}))
```
?

holic: Okay this took a bunch of iteration to get here but here's the idea:

The current approach turns an async generator into a stream of blocks, one emission per yield of the generator. I originally had your suggested approach above, but found we were applying optimistic logs for each block rather than each range of blocks, doing a lot more work than is necessary, especially when hydrating from RPC.

Instead what this does is takes the block range stream (`storedBlock$` but could rename for clarity) and adds an operation before and after the stream: before to wait for + take out a lock and after to apply optimistic logs and release the lock. Concatenating arbitrary things to the nice stream of `storedBlocks$` means we'd have to find/filter them out later, but we can encapsulate that with `ignoreElements()` which basically empties the stream we added to the start/end and only completes/errors the stream.

So it works like this:
- create a block stream/observable for the range of blocks
- before we start evaluating that stream (and thus fetching+storing data), force the stream to wait for the lock to release in case there are existing pending operations
- once unlocked, take out a lock for the duration of the range
- process each block in the range
- once range is done, apply optimistic logs and release the lock before exiting the stream

PR #3216, opened 2024-09-22
chore: add sort-package-json dev dep

holic:
pulled out of https://github.com/latticexyz/mud/pull/3215

adds sort-package-json as a monorepo dev dep, which greatly improved git hook speed when on an unstable internet connection

also renamed to not collide with the dep's bin name


holic:
just a monorepo quality of life change, so merging

PR #3222, opened 2024-09-23
refactor(cli): adjust deploy order

holic:
deploy world then user contracts instead of the other way around, to keep from spending gas on contract deploys in cases where world deploy fails


packages/cli/src/deploy/deploy.ts
```diff
@@ -62,9 +61,29 @@ export async function deploy({
 }: DeployOptions): Promise<WorldDeploy> {
   const deployerAddress = initialDeployerAddress ?? (await ensureDeployer(client));
 
-  await ensureWorldFactory(client, deployerAddress, config.deploy.upgradeableWorldImplementation);
```
holic, line 65: this happens inside `deployWorld`, so no need to double up on it here (or do it when a custom world is used)

PR #3225, opened 2024-09-23
feat(explorer): move filter state to url

karooolis:
Added type-safe URL search params using `nuqs`. Given that the Explorer is going to have more filters, and other state, in the future, it's useful to add it now. Also, table state is now stored in URL search params as well so it can be shared.

### DIFF
--- a/.changeset/sixty-comics-love.md
+++ b/.changeset/sixty-comics-love.md
@@ -1,0 +1,5 @@
+---
+"@latticexyz/explorer": patch
+---
+
+Added support for tracking table filtering, sorting, and additional parameters like selected tableId using type-safe URL search params. This ensures that state changes persist across sessions and can be shared with a URL.
--- a/.changeset/sixty-comics-love.md
+++ b/.changeset/sixty-comics-love.md
@@ -2,4 +2,4 @@
 "@latticexyz/explorer": patch
 ---
 
-Added support for tracking table filtering, sorting, and additional parameters like selected tableId using type-safe URL search params. This ensures that state changes persist across sessions and can be shared with a URL.
+Table filters are now included as part of the URL. This enables deep links and improves navigating between pages without losing search state.
--- a/packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TableSelector.tsx
+++ b/packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TableSelector.tsx
@@ -1,6 +1,8 @@
 import { Check, ChevronsUpDown, Lock } from "lucide-react";
 import { useParams } from "next/navigation";
-import { useState } from "react";
+import { useQueryState } from "nuqs";
+import { Hex } from "viem";
+import { useEffect, useState } from "react";
 import { internalTableNames } from "@latticexyz/store-sync/sqlite";
 import { Button } from "../../../../../../components/ui/Button";
 import {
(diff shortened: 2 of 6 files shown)


packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TableSelector.tsx
```diff
@@ -15,20 +17,28 @@ import { Popover, PopoverContent, PopoverTrigger } from "../../../../../../compo
...
 
-export function TableSelector({ value, options }: Props) {
+export function TableSelector({ tables }: Props) {
+  const [selectedTableId, setTableId] = useQueryState("tableId");
   const [open, setOpen] = useState(false);
   const { worldAddress } = useParams();
 
+  useEffect(() => {
+    if (!selectedTableId && tables.length > 0) {
+      setTableId(tables[0] as Hex);
```
Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TableSelector.tsx:30]: does this operation add another history item? if so that might make the back button a little hard to use, where each time you hit back, it triggers this effect again which adds to history, making back a bit of a no-op unless you can skip past the page where there's no table ID set

ideally this operation would use `replace` here

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TablesViewer.tsx
```diff
@@ -21,22 +20,19 @@ import { Table, TableBody, TableCell, TableHead, TableHeader, TableRow } from ".
...
-  const [rowSelection, setRowSelection] = useState({});
-  const [globalFilter, setGlobalFilter] = useState("");
-  const [showAllColumns, setShowAllColumns] = useState(false);
+export function TablesViewer() {
+  const [selectedTableId] = useQueryState("tableId");
+  const [globalFilter, setGlobalFilter] = useQueryState("filter", parseAsString.withDefault(""));
+  const [showAllColumns, setShowAllColumns] = useQueryState("showAllColumns", parseAsBoolean.withDefault(false));
+  const [sorting, setSorting] = useQueryState(
+    "sort",
+    parseAsArrayOf(parseAsJson<ColumnSort>()).withDefault(useRef([]).current),
```
Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TablesViewer.tsx:29]: does `parseAsJson` not support arrays? why do you need the outer `parseAsArrayOf`?

Comment (holic) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TablesViewer.tsx:29]: also what's this `useRef([]).current` pattern for? why not just `[]`

if you need a stable ref, could lift it out of the component like `const emptyColumnSort = []`

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TablesViewer.tsx
```diff
@@ -21,22 +20,19 @@ import { Table, TableBody, TableCell, TableHead, TableHeader, TableRow } from ".
...
+  );
 
   const { data: schema } = useQuery({
-    queryKey: ["schema", { table: selectedTable }],
+    queryKey: ["schema", { table: selectedTableId }],
     queryFn: async () => {
-      const response = await fetch(`/api/schema?table=${selectedTable}`);
+      const response = await fetch(`/api/schema?table=${selectedTableId}`);
```
Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TablesViewer.tsx:35]: would suggest getting in the habit of encoding URL params to save you some headache later

```suggestion
      const response = await fetch(`/api/schema?table=${encodeURICompoonent(selectedTableId)}`);
```
or

```suggestion
      const response = await fetch(`/api/schema?${new URLSearchParams({ table: selectedTableId })}`);
```

same as all other usages below

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TableSelector.tsx
Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TableSelector.tsx:30]: The default for `nuqs` is `replace`

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TablesViewer.tsx
Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TablesViewer.tsx]: Yeah, I needed a stable array reference to avoid infinite re-render. Lifted out initial value outside the component, was not sure which way is cleaner.

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TablesViewer.tsx
Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TablesViewer.tsx:33]: That's a nice tip, will do that

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TablesViewer.tsx
```diff
@@ -21,22 +19,18 @@ import { Table, TableBody, TableCell, TableHead, TableHeader, TableRow } from ".
 import { bufferToBigInt } from "../utils/bufferToBigInt";
 import { EditableTableCell } from "./EditableTableCell";
 
-type Props = {
-  table: string | undefined;
-};
+const emptyColumnSort: SortingState = [];
```
Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TablesViewer.tsx:22]: could also call this `initialSortingState`, which I use as a pattern quite often and might be clearer than either approach we had before

.changeset/sixty-comics-love.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/explorer": patch
+---
+
+Added support for tracking table filtering, sorting, and additional parameters like selected tableId using type-safe URL search params. This ensures that state changes persist across sessions and can be shared with a URL.
```
holic, line 5:
```suggestion
Table filters are now included as part of the URL. This enables deep links and improves navigating between pages without losing search state.
```

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TablesViewer.tsx
Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TablesViewer.tsx:22]: agreed, a better name, renamed

holic (approved):

PR #3226, opened 2024-09-23
feat(store-sync): add support for live sync from indexer

alvrs:


packages/store-sync/src/createStoreSync.ts
```diff
@@ -199,14 +201,25 @@ export async function createStoreSync({
...
   let lastBlockNumberProcessed: bigint | null = null;
 
-  const storedBlock$ = combineLatest([startBlock$, latestBlockNumber$]).pipe(
+  const indexerLogs$ = indexerUrl
+    ? startBlock$.pipe(
+        mergeMap((startBlock) => {
+          const input = encodeURIComponent(JSON.stringify({ chainId, address, filters }));
+          return fromEventSource<string>(
+            new URL(`/api/logs-live?input=${input}&block_num=${startBlock}&include_tx_hash=true`, indexerUrl),
+          ).pipe(map((messageEvent) => JSON.parse(messageEvent.data) as StorageAdapterBlock));
```
holic, line 210: I find it more robust+readable to build up URL params like
```suggestion
          return fromEventSource<string>(
            new URL(`/api/logs-live?${new URLSearchParams({
              input: JSON.stringify({ chainId, address, filters }),
              block_num: startBlock.toString(),
              include_tx_hash: "true",
            })}`, indexerUrl),
          ).pipe(map((messageEvent) => JSON.parse(messageEvent.data) as StorageAdapterBlock));
```

packages/store-sync/src/createStoreSync.ts
```diff
@@ -215,18 +228,30 @@ export async function createStoreSync({
           ? bigIntMax(range.startBlock, lastBlockNumberProcessed + 1n)
           : range.startBlock,
         toBlock: range.endBlock,
-        storageAdapter,
         logFilter,
       });
```
holic, line 232: one of the reasons I originally coupled the fetch+store operation is so that the fetching wouldn't get ahead of the storing of logs, creating a potential memory explosion

do you know if that's a concern here in the way rxjs processes the stream?

alvrs: https://playcode.io/2019292 -> if storing the logs takes longer than fetching the logs, it would indeed buffer the logs in rxjs and potentially increase memory more than if we only fetch when the previous logs were applied. Wonder if this is actually significant here though - when loading the snapshot from the indexer we already keep the entire snapshot in memory while it's being fed into the storage adapter and the RPC sync wouldn't produce a larger in-memory backlog than that (especially not during live sync, if anything only during the initial sync from RPC, but even in that case it shouldn't be much more than the snapshot).

holic: iirc this was mostly an issue in indexers where you are fetching the entire chain state (many worlds from block 0) and fetching was much faster than storing into the DB, so the process would repeatedly get OOM-killed

alvrs: reverted to `fetchAndStoreLogs` and storing logs separately in the indexer logs stream

packages/store-sync/src/createStoreSync.ts
```diff
@@ -199,14 +201,25 @@ export async function createStoreSync({
   let endBlock: bigint | null = null;
   let lastBlockNumberProcessed: bigint | null = null;
 
-  const storedBlock$ = combineLatest([startBlock$, latestBlockNumber$]).pipe(
+  const indexerLogs$ = indexerUrl
+    ? startBlock$.pipe(
+        mergeMap((startBlock) => {
+          const input = encodeURIComponent(JSON.stringify({ chainId, address, filters }));
+          return fromEventSource<string>(
+            new URL(`/api/logs-live?input=${input}&block_num=${startBlock}&include_tx_hash=true`, indexerUrl),
```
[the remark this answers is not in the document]
alvrs: updated to remove the leading `/`, which will cause the url to consider subpaths (if the `indexerUrl` is provided with trailing `/`)

packages/store-sync/src/createStoreSync.ts
```diff
@@ -199,7 +201,28 @@ export async function createStoreSync({
...
+          );
+          return fromEventSource<string>(url);
+        }),
+        map((messageEvent) => JSON.parse(messageEvent.data) as StorageAdapterBlock),
+        concatMap(async (block) => {
+          await storageAdapter(block);
+          return block;
+        }),
+      )
+    : throwError(() => new Error("No indexer URL provided"));
```
holic, line 223: looks like e2e tests are failing here

packages/store-sync/src/createStoreSync.ts
```diff
@@ -199,7 +201,28 @@ export async function createStoreSync({
...
+            `api/logs-live?${new URLSearchParams({
+              input: JSON.stringify({ chainId, address, filters }),
+              block_num: startBlock.toString(),
+              include_tx_hash: "true",
+            })}`,
+            indexerUrl,
+          );
+          return fromEventSource<string>(url);
+        }),
+        map((messageEvent) => JSON.parse(messageEvent.data) as StorageAdapterBlock),
```
holic, line 217: is this a whole block? not a log?

packages/store-sync/src/createStoreSync.ts
```diff
@@ -215,13 +238,21 @@ export async function createStoreSync({
...
     }),
-    tap(({ blockNumber, logs }) => {
+  );
+
+  const storedBlock$ = storedIndexerLogs$.pipe(
+    catchError((e) => {
+      debug("error streaming logs from indexer:", e);
+      debug("falling back to streaming logs from RPC");
+      return storedRpcLogs$;
+    }),
```
holic, line 254: have we tested the fallback behavior here?

packages/store-sync/src/fromEventSource.ts
```diff
@@ -0,0 +1,10 @@
+import { Observable } from "rxjs";
+
+export function fromEventSource<T>(url: string | URL): Observable<MessageEvent<T>> {
+  return new Observable<MessageEvent>((subscriber) => {
+    const eventSource = new EventSource(url);
+    eventSource.onmessage = (ev): void => subscriber.next(ev);
+    eventSource.onerror = (ev): void => subscriber.error(ev);
```
holic, line 7: I think `onerror` gives you an `Event`, not an `Error`, so I wonder if it's a bad idea to `subscriber.error` with an `Event`. Is there an error embedded in the event we can use instead? or turn it into some error?

packages/store-sync/src/createStoreSync.ts
alvrs: It's a whole block, but it's true that the type is not quite accurate, it comes back as a string from the indexer (same for the `blockNumber` on each log). Thinking about the best way to handle this, would like to avoid having to iterate through every log here to turn the string into a bigint :/

packages/store-sync/src/createStoreSync.ts
alvrs: The fallback behavior already kicks in if no indexer is provided or the existing indexer doesn't support this API, so in that sense it's covered by the existing e2e tests

packages/store-sync/src/createStoreSync.ts
alvrs: never mind, the block number only appears once in the response, so we don't have to iterate through the logs!

holic: if this came back as an RPC log (i.e. hex encoding), we could use viem to convert things like bigints etc.

if this is a "sql api" response, could we use our tools to convert between sql api responses and JS types?

packages/store-sync/src/createStoreSync.ts
alvrs: fixed by not logging the entire error where we catch it but just the error message

packages/store-sync/src/fromEventSource.ts
```diff
@@ -0,0 +1,10 @@
+import { Observable } from "rxjs";
+
+export function fromEventSource<T>(url: string | URL): Observable<MessageEvent<T>> {
+  return new Observable<MessageEvent>((subscriber) => {
+    const eventSource = new EventSource(url);
+    eventSource.onmessage = (ev): void => subscriber.next(ev);
+    eventSource.onerror = (): void => subscriber.error(new Error("Event source closed: " + url));
```
holic, line 7: I am pretty sure `EventSource` attempts to reconnect during an error (I think indefinitely) but unclear if it emits `onerror` during this reconnect flow. We might want to give it a bad URL and see if this emits a bunch of `onerror` and maybe allow a few reconnection attempts before giving up and closing the stream with `subscriber.error`?

packages/store-sync/src/createStoreSync.ts
alvrs: It's the same response that we get from the `/logs` api, reused the same helper to verify the data matches the expected format now

packages/store-sync/src/isStorageAdapterBlock.ts
holic, line 1: should this be in `store-sync/src/sql`? because its specific to sql api?

holic: and could we have a corresponding `toStorageAdapterBlock` that contains the `blockNumber `conversion?

packages/store-sync/src/fromEventSource.ts
alvrs: Just tried that, it does indeed try to reconnect, but it seems like once it reconnects it skips the blocks that happened while the connection was down and just continues with the current block. I think we can solve this on the indexer side by using the block number as event id and also add code to handle reconnections with `last_event_id`, but would leave that as an optimization for later.

packages/store-sync/src/isStorageAdapterBlock.ts
alvrs: it's not specific to the sql api, it's the `/logs` api

alvrs: could rename to `isIndexerStorageAdapterBlock` or `isIndexerLogsResponse` or something 🤷‍♂️

packages/store-sync/src/fromEventSource.ts
holic: yeah exactly, we need to update the indexer SSE stream to include an `id` and use the `Last-Event-ID` header for reconnections

holic: block number would likely need to be block number + log index

packages/store-sync/src/isStorageAdapterBlock.ts
holic: either way 🤷 just something to differentiate from the `StorageAdapterBlock` type (which already uses `bigint`)

alvrs: renamed to `isLogsApiResponse`

packages/store-sync/src/fromEventSource.ts
```diff
@@ -0,0 +1,10 @@
+import { Observable } from "rxjs";
+
+export function fromEventSource<T>(url: string | URL): Observable<MessageEvent<T>> {
+  return new Observable<MessageEvent>((subscriber) => {
+    const eventSource = new EventSource(url);
+    eventSource.onmessage = (ev): void => subscriber.next(ev);
+    eventSource.onerror = (): void => subscriber.error(new Error("Event source failed" + new URL(url).origin));
```
holic, line 7:
```suggestion
    // we immediately close instead of allowing `EventSource` to retry because the logs API doesn't support `Last-Event-ID` yet
    eventSource.onerror = (): void => subscriber.error(new Error("Event source failed" + new URL(url).origin));
```

alvrs: Since `fromEventSource` is technically not specific to the logs API, we should probably add an option to `fromEventSource` to enable/disable retries, so consumers can decide based on the API capabilities. Would leave for a followup

alvrs: #3246

packages/store-sync/src/createStoreSync.ts
```diff
@@ -61,7 +64,7 @@ export async function createStoreSync({
   maxBlockRange,
   initialState,
   initialBlockLogs,
-  indexerUrl,
+  indexerUrl: indexerUrlInput,
```
holic, line 67: small nit: I tend to name these "initial"
```suggestion
  indexerUrl: initialIndexerUrl,
```

packages/store-sync/src/createStoreSync.ts
```diff
@@ -199,7 +202,34 @@ export async function createStoreSync({
...
+            indexerUrl,
+          );
+          return fromEventSource<string>(url);
+        }),
+        map((messageEvent) => {
+          const data = JSON.parse(messageEvent.data);
+          if (!isLogsApiResponse(data)) {
+            throw new Error("Received unexpected from indexer:" + messageEvent.data);
+          }
+          return { ...data, blockNumber: BigInt(data.blockNumber) };
```
holic, line 223: not blocking but might be nice to move these lines into a `toStorageAdapterBlock` since we have this logic in a couple spots

arktype can help a lot with this pattern of parsing+validating+strong types but can save that for a later improvement

holic (dismissed):

PR #3229, opened 2024-09-24
feat(explorer): local indexer inside explorer

karooolis:
Local store indexer is now run as part of the Explorer package. It is skipped if the configured chain is non-anvil.


packages/explorer/src/bin/explorer.ts
```diff
@@ -97,6 +101,30 @@ async function startExplorer() {
...
 }
 
+async function startStoreIndexer() {
+  if (chainId !== anvil.id) {
+    console.log("Skipping SQLite indexer for non-anvil chain ID", chainId);
+    return;
+  }
+
+  if (existsSync(indexerDatabasePath)) {
+    execSync(`shx rm -rf ${indexerDatabasePath}`);
```
holic, line 111: since we're in node, can we use `fs.rm(indexerDatabasePath, {recursive:true, force:true})`

packages/explorer/src/bin/explorer.ts
```diff
@@ -97,6 +101,30 @@ async function startExplorer() {
...
+  }
+
+  console.log("Running SQLite indexer for anvil...");
+  indexerProcess = spawn("sh", ["node_modules/.bin/sqlite-indexer"], {
+    cwd: packageRoot,
+    stdio: "inherit",
+    env: {
```
holic, line 118: should prob pass in a `STORE_ADDRESS` to ensure we're only indexing the targeted world

packages/explorer/src/bin/explorer.ts
```diff
@@ -97,6 +101,30 @@ async function startExplorer() {
   }
 }
 
+async function startStoreIndexer() {
+  if (chainId !== anvil.id) {
+    console.log("Skipping SQLite indexer for non-anvil chain ID", chainId);
+    return;
+  }
```
holic, line 108: thinking it might be cool to accept an RPC URL rather than a chain ID, then we can pass that straight through to the indexer

if we need the chain ID, we can always ask the RPC for it

packages/explorer/src/bin/explorer.ts
```diff
@@ -143,6 +171,9 @@ process.on("SIGINT", () => {
   if (explorerProcess) {
     explorerProcess.kill();
   }
+  if (indexerProcess) {
+    indexerProcess.kill();
+  }
```
holic, line 176:
```suggestion
  explorerProcess?.kill();
  indexerProcess?.kill();
```

packages/explorer/src/bin/explorer.ts
holic: also been trying to use fs/promises where I can rather than regular fs because it lets you parallelize when you need it

packages/explorer/src/bin/explorer.ts
holic: (this prob would require deeper changes in explorer itself to also be able to talk to the provided rpc, so def out of scope of this PR)

karooolis: agreed, will do that separately once we start to focus on any chain support

packages/explorer/src/bin/explorer.ts
```diff
@@ -143,6 +171,9 @@ process.on("SIGINT", () => {
...
   process.exit();
```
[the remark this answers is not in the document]
karooolis: it does not, does indexer spawn subprocesses?

packages/explorer/src/bin/explorer.ts
karooolis: Good point, will keep that in mind. In this case I'm not sure if async version makes sense since we'd have to wait for the database file to be deleted anyway, or am I missing something? So changed to `fs.rmSync`.

packages/explorer/src/bin/explorer.ts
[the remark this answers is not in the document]
holic: By subprocesses I meant the things you're `spawn`ing here. I expect since this explorer bin process is the one spawning, it has a parent-child relationship, and would kill any of its children if the parent process got killed. But maybe that assumption is why a lot of my tools end up with hanging/zombie processes 🙈 

Down to use something like terminate, but usage seems low, so I wonder if there's a more mainstream/battle-tested variant out there.

packages/explorer/src/bin/explorer.ts
karooolis: oh that's neat, added `STORE_ADDRESS` env var as world address

packages/explorer/src/bin/explorer.ts
holic: you can still `await`, I just find it useful to have the promise variants handy when you do want to parallelize certain operations

packages/explorer/src/bin/explorer.ts
holic: since we're parsing the worlds.json, we can also pass in `START_BLOCK` to speed things up (esp helpful when we start supporting remote chains)

packages/explorer/src/bin/explorer.ts
karooolis: interesting, didn't think about that but I see your point, updated

packages/explorer/src/bin/explorer.ts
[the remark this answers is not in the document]
karooolis: Heh, yeah `terminate` package does not seem widely used, and might be an overkill. Just looked into it a bit more and based on my understanding, keeping track of child processes, and then killing them once the parent process is killed is the way to do it. A common danger is spawning child processes but then not killing them explicitly once parent process is killed but we handle that with `process.on("SIGINT")` listener so should be good there. Then the child processes, once stopped, should themselves stop their own child processes but looks like we have no problems there since both `explorer` and `store-indexer` don't seem to leave anything hanging.

packages/explorer/src/bin/explorer.ts
karooolis: could you clarify what you mean, or point to how to obtain the value for `START_BLOCK`, or should it be passed explicitly? Looked into `worlds.json` but it only contains world address.

packages/explorer/src/bin/explorer.ts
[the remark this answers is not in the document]
holic: what about other signals like sigkill?

packages/explorer/src/bin/explorer.ts
holic: oh right, I forgot we don't write the start block for local chains because it creates a lot of noise with git seeing the start block changing after each new deploy (whereas the world address is more stable because its deterministically deployed)

karooolis: cool, will do!

packages/explorer/src/bin/explorer.ts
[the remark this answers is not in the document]
karooolis: Damn, thought I had everything. So turns out `SIGKILL` cannot be captured but `SIGTERM` can. So now capturing `SIGINT` and `SIGTERM` which captures it all to the best of my knowledge. Will do a bit more research into this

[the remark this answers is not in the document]
holic: doesn't help with sigkill but looks like we can listen to both sigint and sigterm with
```ts
process.on('exit', () => {
  ...
})
```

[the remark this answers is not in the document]
karooolis: Just tried it and it's working. Checked the docs and it says `The 'exit' event is emitted when the Node.js process is about to exit as a result of either:` so it does make sense to clean up the child processes here based on the documentation, updated.

.changeset/silent-hairs-heal.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/explorer": patch
+---
+
+The local store indexer has been integrated into the Explorer package and now runs directly as part of it, without any external setup.
```
holic, line 5:
````suggestion
---
"@latticexyz/explorer": patch
"create-mud": patch
---

Explorer now automatically starts a local indexer when using Anvil as the target chain.

If you previously had an `indexer` entry in your `mprocs.yaml` file, it can now be removed.

```diff
-  indexer:
-    cwd: packages/contracts
-    shell: shx rm -rf $SQLITE_FILENAME && pnpm sqlite-indexer
-    env:
-      DEBUG: mud:*
-      RPC_HTTP_URL: "http://127.0.0.1:8545"
-      FOLLOW_BLOCK_TAG: "latest"
-      SQLITE_FILENAME: "indexer.db"
```

````

holic (dismissed):

PR #3244, opened 2024-10-01
chore: enable explorer vercel deployment

alvrs:


package.json
```diff
@@ -57,7 +58,7 @@
   },
   "packageManager": "pnpm@9.6.0",
   "engines": {
-    "node": "^18.20.1",
+    "node": ">=18.0.0 <21.0.0",
```
holic, line 61: fwiw I see this most often formatted like `18.x || 20.x`

karooolis: that would not include v19, is that the intention?

alvrs: i think the semantic difference here is that `18.x || 20.x` ignores `19.x`, do we need to ignore it?

alvrs: we can test if it works with `19.x` by pushing a commit with `19.x` and checking that CI is still happy, if it works i'm in favor of including it

holic: mmm good point, not sure why I've seen it this way - maybe because the even release are considered LTS

package.json
```diff
@@ -20,6 +20,7 @@
     "gas-report": "pnpm run --recursive --parallel gas-report",
     "lint": "pnpm prettier:check && eslint . --ext .ts --ext .tsx",
     "prepare": "husky install && (forge --version || pnpm foundryup)",
+    "prepare-vercel": "(forge --version || pnpm foundryup) && ln -sf /vercel/.foundry/bin/* node_modules/.bin/ && forge --version",
```
holic, line 23: small preference for this to be
```suggestion
    "vercel:prepare": "(forge --version || pnpm foundryup) && ln -sf /vercel/.foundry/bin/* node_modules/.bin/ && forge --version",
```

package.json
karooolis: yeah, looking more into it, looks like that's the reason. I'd be more inclined to use what seems more conventional which is `18.x || 20.x`. It still doesn't stop `19.x` from being used, just with a warning.

alvrs: > even-numbered releases are maintained for longer and are considered more stable, making them better for production environments

wow TIL

package.json
karooolis: oh true, aligns with the rest of the script names, missed it, updated.

alvrs:
✅

karooolis (approved):

package.json
holic: ahh I was hoping for `vercel` as the prefix (as that's the "context" in which its run) rather than `prepare` which kinda feels like part of the other prepare steps (like we do for `build` and `clean` and `test`)

karooolis: ah damn, should have just committed the suggestion and missed it, will change in another PR, my bad

PR #3249, opened 2024-10-03
feat(store): add unwrap() function to ResourceIdInstance

### LINKED ISSUE #3248: Add `unwrap` function to `ResourceId`
When working with resource ids within systems, sometimes users might need to unwrap them into bytes32 (e.g. when storing them as part of a table).

Currently, this is possible by doing `ResourceId.unwrap(id)`, but a simple addition of `unwrap(ResourceId)` to the `ResourceIdInstance` library would allow users to do `id.unwrap()`. Alternatively it could be more explicit, something like `toBytes32`.

vdrg:
Resolves #3248

### DIFF
--- a/.changeset/sour-bats-compare.md
+++ b/.changeset/sour-bats-compare.md
@@ -1,0 +1,5 @@
+---
+"@latticexyz/store": patch
+---
+
+Added an `unwrap` function to the ResourceIdInstance library, so now it is possible to directly unwrap a ResourceId by doing `id.unwrap()`.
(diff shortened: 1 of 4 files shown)


.changeset/sour-bats-compare.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/store": minor
```
holic, line 2:
```suggestion
"@latticexyz/store": patch
```

We tend to not follow semver _exactly_ and sneak new features into patch changes rather than minor and save minor bumps for big changes.

holic (approved):

alvrs (approved):
🎉🎉🎉

PR #3251, opened 2024-10-04
feat(explorer): integrate rejected transactions

karooolis:
CleanShot 2024-10-04 at 13 32 44@2x

### DIFF
--- a/packages/explorer/src/observer/decorator.ts
+++ b/packages/explorer/src/observer/decorator.ts
@@ -37,7 +37,7 @@ export function observer({ explorerUrl = "http://localhost:13690", waitForTransa
       emit("write", {
         writeId,
         address: args.address,
-        from: client.account.address,
+        from: client.account!.address,
         functionSignature: formatAbiItem(functionAbiItem),
         args: (args.args ?? []) as never,
         value: args.value,
--- a/packages/explorer/src/observer/decorator.ts
+++ b/packages/explorer/src/observer/decorator.ts
@@ -37,9 +37,10 @@ export function observer({ explorerUrl = "http://localhost:13690", waitForTransa
       emit("write", {
         writeId,
         address: args.address,
-        from: client.account?.address,
+        from: client.account.address,
         functionSignature: formatAbiItem(functionAbiItem),
         args: (args.args ?? []) as never,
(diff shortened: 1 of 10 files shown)


packages/explorer/src/observer/decorator.ts
```diff
@@ -37,6 +37,7 @@ export function observer({ explorerUrl = "http://localhost:13690", waitForTransa
       emit("write", {
         writeId,
         address: args.address,
+        from: client.account?.address,
```
holic, line 40: `writeContract` throws if no account exists
https://github.com/wevm/viem/blob/44ae1a5549e704486f2643c67f222ab3ce87cd47/src/actions/wallet/writeContract.ts#L184-L187

so I think we can just force this to be defined and not use an optional type here
```suggestion
        from: client.account!.address,
```

karooolis: cool was thinking to do the same, will force it

packages/explorer/src/observer/store.ts
```diff
@@ -10,11 +10,13 @@ export type Write = {
...
   type: MessageType;
   hash?: Hex;
+  from?: Address;
   address: Address;
   functionSignature: string;
   args: unknown[];
   time: number;
   events: Message<Exclude<MessageType, "ping">>[];
+  error?: Error;
 };
```
holic, line 20: just FYI that we should be aware of changes here because it's not guaranteed that the client sending data to the explorer are version-aligned, so although explorer makes some assumptions about new data here, an older client may not be passing it over yet

this currently would only show up if you configured your observer to point to the hosted explorer rather than a local one

holic: that said, doing this as an optional type is a good way to safe-guard against this!

karooolis: yeah great point, will keep this in mind going forward. For now all the new fields are optional so we are safe there

holic: actually my mistake, this is the internal store, which is fine to extend!

the messages definition + decorator is the thing to be mindful of

holic (approved):

PR #3263, opened 2024-10-07
fix(explorer): expand selected transaction table row by hash/writeId

karooolis:


packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/observe/TransactionsTable.tsx
```diff
@@ -97,6 +97,7 @@ export function TransactionsTable() {
     state: {
       expanded,
     },
+    getRowId: (row) => row.hash || row.writeId,
```
Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/observe/TransactionsTable.tsx:100]: might consider `writeId` first since its more stable so that the ID doesn't change once the hash appears?

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/observe/useTransactionWatcher.ts
```diff
@@ -34,7 +32,7 @@ export type WatchedTransaction = {
   write?: Write;
   logs?: Log[];
   error?: BaseError;
-};
+} & ({ hash: Hex; writeId?: string } | { hash?: Hex; writeId: string } | { hash: Hex; writeId: string });
```
Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/observe/useTransactionWatcher.ts:35]: prob don't need that last union?
```suggestion
} & ({ hash: Hex; writeId?: string } | { hash?: Hex; writeId: string });
```

Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/observe/useTransactionWatcher.ts]: Yeah, I'd say it's not strictly accurate anymore though since it can be that both, `hash` and `writeId` are non-null.

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/observe/TransactionsTable.tsx
Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/observe/TransactionsTable.tsx]: I'd lean towards using `writeId || hash` just because it already creates a stable identity for the row. Also, changed the order but somehow can't quite get typescript to infer that either of these types will be non-null, strange that it works in the opposite order. I wonder given the type coercion `row.writeId || row.hash!` if the discriminated union is still necessary, and could just do `hash?: hex; writeId?: string` for type.

Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/observe/TransactionsTable.tsx]: Changed to using `writeId` as the main key. `writeId` is set to `hash` when tx comes from blocks watcher.

holic (dismissed):

holic (approved):

PR #3265, opened 2024-10-07
fix(world): filter out invalid functions from abi

karooolis:


packages/world/ts/getWorldAbi.ts
```diff
@@ -26,8 +26,16 @@ export async function getWorldAbi({
...
-    .filter((abiItem) => !baseFunctionSelectors.includes(toFunctionSelector(abiItem)));
+    .map((func) => {
+      try {
+        return functionSignatureToAbiItem(func.signature);
+      } catch (error) {
+        console.error(error);
+      }
+    })
+    .filter(
+      (abiItem): abiItem is AbiFunction => !!abiItem && !baseFunctionSelectors.includes(toFunctionSelector(abiItem)),
```
alvrs, line 37: tiny nit but for consistency across the codebase, could we use `isDefined` here? (from `@latticexyz/common/utils`)

holic, line 37: `.filter(isDefined)` will narrow so you don't have to manually define this here

PR #3267, opened 2024-10-08
feat(explorer): improve nav click target

holic:
- click target for nav items is bigger and now includes the underline/bottom border
- hovering now shows bottom border

### DIFF
--- a/packages/explorer/src/components/Navigation.tsx
+++ b/packages/explorer/src/components/Navigation.tsx
@@ -14,13 +14,15 @@ function NavigationLink({ href, children }: { href: string; children: React.Reac
   const pathname = usePathname();
   const getLinkUrl = useWorldUrl();
   return (
-    <Link
-      href={getLinkUrl(href)}
-      className={cn("text-sm uppercase underline-offset-[16px]", {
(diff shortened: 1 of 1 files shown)


packages/explorer/src/components/Navigation.tsx
```diff
@@ -14,13 +14,15 @@ function NavigationLink({ href, children }: { href: string; children: React.Reac
...
   const getLinkUrl = useWorldUrl();
   return (
-    <Link
-      href={getLinkUrl(href)}
-      className={cn("text-sm uppercase underline-offset-[16px]", {
-        "font-semibold underline decoration-orange-500 decoration-4": pathname === getLinkUrl(href),
-      })}
-    >
-      {children}
+    <Link href={getLinkUrl(href)} className="group px-3 text-sm font-semibold uppercase">
```
holic, line 17: fyi the nav is now semibold for all links rather than the current/active link, which I think helps the nav stand out relative to other text on the page and the active state is already communicated by the bottom border

karooolis: just checked it out, looks very nice! also should have added the properly sized clickable area right from the start, thanks for doing this

karooolis (approved):

PR #3268, opened 2024-10-08
feat(explorer): add functions filter to query state

karooolis:


packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/interact/Form.tsx
```diff
@@ -14,7 +15,7 @@ import { FunctionField } from "./FunctionField";
 export function Form() {
   const [hash] = useHashState();
   const { data, isFetched } = useWorldAbiQuery();
-  const [filterValue, setFilterValue] = useState("");
+  const [filterValue, setFilterValue] = useQueryState("functions-filter", { defaultValue: "" });
```
Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/interact/Form.tsx:18]: what about just `functions`?

Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/interact/Form.tsx]: True, a bit too verbose. Was also thinking that these should be named rather well or else if we have to update them, it would be a pain to manage outdated URLs.

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/interact/Form.tsx
```diff
@@ -14,7 +15,7 @@ import { FunctionField } from "./FunctionField";
 export function Form() {
   const [hash] = useHashState();
   const { data, isFetched } = useWorldAbiQuery();
-  const [filterValue, setFilterValue] = useState("");
+  const [filterValue, setFilterValue] = useQueryState("functions", { defaultValue: "" });
```
Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/interact/Form.tsx:18]: thinking now it should be just "function" because we're often looking for a specific function
```suggestion
  const [filterValue, setFilterValue] = useQueryState("function", { defaultValue: "" });
```

Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/interact/Form.tsx:18]: ah true, updated

holic (approved):

PR #3274, opened 2024-10-09
feat(explorer): transaction timings

karooolis:
CleanShot 2024-10-11 at 13 33 54


.changeset/eighty-humans-divide.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/explorer": patch
+---
+
+Transactions in Observe tab are now populated with timing metrics when using the observer transport wrapper.
```
holic, line 5:
````suggestion
Transactions in Observe tab are now populated with timing metrics when using the `observer` Viem decorator in local projects.

You can wire up your local project to use transaction timings with:
```
import { observer } from "@latticexyz/explorer/observer";

// Extend the Viem client that is performing writes
walletClient.extend(observer());
```
````

File: packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/observe/TransactionsTable.tsx [hunk 2]
```diff
@@ -85,6 +86,14 @@ export const columns = [
       return <TimeAgo timestamp={timestamp} />;
     },
   }),
+  columnHelper.accessor("write", {
+    header: "Timing",
+    cell: (row) => {
+      const write = row.getValue();
+      if (!write) return <span className="text-white/60">N/A</span>;
```
Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/observe/TransactionsTable.tsx:93]: would prefer to just leave this blank when not filled in to not populate the hosted explorer with a bunch of un-actionable "N/A"s

and ideally combine this with the time column

holic (dismissed):
just a few notes but otherwise lgtm!

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/observe/TransactionsTable.tsx
Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/observe/TransactionsTable.tsx]: yeah good idea, updated so now it's combined with the time column

holic (approved):

PR #3276, opened 2024-10-10
feat(explorer): sql editor

karooolis:


packages/explorer/src/app/(explorer)/globals.css
```diff
@@ -8,6 +8,15 @@ body,
   font-family: var(--font-jetbrains-mono);
 }
 
+.cm-editor {
```
holic, line 11: leftover from codemirror?

karooolis: yeah sorry :| removed

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/useQueryAutocomplete.ts
```diff
@@ -0,0 +1,21 @@
...
+import { SQLAutocomplete, SQLDialect } from "sql-autocomplete";
+import { Address } from "viem";
+import { useMemo } from "react";
+import { Table } from "@latticexyz/config";
+import { useChain } from "../../../../hooks/useChain";
+import { constructTableName } from "../../../../utils/constructTableName";
+
+export function useQueryAutocomplete(table?: Table) {
+  const { id: chainId } = useChain();
+  const { worldAddress } = useParams();
```
Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/useQueryAutocomplete.ts:11]: fyi you can cast params here with
```suggestion
  const { worldAddress } = useParams<{ worldAddress: string }>();
```

maybe even
```suggestion
  const { worldAddress } = useParams<{ worldAddress: Address }>();
```
but we should prob validate that this is an address in the layout like we do for chain names

.changeset/itchy-countries-worry.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/explorer": patch
+---
+
+Enhance Explorer's with SQL syntax highlighting, autocomplete, and query validation.
```
holic, line 5:
```suggestion
Explore page now has a full-featured SQL editor with syntax highlighting, autocomplete, and query validation.
```

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/SQLEditor.tsx
```diff
@@ -1,46 +1,83 @@
...
-import { Form, FormControl, FormField, FormItem } from "../../../../../../components/ui/Form";
-import { Input } from "../../../../../../components/ui/Input";
+import { Form, FormField } from "../../../../../../components/ui/Form";
+import { cn } from "../../../../../../utils";
+import { monacoOptions } from "./consts";
+import { useMonacoSuggestions } from "./useMonacoSuggestions";
+import { useQueryValidator } from "./useQueryValidator";
+
+type Props = {
+  table?: Table;
```
Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/SQLEditor.tsx:17]: does this just aid in autocomplete/validation?

holic (dismissed):

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/SQLEditor.tsx
Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/SQLEditor.tsx:17]: indeed, only need columns and table name from it atm. I suppose it would be somewhat cleaner to just pass these as props explicitly? I remember receiving a similar suggestion in the past.

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/useQueryAutocomplete.ts
Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/useQueryAutocomplete.ts:11]: Ah yeah, that's nice. I'm thinking for world address we should check in layout if it's a valid world. Then in the rest of the app we can safely assume that the world is valid. However, I haven't yet found a way that I like to keep params type-safe, will check it further.

PR #3277, opened 2024-10-12
docs: correct msg.sender in emojimon

three9s:
In Solidity, `msg.sender` is not a function, so the docs are incorrect and this is confusing. This corrects it so it makes sense.


holic (approved):
thank you!

PR #3278, opened 2024-10-12
docs(guides/emojimon): fix git commands

three9s:
If you don't include `origin`, these commands fail with:
```
git reset --hard step-3
fatal: ambiguous argument 'step-3': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]
```

Adding `origin` fixes this:
```
git reset --hard origin/step-3
HEAD is now at 24ccdd0 Step 3: A wild emojimon appears
```


holic (approved):

holic:
Thanks for the fix!

PR #3279, opened 2024-10-13
chore: update to turbo 2.1.3

alvrs:
* attempting to fix vercel deployment issues
* turbo 2.1.3 seems to be slower than our previous version? might be a configuration issue

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,4 +1,5 @@
 {
+  "name": "mud",
   "private": true,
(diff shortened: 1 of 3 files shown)


holic:
*startup is slow, might be related to https://github.com/vercel/turborepo/issues/8801

seems like newer turbo version are doing more work on startup

PR #3280, opened 2024-10-13
ci: explorer vercel deploy improvements

holic:


package.json
```diff
@@ -30,7 +30,7 @@
     "test:setup": "tsx test-setup/startAnvil.ts & pnpm run --filter mock-game-contracts build",
     "type-bench": "pnpm --filter ./test/ts-benchmarks bench",
     "type-stats-repo": "attest stats packages/*",
-    "vercel:prepare": "(forge --version || pnpm foundryup) && ln -sf /vercel/.foundry/bin/* node_modules/.bin/ && forge --version"
+    "vercel:prepare": "ln -sf /vercel/.foundry/bin/* node_modules/.bin/ && forge --version"
```
holic, line 33: noticed we were installing foundry twice because `prepare` is already run automatically as part of `install`, which does the foundry install

so we just need this step to relocate the foundry bins

PR #3282, opened 2024-10-13
docs: use root workspace

holic:
hoping this puts docs into the context of turbo and thus cached builds


holic:
doesn't seem to help much

PR #3285, opened 2024-10-14
feat(explorer): global transactions listener

karooolis:


packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/Providers.tsx
```diff
@@ -36,7 +37,9 @@ export function Providers({ children }: { children: ReactNode }) {
   return (
     <WagmiProvider config={wagmiConfig}>
       <QueryClientProvider client={queryClient}>
-        <RainbowKitProvider theme={darkTheme()}>{children}</RainbowKitProvider>
+        <RainbowKitProvider theme={darkTheme()}>
+          <WorldStoreProvider>{children}</WorldStoreProvider>
```
Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/Providers.tsx:41]: is this necessary?

Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/Providers.tsx:41]: Yeah, set it up using this guide https://zustand.docs.pmnd.rs/guides/nextjs it helps to prevent weird SSR bugs when the store is initiated twice, once on server-side, and next on client-side.

Comment (holic) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/Providers.tsx:41]: yeah since this tooling is all client-only, I have a preference for this being simpler and check if mounted before using to avoid SSR errors

if you really insist on this approach, it might be nice to explore generalizing this into a `ZustandStoreProvider` or something

Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/Providers.tsx]: Yeah I see your point, simplified it to a single `store.ts`.

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/layout.tsx
```diff
@@ -2,12 +2,15 @@
...
 import { Navigation } from "../../../../../components/Navigation";
 import { Providers } from "./Providers";
+import { TransactionsWatcher } from "./observe/TransactionsWatcher";
 
 export default function WorldLayout({ children }: { children: React.ReactNode }) {
   return (
     <Providers>
-      <Navigation />
-      {children}
+      <TransactionsWatcher>
```
Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/layout.tsx:10]: this doesn't seem to do anything as a component, just does hook-like things, so wondering if this can just be

```
useTransactionsWatcher()
```
or similar at the top of this component?

also wondering if we need to include this in some higher-up component to capture messages across all pages, not just a given world?

Comment (holic) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/layout.tsx:10]: ohh I thought this was the observer listener

re: component, could just have it return `null` then you can have it as a sibling instead of a wrapper here

Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/layout.tsx]: Good idea, doesn't have to be a parent indeed.

Also added `observerStore` "initializer" in `WorldsForm.tsx` with `useStore(observerStore)` in order to start observing for local events higher-up the tree as discussed. Not a huge fan of doing it this way but only importing it didn't work, and setting up some kind of `init` function just adds boilerplate. Curious what you think.

packages/explorer/src/app/(explorer)/[chainName]/worlds/WorldsForm.tsx
```diff
@@ -25,6 +27,9 @@ const formSchema = z.object({
 });
 
 export function WorldsForm({ worlds }: { worlds: Address[] }) {
+  // Initialize the observer store to start fetching transactions
+  useStore(store);
```
Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/WorldsForm.tsx:31]: does this cause a rerender of this component every time the store changes?

Comment (holic) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/WorldsForm.tsx:31]: I wonder if the zustand suggested approach of using a provider in next.js is for this reason 🙈 I assumed it was an SSR thing but maybe its a broad solution to this particular issue

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/observe/useTransactionsWatcher.ts
```diff
@@ -0,0 +1,77 @@
...
+  receipt?: TransactionReceipt;
+  status: "pending" | "success" | "reverted" | "rejected" | "unknown";
+  write?: Write;
+  logs?: Log[];
+  error?: BaseError;
+};
+
+export function useTransactionsWatcher() {
+  const { worldAddress } = useParams();
+  const { transactions } = useStore(worldStore);
```
Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/observe/useTransactionsWatcher.ts:36]: afaik this will rerender every time the store changes, not just when `transactions` change

zustand recommends this instead

```suggestion
  const transactions = useStore(worldStore, (state) => state.transactions);
```

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/observe/useTransactionsWatcher.ts
```diff
@@ -0,0 +1,77 @@
...
+  functionData?: DecodeFunctionDataReturnType;
+  value?: bigint;
+  receipt?: TransactionReceipt;
+  status: "pending" | "success" | "reverted" | "rejected" | "unknown";
+  write?: Write;
+  logs?: Log[];
+  error?: BaseError;
+};
+
+export function useTransactionsWatcher() {
```
Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/observe/useTransactionsWatcher.ts:34]: this isn't really watching transactions, but rather finding/merging them?

Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/observe/useTransactionsWatcher.ts:34]: yeah actually good point, it no longer does that, changed to `useObservedTransactions`.

packages/explorer/src/app/(explorer)/[chainName]/worlds/WorldsForm.tsx
Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/WorldsForm.tsx:31]: yeah, unfortunately causes re-renders, it's definitely a code smell to say the least. Using the provider does feel cleaner despite the all the boilerplate. I think the Provider solution is specifically for SSR because I'm not sure how common of a scenario it is to initialize a store in such way. Usually, you'd use one of store's getters/setters, and initialize the store implicitly when it's needed.

Given that the current problem essentially is that the broadcast channel listener does get started, I'm thinking what about moving the broadcast channel event listener into a hook, and initialize it somewhere higher-up the tree that way?

Comment (holic) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/WorldsForm.tsx:31]: do you know if the provider would solve for this?

if so I feel like maybe the provider approach is ~fine and solves for multiple problems (SSR, global mounting/initializing the store and observer)

but curious if we can generalize the store provider boilerplate to avoid repeating the pattern for every store

Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/WorldsForm.tsx:31]: just had a thought: is next.js just looking for a reference to the import in order to consider it used/in scope of the bundle? if so, I wonder if this would solve our problems

```suggestion
  store;
```

Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/WorldsForm.tsx]: haha didn't think about that, def on the weirder side of code that I've seen but yeah, it does work. So perhaps let's stick with that

Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/WorldsForm.tsx]: I'd need to set it up to double-check but I'd think so. Given though that the `store` solution works, even if somewhat hacky, I'm included to stick with that

packages/explorer/src/app/(explorer)/[chainName]/worlds/WorldsForm.tsx
```diff
@@ -25,6 +26,9 @@ const formSchema = z.object({
 });
 
 export function WorldsForm({ worlds }: { worlds: Address[] }) {
+  // Initialize the observer store to start fetching transactions
+  store;
```
Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/WorldsForm.tsx:30]:
```suggestion
  // Reference to imported store observer so the listener is established when this component is used.
  store;
```
also wonder if we could do the same thing outside this component or if it has to be referenced inside to be considered in scope of the page bundle

Comment (holic) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/WorldsForm.tsx:30]: also wondering about doing this inside the top-level `layout.tsx` instead

holic (approved):

PR #3287, opened 2024-10-14
feat(store-sync): add support for watching pending logs

alvrs:


packages/store-sync/src/common.ts
```diff
@@ -115,6 +115,10 @@ export type SyncOptions = {
     blockNumber: bigint;
     logs: readonly StorageAdapterLog[];
   };
+  /**
+   * Optional webSocket URL with `wiresaw_getLogs` and `wiresaw_watchLogs` RPC support for syncing with pending logs
+   */
+  pendingLogsWebSocketRpcUrl?: string;
```
alvrs, line 121: Went back and forth on this. Originally i was just grabbing the websocket url from the public client's `chain` config. Then E2E tests were failing because anvil was configured by default _with_ a websocket URL, but it wasn't the websocket URL used in tests. That made me realise that defaulting to the chain config is probably not a good idea because often a custom RPC URL is passed to the client transport. Unfortunately there isn't a good way to grab the WebSocket URL or underlying socket from a client's transport (especially with things like `fallback` transport), so I think making the pending logs websocket URL explicit is the best option here.

Alternatively we could look for it on the chain config under `rpcUrls.wiresaw` or similar, but i'm not sure if that's better than making it an explicit option on the sync stack.

alvrs: could also rename this property to `experimentalPendingLogsWebSocketRpcUrl` to make it obvious that this is experimental and the prop might be moved in the future.

.changeset/mean-squids-buy.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/store-sync": patch
+---
+
+Add experimental support for syncing state from pending logs to the sync stack.
```
holic, line 5:
```suggestion
Added experimental support for syncing state from pending logs.
```

holic (dismissed):

holic (changes requested):
I think we need to move around the `onProgress` handler so that we can communicate to the client when we're hydrated. But I don't think the current `watchLogs` approach communicates that super well, so unclear how to add this behavior.

packages/store-sync/src/createStoreSync.ts
```diff
@@ -184,24 +186,46 @@ export async function createStoreSync({
...
 
+  let startBlock: bigint | null = null;
   const startBlock$ = initialBlockLogs$.pipe(
     map((block) => bigIntMax(block?.blockNumber ?? 0n, initialStartBlock)),
     // TODO: if start block is still 0, find via deploy event
-    tap((startBlock) => debug("starting sync from block", startBlock)),
+    tap((blockNumber) => {
+      startBlock = blockNumber;
+      debug("starting sync from block", startBlock);
+      return startBlock;
```
holic, line 196: tap is a side effect so no need to return anything
```suggestion
```

packages/store-sync/src/createStoreSync.ts
```diff
@@ -184,24 +186,46 @@ export async function createStoreSync({
...
   );
 
+  let latestBlockNumber: bigint | null = null;
   const latestBlock$ = createBlockStream({ publicClient, blockTag: followBlockTag }).pipe(shareReplay(1));
   const latestBlockNumber$ = latestBlock$.pipe(
     map((block) => block.number),
     tap((blockNumber) => {
+      latestBlockNumber = blockNumber;
       debug("on block number", blockNumber, "for", followBlockTag, "block tag");
+      return latestBlockNumber;
```
holic, line 207:
```suggestion
```

packages/store-sync/src/createStoreSync.ts
```diff
@@ -254,32 +274,38 @@ export async function createStoreSync({
...
 
-  const storedBlock$ = storedIndexerLogs$.pipe(
+  const storedBlock$ = storedPendingLogs$.pipe(
+    catchError((e) => {
+      debug("failed to stream logs from pending log RPC:", e.message);
+      debug("falling back to streaming logs from indexer");
+      return storedIndexerLogs$;
+    }),
     catchError((e) => {
       debug("failed to stream logs from indexer:", e.message);
```
holic, line 284:
```suggestion
    catchError((error) => {
      debug("failed to stream logs from pending log RPC:", error.message);
      debug("falling back to streaming logs from indexer");
      return storedIndexerLogs$;
    }),
    catchError((error) => {
      debug("failed to stream logs from indexer:", error.message);
```

packages/store-sync/src/wiresaw.ts
```diff
@@ -0,0 +1,126 @@
...
+          },
+        })
+      ).result;
+
+      // Listen for wiresaw_watchLogs subscription
+      // Need to use low level methods since viem's socekt client only handles `eth_subscription` messages.
+      // (https://github.com/wevm/viem/blob/f81d497f2afc11b9b81a79057d1f797694b69793/src/utils/rpc/socket.ts#L178)
+      client.socket.addEventListener("message", (message) => {
+        const response = JSON.parse(message.data);
+        if ("error" in response) {
```
holic, line 50:
```suggestion
        if ("error" in response) {
          // if error isn't associated with a request ID, ignore
          // TODO: keep track of our request IDs and ignore others?
          if ("id" in response && response.id != null) return;
```

holic: actually could maybe lift this early return outside the error handler?

holic: I discovered that something sending other random RPC calls (`net_version` specifically, so I think it was Metamask) to the websocket will cause this to close

alvrs: it should currently open a separate websocket connection for this log stream, did you experience any other errors come through here?

holic: yes, I saw `net_version` occasionally sent through this (and not sure where from or how it got a reference to this open socket) and that would fail on our node, causing the stream to close, until it was recently added to the whitelist

holic: Turns out `net_version` request comes from viem's `ping` method used by `keepAlive` option. The right move was to enable this at our node rather than fixing anything here.

And it would be good if viem keeps `net_version` as the ping method (rather than e.g. moving to `eth_chainId`) because we cache `eth_chainId` requests to avoid it mucking with the transaction lifecycle and to reduce end-to-end latency.

holic (approved):

PR #3288, opened 2024-10-14
fix(common): allow overriding fees in writeContract and sendTransaction

alvrs:


holic (approved):

alvrs:
Was wondering why we even did it in this order in the first place and i think it's an artifact from https://github.com/latticexyz/mud/commit/6c8ab471adfb522e841b679072c7ff53fe105034 - where we prepared the request async from sending it and then sent it with the latest fee values. In other words, safe to change the order to allow users override the values i think.

PR #3290, opened 2024-10-14
feat(config): import mud config with tsx

holic:
hoping this solves the race condition in failing tests in main when running tests in parallel, where each test imports a MUD config and thus tries to compile the file to the same path

I tried ts-import and esbuild-register before landing on a working solution with tsx


holic:
blocked-ish on the same world TS issues we keep running into

might be time to transition abi-ts to generate a `.ts` file instead of `.d.ts`

holic:
> blocked-ish on the same world TS issues we keep running into
> 
> might be time to transition abi-ts to generate a `.ts` file instead of `.d.ts`

worked around this for now with https://github.com/latticexyz/mud/pull/3429

PR #3293, opened 2024-10-16
build: improve tsup config for turbo

holic:
This fixes the weirdness around explorer building.

We have turbo configured to run `build` task for each dep as a prerequisite of `dev` task. Before this change, tsup was configured to only run DTS in a special dev mode using `TSUP_SKIP_DTS` (because DTS generation is slow). This meant that explorer couldn't find the DTS files it expects when building and the build would fail.

Instead, we now always build DTS when running a regular tsup build (not watch mode). When tsup is in watch mode, it skips DTS and _also_ skips cleaning, to avoid deleting previously generated DTS files. We need this configured this way otherwise a fast task like building common package would run build + dev, where dev would clean DTS files, then explorer wouldn't be able to find them when it goes to build itself.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,4 +1,5 @@
 {
+  "name": "mud-monorepo",
   "private": true,
   "repository": {
     "type": "git",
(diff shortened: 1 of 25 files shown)


package.json
```diff
@@ -12,7 +13,7 @@
     "build": "turbo run build --concurrency=100%",
     "changelog:generate": "tsx scripts/changelog.ts",
     "clean": "turbo run clean --concurrency=100%",
-    "dev": "TSUP_SKIP_DTS=true turbo run dev --concurrency=100% --filter !@latticexyz/explorer",
+    "dev": "turbo run dev --concurrency=999",
```
holic, line 16: turbo is dumb

holic, line 16: `dev` task uses `persistent: true` to mark it as a long running task (which helps warn when other tasks depend on a thing that will never complete)

thus, we need to manually specify concurrency > number of persistent tasks

PR #3299, opened 2024-10-16
fix(explorer): various fixes

karooolis:
- [x] Not found page if invalid chain name
- [x] Only show selector for worlds if options exist
- [x] Remove "future time" from transactions table
- [x] Improved layout for `Interact` tab
- [x] Wrap long inputs
- [x] Add back logs (regression) 
- [x] Tables query refetching

### DIFF
--- a/packages/explorer/src/app/(explorer)/[chainName]/page.tsx
+++ b/packages/explorer/src/app/(explorer)/[chainName]/page.tsx
@@ -6,6 +6,6 @@ type Props = {
   };
 };
 
-export default async function ChainPage({ params }: Props) {
-  return redirect(`/${params.chainName}/worlds`);
+export default async function ChainPage({ params: { chainName } }: Props) {
+  return redirect(`/${chainName}/worlds`);
 }
--- a/.changeset/soft-bears-rest.md
+++ b/.changeset/soft-bears-rest.md
@@ -7,4 +7,5 @@
 - Remove "future time" from transactions table.
 - Improved layout for Interact tab.
 - Wrap long args in transactions table.
+- New tables polling.
 - Add logs (regression).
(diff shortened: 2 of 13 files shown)


packages/explorer/src/app/(explorer)/[chainName]/page.tsx
```diff
@@ -6,6 +6,6 @@ type Props = {
   };
 };
 
-export default async function ChainPage({ params }: Props) {
-  return redirect(`/${params.chainName}/worlds`);
+export default async function ChainPage({ params: { chainName } }: Props) {
+  return redirect(`/${chainName}/worlds`);
```
Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/page.tsx:10]: thoughts on validating the chain name here and only redirect when valid?

Comment (holic) [reply packages/explorer/src/app/(explorer)/[chainName]/page.tsx:10]: since chain names are known statically, I wonder if its helpful to use `getStaticPaths` on these routes?

Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/page.tsx:10]: `getStaticPaths` only works with `pages` router while we are using `app` router which is the recommended way now and moving forward.

There is something similar in the `app` router called `generateStaticParams` which I think we can use the same way. However, I don't think we should make this page static given that the world addresses which we display in world entry page are dynamic. 

Now `getServerSideProps` inside `app` router is essentially replaced with `async` server-side component with a fetcher function inside. So based on that, the validation also is moved to the component itself. What I think would be best to do here is to actually move the validation inside `layout.tsx`. I already did that for `(explorer)/[chainName]/worlds/[worldAddress]/layout.tsx` but actually thinking that same could be done for `(explorer)/[chainName]/layout.tsx`(which doesn't exist yet but could be created for the purpose of validation). This way there's less repeating logic at least.

holic (approved):

PR #3300, opened 2024-10-16
feat(world-module-erc20): new ERC20 World Module

vdrg:


packages/world-module-erc20/src/Context.sol
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.20;
```
alvrs, line 3:
```suggestion
pragma solidity >=0.8.24;
```
to keep the solidity version aligned across the codebase

packages/world-module-erc20/src/ERC20Module.sol
```diff
@@ -0,0 +1,74 @@
...
+}
+
+function delegateRegister(ERC20RegistryLib lib, IBaseWorld world, ResourceId namespaceId, address token) {
+  (bool success, bytes memory returnData) = address(lib).delegatecall(
+    abi.encodeCall(ERC20RegistryLib.register, (world, namespaceId, token))
+  );
+  if (!success) revertWithBytes(returnData);
+}
+
+using { delegateRegister } for ERC20RegistryLib;
```
alvrs, line 74: Double checking, this is an attempt to work around the library linking issue right? If public library linking was working for systems we could just do `ERC20RegistryLib.register`?

packages/world-module-erc20/src/ERC20Pausable.sol
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: MIT
+// Adapted from OpenZeppelin's ERC20Pausable extension
```
alvrs, line 2: Should we add a gitub permalink to the OZ files for all of these? Might make it easier in the future to compare, especially if OZ releases new contract versions

packages/world-module-erc20/src/Pausable.sol
```diff
@@ -0,0 +1,126 @@
+// SPDX-License-Identifier: MIT
+// Adapted from OpenZeppelin Contracts (utils/Pausable.sol)
+
+pragma solidity ^0.8.20;
+
+import { ResourceId, ResourceIdLib } from "@latticexyz/store/src/ResourceId.sol";
+
+import { Paused as PausedTbl } from "./codegen/tables/Paused.sol";
```
alvrs, line 8: nit: generally trying to avoid abbreviations whenever possible
```suggestion
import { Paused as PausedTable } from "./codegen/tables/Paused.sol";
```

packages/world-module-erc20/src/Pausable.sol
```diff
@@ -0,0 +1,126 @@
+// SPDX-License-Identifier: MIT
+// Adapted from OpenZeppelin Contracts (utils/Pausable.sol)
+
+pragma solidity ^0.8.20;
```
alvrs, line 4:
```suggestion
pragma solidity >=0.8.24;
```

packages/world-module-erc20/src/StoreConsumer.sol
```diff
@@ -0,0 +1,80 @@
...
+  function _encodeResourceId(bytes2 typeId, bytes16 name) internal view virtual override returns (ResourceId) {
+    return ResourceIdLib.encode(typeId, name);
+  }
+}
+
+abstract contract WithWorld is WithStore {
+  bytes14 public immutable NAMESPACE;
+
+  error NamespaceAlreadyExists(bytes14 namespace);
+  error CallerHasNoNamespaceAccess();
```
alvrs, line 51: A pattern we use in World and Store is to prefix the error with the contract/library name, ie `WithWorld_NamespaceAlreadyExists` - makes it a little easier to know where the error originated from in revert messages.

https://github.com/latticexyz/mud/blob/3a80bed31b97d439025f68b8e4ded27354e102f1/packages/world/src/IWorldErrors.sol#L50

packages/world-module-erc20/src/StoreConsumer.sol
```diff
@@ -0,0 +1,80 @@
...
+
+  modifier onlyNamespace() {
+    address sender = _msgSender();
+    if (!ResourceAccess.get(getNamespaceId(), sender)) {
+      revert CallerHasNoNamespaceAccess();
+    }
```
alvrs, line 57: could technically reuse the `AccessControl` library here: https://github.com/latticexyz/mud/blob/3a80bed31b97d439025f68b8e4ded27354e102f1/packages/world/src/AccessControl.sol#L51-L56

packages/world-module-erc20/src/StoreConsumer.sol
```diff
@@ -0,0 +1,80 @@
...
+  constructor(IBaseWorld world, bytes14 namespace) WithStore(address(world)) {
+    NAMESPACE = namespace;
+
+    ResourceId namespaceId = getNamespaceId();
+
+    if (ResourceIds.getExists(namespaceId)) {
+      revert NamespaceAlreadyExists(namespace);
+    }
+
+    world.registerNamespace(namespaceId);
```
alvrs, line 70: `registerNamespace` will already revert if the namespace exists!

https://github.com/latticexyz/mud/blob/3a80bed31b97d439025f68b8e4ded27354e102f1/packages/world/src/modules/init/implementations/WorldRegistrationSystem.sol#L49-L58

alvrs (commented):
overall very nice, very excited about this!

.changeset/giant-birds-argue.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/world-module-erc20": patch
+---
+
+feat(world-module-erc20): New ERC20 World Module
```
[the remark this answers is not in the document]
holic: for context, the content of these changesets become part of the package's changelog and also the global changelog: https://mud.dev/changelog

packages/world-module-erc20/CHANGELOG.md
holic, line 1: should delete this, changesets will create this automatically

packages/world-module-erc20/src/StoreConsumer.sol
holic: also we tend to lift this out into an interface and extend the interface when we need to ensure the errors propagate out to the ABI

for example, errors defined libraries where we make a library call in a contract won't get the errors from the library in the ABI

I _think_ this is fixed in newer solc versions but haven't checked in a while

packages/world-module-erc20/src/Constants.sol
```diff
@@ -0,0 +1,36 @@
...
+  bytes14 constant NAMESPACE = "erc20-module";
+  bytes16 constant REGISTRY_TABLE_NAME = "ERC20_REGISTRY";
+
+  function namespaceId() internal pure returns (ResourceId) {
+    return WorldResourceIdLib.encodeNamespace(NAMESPACE);
+  }
+
+  function registryTableId() internal pure returns (ResourceId) {
+    return WorldResourceIdLib.encode(RESOURCE_TABLE, NAMESPACE, REGISTRY_TABLE_NAME);
+  }
```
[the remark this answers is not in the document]
vdrg: so these are only used in the module installation, the reason I added them as functions using WorldResourceIdLib instead of manual encoding is so that if at any point the `encode` implementation changes then it would break those other constants, but I can move if that's ok.

packages/world-module-erc20/src/ERC20Module.sol
vdrg: yeah exactly

packages/world-module-erc20/src/ERC20Pausable.sol
vdrg: will do 🫡

packages/world-module-erc20/src/StoreConsumer.sol
vdrg: but this would make it harder to see where the error comes from 🤔 I'm fine with both, which ever is more standard across the codebase

packages/world-module-erc20/src/StoreConsumer.sol
```diff
@@ -0,0 +1,74 @@
...
+      StoreCore.registerInternalTables();
+    }
+  }
+
+  function _encodeResourceId(bytes2 typeId, bytes16 name) internal view virtual override returns (ResourceId) {
+    return ResourceIdLib.encode(typeId, name);
+  }
+}
+
+abstract contract WithWorld is WithStore {
```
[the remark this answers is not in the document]
vdrg: makes sense, not sure why I didn't, I guess it was just easier to import from the same place. Will change

packages/world-module-erc20/gas-report.json
```diff
@@ -0,0 +1,224 @@
+[
+  {
+    "file": "test/ERC20BaseTest.sol",
+    "test": "testApprove",
+    "name": "approve",
+    "gasUsed": 70832
```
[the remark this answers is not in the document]
vdrg: so we should be getting it for both, but just realized the problem is that the tests for both cases are in the same file for each extension, so I'm guessing they get replaced by the latest one with the same test name / report name. I can split them into files, I just liked how things looked without splitting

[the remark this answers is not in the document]
vdrg: actually will instead add a function to get a different report name depending on the store being used 👍

packages/world-module-erc20/src/StoreConsumer.sol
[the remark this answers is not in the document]
vdrg: done

packages/world-module-erc20/README.md
```diff
@@ -0,0 +1,129 @@
+# ERC20 World Module
+
```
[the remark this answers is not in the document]
vdrg: done

alvrs (approved):

packages/world-module-erc20/ts/exports/index.ts
```diff
@@ -0,0 +1,25 @@
+import { encodeAbiParameters, stringToHex } from "viem";
+
+export type DefineERC20ConfigInput = {
+  namespace: string;
+  name: string;
+  symbol: string;
+};
+
+export function defineERC20Config({ namespace, name, symbol }: DefineERC20ConfigInput) {
```
holic, line 9: how about `defineERC20Module` since it returns module input?

and can we specify the return type here so we know we're always getting config-compatible module input

packages/world-module-erc20/ts/exports/index.ts
holic, line 1: we should save this file for doing the exports, not where the logic lives

so I would move this file to `ts/defineERC20Module.ts` then this file becomes

```
export { defineERC20Module } from "../defineERC20Module";
```

PR #3301, opened 2024-10-16
fix(store-sync): reconnect unresponsive watchLogs socket

alvrs:


packages/store-sync/src/wiresaw.ts
```diff
@@ -16,70 +18,98 @@ type WatchLogsResult = {
...
-            const blockNumber = parsedLogs[0].blockNumber;
-            subscriber.next({ blockNumber, logs: parsedLogs });
-          } else {
-            logBuffer.push(...parsedLogs);
+    function setupClient(): void {
+      // Buffer the live logs received until the gap from `startBlock` to `currentBlock` is closed
+      let caughtUp = false;
+      const logBuffer: StoreEventsLog[] = [];
+
+      getWebSocketRpcClient(url).then(async (_client) => {
```
holic, line 35: looks like this is a fairly lightweight wrapper around `getSocketRpcClient` that implements the ping and such: https://github.com/wevm/viem/blob/main/src/utils/rpc/webSocket.ts

wonder if we should use this instead? and replace their ping implementation?

alvrs: We would also have to reimplement the remaining stuff in `getWebSocketRpcClient` no? We could also just pass `keepAlive: false` sice keep alive is handled outside, but still keep the reconnect and other logic in `getWebSocketRpcClient`

packages/store-sync/src/wiresaw.ts
```diff
@@ -16,70 +17,102 @@ type WatchLogsResult = {
...
+          subscriber.error("Could not fetch initial wiresaw logs");
         }
+
+        // Keep websocket alive and reconnect if it's not alive anymore
+        keepAliveInterval = setInterval(async () => {
+          const result = await Promise.race([
+            client.requestAsync({ body: { method: "net_version" } }),
+            new Promise<void>((resolve) => {
+              setTimeout(resolve, 2000);
+            }),
```
holic, line 96: could also do
```suggestion
            new Promise<never>((resolve, reject) => {
              setTimeout(reject, 2000);
            }),
```

packages/store-sync/src/wiresaw.ts
holic: was thinking we could let the keep alive stay as is and not reimplement that and instead do something different in the `ping` method

but I am seeing now that lower level stuff doesn't do the request/response lifecycle, just sends stuff to the socket and something else handles data coming back, so maybe not a great candidate for it

PR #3307, opened 2024-10-18
fix(explorer): format default sql query

karooolis:
CleanShot 2024-10-24 at 09 33 10


packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/Explorer.tsx
```diff
@@ -31,7 +32,10 @@ export function Explorer() {
       if (indexer.type === "sqlite") {
         setQuery(`SELECT * FROM "${tableName}"`);
       } else {
-        setQuery(`SELECT ${Object.keys(table.schema).join(", ")} FROM ${tableName}`);
+        const columns = Object.keys(table.schema).map((column) =>
+          postgresKeywords.includes(column.toLowerCase()) ? `"${column}"` : column,
+        );
```
Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/Explorer.tsx:37]: any reason we can't just double quote all column names instead of just reserved keywords?

PR #3308, opened 2024-10-18
fix(cli): support public library methods in modules

vdrg:


packages/cli/src/deploy/resolveConfig.ts
```diff
@@ -22,7 +24,20 @@ export async function resolveConfig({
...
+  const requirePath = await findUp("package.json", { cwd: process.cwd() });
+  if (!requirePath) throw new Error("Could not find package.json to import relative to.");
+  const require = createRequire(requirePath);
+
+  const moduleOutDirs = config.modules.flatMap((mod) => {
+    if (mod.artifactPath == undefined) {
+      return [];
+    }
+
+    const moduleOutDir = path.join(require.resolve(mod.artifactPath), "../../");
```
holic, line 36: this `../..` feels a little magical, can we describe it with a comment?

packages/cli/src/deploy/resolveConfig.ts
```diff
@@ -22,7 +24,20 @@ export async function resolveConfig({
   readonly systems: readonly System[];
   readonly libraries: readonly Library[];
 }> {
-  const libraries = findLibraries(forgeOutDir).map((library): Library => {
+  const requirePath = await findUp("package.json", { cwd: process.cwd() });
```
holic, line 27:
```suggestion
  const requirePath = await findUp("package.json");
```
I believe the `cwd` default is `process.cwd()`

packages/cli/src/utils/getContractArtifact.ts
```diff
@@ -11,18 +11,24 @@ export type GetContractArtifactResult = {
   deployedBytecodeSize: number;
 };
 
+function HexNotStrict(value: string): value is Hex {
```
holic, line 14:
```suggestion
function isBytecode(value: string): value is Hex {
```

holic: there's a better version of this in `world/ts/node/types.ts` if we wanna export that and reuse it (or lift it into common package0

vdrg: Right! Saw that but wanted to keep the changes as small as possible,  but happy to include that here if you think that's better

vdrg: for now will keep this so we can refactor and unify using the arktype one in a separate PR

packages/cli/src/deploy/resolveConfig.ts
```diff
@@ -22,18 +24,34 @@ export async function resolveConfig({
...
+  const requirePath = await findUp("package.json");
+  if (!requirePath) throw new Error("Could not find package.json to import relative to.");
+  const require = createRequire(requirePath);
+
+  const moduleOutDirs = config.modules.flatMap((mod) => {
+    if (mod.artifactPath == undefined) {
+      return [];
+    }
+
+    // Navigate up two dirs to get the contract output directory
```
holic, line 36: I didn't mean to explain what `../..` does but _why_ we're specifically navigating up two dirs

```suggestion
    // forge artifacts are organized by `{outDir}/{sourceFilename}/{contractName}.json`, so we navigate up two directories to get to `outDir`
```

.changeset/gentle-carrots-kneel.md
```diff
@@ -0,0 +1,7 @@
+---
+"@latticexyz/cli": patch
+"@latticexyz/world-module-erc20": patch
+"@latticexyz/world-modules": patch
+---
+
+World Modules (and contracts imported by them) can now use public library methods.
```
holic, line 7: I would probably separate these two changesets, with the CLI being something like 

> Added support for deploying public libraries used within modules.

vdrg: done!

packages/cli/src/deploy/resolveConfig.ts
```diff
@@ -22,18 +24,34 @@ export async function resolveConfig({
...
-      name: library.name,
-      abi: contractData.abi,
-      prepareDeploy: createPrepareDeploy(contractData.bytecode, contractData.placeholders),
-      deployedBytecodeSize: contractData.deployedBytecodeSize,
-    };
+  const requirePath = await findUp("package.json");
+  if (!requirePath) throw new Error("Could not find package.json to import relative to.");
+  const require = createRequire(requirePath);
+
+  const moduleOutDirs = config.modules.flatMap((mod) => {
```
holic, line 31: maybe wrap this in a set so if we install multiple ERC20s, we only need to load the libs once?

```suggestion
  const moduleOutDirs = new Set(config.modules.flatMap((mod) => {
```

or ideally, we add support for multiple out dirs in `findLibraries`

holic (approved):
nice! thanks for the fix!

packages/cli/src/deploy/resolveConfig.ts
vdrg: ah makes sense, I had actually thought of deduplicating but then forgot. Will try it withing findLibraries 👍

vdrg: https://github.com/latticexyz/mud/pull/3321/files

Let me know if this looks good to you 🙏

PR #3309, opened 2024-10-19
feat: Add StrictMode to react template

roninjin10:
Adding strict mode will create more robust applications including
- Less of a chance of a hiddent time bomb of an infinite loop that jus thappenes to not be getting triggered yet
- Less of a chance of having bugs related to non hermetic useEffects etc.


holic:
thank you!

- linter is complaining
- can you apply this change to the other templates using React and also the examples/e2e tests?

PR #3311, opened 2024-10-21
feat(explorer): multi-line sql editor

karooolis:
CleanShot 2024-10-23 at 10 24 39


packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/SQLEditor.tsx
```diff
@@ -39,44 +42,54 @@ export function SQLEditor({ table }: Props) {
...
+                  editor.onDidBlurEditorText(() => setIsFocused(false));
+                }}
+                loading={null}
+              />
+            </div>
           )}
         />
 
-        <Button className="absolute right-1 top-1 h-8 px-4" type="submit">
+        <Button className="absolute bottom-1 right-1 h-8 px-4" type="submit">
```
Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/SQLEditor.tsx:92]: no change needed here but just wanted to note that I've found using position relative/absolute to create stacking contexts ends up being more trouble than its worth, because you end up fighting z-index or weird layering of things like modals etc.

now that we have grids, I now tend to use an approach like
```
<div className="grid">
  <div className="col-start-1 row-start-1">
    ...
  </div>
  <div className="col-start-1 row-start-1">
    ...
  </div>
</div>
```

obviously you can get more nuanced with this where you define an actual grid with template rows/cols and allow the top layer to take up only as much room as needed, but I only reach for that when I need to be specific about placement/alignment/spacing

another thing to note about this strategy is that the top layer (if using a full-size layer) will take over as the click target, so also need to use something like `pointer-events-none` on the layer's container and `pointer-events-auto` in each child

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/useQueryValidator.ts
```diff
@@ -10,26 +10,49 @@ import { useMonacoErrorMarker } from "./useMonacoErrorMarker";
...
+      break;
+    }
+    currentPosition += lines[i].length + 1; // +1 for newline character
+  }
+
+  return {
+    startLineNumber,
+    endLineNumber: startLineNumber,
+    startColumn,
+    endColumn: startColumn + erroredQueryStr.length,
```
Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/useQueryValidator.ts:32]: noticed that below we also add 1 to `endColumn` value, do we need this here too?

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/useQueryValidator.ts
```diff
@@ -10,26 +10,49 @@ import { useMonacoErrorMarker } from "./useMonacoErrorMarker";
 
 const sqlParser = new Parser();
 
+function findErrorPosition(query: string, erroredQueryStr: string) {
```
Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/useQueryValidator.ts:13]:
```suggestion
function findErrorPosition(query: string, erroredQueryString: string) {
```

or

```suggestion
function findErrorPosition(query: string, target: string) {
```

holic (approved):
looks like there's a white left border/line where there's indentation? any way to darken that a tad?

otherwise lgtm!

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/useQueryValidator.ts
Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/useQueryValidator.ts]: it ends up with +1 being added a few lines above (`startLineNumber = i + 1`). Double-checked that columns are marked correctly just in case

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/useQueryValidator.ts
Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/useQueryValidator.ts]: good suggestion, both are better than what was before, chose `target`

PR #3313, opened 2024-10-21
feat(explorer): bundled transactions support

karooolis:
CleanShot 2024-10-28 at 13 01 39

CleanShot 2024-10-27 at 10 24 12


packages/explorer/src/observer/decorator.ts
```diff
@@ -22,49 +31,225 @@ export function observer({ explorerUrl = "http://localhost:13690", waitForTransa
...
+    //       data: decoded.args[2],
+    //     });
+
+    //     console.log("observerWrite decodedWorldCall", decodedWorldCall);
+    //   } catch (error) {
+    //     console.error("observerWrite error", error);
+    //   }
+    // }
+
+    if (client.type === "bundlerClient") {
```
holic, line 63: shouldn't need to and prob shouldn't conditionally extend the client based on the type

packages/explorer/src/observer/decorator.ts
```diff
@@ -22,49 +31,225 @@ export function observer({ explorerUrl = "http://localhost:13690", waitForTransa
 ) => Pick<WalletActions<chain, account>, "writeContract"> {
```
holic, line 31:
```suggestion
) => Pick<WalletActions<chain, account>, "writeContract"> & Pick<BundlerActions<chain, account>, "sendUserOperation"> {
```

packages/explorer/src/observer/decorator.ts
karooolis: `sendUserOperation` cannot be extended on a wallet client, hence why. But this might still be worked around differently

.changeset/funny-peas-care.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/explorer": patch
+---
+
+The Explorer now supports bundled transactions, monitoring them by either listening to chain blocks or using the `observer` transport wrapper. Each user operation within a bundled transaction is displayed as an individual transaction in the Observe tab.
```
holic, line 5:
```suggestion
Added support for ERC-4337 bundled transactions, monitoring them by either listening to chain blocks or using the `observer` transport wrapper. Each user operation within a bundled transaction is displayed as an individual transaction in the Observe tab.
```

packages/explorer/src/observer/decorator.ts
```diff
@@ -19,52 +21,100 @@ export function observer({ explorerUrl = "http://localhost:13690", waitForTransa
...
-      write.then((hash) => {
-        const receipt = getAction(client, waitForTransactionReceipt, "waitForTransactionReceipt")({ hash });
-        emit("waitForTransactionReceipt", { writeId, hash });
-        Promise.allSettled([receipt]).then(([result]) => {
-          emit("waitForTransactionReceipt:result", { ...result, writeId });
+  return (client) => {
+    return {
+      async sendUserOperation(args) {
+        const writeId = `${client.uid}-${++writeCounter}`;
+        const write = getAction(client, sendUserOperation, "sendUserOperation")(args);
```
holic, line 31: I was looking at this and finding the many instances of an early return with the same value (`return write`) a little funky. I do like early returns, but returning _with_ the same value each time felt weird. That made me wonder if `finally` could override the return value, and I think the answer is _yes_ (TIL!)

```ts
const write = getAction(...)(args);

try {
  if (someCondition) return;
  if (anotherCondition) return;
}
finally {
  return write;
}
```

not sure if I love this or hate this but it sure is interesting haha

an alternative is a self-invoking function
```ts
const write = getAction(...)(args);

(() => {
  if (someCondition) return;
  if (anotherCondition) return;
})()

return write;
```

anyway, no action needed, just was an interesting thought experiment that taught me something about JS

packages/explorer/src/observer/decorator.ts
```diff
@@ -19,52 +21,100 @@ export function observer({ explorerUrl = "http://localhost:13690", waitForTransa
...
-        emit("write:result", { ...result, writeId });
-      });
-
-      write.then((hash) => {
-        const receipt = getAction(client, waitForTransactionReceipt, "waitForTransactionReceipt")({ hash });
-        emit("waitForTransactionReceipt", { writeId, hash });
-        Promise.allSettled([receipt]).then(([result]) => {
-          emit("waitForTransactionReceipt:result", { ...result, writeId });
+  return (client) => {
+    return {
```
holic, line 28:
```suggestion
  return (client) => ({
```
small thing to help with the diff + reduce nesting

holic (approved):
curious if there's a better way to abstract over some of this complexity, but otherwise lgtm!

karooolis:
> curious if there's a better way to abstract over some of this complexity, but otherwise lgtm!

That's what I've been trying to figure out, would love to simplify it. Going to keep this at the top of my head since this is now probably the most complex part of the app.

packages/explorer/src/observer/decorator.ts
karooolis: Yeah, not very fond of the multiple early returns either. TIL about `finally` as well haha. Although in this case I prefer to return `write` multiple times just because it makes the intent clearer. I'd love to do it cleaner here but the `calls` type is unknown, so have to do all these checks to make sure it's actually the `userOp` we are expecting.

PR #3320, opened 2024-10-22
feat(stash): add useStash and improve other helpers

holic:


packages/stash/src/actions/getRecord.ts
```diff
@@ -1,16 +1,28 @@
...
-import { Key, Stash, TableRecord } from "../common";
+import { Key, Stash, State, TableRecord } from "../common";
 import { encodeKey } from "./encodeKey";
 
-export type GetRecordArgs<table extends Table = Table> = {
-  stash: Stash;
+export type GetRecordArgs<
+  table extends Table = Table,
+  defaultValue extends Omit<TableRecord<table>, keyof Key<table>> | undefined = undefined,
+> = propwiseXor<{ stash: Stash }, { state: State }> & {
```
alvrs, line 9: do you think it's still useful to have both variants (with `stash` or with `state`) or should we just change all actions to take in `state` for simplicity?

holic: I think it's still useful and not much work to maintain both in this case.

holic: Another thing on my mind is that we'll likely have a corresponding `useRecord` and `useRecords` hooks to mirror the state variant of `getRecord` and `getRecords` as shorthands for common usage in the React context:

```
useStash(stash, (state) => getRecord({ state, table, key })
useRecord({ stash, table, key })
```

The nice things about building with `useRecord` and `useRecords` hooks instead of `useStash` directly is that you avoid the foot gun that is infinite rendering due to unstable return values of a selector without also specifying an equality function - this can all be abstracted away.

PR #3326, opened 2024-10-24
fix(explorer): handle empty queries

karooolis:


packages/explorer/src/app/(explorer)/queries/useTableDataQuery.ts
```diff
@@ -51,16 +51,16 @@ export function useTableDataQuery({ table, query }: Props) {
...
         })
         .filter((key) => schemaKeys.includes(key));
-      const rows = result.slice(1).map((row) => Object.fromEntries(columnKeys.map((key, index) => [key, row[index]])));
+      const rows = result.slice(1)?.map((row) => Object.fromEntries(columnKeys.map((key, index) => [key, row[index]])));
 
       return {
-        columns: columnKeys,
-        rows,
+        columns: columnKeys || [],
+        rows: rows || [],
```
holic, line 63: any way we can improve the types of the above so that we have a better sense for when this happens?

holic (dismissed):

packages/explorer/src/app/(explorer)/queries/useTableDataQuery.ts
karooolis: nice, `noUncheckedIndexedAccess` is actually very helpful here. Added to the explorer, and fixed all TS errors that that bubbled up. Also adjusted `useTableDataQuery` to reflect better on what should happen with rows data given column keys.

PR #3336, opened 2024-10-26
fix(cli): throw error on schema changes

### LINKED ISSUE #3207: deployer should throw if table ID exists with diff schema
we currently skip existing tables

https://github.com/latticexyz/mud/blob/5a6c03c6bc02c980ca051dadd8e20560ac25c771/packages/cli/src/deploy/ensureTables.ts#L32

which is good if the table is the shape we expect, because they're considered immutable

but if you were to override a table schema in your config, this should throw an error rather than skipping, to help teach that table schemas are immutable

holic:
image

closes #3207

### DIFF
--- a/packages/cli/src/deploy/ensureTables.ts
+++ b/packages/cli/src/deploy/ensureTables.ts
@@ -8,12 +8,12 @@ import {
   getSchemaTypes,
   getValueSchema,
   getKeySchema,
-  KeySchema,
 } from "@latticexyz/protocol-parser/internal";
 import { debug } from "./debug";
 import { getTables } from "./getTables";
 import pRetry from "p-retry";
 import { Table } from "@latticexyz/config";
+import { isDefined } from "@latticexyz/common/utils";
 
 export async function ensureTables({
   client,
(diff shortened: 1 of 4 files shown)


packages/cli/src/deploy/ensureTables.ts
```diff
@@ -24,15 +24,56 @@ export async function ensureTables({
...
+
+    const schemaErrors = existingTables
+      .map((table) => {
+        const configTable = configTables.get(table.tableId)!;
+        if (table.keySchemaHex !== configTable.keySchemaHex || table.valueSchemaHex !== configTable.valueSchemaHex) {
+          return [
+            `"${resourceToLabel(table)}" table:`,
+            `  Registered schema: ${JSON.stringify({ schema: getSchemaTypes(table.schema), key: table.key })}`,
+            `      Config schema: ${JSON.stringify({ schema: getSchemaTypes(configTable.schema), key: configTable.key })}`,
+          ].join("\n");
```
alvrs, line 59: oh that's a nice approach for formatted multi line strings

alvrs (approved):

Issue #3342, opened 2024-10-28
deployment of contracts to an existing world fails due to over rate limit error 


0xxlegolas:
Deploying a contract to an existing world in a local anvil node is generally successful; however, attempting to deploy contracts to an existing world to a public/remote chain consistently results in a rate limit error.

Error Details: 
```
 details: '{"code":-32016,"message":"over rate limit"}',
      docsPath: undefined,
      metaMessages: [Array],
      shortMessage: 'HTTP request failed.',
      version: '2.21.19',
      body: [Object],
      headers: [HeadersList],
      status: 429,
      url: 'https://rpc.garnetchain.com'
    },
```

![image](https://github.com/user-attachments/assets/c1bf5830-330a-41d9-842e-6d99f9d201e7)

**Environment:**

RPC Endpoint: https://rpc.garnetchain.com
Contract Address: 0x94b6ec3d701d10657e890b7ca51a5b059c9cc846 (deployed in Garnet)

**Steps to Reproduce:**
Deploy any contract to the existing world at 0x94b6ec3d701d10657e890b7ca51a5b059c9cc846 using the public Garnet endpoint.

0xxlegolas:
@holic

ludns:
I think we could turn off rate limiting till we switch to Peroxide and have a better sense of how to rate limit this properly. ProxyD is hard to work with and I am pretty confused as to why it rate limits so heavily in some cases but not others. We could try bumping the values up again.

PR #3350, opened 2024-10-31
feat(cli): use fetchLogs for resource tags

holic:
fixes https://github.com/latticexyz/mud/issues/3342


packages/store-sync/src/index.ts
```diff
@@ -7,4 +7,5 @@ export * from "./logToTable";
 export * from "./tablesWithRecordsToLogs";
 export * from "./tableToLog";
 export * from "./recordToLog";
-export * from "./logToRecord";
+
+export { logToRecord, type LogToRecordArgs } from "@latticexyz/store/internal";
```
holic, line 11: feels weird to export an internal thing but this let me move it to store

maybe we just remove this export since its intent is internal?

PR #3366, opened 2024-11-07
feat(explorer): get observer url from chain config

holic:


packages/common/src/chains/rhodolite.ts
```diff
@@ -11,9 +17,27 @@ export const rhodolite = {
...
   rpcUrls: {
+    default: defaultRpcUrls,
+    bundler: defaultRpcUrls,
+    quarryPassIssuer: defaultRpcUrls,
+    wiresaw: defaultRpcUrls,
+  },
+  contracts: {
+    ...chainConfig.contracts,
+    quarryPaymaster: {
+      // https://github.com/latticexyz/quarry-paymaster/blob/6542e2f9cf1b61340565d5d840ef2c63a8822c0e/packages/contracts/worlds.json#L3
```
alvrs, line 28:
```suggestion
```

packages/explorer/src/observer/decorator.ts
```diff
@@ -15,106 +15,111 @@ export type ObserverOptions = {
...
+        return write;
+      },
+
+      async writeContract(args) {
+        const writeId = `${client.uid}-${++writeCounter}`;
+        const write = getAction(client, writeContract, "writeContract")(args);
+
+        emit("write", {
+          writeId,
+          from: client.account!.address,
```
alvrs, line 87: missed in #3364 but i think we should also add the `client.userAddress` check here
```suggestion
          // TODO: type as SessionClient once available from entrykit
          // eslint-disable-next-line @typescript-eslint/no-explicit-any
          from: (client as any).userAddress ?? client.account!.address,
```

alvrs (dismissed):

PR #3369, opened 2024-11-11
fix(explorer): get block when transaction hashes are missing

holic:
see https://github.com/wevm/viem/issues/2995 for details

image


holic:
confirmed via vercel preview that this fixes the errors

PR #3374, opened 2024-11-19
feat: system libraries

vdrg:


packages/store/ts/codegen/tablegen.ts
```diff
@@ -65,4 +66,6 @@ export async function tablegen({ rootDir, config }: TablegenOptions) {
       }
     }),
   );
+
+  debug("Generated tables");
```
alvrs, line 70: would it be useful for this log to include the tables that were generated?

packages/world/ts/node/render-solidity/renderSystemLibrary.ts
```diff
@@ -0,0 +1,260 @@
...
+    },
+  ];
+
+  const callingFromRootSystemErrorName = `${libraryName}_CallingFromRootSystem`;
+  const errors = [{ name: callingFromRootSystemErrorName, parameters: [] }, ...systemErrors];
+
+  const camelCaseSystemLabel = systemLabel.charAt(0).toLowerCase() + systemLabel.slice(1);
+  const userTypeName = `${systemLabel}Type`;
+
+  // Remove view functions for root systems
```
alvrs, line 59: should we add some more details to this comment for why we have to remove view function for root systems here? (similar to the comment in `SystemCall.staticcall`)

packages/world/ts/node/render-solidity/renderSystemLibrary.ts
```diff
@@ -0,0 +1,260 @@
...
+  }
+}
+
+function renderEncodeSystemCall(interfaceName: string, functionName: string, parameters: string[]) {
+  const paramNames = parameters.map((param) => param.split(" ").slice(-1)[0]).join(", ");
+  return `abi.encodeCall(${interfaceName}.${functionName}, (${paramNames}))`;
+}
+
+function renderAbiDecode(returnParameters: string[]) {
+  if (returnParameters.length === 0) return "result;";
```
alvrs, line 250: should this be `return result;`? 🤔 

```suggestion
  if (returnParameters.length === 0) return "return result;";
```

vdrg: It is so that the compiler doesn't complain about an unused variable. I guess we could just not assign it if not needed

test/system-libraries/src/namespaces/a/codegen/libraries/ASystemLib.sol
```diff
@@ -0,0 +1,157 @@
...
+import { revertWithBytes } from "@latticexyz/world/src/revertWithBytes.sol";
+import { IWorldCall } from "@latticexyz/world/src/IWorldKernel.sol";
+import { SystemCall } from "@latticexyz/world/src/SystemCall.sol";
+import { Systems } from "@latticexyz/world/src/codegen/tables/Systems.sol";
+import { ResourceId } from "@latticexyz/store/src/ResourceId.sol";
+import { StoreSwitch } from "@latticexyz/store/src/StoreSwitch.sol";
+
+type ASystemType is bytes32;
+
+ASystemType constant aSystem = ASystemType.wrap(0x737961000000000000000000000000004153797374656d000000000000000000);
```
alvrs, line 16: should we render a comment above this that explains the hex? (ie `// equivalent to WorldResourceIdLib.encode({ 
typeId: RESOURCE_SYSTEM, namespace: "a", "name: "ASystem" }))`

alvrs (dismissed):
This is great! Let's add a changeset that includes a short explainer for how to use these. I'd probably also just add that short explainer to https://mud.dev/world/systems right away (maybe with a small disclaimer that this is a new feature and might slightly change based on feedback).

packages/world/ts/node/render-solidity/renderSystemLibrary.ts
vdrg: this was not necessary, will add those functions back to the root system libs

packages/world/ts/node/render-solidity/renderSystemLibrary.ts
vdrg: fixed by only assigning the result if needed

packages/world/ts/config/v2/defaults.ts
```diff
@@ -26,6 +26,8 @@ export type MODULE_DEFAULTS = typeof MODULE_DEFAULTS;
 export const CODEGEN_DEFAULTS = {
   worldInterfaceName: "IWorld",
   worldgenDirectory: "world",
+  systemLibrariesDirectory: "libraries",
```
holic, line 29: imo this should be in 

{namespace}/codegen/systems

to match tables

holic: or maybe we're already generating them that way and this isn't used anymore?

(on mobile so it's hard to tell)

alvrs: currently the `sytstemLibrariesDirectory` option is the name of the subdirectory in `codegen` at the namespace level: https://github.com/latticexyz/mud/blob/635233dc9bce159757bb28b412a3734a2b10aab5/packages/world/ts/node/render-solidity/worldgen.ts#L34-L42

packages/world/ts/node/render-solidity/renderSystemLibrary.ts
```diff
@@ -0,0 +1,270 @@
...
+  const userTypeName = `${systemLabel}Type`;
+
+  return `
+    ${renderedSolidityHeader}
+
+    ${renderImports(imports)}
+
+    type ${userTypeName} is bytes32;
+
+    // equivalent to WorldResourceIdLib.encode({  typeId: RESOURCE_SYSTEM, namespace: "${namespace}", name: "${systemName}" }))
```
alvrs, line 67: one space too many

```suggestion
    // equivalent to WorldResourceIdLib.encode({ typeId: RESOURCE_SYSTEM, namespace: "${namespace}", name: "${systemName}" }))
```

alvrs (dismissed):
very nice ✨✨

packages/world/ts/config/v2/defaults.ts
holic: in that case I think this should be `systems`? because tables are libraries too and this makes it confusing

vdrg: done

PR #3376, opened 2024-11-20
feat(stash): various improvements

alvrs:


packages/stash/src/actions/applyUpdates.ts
```diff
@@ -0,0 +1,91 @@
...
+      table,
+      key,
+      previous: prevRecord,
+      current: nextRecord,
+    });
+  }
+
+  queueMicrotask(() => {
+    notifySubscribers(stash);
+  });
```
holic, line 71:
```suggestion
  // TODO: wrap in queueMicrotask here once we figure out why it affects sync fallback
  notifySubscribers(stash);
```

.changeset/popular-lies-teach.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/stash": patch
+---
+
+Consolidated how state changes are applied and subscribers notified.
```
holic, line 5:
```suggestion
Consolidated how state changes are applied and subscribers notified. Stash subscribers now receive an ordered list of state updates rather than an object.
```

holic (approved):
lgtm! thanks for converting to an array of state changes 🙏 

still unclear if `queueMicrotask` is safe to do here and wondering if we tested this against a real app, or if we should remove for now

packages/stash/src/actions/applyUpdates.ts
holic: tested locally with a fresh template and everything seems fine

making a mental note to come back to this if we see weird issues with syncing, esp large chunks of data

PR #3378, opened 2024-11-20
fix: turbo clean dependencies + gas report underflow + test snapshots

vdrg:
we need these three fixes to fix the workflow


turbo.json
```diff
@@ -17,6 +17,23 @@
...
+      ],
+      "cache": false
+    },
+    "@latticexyz/world#clean": {
+      "dependsOn": [
+        "@latticexyz/world-module-metadata#clean",
+        "@latticexyz/world-module-erc20#clean",
+        "@latticexyz/world-modules#clean"
+      ],
+      "cache": false
```
holic, line 36: just so I understand, this is like an "inverse dependency"? cleaning the stuff that depends on store/world first before cleaning store/world

vdrg: yes exactly

packages/gas-report/src/GasReporter.sol
```diff
@@ -31,7 +31,7 @@ contract GasReporter is Test {
     if (!vm.envOr("GAS_REPORTER_ENABLED", false)) return;
     // subtract 160 gas used by the GasReporter itself
     // add 1 gas so we can later check that this is set
-    uint256 gasUsed = __currentGasReportValue - gas - 160 + 1;
+    uint256 gasUsed = __currentGasReportValue - gas - 159 + 1;
```
alvrs, line 34: don't forget the comment above

holic (approved):

packages/gas-report/src/GasReporter.sol
vdrg: updated

PR #3381, opened 2024-11-22
feat(world): add SystemCall.staticcall

vdrg:


packages/world/src/SystemCall.sol
```diff
@@ -74,6 +74,57 @@ library SystemCall {
...
+    ResourceId systemId,
+    bytes memory callData
+  ) internal view returns (bool success, bytes memory data) {
+    // Load the system data
+    (address systemAddress, bool publicAccess) = Systems._get(systemId);
+
+    // Check if the system exists
+    if (systemAddress == address(0)) revert IWorldErrors.World_ResourceNotFound(systemId, systemId.toString());
+
+    // Staticcalls are not supported for root systems
```
alvrs, line 97: Should we add some context for why they're not supported for root systems in this comment? Might not be obvious on first sight

alvrs (dismissed):

PR #3384, opened 2024-11-28
chore(store-consumer): skip dev script

alvrs:
there are no typescript files in `store-consumer` so `tsup` errors when running `dev`


vdrg (approved):
🤦 damn I missed that when copying from erc20 module, lgtm

vdrg (approved):

PR #3385, opened 2024-11-28
feat: add `getRecords` util to fetch records from indexer, update deployer and explorer to use indexer

### LINKED ISSUE #3355: investigate garnet rpc timing out when deploying to long running worlds
getting reports that the deployer times out when deploying to world at `0x94b6ec3d701d10657e890b7ca51a5b059c9cc846` on garnet (see #3342 for more context)

sounds like this happens even deploying a fresh MUD template to this world

alvrs:
fixes #3355

- [x] add `indexerUrl` as param to deployer
- [x] fall back to using rpc if no `indexerUrl` is provided
- [x] refactor to move `getFunctions` and `getWorldAbi` to `store-sync` to avoid circular dependency


packages/cli/src/deploy/getTables.ts
```diff
@@ -26,45 +18,29 @@ type DeployedTable = Omit<Table, "label" | "namespaceLabel"> & {
...
-        address: worldDeploy.address,
-        fromBlock,
-        toBlock,
-        tableId: storeConfig.namespaces.store.tables.Tables.tableId,
-      });
-    },
   });
-  const logs = flattenStoreLogs(blockLogs.flatMap((block) => block.logs));
 
   // TODO: combine with store-sync logToTable and export from somewhere
```
vdrg, line 36: is this comment outdated?

packages/cli/src/deploy/common.ts
```diff
@@ -119,3 +119,10 @@ export type Module = DeterministicContract & {
    */
   readonly optional?: boolean;
 };
+
+export type CommonDeployOptions = {
+  readonly client: Client<Transport, Chain | undefined, Account>;
```
vdrg, line 124: in which cases do we have an undefined chain? couldn't we enforce it to be defined so we can remove chainId as an additional arg? (and just get it with client.chain.id) not sure if this makes sense

vdrg (approved):
lgtm just a few questions

holic:
Deviating from the native viem types for public clients and wallet clients make types really annoying to work with and pass around clients.On Dec 4, 2024, at 1:54 AM, V ***@***.***> wrote:﻿
@vdrg approved this pull request.

lgtm just a few questions

In packages/cli/src/deploy/getTables.ts:
>    });
-  const logs = flattenStoreLogs(blockLogs.flatMap((block) => block.logs));
 
   // TODO: combine with store-sync logToTable and export from somewhere

is this comment outdated?



In packages/cli/src/deploy/common.ts:
> @@ -119,3 +119,10 @@ export type Module = DeterministicContract & {
    */
   readonly optional?: boolean;
 };
+
+export type CommonDeployOptions = {
+  readonly client: Client<Transport, Chain | undefined, Account>;

in which cases do we have an undefined chain? couldn't we enforce it to be defined so we can remove chainId as an additional arg? (and just get it with client.chain.id) not sure if this makes sense

packages/cli/src/deploy/getTables.ts
alvrs: yep good catch, removed

PR #3390, opened 2024-12-06
Fix: Update buildSystemsManifest.ts to support windows paths

juanfranblanco:
```mud build``` is not supported in the git bash in windows because the path comparison is failing when checking for artifacts, this fixes this by changing the slashes in a common format


juanfranblanco:
Yes it should work, much better if you have it somewhere else.

Issue #2316, opened 2024-02-27
store sync should fetch block of latest block number alongside logs


holic:
see https://github.com/latticexyz/mud/pull/2315#discussion_r1504071416

load balanced RPCs may be slightly out of sync, so when we get a notif for latest block number from a subscription, then fetch logs for that block number, the underlying rpc may not yet have the logs and return empty array

ideally we should fetch the block alongside the logs for that block to confirm the rpc has it, otherwise retry

(and need to handle/account for rpcs that don't support batch calling in this way)

PR #3394, opened 2024-12-11
feat(store-sync): add flag to validate block ranges when fetching from load balanced RPCs

alvrs:
Context on the issue in: https://indexsupply.com/shovel/docs/#unsynchronized-ethereum-nodes

> 3rd party Ethereum API providers may load balance RPC requests across a set of unsynchronized nodes. In rare cases, the eth_getLogs request will be routed to a node that doesn’t have the latest block.
>
> Shovel mitigates this problem by requesting logs using a batch request that includes: eth_getLogs and eth_getBlockByNumber. Shovel tests the eth_getBlockByNumber response to ensure the node serving the eth_getLogs request has processed the requested block. This solution assumes the node provider does not separate the batch.

Unfortunately, [Viem doesn't allow us to do a direct batch call via transport request](https://github.com/wevm/viem/discussions/3274) as batching is abstracted away. There's also no way to detect if a transport has batching enabled. Ideally we'd still allow for passing in a `Client` or `Transport`, where the user can configure their fallback/retry logic, and do a direct batch call with that client/transport or detect if batching is enabled and then do two parallel requests.

Instead, we add an option to fetch/sync with a `Chain` rather than a `Client`, which internally uses the chain's RPC URLs to make requests when `internal_validateBlockRange` is enabled. This means we miss out on some "passthrough" client config (fallback, retry, etc.), but I now kinda like the idea that we'll have more nuanced control over how we talk to the RPC given one or more HTTP URLs and zero or more WS URLs. Maybe `Chain` is a better way forward? Especially as we expand features (like wiresaw), we don't have to ask folks to extend/wrap their transports, etc.

fixes https://github.com/latticexyz/mud/issues/2316
closes https://github.com/latticexyz/mud/pull/2901


dhvanipa:
For the sync stack, the `fromBlock` is decided from a local variable and the `toBlock` is decided by a [subscription](https://viem.sh/docs/actions/public/watchBlocks) to the latest block number.

So we still need an additional filter from the latest block number subscription to not emit when the toBlock is behind the fromBlock, which happens with unsynchronized RPCs.

There is currently silent failure when this happens as when toBlock is less than fromBlock, no error is thrown, and no logs are fetched either

holic:
> So we still need an additional filter from the latest block number subscription to not emit when the toBlock is behind the fromBlock, which happens with unsynchronized RPCs.

As far as I know, this shouldn't happen. Latest block number subscription is derived from Viem's `watchBlocks, which has an explicit check to not emit blocks in the past.
https://github.com/wevm/viem/blob/e1fa711a133629120fb61577caf0288744150054/src/actions/public/watchBlocks.ts#L183-L185

packages/store-sync/src/react/SyncProvider.tsx
```diff
@@ -19,17 +18,18 @@ export function SyncProvider({ chainId, adapter, children, ...syncOptions }: Pro
...
-    queryFn: async () => {
-      const client = getClient(config, { chainId });
-      if (!client) {
-        throw new Error(`Unable to retrieve Viem client for chain ${chainId}.`);
-      }
-      return await adapter({ publicClient: client, ...syncOptions });
-    },
+    queryFn: () =>
+      // TODO: decide if we want to pass in publicClient when not using `internal_validateBlockRange`
+      //       so that the rpc client inherits transports from wagmi config
```
holic, line 31: let's go back to `publicClient` by default because viem plans to add batched RPC call support in v3

only when we're inside the flag should we change to using chain and our own client

packages/block-logs-stream/src/getLogs.ts
```diff
@@ -0,0 +1,77 @@
...
+        toBlock: numberToHex(opts.toBlock),
+      },
+    ],
+  });
+
+  if (!opts.internal_validateBlockRange) {
+    const logs = await getRpcClient(opts).request(logsRequest);
+    return logs.map((log) => formatLog(log));
+  }
+
```
holic, line 40: TODO: add tests for this branch

packages/block-logs-stream/src/fetchLogs.ts
```diff
@@ -158,6 +160,18 @@ export async function* fetchLogs<abiEvents extends readonly AbiEvent[]>({
...
       }
 
+      if (
+        retryCount < maxRetryCount &&
+        (MISSING_BLOCK_ERRORS.some((e) => error.message.includes(e)) ||
+          error.message === new BlockNotFoundError({ blockNumber: toBlock }).message)
+      ) {
+        const seconds = 2 * retryCount;
+        debug(`missing block, retrying in ${seconds}s`);
+        await wait(1000 * seconds);
```
alvrs, line 170: the main reason we'd run into this is a load balanced RPC, in which case much less than 2s wait time should be enough for the RPCs to be back in sync. Should we do the first retry after like 500ms?

alvrs (commented):
✅

packages/block-logs-stream/src/fetchLogs.ts
holic: the first retry happens immediately (retryCount starts at zero)

PR #3395, opened 2024-12-12
fix(world): library imports and overloaded system functions

vdrg:
- Fixes relative system imports in libraries
- Add support for overloaded system functions


packages/world/ts/node/render-solidity/renderSystemLibrary.ts
```diff
@@ -248,9 +249,31 @@ function renderRootCallWrapperFunction(
...
+  `;
+}
+
+function functionInterfaceName(contractFunction: ContractInterfaceFunction) {
+  const { name, parameters } = contractFunction;
+  const paramTypes = parameters
+    .map((param) => param.split(" ")[0])
+    .map((type) => type.replace("[]", "Array"))
+    .join("_");
+  return `_I${name}${paramTypes.length === 0 ? "" : `_${paramTypes}`}`;
```
holic, line 270: is this `I` prefix necessary when we're kinda "hiding" it already with `_` prefix?

holic (dismissed):

packages/world/ts/node/render-solidity/renderSystemLibrary.ts
vdrg: yeah not necessary, do you think it would be better to just do `ISetValue` or maybe `_ISetValue`? I just don't want to run into conflicts eventually, like if a system imports an `ISomeInterface` and the system also has a `someInterface` function.

holic: it's really just to have a function signature defined so we can `encodeCall` right? and not meant to be used anywhere else downstream?

since this is all codegen, is it beneficial to have these defined this way with `encodeCall` vs something like `encodeWithSignature`?

.changeset/real-pigs-bake.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/world": patch
+---
+
+Fixes relative system imports in system libraries, and adds support for overloaded system functions.
```
holic, line 5:
```suggestion
Fixes relative system imports in system libraries and adds support for overloaded system functions.
```

packages/world/ts/node/render-solidity/renderSystemLibrary.ts
vdrg: the benefit is being able to use the system imports / types as parameters directly, which simplifies the code a lot. For example if a system function receives a struct as parameter, generating the signature for it becomes much more complex than just using encodeCall, as you need to import symbol definitions, etc.

For context, @Kooshaba was implementing this with `encodeWithSelector` here: https://github.com/latticexyz/mud/pull/3397/files, but we ended up going with the interfaces as it was simpler.

holic: got it, that all makes sense

in that case, I'd prob just nix the `I` prefix since its not meant to be used outside of the internals of the codegen methods, to make it a bit more readable

and perhaps a comment above the interfaces to explain what they're there for

all our codegen is fairly well documented within code so that its still human readable/usable, easy to dig into to understand what's going on under the hood, and has nice natspec when verifying

PR #3397, opened 2024-12-15
fix: support overloaded system functions in system libraries

Kooshaba:
`encodeCall` was unable to differentiate between overloaded system functions, so i needed to manually construct the selector with `encodeWithSelector`.

i had to add the ability for codegen to parse struct definitions from source files in order to generate full function signatures. this adds one restriction when writing systems: you can't import types from the generated libraries themselves. if this is done, when generating for the first time the system libraries won't be available and it will fail when attempting to find type definitions.

i also punted on being able to import structs from non-relative paths and use them in function signatures. i didn't need it and it was a bit more complicated, but lmk if you think that is important to add.


vdrg:
closing in favor of #3395

PR #3400, opened 2024-12-21
feat(store): improve memory copy

dk1a:
`Memory.copy` behavior is unchanged (I changed dev comment however to better reflect `mcopy`'s description, rather than the previous workaround)
Only the relevant bytecode is changing (which I think what `getStoreLogs.test.ts` is about)

I wanted to use `mcopy` from the beginning, but back in 2023 that eip wasn't implemented yet, so the lengthy code was a less efficient workaround
With solidity 0.8.24 the opcode is usable


dk1a:
What do you do with store logs? The remaining failing tests seem unrelated and happen for me even on main

vdrg:
@dk1a nice! I think you just need to update the vitest snapshots, go to the store package and run `pnpm test --update`. Also, you need to generate the docs `with pnpm docs:generate:api`.

dk1a:
Reinstalling everything locally seems to have helped with docs - I'd run them before but had different results
Also got confused by test's expectations about setup (just running vitest update wasn't enough), but I think I got it correctly now

holic:
I'm curious why we see gas improvements in store but not world?

alvrs:

PR #3404, opened 2025-01-03
fix(cli): only deploy public libraries used by modules or systems

vdrg:


vdrg:
Dropping for now, will try a different approach, it was getting a bit messy

PR #3411, opened 2025-01-06
feat(store-sync): apply logs as one stash update

holic:

like zustand storage adapter, this applies all logs as one `applyUpdates` operation, resulting in one subscriber emission instead of one per log

### DIFF
--- a/packages/store-sync/src/stash/createStorageAdapter.ts
+++ b/packages/store-sync/src/stash/createStorageAdapter.ts
@@ -1,4 +1,4 @@
-import { Stash, deleteRecord, getRecord, setRecord } from "@latticexyz/stash/internal";
+import { Stash, PendingStashUpdate, TableRecord, applyUpdates, getRecord } from "@latticexyz/stash/internal";
 import {
   decodeKey,
   decodeValueArgs,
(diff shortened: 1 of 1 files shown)


packages/store-sync/src/stash/createStorageAdapter.ts
```diff
@@ -23,20 +23,31 @@ export function createStorageAdapter({ stash }: CreateStorageAdapter): StorageAd
       .map((table) => [table.tableId, table]),
   );
 
+  function getRecordId(tableId: Hex, keyTuple: readonly Hex[]): string {
+    return `${tableId}:${concatHex(keyTuple)}`;
+  }
+
   return async function storageAdapter({ logs }: StorageAdapterBlock): Promise<void> {
+    const pendingRecords: Record<string, PendingStashUpdate> = {};
+    const updates: PendingStashUpdate[] = [];
```
holic, line 32: feels a little weird to do all this tracking here, might be nice if in the future stash would export a util for this like

```
mutate(({ getRecord, setRecord }) => {
  getRecord(...)
  setRecord(...)
});
```

where it internally keeps track of the pending updates/mutations and applies them all at the end of the function execution

alvrs (approved):

PR #3412, opened 2025-01-06
test: fix examples

holic:


holic:
I added and then removed `tsc` from minimal example's contracts package because it had issues with TS types inside the linked world package.

If we don't fix the world package types themselves, we can work around this by having a separate tsconfig file for running tests without the TS paths, to mimic the behavior of running this with the prebuilt DTS files as if pulled from npm, rather than the source files.

PR #3418, opened 2025-01-07
feat(explorer): show event logs for interact function

karooolis:
CleanShot 2025-01-07 at 13 43 33@2x


alvrs (approved):

karooolis:
> as a followup, should we decode the logs if it's a store log? (to show which table, column etc changed?)

Oh that's a nice idea, added to Explorer polish list.

PR #3419, opened 2025-01-07
feat(entrykit): initial release

holic:


packages/entrykit/package.json
```diff
@@ -0,0 +1,98 @@
+{
+  "name": "@latticexyz/entrykit",
+  "version": "0.0.0",
```
alvrs, line 3:
```suggestion
  "version": "2.2.14",
```
for consistency with the other MUD packages (this might be updated automatically on the next release)

packages/entrykit/playground/wagmiConfig.ts
```diff
@@ -0,0 +1,40 @@
...
+
+const chains = [
+  {
+    ...anvil,
+    rpcUrls: {
+      ...anvil.rpcUrls,
+      // TODO: automatically grant allowance in anvil instead of requiring the service
+      // quarryPassIssuer: {
+      //   http: ["http://127.0.0.1:3003/rpc"],
+      // },
```
alvrs, line 14:
```suggestion
```
is this solved by using the generous paymaster (or anvil top up) locally?

packages/entrykit/src/AccountModalErrorBoundary.tsx
```diff
@@ -0,0 +1,51 @@
...
+  return (
+    <ErrorBoundary
+      fallbackRender={({ error, resetErrorBoundary }) => (
+        <div className={twMerge("flex-grow flex flex-col justify-center p-5 gap-2")}>
+          <ErrorNotice error={error instanceof Error ? error.stack ?? error.message : error} />
+          {retries > 0 ? (
+            <button
+              type="button"
+              onClick={async (event) => {
+                // fake pending state while we retry so that if the same error occurs,
```
alvrs, line 28: the comment's end seems to be missing

packages/entrykit/src/quarry/transports/methods/estimateUserOperationGas.ts
```diff
@@ -0,0 +1,20 @@
...
+  _params: rpcMethod["Parameters"],
+): Promise<rpcMethod["ReturnType"]> {
+  return formatUserOperationRequest({
+    callGasLimit: 20_000_000n,
+    preVerificationGas: 200_000n,
+    verificationGasLimit: 2_000_000n,
+    paymasterVerificationGasLimit: 200_000n,
+    paymasterPostOpGasLimit: 200_000n,
+  });
+}
```
alvrs, line 20: wonder if for now we should do the rpc call and accept the additional round trip rather than hardcoding by default

alvrs (dismissed):
nothing blocking!

packages/entrykit/package.json
holic: it does get automatically updated on release (snapshots included), so this initial value doesn't really matter, but will update for consistency

packages/entrykit/playground/wagmiConfig.ts
holic: the entrykit playground is _our_ development environment, nothing shipped to users, so leaving this will let us play with quarry stuff locally

there tends to be a lot of things to configure in various spots, so commenting this out here is sort of a breadcrumb trail for myself to remember all the things to turn on before getting things wired up properly

packages/entrykit/src/quarry/transports/methods/estimateUserOperationGas.ts
holic: iirc an rpc call is done automatically when this isn't defined here, and there were issues estimating gas when lazily creating the smart account (both passkey and session account)

I can remove for now to see if it works, and if so, revisit this when we get back to passkey smart accounts

PR #3423, opened 2025-01-07
feat(explorer): copy button

karooolis:
Small QoL improvement to easier copy values. Not yet clear on all the places where it makes sense to add it so only sprinkled it conservatively, and will add as needed.

CleanShot 2025-01-07 at 18 29 58@2x


packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/observe/TransactionTableRow.tsx
```diff
@@ -85,14 +86,14 @@ export function TransactionTableRow({ row }: { row: Row<ObservedTransaction> })
                     <div className="flex items-start gap-x-4">
                       <h3 className="w-[45px] flex-shrink-0 text-2xs font-bold uppercase">Inputs</h3>
 
-                      <div className="flex w-full flex-col gap-y-4">
+                      <div className="rela flex w-full flex-col gap-y-4">
```
Comment (alvrs) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/observe/TransactionTableRow.tsx:89]: is `rela` a tailwind class name? 🤔

Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/observe/TransactionTableRow.tsx]: Good catch! This was a typo, fixed now.

packages/explorer/src/utils.ts
```diff
@@ -2,6 +2,12 @@ import { formatEther } from "viem";
 import { type ClassValue, clsx } from "clsx";
 import { twMerge } from "tailwind-merge";
 
+Object.defineProperty(BigInt.prototype, "toJSON", {
+  get() {
+    return () => String(this);
+  },
+});
```
holic, line 9: no no no no no no, please don't do this, the web has a long long history of learning this lesson over and over and this is always a bad idea

viem has a `stringify` util to stringify JSON with bigints if you wanna use that instead: https://github.com/wevm/viem/blob/main/src/utils/stringify.ts

karooolis: yeah my bad, I was taking the approach here of "master the rules before breaking them" but probably it's like with "don't roll your own crypto", just don't do it. `stringify` looks perfect, thanks, updating.

holic: specifically extending native prototypes that is the bad idea!

karooolis: yeah actually looking back it was a rather terrible idea, as you said should just never do it. Wish I knew about `stringify` before, very useful! Added it.

holic (approved):

PR #3424, opened 2025-01-07
fix(cli): set stateBlock to latest block instead of the deploy block

JackieXu:
follow up to https://github.com/latticexyz/mud/pull/3417

### DIFF
--- a/packages/cli/src/deploy/getWorldDeploy.ts
+++ b/packages/cli/src/deploy/getWorldDeploy.ts
@@ -14,11 +14,14 @@ export async function getWorldDeploy(
 ): Promise<WorldDeploy> {
   const address = getAddress(worldAddress);
 
+  // Fetch latest block to set the range of blocks for logs retrieval.
+  const stateBlock = await getBlock(client, { blockTag: "latest" });
+
   let deploy = deploys.get(address);
   if (deploy != null) {
     return {
       ...deploy,
-      stateBlock: (await getBlock(client, { blockTag: "latest" })).number,
+      stateBlock: stateBlock.number,
     };
   }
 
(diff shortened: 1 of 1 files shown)


packages/cli/src/deploy/getWorldDeploy.ts
```diff
@@ -41,9 +41,12 @@ export async function getWorldDeploy(
     throw new Error("could not find `HelloWorld` or `HelloStore` event");
   }
 
+  // Fetch latest block to set the range of blocks for logs retrieval.
+  const stateBlock = await getBlock(client, { blockTag: "latest" });
```
holic, line 45: maybe move this up to dedupe the two calls above to the latest block?

packages/cli/src/deploy/getWorldDeploy.ts
```diff
@@ -24,9 +24,12 @@ export async function getWorldDeploy(
 
   debug("looking up world deploy for", address);
 
+  // Fetch latest block to set the range of blocks for logs retrieval.
+  const stateBlock = await getBlock(client, { blockTag: "latest" });
```
holic, line 28: there's one more usage above this too!

packages/cli/src/deploy/getWorldDeploy.ts
```diff
@@ -14,19 +14,22 @@ export async function getWorldDeploy(
 ): Promise<WorldDeploy> {
   const address = getAddress(worldAddress);
 
+  // Fetch latest block to set the range of blocks for logs retrieval.
```
alvrs, line 17:
```suggestion
```

alvrs (approved):

PR #3426, opened 2025-01-08
feat(explorer): tables viewer pagination

karooolis:
Paginate explorer's table by using SQL's offset/limit clauses.

https://github.com/user-attachments/assets/a9bbfdb7-6eea-4ffd-a115-e8c392c905ec


examples/local-explorer/packages/contracts/worlds.json
holic, line 1: why was this removed?

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/Explorer.tsx
```diff
@@ -18,33 +18,32 @@ export function Explorer() {
...
   useEffect(() => {
     if (table && (!query || prevSelectedTableId !== selectedTableId)) {
       const tableName = constructTableName(table, worldAddress as Hex, chainId);
-
       if (indexer.type === "sqlite") {
         setQuery(`SELECT * FROM "${tableName}";`);
       } else {
         const columns = Object.keys(table.schema).map((column) => `"${column}"`);
-        setQuery(`SELECT ${columns.join(", ")} FROM "${tableName}";`);
+        setQuery(`SELECT ${columns.join(", ")} FROM ${tableName} LIMIT ${pageSize} OFFSET 0;`);
```
Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/Explorer.tsx:35]: didn't we just get done adding quotes to table names? and now we're removing em? 🤔

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TablesViewer.tsx
```diff
@@ -33,12 +46,47 @@ type Props = {
...
   const [globalFilter, setGlobalFilter] = useQueryState("filter", parseAsString.withDefault(""));
   const [sorting, setSorting] = useQueryState("sort", parseAsJson<SortingState>().withDefault(initialSortingState));
+  const [page, setPage] = useQueryState("page", parseAsInteger.withDefault(0));
+  const [pageSize, setPageSize] = useQueryState("pageSize", parseAsInteger.withDefault(10));
+  const { data: tableData, isPending, isFetching, isError, error } = useTableDataQuery({ table, query, isLiveQuery });
+  const isLoading = isPending || (isFetching && !isLiveQuery);
+  const [pagination, setPagination] = useState<PaginationState>({
+    pageIndex: page,
+    pageSize,
+  });
```
Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TablesViewer.tsx:62]: why the extra state here when the state already lives in the url? why not just use query as source of truth instead of carrying it around twicE?

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TablesViewer.tsx
```diff
@@ -33,12 +46,47 @@ type Props = {
...
+  }, [page, pageSize]);
+
+  const handlePaginationChange: OnChangeFn<PaginationState> = useCallback(
+    (updater) => {
+      const newPaginationState = typeof updater === "function" ? updater(pagination) : updater;
+      const { pageIndex: newPageIndex, pageSize: newPageSize } = newPaginationState;
+      setPage(newPageIndex);
+      setPageSize(newPageSize);
+
+      const decodedQuery = decodeURIComponent(query);
```
Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TablesViewer.tsx:79]: the thing we use for query params doesn't give it back to use decoded already?

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/utils/getLimitOffset.ts
```diff
@@ -0,0 +1,32 @@
+import { Parser } from "node-sql-parser";
+
+const sqlParser = new Parser();
+const opt = {
+  database: "Postgresql",
+};
+
+export function getLimitOffset(query: string) {
+  const decodedQuery = decodeURIComponent(query);
```
Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/utils/getLimitOffset.ts:9]: same question as above, I am confused why this isn't already decoded

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/Explorer.tsx
Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/Explorer.tsx]: ah goddamn, probably lost during rebase, brought back.

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TablesViewer.tsx
Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TablesViewer.tsx]: Initially I didn't like the URL query string it produced but actually it probably doesn't matter that much, and doing it like that is def nicer code, adjusted.

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/utils/getLimitOffset.ts
Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/utils/getLimitOffset.ts:9]: It is but if we don't `decodeURIComponent` and `encodeURIComponent`, it loses new line characters inside the query. Found this to work reliably.

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TablesViewer.tsx
Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TablesViewer.tsx:66]: answered below

examples/local-explorer/packages/contracts/worlds.json
karooolis: Ah sorry, not the first time I commit it due to address change when, in fact, it's not related to the PR. Will set up personal `.gitignore` to ignore `worlds.json`.

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/utils/getLimitOffset.ts
Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/utils/getLimitOffset.ts:9]: Actually, a good time to put this into a reusable hook, added `useSQLQueryState.ts`

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TablesViewer.tsx
Comment (holic) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TablesViewer.tsx]: share more? what was the before/after? I didn't run this myself so I can't tell

Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TablesViewer.tsx]: Now it looks like this `&pagination=%7B%22pageSize%22:10,%22pageIndex%22:4%7D` and before it was `&page=2&pageSize=10`. But I suppose it's a minor difference given most of the rest of the URL query is quite unreadable anyhow.

examples/local-explorer/packages/contracts/worlds.json
karooolis: Added to `.gitignore`

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TablesViewer.tsx
Comment (holic) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TablesViewer.tsx]: ahh okay, I much prefer the nice/user friendly params

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/hooks/usePaginationState.ts
```diff
@@ -0,0 +1,12 @@
...
+import { PaginationState } from "@tanstack/react-table";
+
+export function usePaginationQueryState(defaultPageSize = 10) {
+  return useQueryState(
+    "pagination",
+    parseAsJson<PaginationState>().withDefault({
+      pageIndex: 0,
+      pageSize: defaultPageSize,
+    }),
+  );
```
Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/hooks/usePaginationState.ts:11]: would this make sense to get nicer query params?
```suggestion
  const [page, setPage] = useQueryState("page", parseAsNumber().withDefault(0));
  const [pageSize, setPageSize] = useQueryState("pageSize", parseAsNumber().withDefault(defaultPageSize));
  return { page, setPage, pageSize, setPageSize };
```

Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/hooks/usePaginationState.ts]: Changed it slightly so it returns a tuple like `useQueryState` but yeah, this works nicely 👌

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/utils/getLimitOffset.ts
Comment (holic) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/utils/getLimitOffset.ts:9]: this sounds like `useQueryState` is broken then, using native `URLSearchParams` should correctly encode/decode without any data loss

image

holic (approved):

PR #3427, opened 2025-01-08
feat(explorer): various improvements

karooolis:
Added various QoL improvements.
CleanShot 2025-01-08 at 15 46 08@2x

### DIFF
--- a/.changeset/eleven-days-teach.md
+++ b/.changeset/eleven-days-teach.md
@@ -1,0 +1,8 @@
+---
+"@latticexyz/explorer": patch
+---
+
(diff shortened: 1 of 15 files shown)


ludns:
One thing I would love to see (but is not crazy important for now) is the time it takes to execute a query after it has finished running. Having time taken + rows returned  (and later maybe bytes scanned if the API returns it) when you run a query is something a bunch of SQL GUIs do.

Also enabling queries to re-run if the SQL input has not changed (currently it won't let you re-run a query if it hasn't changed, which is unintuitive given the same query can return different results over time).

PR #3428, opened 2025-01-08
feat(common): set chain fees

holic:

not sure if this is actually used, but seems more complete/accurate than without? this may also be technically incorrect if we have a period of heavy load/traffic, where gas prices spike past 100k wei


holic:
closing in favor of our polling fee approach

PR #3436, opened 2025-01-09
fix(world): use WorldContextConsumerLib._msgSender() instead of msg.sender

vdrg:


.changeset/heavy-eggs-smell.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/world": patch
+---
+
+Use WorldContextConsumerLib._msgSender() instead of msg.sender in system libraries.
```
holic, line 5:
```suggestion
Use `WorldContextConsumerLib._msgSender()` instead of `msg.sender` in system libraries.
```

alvrs (approved):

PR #3439, opened 2025-01-10
feat(create-mud): new react template with stash/entrykit

holic:


templates/react/package.json
```diff
@@ -27,5 +27,29 @@
   "engines": {
     "node": "^18",
     "pnpm": "^8 || ^9"
+  },
+  "pnpm": {
```
holic, line 31: this is here for the same reasons as we had before where the wagmi instance used by rainbowkit isn't resolved to the same place as the wagmi instance used by the template/app, so we get issues with wagmi config not found

it means that you don't get live updates when developing the template and to get the latest version of MUD deps, you have to do something like
```
rm -rf node_modules packages/*/node_modules; pnpm install
```
to forcefully remove all deps and reinstall (even `pnpm install --force` isn't enough)

holic: I found another approach that works a bit better and has less downsides.

templates/react/package.json
```diff
@@ -27,5 +27,12 @@
...
+    "overrides": {
+      "@tanstack/react-query": "link:../../packages/entrykit/node_modules/@tanstack/react-query",
+      "@types/react": "link:../../packages/entrykit/node_modules/@types/react",
+      "wagmi": "link:../../packages/entrykit/node_modules/wagmi"
+    }
```
holic, line 36: This is how we work around the linking issue where different instances of Wagmi exist inside the dependency tree (even if pinned to the same version), which leads to things like Rainbowkit inside EntryKit not being able to find the Wagmi config provided the template's app.

The downside is that we need to keep versions aligned in the template + EntryKit to not have failing tests or weird edge cases. But that seems like a reasonable trade-off for having a working local environment for templates.

Some potential solutions we can look into in the future:
- move templates into the monorepo root workspace (would mean "flattening" the templates to not use workspaces) so the templates can share common pnpm dependencies with MUD packages
- replace Rainbowkit in EntryKit with our own wallet connection UI (the underlying error seems to be related to transitive deps across linked deps in pnpm workspace)

templates/react/packages/client/src/App.tsx
```diff
@@ -1,105 +1,46 @@
...
+  const onMove = useMemo(
+    () =>
+      sync.data && worldContract
+        ? async (direction: Direction) => {
+            const tx = await worldContract.write.app__move([mudConfig.enums.Direction.indexOf(direction)]);
+            await sync.data.waitForTransaction(tx);
+          }
+        : undefined,
+    [sync.data, worldContract],
+  );
```
holic, line 26: This still feels a little clunky to me. I tried to streamline this by doing the sync data check inside `useWorldContract` so that we could have one hook to return both `worldContract` and `waitForTransaction` (since you're often doing these together), but I was constantly battling TS depth issues where editing anything in `useWorldContract` caused it to spin due to the huge world ABI.

A potential next step to improve this would be better utils to call individual systems via their individual ABIs rather than a monolithic world contract, but wagmi doesn't make it easy to call using a separate client (session client, in our case).

.changeset/five-bats-live.md
```diff
@@ -0,0 +1,5 @@
+---
+"create-mud": patch
+---
+
+Updated React template with Stash client library, EntryKit for wallet support, and a cleaned up app structure.
```
alvrs, line 5:
```suggestion
Updated React template with Stash client state library, EntryKit for wallet support, and a cleaned up app structure.
```

alvrs (approved):
Very nice!

PR #3440, opened 2025-01-10
fix(entrykit): don't call decorator

holic:
I was seeing errors like `extendFn is not a function` that stemmed from assuming `smartAccountActions` was meant to be called/instantiated. Not sure how this didn't show up before, but I seem to recall seeing this sporadically. Might be related to semver (as in a prior patch version had diff behavior), so I also pinned the permissionless version.

### DIFF
--- a/packages/entrykit/src/getSessionClient.ts
+++ b/packages/entrykit/src/getSessionClient.ts
@@ -18,13 +18,14 @@ export async function getSessionClient<chain extends Chain>({
   worldAddress: Address;
 }): Promise<SessionClient<chain>> {
   const bundlerTransport = getBundlerTransport(client.chain);
-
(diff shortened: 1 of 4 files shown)


packages/entrykit/src/getSessionClient.ts
```diff
@@ -18,13 +18,14 @@ export async function getSessionClient<chain extends Chain>({
...
+  const bundlerClient = createBundlerClient({
     transport: bundlerTransport,
     client,
     account: sessionAccount,
-  })
-    .extend(smartAccountActions())
+  });
+
+  const sessionClient = bundlerClient
+    .extend(smartAccountActions)
```
holic, line 28: this was the main fix

PR #3444, opened 2025-01-13
feat(explorer): query execution time

karooolis:
CleanShot 2025-01-13 at 17 16 55


packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/SQLEditor.tsx
```diff
@@ -99,6 +105,18 @@ export function SQLEditor({ table }: Props) {
...
+          {tableData && (
+            <span className="mr-4 flex items-center gap-1.5 text-xs text-white/60">
+              <Tooltip text="Execution time for the SQL query">
+                <span className="flex items-center gap-1.5">
+                  <span className="inline-block h-[6px] w-[6px] animate-pulse rounded-full bg-success" />
+                  <span>{Math.round(tableData.queryDuration)}ms</span>
+                </span>
+              </Tooltip>
+              ·<span> {tableData.rows.length} rows</span>
+            </span>
```
Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/SQLEditor.tsx:117]: instead of a bunch of UI indicators, tooltips, etc. what if this just reads
> 200 rows in 200ms

Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/SQLEditor.tsx:117]: I'm leaning towards leaving it with UI indicators of which there is only one primarily i.e. the blinking green dot to show that the query is running. The goal is to indicate clearly that the results are "live". When merging with `pause/resume` live query PR, I'll also change color to indicate if it's paused as well. I'd love to animate it a little better still but the current solution is probably the best bang for the buck, and will leave more fancy animation as a pleasant side quest.

Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/SQLEditor.tsx:117]: They can work independently but yeah, do need to fix merge conflicts

holic (approved):

packages/explorer/src/app/(explorer)/queries/useTableDataQuery.ts
```diff
@@ -76,13 +80,14 @@ export function useTableDataQuery({ table, query, isLiveQuery }: Props) {
...
         rows,
+        queryDuration: data.queryDuration,
       };
     },
     retry: false,
     enabled: !!table && !!query,
     refetchInterval: (query) => {
       if (query.state.error) return false;
-      else if (isLiveQuery) return false;
+      else if (!isLiveQuery) return false;
```
holic, line 90: is this technically a bug fix?

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TablesViewer.tsx
```diff
@@ -110,7 +114,11 @@ export function TablesViewer({ table, query, isLiveQuery }: Props) {
   });
 
   return (
-    <div className="!-mt-10 space-y-4">
+    <div
+      className={cn("space-y-4", {
+        "!-mt-10": indexer.type === "hosted",
```
Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TablesViewer.tsx:119]: what does the `!-` prefix mean?

Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TablesViewer.tsx:119]: `!important`. I've used it here because it clashes with `space-y-4`. I know generally `!important` is a bad practice but in this case it's only isolated to this `div` due to layout "exception" that happens when SQL editor is not shown when local indexer is used.

PR #3446, opened 2025-01-13
feat(explorer): indicate unsaved query changes

karooolis:
CleanShot 2025-01-13 at 22 34 27@2x

### DIFF
--- a/.changeset/curly-carrots-sit.md
+++ b/.changeset/curly-carrots-sit.md
@@ -1,0 +1,5 @@
+---
+"@latticexyz/explorer": patch
+---
+
+The SQL editor now displays an indicator for unsaved changes.
--- a/.changeset/curly-carrots-sit.md
+++ b/.changeset/curly-carrots-sit.md
@@ -2,4 +2,4 @@
 "@latticexyz/explorer": patch
 ---
 
-The SQL editor now displays an indicator for unsaved changes.
+The SQL editor now shows an indicator if the query has changed since it was last run.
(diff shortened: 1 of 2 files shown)


.changeset/curly-carrots-sit.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/explorer": patch
+---
+
+The SQL editor now displays an indicator for unsaved changes.
```
holic, line 5: what does "unsaved" mean? makes me think that can we save queries, but I don't think that's true? maybe rephrase a bit?

```suggestion
The SQL editor now shows an indicator when the query has changed but has not been run.
```

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/SQLEditor.tsx
```diff
@@ -31,6 +31,8 @@ export function SQLEditor({ table }: Props) {
       query,
     },
   });
+  const currentQuery = form.watch("query");
+  const hasUnsavedChanges = currentQuery !== query;
```
Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/SQLEditor.tsx:35]: similar to above, maybe rename?
```suggestion
  const hasQueryChanged = currentQuery !== query;
```
or just inline this ternary below since we don't use it anywhere else

.changeset/curly-carrots-sit.md
karooolis: Yeah this sounds better, thanks!

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/SQLEditor.tsx
Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/SQLEditor.tsx:35]: Yeah agreed, inlining is better in this case, adjusted

holic (approved):

PR #3447, opened 2025-01-14
feat(explorer): key columns indicator

karooolis:
CleanShot 2025-01-14 at 15 38 56

### DIFF
--- a/.changeset/fast-nails-build.md
+++ b/.changeset/fast-nails-build.md
@@ -1,0 +1,5 @@
+---
+"@latticexyz/explorer": patch
+---
+
+Indicate MUD schema key columns in the table view of the Explore tab.
--- a/.changeset/fast-nails-build.md
+++ b/.changeset/fast-nails-build.md
@@ -2,4 +2,4 @@
 "@latticexyz/explorer": patch
 ---
 
-Indicate MUD schema key columns in the table view of the Explore tab.
+Indicated MUD schema key columns in the table view of the Explore tab.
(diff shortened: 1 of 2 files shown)


holic:
what about, instead of an icon, we bold the keyed fields or add a subtle background? similar to how table data is visualized here: https://mud.dev/store/data-model

karooolis:
> what about, instead of an icon, we bold the keyed fields or add a subtle column background? similar to how table data is visualized here: https://mud.dev/store/data-model

I took inspiration from Supabase, somehow found it quite nice and visually non-disruptive. For now would prefer to leave it as it is just because of other priorities and also it was easy/quick change to make but noted it in to recheck on how it would look like, could look nice. 
CleanShot 2025-01-14 at 17 26 19

.changeset/fast-nails-build.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/explorer": patch
+---
+
+Indicate MUD schema key columns in the table view of the Explore tab.
```
holic, line 5:
```suggestion
Indicated MUD schema key columns in the table view of the Explore tab.
```

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TablesViewer.tsx
```diff
@@ -48,6 +49,7 @@ export function TablesViewer({ table, query }: { table?: TableType; query?: stri
               className="-ml-4"
               onClick={() => column.toggleSorting(column.getIsSorted() === "asc")}
             >
+              {Object.keys(keySchema).includes(name) && <KeyIcon className="mr-2 h-3 w-3" />}
```
Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TablesViewer.tsx:52]: you could do `keySchema[name]` or `name in keySchema` here to avoid converting this to an array then iterating through it

you could also `table.schema.key.includes(name)`, as in no need to generate a key schema from the table since the schema already lists the key field names

holic (approved):

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TablesViewer.tsx
Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TablesViewer.tsx:52]: oh for real, nicely spotted! not sure why I did it that way, used `name in keySchema` in the end, like the descriptiveness of it.

holic (approved):

PR #3450, opened 2025-01-15
feat(stash): add useRecord and useRecords hooks

holic:
pulled out of https://github.com/latticexyz/mud/pull/3439

### DIFF
--- a/packages/stash/src/react/isArrayEqual.ts
+++ b/packages/stash/src/react/isArrayEqual.ts
@@ -1,0 +1,8 @@
+export function isArrayEqual(a: readonly unknown[], b: readonly unknown[]): boolean {
+  if (a === b) return true;
+  if (a.length !== b.length) return false;
+  for (let i = a.length - 1; i >= 0; i--) {
+    if (a[i] !== b[i]) return false;
+  }
+  return true;
+}
--- a/packages/stash/src/react/useRecords.ts
+++ b/packages/stash/src/react/useRecords.ts
@@ -1,0 +1,22 @@
+import { Table } from "@latticexyz/config";
+import { Key, Stash, TableRecord } from "../common";
+import { useStash } from "./useStash";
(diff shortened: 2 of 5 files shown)


packages/stash/src/react/useRecords.ts
```diff
@@ -0,0 +1,22 @@
...
+import { getRecords } from "../actions/getRecords";
+import { isArrayEqual } from "./isArrayEqual";
+
+export type UseRecordsOptions<table extends Table = Table> = {
+  stash: Stash;
+  table: table;
+  keys?: readonly Key<table>[];
+};
+
+export type UseRecordsResult<table extends Table = Table> = readonly TableRecord<table>[];
```
holic, line 13: In almost all cases, I want to work with an array of records, not an object.

Doing this inside a `useStash` selector means we can memoize the results and avoid things like `Object.values` everywhere.

packages/stash/src/react/isArrayEqual.ts
```diff
@@ -0,0 +1,8 @@
+export function isArrayEqual(a: readonly unknown[], b: readonly unknown[]): boolean {
+  if (a === b) return true;
+  if (a.length !== b.length) return false;
+  for (let i = a.length - 1; i >= 0; i--) {
```
holic, line 4: one nice thing about going backwards is that new entries in objects would be at the end once converted to an array, so this should fairly quickly return in cases where a new record is added

packages/stash/src/react/isArrayEqual.ts
```diff
@@ -0,0 +1,8 @@
+export function isArrayEqual(a: readonly unknown[], b: readonly unknown[]): boolean {
```
holic, line 1: this is specifically optimized for comparing `Object.values(getRecords(...))` from stash, where we can expect stable references to objects, but not necessarily the outer records map

alvrs (approved):

PR #3455, opened 2025-01-17
fix(cli): don't overload system abi type

holic:
pulled out of https://github.com/latticexyz/mud/pull/3290

not sure why this appeared there but not in main, but `abi` on `System` type was being overloaded (`Abi` set by `DeterministicContract` and then `string[]` set by `System`) throwing type errors

ideally one of these becomes authoritative (e.g. manifest) but didn't wanna think through that at the moment, so I moved the metadata stuff aside

### DIFF
--- a/.changeset/seven-seahorses-roll.md
+++ b/.changeset/seven-seahorses-roll.md
@@ -1,0 +1,5 @@
+---
+"@latticexyz/cli": patch
+---
+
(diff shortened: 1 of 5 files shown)


holic:
looks like this wasn't showing up due to `moduleResolution: "Node"`

Issue #3389, opened 2024-12-05
mud cli do not work with node 22


wighawag:
Just installed @latticexyz/cli and made a simple config file

exeuting `pnpm mud tablegen` give me the following:

```
file:///home/wighawag/dev/local/mud/test-cli/node_modules/.pnpm/@latticexyz+cli@2.2.14/node_modules/@latticexyz/cli/dist/commands-B5O3X52W.js:9
  \u26A0\uFE0F CREATE2 deployer created at ${a.contractAddress} does not match the CREATE2 determinstic deployer we expected (${Se})`),a.contractAddress}import{waitForTransactionReceipt as gr}from"viem/actions";import{concatHex as Yt,getCreate2Address as Gt}from"viem";import{getBytecode as Qt}from"viem/actions";import{padHex as _t}from"viem";import Kt from"@latticexyz/world/out/IBaseWorld.sol/IBaseWorld.abi.json"assert{type:"json"};import{helloStoreEvent as Jt}from"@latticexyz/store";import{helloWorldEvent as qt}from"@latticexyz/world";var S=_t("0x",{size:32}),Q=parseInt("6000",16),ie=[Jt,qt],D=Kt,Ge=["2.0.0","2.0.1","2.0.2"],Qe=["2.0.0","2.0.1","2.0.2"];import{sendTransaction as Zt}from"@latticexyz/common";import Xt from"p-retry";async function Ze({client:e,deployerAddress:o,bytecode:t,deployedBytecodeSize:r,debugLabel:n="contract"}){if(t.includes("__$"))throw new Error(`Found unlinked public library in ${n} bytecode`);let s=Gt({from:o,salt:S,bytecode:t});return await Qt(e,{address:s,blockTag:"pending"})?(m("found",n,"at",s),[]):(r!=null&&(r>Q?console.warn(`
                                                                                                                                                                                                                                                                                                                                                                                                                                ^^^^^^

SyntaxError: Unexpected identifier 'assert'
    at compileSourceTextModule (node:internal/modules/esm/utils:338:16)
    at ModuleLoader.moduleStrategy (node:internal/modules/esm/translators:102:18)
    at #translate (node:internal/modules/esm/loader:437:12)
    at ModuleLoader.loadAndTranslate (node:internal/modules/esm/loader:484:27)
    at async ModuleJob._link (node:internal/modules/esm/module_job:115:19)

Node.js v22.12.0
```


reproduction repo: https://github.com/bug-reproduction/mud-node-22

[...]

PR #3456, opened 2025-01-17
feat: bump to node 20

holic:
node 18 EOL is april 2025: https://endoflife.date/nodejs

```
pnpm clean
rm -rf node_modules packages/*/node_modules test/*/node_modules
nvm install 20
nvm use default
npm install -g pnpm
pnpm install
pnpm build
```


holic:
was hoping bumping to node 20 would fix https://github.com/latticexyz/mud/issues/3389 but needs node 22

might see about skipping 20 in favor of 22

holic:
tried node 22 but something about our TS->JS build step prob has to change to make JSON imports work

- using `assert` after json import leads to `SyntaxError: Unexpected identifier 'assert'` when using built files from other packages (e.g. cli bin)
- using `with` after json import leads to `Expected ";" but found "with"` when building ESM/TS file
- not using either leads to `TypeError [ERR_IMPORT_ATTRIBUTE_MISSING]: Module "[path].abi.json" needs an import attribute of "type: json"` when using built files from other packages (e.g. cli bin)

holic:
after bumping tsup to resolve the `with` parse errors during TS builds, I now get issues with create-mud

```
create-mud:build: node:internal/modules/esm/resolve:314
create-mud:build:   return new ERR_PACKAGE_PATH_NOT_EXPORTED(
create-mud:build:          ^
create-mud:build:
create-mud:build: Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: No "exports" main defined in node_modules/.pnpm/npm-run-path@6.0.0/node_modules/unicorn-magic/package.json
create-mud:build:     at exportsNotFound (node:internal/modules/esm/resolve:314:10)
create-mud:build:     at packageExportsResolve (node:internal/modules/esm/resolve:605:13)
create-mud:build:     at resolveExports (node:internal/modules/cjs/loader:638:36)
create-mud:build:     at Function._findPath (node:internal/modules/cjs/loader:743:31)
```

likely stemming from create-create-app being super old and outdated, and maybe we need to revisit node 22 when we go change our template strategy

PR #3462, opened 2025-01-17
fix(store): add missing ts codegen exports

dk1a:
Some files (one of which I needed) haven't been added to the export endpoint
I also reordered them to be alphabetic


holic:
these aren't really meant to be externally consumed, so I am curious to hear how you're using them

dk1a:
A codegen-ey plugin, though it's not really a plugin since it's a parallel pipeline entirely
Some of mud's codegen can be reused for similar purposes, in my case autogenerating a unique index (sql-like) based on table config and a list of indexed keys/values
I'll link the repo in discord once something viable comes together

PR #3464, opened 2025-01-20
chore(explorer): handle tuples in interact form

### LINKED ISSUE #3359: "Cannot convert undefined to a BigInt" error when using World Explorer write
When calling the drop function on the following MUD world with these params: 37, 98, (1268, -93, -11), the explorer says: "Cannot convert undefined to a BigInt"

https://explorer.mud.dev/redstone/worlds/0xf75b1b7bdb6932e487c4aa8d210f4a682abeacf0/interact?function=drop 

Screenshot 2024-11-01 at 11 39 59 PM

JamesLefrere:
Fixes #3359 by doing what the user intended and making it clear what to input.

image

### DIFF
--- a/packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/interact/FunctionField.tsx
+++ b/packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/interact/FunctionField.tsx
@@ -45,13 +45,24 @@ const getInputLabel = (input: AbiParameter): string => {
     return input.type;
   }
 
-  const componentsString = input.components.map(getInputLabel).join(", ");
   if (input.type === "tuple") {
-    return `[${componentsString}]`;
+    return input.name || input.type;
   } else if (input.type === "tuple[]") {
-    return `[${componentsString}][]`;
+    return `${input.name}[]`;
   }
   return input.type;
+};
+
+const getInputPlaceholder = (input: AbiParameter): string => {
+  if (!("components" in input)) {
+    return input.type;
+  }
+
+  const componentsString = input.components.map(getInputLabel).join(", ");
+  if (input.type === "tuple[]") {
+    return `[${componentsString}][]`;
+  }
+  return `[${componentsString}]`;
 };
 
 export function FunctionField({ worldAbi, functionAbi }: Props) {
@@ -153,7 +164,7 @@ export function FunctionField({ worldAbi, functionAbi }: Props) {
                 <FormItem>
                   <FormLabel>{input.name}</FormLabel>
                   <FormControl>
-                    <Input placeholder={inputLabels[index]} {...field} />
+                    <Input placeholder={getInputPlaceholder(input)} {...field} />
                   </FormControl>
                   <FormMessage />
                 </FormItem>
--- a/packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/interact/FunctionField.tsx
+++ b/packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/interact/FunctionField.tsx
@@ -4,7 +4,7 @@ import { Coins, ExternalLinkIcon, Eye, LoaderIcon, Send } from "lucide-react";
 import Link from "next/link";
 import { useParams } from "next/navigation";
 import { toast } from "sonner";
-import { Abi, AbiFunction, Address, Hex, decodeEventLog, stringify } from "viem";
+import { Abi, AbiFunction, AbiParameter, Address, Hex, decodeEventLog, stringify } from "viem";
 import { useAccount, useConfig } from "wagmi";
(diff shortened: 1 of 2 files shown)


karooolis:
Hey @JamesLefrere , thanks for the contribution, this looks great! We have plans to improve the tuples/arrays support further but this is already very useful. Only issue is that it doesn't seem to work for all tuples, for example `batchCall` (see the attached screenshot). Seems to happen when the function parameter is an array of tuples.

CleanShot 2025-01-21 at 12 08 53@2x

JamesLefrere:
Ah thanks @karooolis , you're right, looks like we need to make that recursive perhaps and the user could enter `[[a, b, c], [d, e]]`.

karooolis:
> Ah thanks @karooolis , you're right, looks like we need to make that recursive perhaps and the user could enter `[[a, b, c], [d, e]]`.

Yeah good shoutout, a tuple could contain another tuple, and so on.

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/interact/FunctionField.tsx
```diff
@@ -56,6 +56,13 @@ export function FunctionField({ worldAbi, functionAbi }: Props) {
   const [txHash, setTxHash] = useState<Hex>();
   const txUrl = blockExplorerTransactionUrl({ hash: txHash, chainId });
 
+  // Tuples are mapped to e.g. "[int16, int16, int16]" to inform the user what to input
+  const inputLabels = functionAbi.inputs.map((input) =>
+    input.type === "tuple"
+      ? `[${(input as unknown as { components: AbiParameter[] }).components.map((c) => c.type).join(", ")}]`
```
Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/interact/FunctionField.tsx:62]: ooc why do we have to cast this type?

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/interact/FunctionField.tsx
```diff
@@ -111,14 +120,13 @@ export function FunctionField({ worldAbi, functionAbi }: Props) {
...
-  const inputsLabel = functionAbi?.inputs.map((input) => input.type).join(", ");
   return (
     <div className="pb-6">
       <Form {...form}>
         <form onSubmit={form.handleSubmit(onSubmit)} id={functionAbi.name} className="space-y-4">
           <h3 className="font-semibold">
-            <span className="text-orange-500">{functionAbi?.name}</span>
-            <span className="opacity-50">{inputsLabel && ` (${inputsLabel})`}</span>
+            <span className="text-orange-500">{functionAbi.name}</span>
+            <span className="opacity-50">{`(${inputLabels.join(", ")})`}</span>
```
Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/interact/FunctionField.tsx:129]:
```suggestion
            <span className="opacity-50">({inputLabels.join(", ")})</span>
```

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/interact/FunctionField.tsx
Comment (JamesLefrere) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/interact/FunctionField.tsx]: I missed `readonly components` 👍

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/interact/FunctionField.tsx
```diff
@@ -56,6 +56,13 @@ export function FunctionField({ worldAbi, functionAbi }: Props) {
   const [txHash, setTxHash] = useState<Hex>();
   const txUrl = blockExplorerTransactionUrl({ hash: txHash, chainId });
 
+  // Tuples are mapped to e.g. "[int16, int16, int16]" to inform the user what to input
+  const inputLabels = functionAbi.inputs.map((input) =>
+    input.type === "tuple"
+      ? `[${(input as { components: readonly AbiParameter[] }).components.map((c) => c.type).join(", ")}]`
```
Comment (holic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/interact/FunctionField.tsx:62]: I am still confused why casting is necessary at all, as this type should be narrowed by the `input.type === "tuple"` check

Comment (JamesLefrere) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/interact/FunctionField.tsx]: I was too. Probably because of union & intersection with `components` not in the base type? Either way, this is fixable by checking for `components` as well, no cast needed then.

karooolis:
Having implemented recursive handling of tuples I'm thinking next to the function name to not show tuples types but rather the tuple's name. And then inside inputs show the types. Added screenshot for the reference on how it currently looks.

CleanShot 2025-01-23 at 20 12 58@2x

karooolis:
In the end came up with this for now:
CleanShot 2025-01-31 at 17 48 35@2x
CleanShot 2025-01-31 at 17 48 57@2x

karooolis (approved):

karooolis (approved):

PR #3467, opened 2025-01-20
chore: standardise workspace dependencies

### LINKED ISSUE #2962: move common dev deps to workspace root
things like eslint, TS, arktype, tsx, etc. can all be moved to workspace root to avoid installing in multiple places and managing versions everywhere

JamesLefrere:
Fixes #2962 

- Replace common workspace package devDependencies (where possible without major modifications) with the workspace root's devDependencies
- Update some error message snapshots for workspace vitest version
- Caveats:
  - `execa` wasn't updated to the workspace version in all cases, because there are some breaking changes going from 6/7 to 9
  - Some optional peer dependencies have version mismatches still. The React hooks tests just need an update from the discontinued `react-hooks-testing-library`.
  - There are two versions of `forge-std` in use still; unclear why these were chosen, so I didn't change them
  - No packages were added to the devDeps, but they are used a lot (e.g. `solhint`). It probably makes sense to add more, but this is a large PR already.

### DIFF
--- a/packages/abi-ts/package.json
+++ b/packages/abi-ts/package.json
@@ -42,9 +42,7 @@
   },
   "devDependencies": {
     "@types/debug": "^4.1.7",
-    "@types/yargs": "^17.0.10",
-    "tsup": "^6.7.0",
-    "vitest": "0.34.6"
+    "@types/yargs": "^17.0.10"
   },
   "publishConfig": {
     "access": "public"
--- a/packages/block-logs-stream/package.json
+++ b/packages/block-logs-stream/package.json
@@ -38,9 +38,7 @@
     "rxjs": "7.5.5"
   },
   "devDependencies": {
-    "@types/debug": "^4.1.7",
-    "viem": "2.21.19",
-    "vitest": "0.34.6"
+    "@types/debug": "^4.1.7"
   },
   "peerDependencies": {
     "viem": "2.x"
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -83,7 +83,6 @@
     "@types/toposort": "^2.0.6",
     "@types/yargs": "^17.0.10",
     "ds-test": "https://github.com/dapphub/ds-test.git#e282159d5170298eb2455a6c05280ab5a73a4ef0",
-    "forge-std": "https://github.com/foundry-rs/forge-std.git#74cfb77e308dd188d2f58864aaf44963ae6b88b1",
-    "vitest": "0.34.6"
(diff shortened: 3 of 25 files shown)


packages/abi-ts/package.json
```diff
@@ -34,17 +34,17 @@
     "test:ci": "pnpm run test"
   },
   "dependencies": {
-    "chalk": "^5.3.0",
+    "chalk": "*",
     "debug": "^4.3.4",
-    "execa": "^7.0.0",
-    "glob": "^10.4.2",
+    "execa": "*",
+    "glob": "*",
```
holic, line 40: when publishing, do these version wildcards get swapped for the workspace semver?

JamesLefrere: Great point...from my testing, no they don't 😭 

Seems like this a non-starter for mud in that case, we can't run this PR as-is. 

Some possible paths:

1. Version 10 of pnpm introduces [Catalogs](https://pnpm.io/catalogs#the-catalog-protocol-catalog). I just tried this locally and it works as intended; define catalog versions in `pnpm-workspace.yaml`, specify `catalog:` in workspace packages, and then packing/publishing any workspace package will set its version directly in the resulting `package.json`. We could try an upgrade (separate PR) then change this to use catalogs.
2. Use a tool like [syncpack](https://jamiemason.github.io/syncpack/) to manage workspace package versions.

Not sure what your plans are for v10 at the moment [(release notes here)](https://github.com/pnpm/pnpm/releases/tag/v10.0.0). Sounds... maybe doable?

holic: I recently got rid of our usage of catalogs as it made some other things harder to work with: https://github.com/latticexyz/mud/pull/3405

I have already made a habit of trying to keep versions aligned across the repo. It's not perfect, but hasn't been a huge issue.

The issue you're trying to solve (https://github.com/latticexyz/mud/issues/2962) is specifically for common dev deps, so we can still move those up to the workspace root (and no need to declare them in each individual package's dev deps).

JamesLefrere: Got it, shame. Ok then, I'll just simply remove the wildcards from all `dependencies` in workspace packages, leave the wildcards for `devDependencies`, and that should do the trick.

holic: IMO we should lift the common dev deps into the workspace root's dev deps, and remove them from each package, rather than a wildcard

JamesLefrere:
Before:

```
Packages: +2285
...
1.7G    node_modules

```

After:

```
Packages: +2157
...
1.5G    node_modules
```

PR #3470, opened 2025-01-20
feat(explorer): save searched namespace

karooolis:
Save searched namespace so that context is not lost when re-opening namespace search menu.

CleanShot 2025-01-20 at 17 41 59


packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TableSelector.tsx
```diff
@@ -44,6 +44,16 @@ export function TableSelector({ tables }: { tables?: Table[] }) {
     }
   }, [selectedTableId, setTableId, tables]);
 
+  useLayoutEffect(() => {
+    if (open && selectedTableId && selectedTableConfig) {
+      setTimeout(() => {
```
Comment (karooolis) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TableSelector.tsx:49]: Don't like it but couldn't get it to work any other way. Need to wait for dropdown to get rendered in order for it to be queried.

holic (approved):

PR #3471, opened 2025-01-20
feat(explorer): enable re-executing query

karooolis:
Until now, it was only possible to execute query if it changed because the `useTableDataQuery` hook would only rerun in case `query` changed. Now calling `refetch()` explicitly to trigger new table data fetch.


holic (approved):

holic:
need a changeset

holic (approved):

PR #3472, opened 2025-01-20
chore: update node 22

JamesLefrere:
Picks commits from #3456 

- Updates engine and node types to v22, dropping support for older versions
- Resolves issue with outdated `unicorn-magic` dependency by overriding it with a fixed version


holic:
one question I have is whether the packages built under node 22 will still be usable in node 18 + 20

JamesLefrere:
> one question I have is whether the packages built under node 22 will still be usable in node 18 + 20

I have just tested this with the vanilla example with node 18, and it seems fine. Likewise, we only have the engines requirement on the root workspace, so all good there.

PR #3475, opened 2025-01-21
chore(indexer): add v2 api route, don't include schemas table by default

### LINKED ISSUE #3386: Add option to indexer API to not include `store.Tables` logs in `/api/logs` response

JamesLefrere:
Fixes #3386 

- Add an `/api/2/logs` route and deprecate the previous version (which remains in operation under the current route `/api/logs` for backwards compatibility)
- Intentionally repeat the code verbatim (but without the unwanted filters) so that future modifications do not break third party uses of the v1 route.
- Use the v2 api for logs in the indexer client
- Always add the schemas table to the filters in `createStoreSync`, as this is no longer included by default in the indexer v2 api
- Remove redundant filter for schemas table in `getRecords`
- The only other instance of getting indexer logs that requires the schemas table—the public SQL API's `getSnapshot`—calls `createStoreSync` internally, so this can remain unchanged.

**TODO**
- [x] Build and tests pass locally
- [ ] Test with example apps, ensure tables are created in different storage types

### DIFF
--- a/packages/store-indexer/src/bin/postgres-frontend.ts
+++ b/packages/store-indexer/src/bin/postgres-frontend.ts
@@ -9,7 +9,8 @@ import { drizzle } from "drizzle-orm/postgres-js";
 import postgres from "postgres";
 import { frontendEnvSchema, parseEnv } from "./parseEnv";
 import { createQueryAdapter } from "../postgres/deprecated/createQueryAdapter";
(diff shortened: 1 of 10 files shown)


ludns:
Could you use v2 instead of 2 in the url? ie: `/api/v2/logs`

PR #3476, opened 2025-01-21
chore(cli): print mud version in deploy

### LINKED ISSUE #3341: print mud version during deploy
we often ask for this when debugging deploy issues and would be easier if it was just in the logs

JamesLefrere:
Fixes #3341

### DIFF
--- a/packages/cli/src/runDeploy.ts
+++ b/packages/cli/src/runDeploy.ts
@@ -1,5 +1,6 @@
 import path from "node:path";
 import { existsSync, mkdirSync, readFileSync, writeFileSync } from "node:fs";
+import packageJson from "../package.json";
 import { InferredOptionTypes, Options } from "yargs";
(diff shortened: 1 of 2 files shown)


packages/cli/src/runDeploy.ts
```diff
@@ -71,6 +73,10 @@ export async function runDeploy(opts: DeployOptions): Promise<WorldDeploy> {
   const config = (await loadConfig(configPath)) as WorldConfig;
   const rootDir = path.dirname(configPath);
 
+  if (opts.printVersion) {
```
alvrs, line 76: what would be a reason why someone would want to skip this log specifically? 🤔

JamesLefrere: Quite right, done 👍

holic (approved):

PR #3479, opened 2025-01-21
feat(create-mud): rip out create-create-app

holic:
I ripped out create-create-app because it's getting in the way of new templates and we were only using it to 1) recursively copy files and 2) inject the MUD version.


packages/create-mud/src/bin/cli.ts
```diff
@@ -0,0 +1,62 @@
...
+        const targetFile = path.join(targetDir, filename);
+
+        await fs.mkdir(path.dirname(targetFile), { recursive: true });
+
+        if (/package\.json$/.test(sourceFile)) {
+          const source = await fs.readFile(sourceFile, "utf-8");
+          await fs.writeFile(targetFile, source.replaceAll(/{{mud-version}}/g, args.mudVersion), "utf-8");
+        } else {
+          await fs.copyFile(sourceFile, targetFile);
+        }
```
holic, line 55: If we didn't have to do a string replace on specific files, this could be simplified to

```
fs.cp(sourceDir, targetDir, { recursive: true })
```

alvrs (dismissed):

PR #3484, opened 2025-01-22
feat(explorer): blocks API endpoint + prefetch transactions

karooolis:


ludns:
Are we going to ever merge this (more to calibrate myself as I really want this lol. Not meant as a passive aggressive question)? 
When looking at what builders do with Eve Frontier this would be a really useful tool. Combined with the ability for filtering by namespace and/or system, it would effectively give you a mini explorer for each third-party developer which would be awesome.

karooolis:
> Are we going to ever merge this (more to calibrate myself as I really want this lol. Not meant as a passive aggressive question)? When looking at what builders do with Eve Frontier this would be a really useful tool. Combined with the ability for filtering by namespace and/or system, it would effectively give you a mini explorer for each third-party developer which would be awesome.

Would love to have this merged too. It relies on a running Shovel instance, I'll see what can be done to get it synced. And was thinking about filtering too but didn't come to that yet, will add!

PR #3485, opened 2025-01-22
feat(create-mud): new react-ecs template

holic:
based on the new react template from https://github.com/latticexyz/mud/pull/3478

PR #3486, opened 2025-01-22
feat(store-sync): recs sync adapter

holic:
pulled out of https://github.com/latticexyz/mud/pull/3485

- consolidated how we derive components from config and tables
- updated storage adapter to lean on components already registered in RECS world
- lifted the registration of components out of storage adapter, so they're accessible outside async context


packages/store-sync/src/recs/defineInternalComponents.ts
```diff
@@ -5,6 +5,7 @@ export type InternalComponents = ReturnType<typeof defineInternalComponents>;
 
 export function defineInternalComponents(world: World) {
   return {
+    /** @deprecated */
     RegisteredTables: defineComponent<{ table: Type.T }, Metadata, Table>(
```
holic, line 9: was torn on what to do with this! this is no longer needed as part of the storage adapter because we now use the `component.metadata.table` for things like schemas

I removed the populating of this table/component from the storage adapter and deprecated this (rather than removed) because it used to be part of the `components` return type of `syncToRecs`

I could just remove because it was considered "internal" and doesn't appear to be used: https://github.com/search?q=RegisteredTables&type=repositories

thoughts?

alvrs: i would be fine with removing

alvrs (approved):

PR #3491, opened 2025-01-23
feat: move CallWithSignature module to its own package

holic:
Moves `Unstable_CallWithSignatureModule` into its own package (for easier auditing and isolation) and installs it by default during deploy.

For users with the old module in their config, they'll see this message:
image


packages/cli/src/deploy/compat/excludeUnstableCallWithSignatureModule.ts
```diff
@@ -0,0 +1,19 @@
...
+  ) {
+    console.warn(
+      [
+        "",
+        `⚠️ Your \`mud.config.ts\` is using \`Unstable_CallWithSignatureModule\`. This module can be removed from your config as it is now installed by default during deploy.`,
+        "",
+      ].join("\n"),
+    );
+    return false;
+  }
```
alvrs, line 17: :chefkiss:

alvrs (dismissed):

PR #3492, opened 2025-01-23
fix(create-mud): ensure gitignore is copied

holic:
npm excludes `.gitignore` when packaging and publishing, so although these were getting copied with build, they were missing from npm artifacts:
https://www.npmjs.com/package/create-mud/v/2.2.17-86f455c0a09d5d8af64d030669ebf30ffa71420b?activeTab=code

we resolve this by moving the file aside and moving it back when creating the template

no changeset because these new templates aren't published yet and this fix is captured under existing changesets


vdrg (approved):
lgtm!

PR #3496, opened 2025-01-23
feat(world): add CallWithSignatureSystem

### LINKED ISSUE #3437: finalize CallWithSignature
- [ ] audit module
- [ ] consider making it a default module in MUD as it's needed by entrykit

alvrs:
closes #3437


packages/world/src/modules/init/implementations/CallWithSignatureSystem/ECDSA.sol
```diff
@@ -0,0 +1,174 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts (last updated v5.0.0) (utils/cryptography/ECDSA.sol)
```
holic, line 2: did we decide it was better to copy/paste these here instead of importing from OZ as a dep?

packages/world/src/modules/init/implementations/CallWithSignatureSystem/getSignedMessageHash.sol
```diff
@@ -0,0 +1,53 @@
...
+    keccak256(
+      abi.encodePacked(
+        "\x19\x01",
+        domainSeperator,
+        keccak256(
+          abi.encode(
+            CALL_TYPEHASH,
+            signer,
+            keccak256(abi.encodePacked(WorldResourceIdLib.toTrimmedString(systemId.getNamespace()))),
+            keccak256(abi.encodePacked(WorldResourceIdLib.toTrimmedString(systemId.getName()))),
```
holic, line 46: any idea why we `abi.encodePacked` instead of casting to `bytes`?

packages/world/src/modules/init/implementations/CallWithSignatureSystem/ECDSA.sol
alvrs: I think the original reason was that the OZ package was a pretty heavy dependency and we only needed very few files from it. At this point we do have a couple vendored OZ files in the world package (`ERC165Checker`, `IERC1967`, `Proxy`, `StorageSlot`) but I don't think it's worth changing our approach now. MUD is audited as a whole and the auditors can easily check that the files match the original sources. Changing paths of a Solidity package is always annoying because it would break things downstream if the files are imported since there are no explicit exports/remappings.

packages/world/src/modules/init/constants.sol
```diff
@@ -37,3 +37,11 @@ ResourceId constant BATCH_CALL_SYSTEM_ID = ResourceId.wrap(
 ResourceId constant REGISTRATION_SYSTEM_ID = ResourceId.wrap(
   bytes32(abi.encodePacked(RESOURCE_SYSTEM, ROOT_NAMESPACE, bytes16("Registration")))
 );
+
+/**
+ * @dev Resource ID for the call with signature system.
+ * @dev This ID is derived from the RESOURCE_SYSTEM type, the ROOT_NAMESPACE, and the system name.
+ */
+ResourceId constant CALL_WITH_SIGNATURE_SYSTEM_ID = ResourceId.wrap(
+  (bytes32(abi.encodePacked(RESOURCE_SYSTEM, ROOT_NAMESPACE, "CallWithSignature")))
```
holic, line 46:
```suggestion
  (bytes32(abi.encodePacked(RESOURCE_SYSTEM, ROOT_NAMESPACE, bytes16("CallWithSignatur"))))
```
`bytes16` to keep us honest

and snipped off the last char because the full name didn't fit into bytes16

packages/world/src/modules/init/implementations/CallWithSignatureSystem/ECDSA.sol
holic: I _hope_ no one has imported these directly 🙈 wonder if we should start putting not-for-external-use source files into `internal` dirs, like we've started doing for TS exports

alvrs:

PR #3503, opened 2025-01-24
fix(cli): use system label when verifying

jaxernst:
Usage of system.name can result in an incorrectly abbreviated contract path. Use system.label to ensure the full path is correctly resolved.

### DIFF
--- a/.changeset/silly-kids-enjoy.md
+++ b/.changeset/silly-kids-enjoy.md
@@ -1,0 +1,5 @@
+---
+"@latticexyz/cli": patch
+---
+
+Fixed an issue where 'mud verify' will crash after failing to resolve the full contract artifact paths.
--- a/.changeset/silly-kids-enjoy.md
+++ b/.changeset/silly-kids-enjoy.md
@@ -2,4 +2,4 @@
 "@latticexyz/cli": patch
 ---
 
-Fixed an issue where 'mud verify' will crash after failing to resolve the full contract artifact paths.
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/verify.ts
```diff
@@ -72,7 +72,7 @@ const commandModule: CommandModule<Options, Options> = {
     // TODO: replace with `resolveConfig` and support for linked libs
     const configSystems = await resolveSystems({ rootDir, config });
     const systems = configSystems.map((system) => {
-      const contractData = getContractData(`${system.name}.sol`, system.name, outDir);
+      const contractData = getContractData(`${system.label}.sol`, system.label, outDir);
       return {
         name: system.name,
```
holic, line 77:
```suggestion
        name: system.label,
```

.changeset/silly-kids-enjoy.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/cli": patch
+---
+
+Fixed an issue where 'mud verify' will crash after failing to resolve the full contract artifact paths.
```
holic, line 5: Fixed an issue with `mud verify` where system contract artifacts were being resolved incorrectly.

holic, line 5:
```suggestion
Fixed an issue with `mud verify` where system contract artifacts were being resolved incorrectly.
```

PR #3507, opened 2025-01-24
fix(protocol-parser): add padding for encodedLengths

alvrs:


packages/store-sync/src/stash/createStorageAdapter.ts
holic, line 1: do we need this? feels a bit of a one-off, like why aren't we adding this to all the adapters?

packages/protocol-parser/src/decodeValueArgs.ts
```diff
@@ -13,7 +13,7 @@ export function decodeValueArgs<TSchema extends ValueSchema>(
     valueSchema,
     concatHex([
       readHex(staticData, 0, staticDataLength(Object.values(valueSchema).filter(isStaticAbiType))),
-      encodedLengths,
+      padHex(encodedLengths, { size: 32 }),
```
holic, line 16: do we wanna add a test somewhere that captures the before/after of this change to demonstrate the bug and the fix?

holic (approved):

.changeset/shaggy-shirts-work.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/protocol-parser": patch
+---
+
+Added padding to make sure the `encodedLength` field of an store log is always 32 bytes.
```
holic, line 5:
```suggestion
Added padding to make sure the `encodedLengths` field of an store log is always 32 bytes.
```

Issue #3508, opened 2025-01-24
local explorer should set sqlite indexer port


holic:
and prob have a separate flag + env var for it, otherwise it defaults to `PORT` which explorer bin itself already uses

PR #3511, opened 2025-01-24
fix(explorer): create a separate flag and port for the indexer created by the explorer

dhvanipa:
Solves #3508


packages/explorer/src/bin/explorer.ts
```diff
@@ -20,6 +20,12 @@ const argv = yargs(process.argv.slice(2))
       type: "number",
       default: process.env.PORT || 13690,
     },
+    indexerPort: {
+      alias: "pI",
```
karooolis, line 24:
```suggestion
      alias: "pi",
```

karooolis (approved):

PR #3514, opened 2025-01-26
fix(cli): support verifying systems with linked libraries

jaxernst:
The `verify` command currently looks up system addresses using the bytecode artifact without considering library link references. This fix uses the same lookup logic as the `deploy` command to inject these references into the bytecode used for create2 address lookup.

### DIFF
--- a/.changeset/famous-doors-listen.md
+++ b/.changeset/famous-doors-listen.md
@@ -2,4 +2,4 @@
 "@latticexyz/cli": patch
 ---
 
-The `verify` cli command will be able to verify systems with library link references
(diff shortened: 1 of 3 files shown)


.changeset/famous-doors-listen.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/cli": patch
+---
+
+The `verify` cli command will be able to verify systems with library link references
```
holic, line 5:
```suggestion
The `verify` command should now be able to correctly verify systems using public libraries.
```

holic (approved):

holic:
great find and great PR, thank you!

Issue #3516, opened 2025-01-27
Incorrect transaction status in `waitForTransaction`


Wertusser:
Transaction status differs from `waitForTransaction`(pic2) and Explorer's status (pic1)

<img width="1367" alt="Image" src="https://github.com/user-attachments/assets/eeafc2b2-0738-43fd-8a19-3d0e942cc62c" />

![Image](https://github.com/user-attachments/assets/d0ea2900-46da-4dd3-a6d6-6b25b0a0ad77)

Seems like `waitForTransaction` always return `success` status if transaction is included in block, even if internal `UserOp` was failed to execute.

Example of reverted user op:
<img width="783" alt="Image" src="https://github.com/user-attachments/assets/44853409-7956-42cf-85bf-351dedf2ba16" />

Wertusser:
Missing `waitForUserOperationReceipt` check here

https://github.com/latticexyz/mud/blob/452d3e53db2768f2e296ea7560c9bdd6560cd797/packages/store-sync/src/createStoreSync.ts#L348-L351

PR #3518, opened 2025-01-28
fix(store-sync): handle user ops in waitForTransaction

holic:
fixes https://github.com/latticexyz/mud/issues/3516

I don't love that we're kinda lying about the `status` of a transaction in the return value of this, but given that it's our own `WaitForTransactionResult`, this seems fine?

### DIFF
--- a/packages/store-sync/src/createStoreSync.ts
+++ b/packages/store-sync/src/createStoreSync.ts
@@ -1,5 +1,6 @@
 import { storeEventsAbi } from "@latticexyz/store";
-import { GetTransactionReceiptErrorType, Hex } from "viem";
+import { GetTransactionReceiptErrorType, Hex, parseEventLogs } from "viem";
+import { entryPoint07Abi } from "viem/account-abstraction";
 import {
   StorageAdapter,
   StorageAdapterBlock,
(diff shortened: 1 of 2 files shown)


packages/store-sync/src/createStoreSync.ts
```diff
@@ -344,23 +346,49 @@ export async function createStoreSync({
...
-          const { status, blockNumber, transactionHash } = await getAction(
-            publicClient,
-            getTransactionReceipt,
-            "getTransactionReceipt",
-          )({ hash: tx });
-          if (lastBlock.blockNumber >= blockNumber) {
-            return { status, blockNumber, transactionHash };
+          const receipt = await getAction(publicClient, getTransactionReceipt, "getTransactionReceipt")({ hash: tx });
+
+          if (lastBlock.blockNumber >= receipt.blockNumber) {
```
alvrs, line 360: why do we need this check? i see we also had it before, before we were not returning anything if the check didn't pass, now it seems like we're still returning the receipt in that case (but are not performing the user op checks)?

alvrs (approved):

packages/store-sync/src/createStoreSync.ts
```diff
@@ -339,29 +341,56 @@ export async function createStoreSync({
...
+            if (userOpEvents.length > 1) {
+              const issueLink = `https://github.com/latticexyz/mud/issues/new${new URLSearchParams({
+                title: "waitForTransaction found more than one user op",
+                body: `MUD version: ${packageJson.version}\nChain ID: ${chainId}\nTransaction: ${tx}\n`,
+              })}`;
+              console.warn(
+                // eslint-disable-next-line max-len
+                `Receipt for transaction ${tx} had more than one \`UserOperationEvent\`. This may have unexpected behavior.\n\nIf you encounter this, please open an issue:\n${issueLink}`,
+              );
+            }
```
alvrs, line 380: :chefskiss

PR #3521, opened 2025-01-28
fix(store): prevent delete out of bounds in `spliceDynamicData`

alvrs:
In `spliceDynamicData`, we’re checking that `start` is within the bounds of the previous field length but aren’t considering `deleteCount` in that check. There is another check that checks that `start + deleteCount` lines up with the previous length of the field if the total length of the field changed, but this only applies if the length changed. That means if the length of the data to insert is the same as `deleteCount`, it is possible to “insert data after the length of the field” (ie by setting `start` to the end of the field). I put “insert data after the length of the field” in quotes, since the length of the field is not actually changed, which means when retrieving the whole field onchain the data appended at the end would not be included, similar to how items that are pop’ed from a dynamic field are not actually cleared from storage but just the field length is reduced.

But means indexers/clients need to be aware of this nuance and use `encodedLengths` as source of truth (like we do onchain).

We can remove this edge case by changing the check to `if(startWithinField > previousFieldLength - deleteCount)`.

When using our table libraries this does not happen since they don't call `spliceDynamicData` with an invalid `start` value, but it’s possible to trigger this by calling `world.spliceDynamicData` manually.


alvrs:

PR #3523, opened 2025-01-28
feat(common): export getFeeRef

dhvanipa:
`getFeeRef` handles auto refreshing the latest fees so should be exposed to clients so they can also make use of them.

For example, in Biomes we set the priority fee based on a base fee threshold.


holic:
closing for now as we don't wanna expose this, will revisit with an updated approach

PR #3527, opened 2025-01-29
feat(world): strip world prefix from function names when encoding system calls

### LINKED ISSUE #2463: Function Selector not Rerouted in Namespaced Systems
Using the encodeSystemCall helper function doesn't work in namespaced functions when passing in the World ABI as the abi parameter. This is because the function selector in the system (SpawnSystem.spawn.selector) doesn't match the one used by viem.encodeFunctionData (IWorld.Primodium__spawn). This makes any call or batch call revert.

holic:
fixes #2463

### DIFF
--- a/packages/world/ts/encodeSystemCall.ts
+++ b/packages/world/ts/encodeSystemCall.ts
@@ -1,13 +1,13 @@
 import { Abi, EncodeFunctionDataParameters, Hex, encodeFunctionData, type ContractFunctionName } from "viem";
 import type { AbiParametersToPrimitiveTypes, ExtractAbiFunction } from "abitype";
 import { worldCallAbi } from "./worldCallAbi";
+import { internal_normalizeSystemFunctionName } from "./normalizeSystemFunctionName";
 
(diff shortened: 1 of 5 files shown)


packages/world/ts/encodeSystemCall.ts
```diff
@@ -1,13 +1,22 @@
...
+  readonly abi: abi;
+  /**
+   * System's resource ID
+   */
   readonly systemId: Hex;
-};
+  /**
+   * System function name to call
+   */
+  readonly functionName: functionName;
```
holic, line 18: pulled these out for stronger types in the function body and so we can add TSDoc to clarify the `abi` param

packages/world/ts/normalizeSystemFunctionName.ts
```diff
@@ -0,0 +1,16 @@
+import { hexToResource, resourceToLabel } from "@latticexyz/common";
+import { Hex } from "viem";
+
+/** @internal */
+export function internal_normalizeSystemFunctionName(systemId: Hex, functionName: string) {
```
holic, line 5: I don't wanna propagate the usage of this so I prefixed this with `internal_` to make it clear

alvrs (approved):

PR #3529, opened 2025-01-29
feat(entrykit): deploy prereqs to any chain

holic:
```sh
FOUNDRY_PROFILE=garnet pnpm entrykit-deploy
#
RPC_HTTP_URL=http://rpc.garnetchain.com pnpm entrykit-deploy
```


packages/cli/src/build.ts
```diff
@@ -15,11 +15,7 @@ type BuildOptions = {
   config: WorldConfig;
 };
 
-export async function build({
-  rootDir,
-  config,
-  foundryProfile = process.env.FOUNDRY_PROFILE,
```
holic, line 21: this automatically defaults in foundry

PR #3115, opened 2024-09-01
fix(store-sync): handle TransactionReceiptNotFoundError

CryptoSpaces:
fixes https://github.com/latticexyz/mud/issues/1833

Issue #1833, opened 2023-10-27
improve `TransactionReceiptNotFoundError` error catching in `waitForTransaction`


holic:
https://github.com/latticexyz/mud/blob/f6d214e3d79f9591fddd3687aa987a57f417256c/packages/store-sync/src/createStoreSync.ts#L293

This check can fail if the project is using a different viem version from the MUD packages (see also #1832). For this particular check, we should consider other approaches (e.g. constructor name) that are more resilient to viem version misalignment.

holic:
Sounds like we should check `error.name` instead: https://viem.sh/docs/error-handling.html

holic:
To workaround this, make sure your viem version is aligned with the viem version used by your MUD version. You can check your `pnpm-lock.yaml` and if you see different versions of viem in there, that's probably why you see this error.

PR #3534, opened 2025-01-29
fix(common): improve nonce error handling

dhvanipa:
Similar to https://github.com/latticexyz/mud/pull/3115 except for `viem` nonce errors

Related issue: https://github.com/latticexyz/mud/issues/1833

### DIFF
--- a/packages/common/src/createNonceManager.ts
+++ b/packages/common/src/createNonceManager.ts
@@ -77,7 +77,8 @@ export function createNonceManager({
 
   function shouldResetNonce(error: Error): boolean {
     return (
-      error.name === "BaseError" &&
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      typeof (error as any).walk === "function" && // Check if it has a `walk` method (indicating it's a `BaseError`)
       (error as BaseError).walk(
         (e) => e instanceof Error && (e.name === "NonceTooLowError" || e.name === "NonceTooHighError"),
       ) != null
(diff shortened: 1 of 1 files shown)


packages/common/src/createNonceManager.ts
```diff
@@ -75,11 +75,8 @@ export function createNonceManager({
     await ref.noncePromise;
   }
 
-  function shouldResetNonce(error: unknown): boolean {
-    return (
-      error instanceof BaseError &&
-      error.walk((e) => e instanceof NonceTooLowError || e instanceof NonceTooHighError) != null
-    );
+  function shouldResetNonce(error: Error): boolean {
+    return error.name === "NonceTooLowError" || error.name === "NonceTooHighError";
```
holic, line 79: did you test this? I think we still need to use `BaseError` and `error.walk` because these nonce errors are deeply nested

dhvanipa: ahh right yes, need to `walk`. Updated!

packages/common/src/createNonceManager.ts
```diff
@@ -75,10 +75,13 @@ export function createNonceManager({
     await ref.noncePromise;
   }
 
-  function shouldResetNonce(error: unknown): boolean {
+  function shouldResetNonce(error: Error): boolean {
     return (
-      error instanceof BaseError &&
-      error.walk((e) => e instanceof NonceTooLowError || e instanceof NonceTooHighError) != null
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      typeof (error as any).walk === "function" && // Check if it has a `walk` method (indicating it's a `BaseError`)
```
dhvanipa, line 81: Not sure if there's a better way to check if the error is a `BaseError`

packages/common/src/createNonceManager.ts
dhvanipa: Tested locally by setting nonce + 1 and it works now

PR #3535, opened 2025-01-29
fix(block-logs-stream): add missing check for validateBlockRange

dhvanipa:
There's a check for `internal_clientOptions` being defined but not if `validateBlockRange` is set to true in `getLogs`

### DIFF
--- a/packages/block-logs-stream/src/getLogs.ts
+++ b/packages/block-logs-stream/src/getLogs.ts
@@ -33,7 +33,7 @@ export async function getLogs(opts: GetLogsOptions): Promise<Log[]> {
     ],
   });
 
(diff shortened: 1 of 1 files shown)


packages/block-logs-stream/src/getLogs.ts
```diff
@@ -33,7 +33,7 @@ export async function getLogs(opts: GetLogsOptions): Promise<Log[]> {
     ],
   });
 
-  if (!opts.internal_clientOptions) {
+  if (!opts.internal_clientOptions || !opts.internal_clientOptions.validateBlockRange) {
```
holic, line 36:
```suggestion
  if (!opts.internal_clientOptions?.validateBlockRange) {
```

dhvanipa: cleannnnnnnnnn

PR #3544, opened 2025-01-30
refactor(world-module-erc20): change erc20 module table names to pascal case

vdrg:
This makes it consistent with other modules.

### DIFF
--- a/.changeset/ninety-beers-behave.md
+++ b/.changeset/ninety-beers-behave.md
@@ -2,4 +2,4 @@
 "@latticexyz/world-module-erc20": patch
 ---
 
-Changed erc20 module table names to pascal case for consistency.
+Updated table names to pascal case for consistency.
--- a/.changeset/ninety-beers-behave.md
+++ b/.changeset/ninety-beers-behave.md
@@ -1,0 +1,5 @@
+---
+"@latticexyz/world-module-erc20": patch
+---
+
+Changed erc20 module table names to pascal case for consistency.
--- a/packages/world-module-erc20/src/experimental/Constants.sol
+++ b/packages/world-module-erc20/src/experimental/Constants.sol
@@ -6,7 +6,7 @@ import { ResourceId, WorldResourceIdLib } from "@latticexyz/world/src/WorldResou
 
 library ModuleConstants {
   bytes14 internal constant NAMESPACE = "erc20-module";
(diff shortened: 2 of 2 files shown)


packages/world-module-erc20/src/experimental/Constants.sol
```diff
@@ -18,19 +18,19 @@ library ModuleConstants {
 }
 
 library ERC20TableNames {
-  bytes16 internal constant TOTAL_SUPPLY = "TOTAL_SUPPLY";
+  bytes16 internal constant TOTAL_SUPPLY = "TotalSupply";
```
holic, line 21: thoughts on `ERC20TotalSupply` etc? thinking that maybe we'll add ERC721 tables and name conflicts would be sad, esp if the schemas need to be different for some reason

vdrg: We can, though they will be in separate namespaces, I'm not that familiar with how it would be visualized in the explorer. There will still be conflicts with other things like "Paused".

holic: is it required that there be one namespace per token? multiple tokens can't live in the same namespace?

vdrg: yes, each token registers its own tables and namespace

holic: so just to confirm, this would fail?
```ts
export default defineWorld({
  ...
  modules: [
    defineERC20Module({ namespace: "omg", name: "omg token", ... }),
    defineERC20Module({ namespace: "omg", name: "omg another token", ... }),
  ],
});
```

vdrg: yeah that should fail. Regarding prefixing: I think that would make the implementation more complex and also make it harder to visualize/query no? instead of just using the specific token namespace. Also having in separate namespaces makes it easier to isolate IMO, and you can still grant access to the namespace to system in other namespaces.

.changeset/ninety-beers-behave.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/world-module-erc20": patch
+---
+
+Changed erc20 module table names to pascal case for consistency.
```
holic, line 5:
```suggestion
Updated table names to pascal case for consistency.
```

holic (dismissed):

PR #3546, opened 2025-01-30
refactor(store-consumer): adapt WithWorld to be a System

vdrg:
Now `WithWorld` is also a System, and we modify the `onlyNamespace` modifier so that protected functions can only be called by calling the system through the world.


packages/store-consumer/src/experimental/Context.sol
```diff
@@ -4,7 +4,7 @@ pragma solidity >=0.8.24;
 // @dev Provides information about the current execution context
 // We only use it in these contracts in case we want to extend it in the future
 abstract contract Context {
-  function _msgSender() internal view virtual returns (address) {
+  function _msgSender() public view virtual returns (address) {
```
holic, line 7: ooc why public for such a small thing?

vdrg: because `_msgSender` in `WorldContext.sol` is public so we need to be able to override

vdrg:
> This means you can use stuff like `batchCall` and account delegation all works right? Since it'd be a system instead of a standalone contract?

Yeah I think those should work, but also so that you can restrict access to systems within the same namespace.

packages/store-consumer/src/experimental/Context.sol
alvrs: for context: it's public to have a unique interface id we can check for in `supportsInterface` so only contracts that are intended to be used as systems can be registered as systems

packages/store-consumer/src/experimental/WithWorld.sol
```diff
@@ -6,20 +6,31 @@ import { ResourceId } from "@latticexyz/store/src/ResourceId.sol";
...
+      revert WithWorld_CallerIsNotWorld();
+    }
+    _;
+  }
 
   modifier onlyNamespace() {
-    address sender = _msgSender();
-    if (!ResourceAccess.get(getNamespaceId(), sender)) {
+    // We use WorldContextConsumer directly as we already know the world is the caller
+    if (!_callerIsWorld() || !ResourceAccess.get(getNamespaceId(), WorldContextConsumer._msgSender())) {
```
alvrs, line 33: so if my EOA has access to the namespace I still have to call the token system through the world correct?

packages/store-consumer/test/StoreConsumer.t.sol
```diff
@@ -77,23 +82,83 @@ contract StoreConsumerTest is Test, GasReporter {
...
+    MockWithWorld mock = new MockWithWorld(world, namespace, true);
+    mock.transferNamespaceOwnership(address(this));
+
+    // Register the mock as a system with PRIVATE access
+    world.registerSystem(systemId, mock, false);
+
+    address alice = address(0x1234);
+
+    vm.prank(alice);
+    vm.expectRevert();
```
alvrs, line 103: should we check for the specific revert we're expecting here? (I guess `World_AccessDenied`?)

packages/store-consumer/test/StoreConsumer.t.sol
```diff
@@ -77,23 +82,83 @@ contract StoreConsumerTest is Test, GasReporter {
...
+    MockWithWorld mock = new MockWithWorld(world, namespace, true);
+    mock.transferNamespaceOwnership(address(this));
+
+    // Register the mock as a system with PUBLIC access
+    world.registerSystem(systemId, mock, true);
+
+    address alice = address(0x1234);
+
+    vm.prank(alice);
+    vm.expectRevert();
```
alvrs, line 128: same here, should we check the specific revert we're expecting here?

packages/store-consumer/test/StoreConsumer.t.sol
```diff
@@ -77,23 +82,83 @@ contract StoreConsumerTest is Test, GasReporter {
...
+    world.registerSystem(systemId, mock, true);
+
     address alice = address(0x1234);
 
     vm.prank(alice);
     vm.expectRevert();
     mock.onlyCallableByNamespace();
 
+    vm.prank(alice);
+    vm.expectRevert();
```
alvrs, line 156: same here

packages/store-consumer/test/StoreConsumer.t.sol
vdrg: yep, just wanted to confirm we like this approach before polishing the final details

packages/store/ts/getStoreLogs.test.ts
```diff
@@ -158,8 +158,8 @@ describe("getStoreLogs", async () => {
         "Store_SpliceStaticData  store__ResourceIds  (0x746200000000000000000000000000005465727261696e000000000000000000)",
         "Store_SpliceStaticData  store__ResourceIds  (0x737900000000000000000000000000004d6f766553797374656d000000000000)",
         "Store_SetRecord  world__Systems  (0x737900000000000000000000000000004d6f766553797374656d000000000000)",
-        "Store_SpliceStaticData  world__SystemRegistry  (0x00000000000000000000000008f2b45d8787be8a81869d9968f25323861352b0)",
-        "Store_SpliceStaticData  world__ResourceAccess  (0x6e73000000000000000000000000000000000000000000000000000000000000,0x00000000000000000000000008f2b45d8787be8a81869d9968f25323861352b0)",
+        "Store_SpliceStaticData  world__SystemRegistry  (0x0000000000000000000000006eb355773196079fe643ed78724140adb1f86c11)",
+        "Store_SpliceStaticData  world__ResourceAccess  (0x6e73000000000000000000000000000000000000000000000000000000000000,0x0000000000000000000000006eb355773196079fe643ed78724140adb1f86c11)",
```
alvrs, line 162: do you know why this changed? 🤔

alvrs (approved):

packages/store/ts/getStoreLogs.test.ts
vdrg: I assumed it was due to a rebase, will rebase again as I believe @holic updated the snapshot recently as well

PR #3562, opened 2025-02-04
fix(store-sync): skip invalid utf-8 characters in strings before inserting into postgres

alvrs:
Currently an invalid UTF-8 character in a `string` type column makes the decoded postgres indexer error with `PostgresError: invalid byte sequence for encoding "UTF8": 0x00`. This PR adds a cleanup step before upserting values into postgres to remove `0x00` characters.

### DIFF
--- a/packages/store-sync/src/postgres-decoded/cleanStrings.ts
+++ b/packages/store-sync/src/postgres-decoded/cleanStrings.ts
@@ -1,0 +1,10 @@
+import { SchemaToPrimitives, ValueSchema } from "@latticexyz/protocol-parser/internal";
+
+export function cleanStrings<TSchema extends ValueSchema>(
+  schema: TSchema,
+  value: SchemaToPrimitives<TSchema>,
+): SchemaToPrimitives<TSchema> {
+  return Object.fromEntries(
+    Object.entries(value).map(([key, value]) => [key, schema[key] === "string" ? value.replaceAll("\x00", "") : value]),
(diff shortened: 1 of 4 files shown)


holic:
wouldn't this corrupt the data?

alvrs:
> wouldn't this corrupt the data if a \0 is in the middle of the string somewhere?

Postgres doesn't support `\x00` bytes in strings, so the only alternative to removing the bytes here is skipping the entire record or throwing an error, both of which are worse my opinion (especially because this is often user provided data, so a user could make the indexer break for everyone by providing a malformed string).

Seems like the discussion in https://github.com/ponder-sh/ponder/issues/1456 also tends towards this behaviour as default. I don't think we need to add a flag to disable this behaviour until someone requests it.

packages/store-sync/src/postgres-decoded/cleanStrings.ts
```diff
@@ -0,0 +1,10 @@
+import { SchemaToPrimitives, ValueSchema } from "@latticexyz/protocol-parser/internal";
+
+export function cleanStrings<TSchema extends ValueSchema>(
```
holic, line 3: I like Ponder's naming
```suggestion
export function removeNullCharacters<TSchema extends ValueSchema>(
```

holic:
ohh I was worried that these fields were being used to re-encode the data before mutating via event, but I don't think that's a concern here because we still store the raw bytes and use that as the source of truth

holic (approved):

PR #3581, opened 2025-02-06
refactor(world): add default-reverting methods to Module

holic:
mostly a clean up thing that has been bugging me

this means module authors don't need to implement this revert on their own, they just have to implement the install method that is appropriate


vdrg (dismissed):
nice! changeset missing

alvrs (approved):

Issue #3560, opened 2025-02-02
`mud test`/`mud deploy` swallows forge output


holic:
run `mud test` on a project with failing tests using foundry nightly

```
% pnpm mud test

...
running "forge test --fork-url http://127.0.0.1:4242"
Error: Warning: This is a nightly build of Foundry. It is recommended to use the latest stable version. Visit https://book.getfoundry.sh/announcements for more information.
To mute this warning set `FOUNDRY_DISABLE_NIGHTLY_WARNING` in your environment.

Error running "forge test --fork-url http://127.0.0.1:4242"
    [stacktrace]
 ELIFECYCLE  Test failed. See above for more details.
```

```
% FOUNDRY_DISABLE_NIGHTLY_WARNING=1 pnpm mud test

...
running "forge test --fork-url http://127.0.0.1:4242"
Error: Command failed with exit code 1: forge test --fork-url 'http://127.0.0.1:4242'

No files changed, compilation skipped

Ran 1 test for src/World.t.sol:WorldTest
[PASS] testWorldDeployed() (gas: 4936)
Suite result: ok. 1 passed; 0 failed; 0 skipped; finished in 1.93ms (348.29µs CPU time)

Ran 1 test for src/SettleSystem.t.sol:SettleSystemTest
[FAIL: revert: Not enough energy.] testSettle() (gas: 191765)
Suite result: FAILED. 0 passed; 1 failed; 0 skipped; finished in 5.41ms (3.83ms CPU time)

Ran 1 test for src/MoveSystem.t.sol:MoveSystemTest
[PASS] testMove() (gas: 251111)
Suite result: ok. 1 passed; 0 failed; 0 skipped; finished in 5.68ms (4.10ms CPU time)

Ran 3 test suites in 139.63ms (13.02ms CPU time): 2 tests passed, 1 failed, 0 skipped (3 total tests)

Failing tests:
Encountered 1 failing test in src/SettleSystem.t.sol:SettleSystemTest
[FAIL: revert: Not enough energy.] testSettle() (gas: 191765)

Encountered a total of 1 failing tests, 2 tests succeeded
Error running "forge test --fork-url http://127.0.0.1:4242"
    [stacktrace]
 ELIFECYCLE  Test failed. See above for more details.
```

in general, we should probably replace our `forge` wrapper with plain calls to `execa` and better piping to preserve terminal colors, etc

holic:
[...]

PR #3582, opened 2025-02-06
fix(cli): use execa directly instead of custom foundry wrappers

holic:
fixes https://github.com/latticexyz/mud/issues/3560

and improves output (color is preserved, commands are easier to identify)
image
image


vdrg (approved):
lgtm!

PR #3584, opened 2025-02-06
feat(world-consumer): convert store-consumer package into world-consumer

vdrg:
Simplifies the existing `store-consumer` package to just support consuming from a World. Updates the `world-module-erc20` contracts to inherit from the new `WorldConsumer` contract.


packages/world-consumer/src/experimental/WorldConsumer.sol
```diff
@@ -0,0 +1,107 @@
...
+
+abstract contract WorldConsumer is System {
+  bytes14 public immutable namespace;
+
+  error WithWorld_RootNamespaceNotAllowed();
+  error WithWorld_NamespaceAlreadyExists(bytes14 namespace);
+  error WithWorld_NamespaceDoesNotExists(bytes14 namespace);
+  error WithWorld_CallerHasNoNamespaceAccess(bytes14 namespace, address caller);
+  error WithWorld_CallerIsNotWorld(address caller);
+  error WithWorld_DirectEthNotAllowed();
```
holic, line 22:
```suggestion
  error WithWorld_ValueNotAllowed();
```

packages/world-consumer/src/experimental/WorldConsumer.sol
```diff
@@ -0,0 +1,107 @@
...
+import { ResourceAccess } from "@latticexyz/world/src/codegen/tables/ResourceAccess.sol";
+import { IBaseWorld } from "@latticexyz/world/src/codegen/interfaces/IBaseWorld.sol";
+import { WorldResourceIdLib } from "@latticexyz/world/src/WorldResourceId.sol";
+import { WorldContextConsumer } from "@latticexyz/world/src/WorldContext.sol";
+import { System } from "@latticexyz/world/src/System.sol";
+
+abstract contract WorldConsumer is System {
+  bytes14 public immutable namespace;
+
+  error WithWorld_RootNamespaceNotAllowed();
```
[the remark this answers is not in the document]
vdrg: yess still wip

packages/world-consumer/src/experimental/WorldConsumer.sol
vdrg: I like ValueNotAllowed better

packages/world-consumer/src/experimental/WorldConsumer.sol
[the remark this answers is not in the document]
holic: oops sorry for jumping in too early 🙈

.changeset/shy-terms-walk.md
```diff
@@ -0,0 +1,7 @@
+---
+"@latticexyz/world-consumer": patch
+"@latticexyz/world-module-erc20": patch
+---
+
+Renamed `store-consumer` package to `world-consumer`. The `world-consumer` package now only includes a single `WorldConsumer` contract that always attaches to a World.
+The `world-module-erc20` package was updated to reflect this refactor.
```
holic, line 7: I'd prob move this to its own changeset like
> Migrated from `store-consumer` to `world-consumer`.

packages/world-consumer/README.md
```diff
@@ -0,0 +1,8 @@
+# Store Consumer Contracts
```
holic, line 1:
```suggestion
# World Consumer
```

.changeset/shy-terms-walk.md
vdrg: done!

packages/world-consumer/src/experimental/WorldConsumer.sol
```diff
@@ -0,0 +1,107 @@
...
+
+abstract contract WorldConsumer is System {
+  bytes14 public immutable namespace;
+
+  error WorldConsumer_RootNamespaceNotAllowed();
+  error WorldConsumer_NamespaceAlreadyExists(bytes14 namespace);
+  error WorldConsumer_NamespaceDoesNotExists(bytes14 namespace);
+  error WorldConsumer_CallerHasNoNamespaceAccess(bytes14 namespace, address caller);
+  error WorldConsumer_CallerIsNotWorld(address caller);
+  error WorldConsumer_ValueNotAllowed();
```
holic, line 22: I wonder if its worth adding a `worldAddress` to each? might make it easier to see what world is trying to be called if you end up configuring or calling this wrong

I also wonder about `bytes14 namespace` vs `ResourceId namespaceId` since that's generally what we pass around inside world

vdrg: ah let me check other errors

vdrg: mm so we have:

```
  error World_AccessDenied(string resource, address caller);
  error World_InvalidNamespace(bytes14 namespace);
```

As the current error is just referring to a namespace, I feel it might be more consistent to keep it as just `bytes14` but not sure.

Adding the world address makes sense.

vdrg: Added it as a `IBaseWorld world`

holic: I think we can keep it `address world` to simplify if you wanted to avoid casting 

I also wonder if we should use `_world()` and `WorldContext._world()`
https://github.com/latticexyz/mud/blob/b7901812d4035faa3ec9bb75f31ffe7af398bdf2/packages/world/src/WorldContext.sol#L95-L97

Instead of reimplementing + renaming as `getWorld()`?

.changeset/shy-terms-walk.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/world-consumer": patch
+---
+
+Renamed `store-consumer` package to `world-consumer`. The `world-consumer` package now only includes a single `WorldConsumer` contract that always attaches to a World.
```
holic, line 5:
```suggestion
Renamed `store-consumer` package to `world-consumer`. The `world-consumer` package now only includes a single `WorldConsumer` contract that is bound to a `World`.
```

packages/world-consumer/src/experimental/WorldConsumer.sol
vdrg: ah good point! will do, this was from before using world context

packages/world-consumer/src/experimental/WorldConsumer.sol
```diff
@@ -0,0 +1,94 @@
...
+    return WorldContextConsumer._msgValue();
+  }
+
+  function _encodeTableId(bytes16 name) internal view returns (ResourceId) {
+    return WorldResourceIdLib.encode(RESOURCE_TABLE, namespace, name);
+  }
+
+  function _encodeOffchainTableId(bytes16 name) internal view returns (ResourceId) {
+    return WorldResourceIdLib.encode(RESOURCE_OFFCHAIN_TABLE, namespace, name);
+  }
```
holic, line 93: I don't know if I like the extra surface area to maintain with this. Seems simple enough to encode these as needed?

But I do think this points to a need to have more helpers in `WorldResourceIdLib` to derive various resource IDs based on an existing resource ID.

packages/world-consumer/src/experimental/WorldConsumer.sol
```diff
@@ -0,0 +1,94 @@
...
+      }
+      _world.registerNamespace(namespaceId);
+    } else if (!namespaceExists) {
+      revert WorldConsumer_NamespaceDoesNotExists(worldAddress, _namespace);
+    }
+  }
+
+  function getNamespaceId() public view returns (ResourceId) {
+    return WorldResourceIdLib.encodeNamespace(namespace);
+  }
```
holic, line 72: any benefit to a separate getter here vs storage?

```solidity
bytes14 public immutable namespace;
ResourceId public immutable namespaceId;

constructor(...) {
  ...
  namespace = _namespace;
  namespaceId = WorldResourceIdLib.encodeNamespace(_namespace);
}
```

alvrs (approved):

packages/world-consumer/src/experimental/WorldConsumer.sol
vdrg: no actually when using immutable it is not a storage read, it becomes part of the bytecode same as a constant. I guess I just didn't want to have redundant stuff that can be derived from existing data, but I think it makes sense from a cost perspective.

packages/world-consumer/src/experimental/WorldConsumer.sol
vdrg: done, will create an issue for better helpers

packages/world-consumer/src/experimental/WorldConsumer.sol
```diff
@@ -0,0 +1,82 @@
...
+import { ResourceIds } from "@latticexyz/store/src/codegen/tables/ResourceIds.sol";
+import { ResourceId } from "@latticexyz/store/src/ResourceId.sol";
+import { ResourceAccess } from "@latticexyz/world/src/codegen/tables/ResourceAccess.sol";
+import { IBaseWorld } from "@latticexyz/world/src/codegen/interfaces/IBaseWorld.sol";
+import { WorldResourceIdLib } from "@latticexyz/world/src/WorldResourceId.sol";
+import { WorldContextConsumer } from "@latticexyz/world/src/WorldContext.sol";
+import { System } from "@latticexyz/world/src/System.sol";
+
+abstract contract WorldConsumer is System {
+  bytes14 public immutable namespace;
```
holic, line 15: do we need this anymore if we can now do `namespaceId.getNamespace()`?

holic: oh does immutable not have the same read gas cost?

vdrg: yeah immutable actually works similar to constants, the data gets added to the bytecode so you are not reading from storage. That reminds me that immutable vars shouldn't be readable from within the constructor, so not sure why the tests are passing 🤔

holic: image

TIL how `immutable` works

vdrg: Ah since solidity 0.8.21 they can be read in the constructor, all good

holic (approved):

Issue #3569, opened 2025-02-05
add option to register delegation during module install


holic:
add something like `useDelegation: bool` to `ModuleInput` then, when deployer goes to install the module, it wraps the `install` or `installRoot` call in a `registerDelegation` and `unregisterDelegation`, giving the module permission to act on behalf of the caller
(would be cool to do this operation atomically with `batchCall` if possible)

this solves for issues mentioned in #3541 where modules truly become onchain scripts that can act on behalf of the caller, like creating namespaces or adding to existing namespaces owned by the caller

this also unlocks idempotent modules where a module can implement its own upgrade behavior, and we can prob move away from the "module already installed" approach we have now (https://github.com/latticexyz/mud/issues/2420, https://github.com/latticexyz/mud/issues/3035)

Issue #2957, opened 2024-07-18
revisit module args in config


holic:
they're currently hard to work with

PR #3586, opened 2025-02-06
feat: install module with delegation

holic:
fixes https://github.com/latticexyz/mud/issues/3569


packages/world-module-metadata/src/MetadataModule.sol
holic, line 1: Wait, what? How is the diff of this file so small? Aren't we using `callFrom` for every call now?

I'm glad you asked! Take a peek at `DelegatorContext` and how it's used here :)))

packages/store/ts/getStoreLogs.test.ts
holic, line 1: so cool to see the delegation in action via these store log changes for the mock game contracts

packages/cli/src/deploy/ensureResourceTags.ts
```diff
@@ -85,7 +83,7 @@ export async function ensureResourceTags<const value>({
       {
         optional: true,
         name: "MetadataModule",
-        installAsRoot: false,
+        installStrategy: "delegation",
```
alvrs, line 86: Do we want to install the metadata module via a delegation? (Do we want to give the installer ownership over the metadata tables or keep it like anyone can install it but it's immutable after it's been installed?)

packages/world/ts/config/v2/output.ts
```diff
@@ -4,13 +4,14 @@ import { DynamicResolution, ValueWithType } from "./dynamicResolution";
...
-   */
-  readonly name?: string;
   /** Should this module be installed as a root module? */
   readonly root: boolean;
+  /**
+   * Allow the module to call systems on your behalf.
+   * Useful when modules need to create namespaces or
+   * upgrade systems in namespaces you own.
+   */
+  readonly useDelegation: boolean;
```
alvrs, line 14: Do we throw an error/warning if both root and useDelegation are set to true?

holic: would be nice but we have no module config validation yet, nor super strong types, so didn't feel like adding it

what happens in this case is `root` is picked first, then `useDelegation`, then regular/default path

alvrs (approved):

packages/world/ts/config/v2/output.ts
holic: these config options will ideally be abstracted away into helpers provided by the module like `defineERC20Module`

packages/cli/src/deploy/ensureResourceTags.ts
holic: chatted in discord:

this doesn't change the current behavior, which passed namespace ownership back to the deployer: https://github.com/latticexyz/mud/pull/3586/files#diff-4b5b7ac0dc35d7914ce67b4dfcee8d54a4814b759a87d0ddce84efe17d3ade08L55

can separately decide what we wanna do going forward

packages/world/ts/config/v2/output.ts
holic: could make sense to use the same `installStrategy: "root" | "delegation" | "default"` approach in config but will tackle that later when we clean up module config (needs a rethink for args anyway: https://github.com/latticexyz/mud/issues/2957)

PR #3587, opened 2025-02-07
feat(world): generate system libs

holic:
pulled out of https://github.com/latticexyz/mud/pull/3586


packages/world/ts/node/render-solidity/renderWorldInterface.ts
```diff
@@ -20,7 +20,7 @@ export function renderWorldInterface({
   const imports = [
     ...(interfaceName === "IBaseWorld"
       ? [
-          { symbol: "IStore", path: `${storeImportPath}/IStore.sol` },
+          { symbol: "IStoreKernel", path: `${storeImportPath}/IStoreKernel.sol` },
```
alvrs, line 23: Wonder if this breaks functions that expect `IStore` as input, which `IBaseWorld` previously matched but now doesn't anymore? But not sure how common that is, if it happens would be easy enough to work around by casting to `IStore`. If CI is happy, I'm happy.

alvrs (approved):

packages/world/ts/node/render-solidity/renderWorldInterface.ts
holic: just discovered that an fixing it now

packages/world/src/modules/init/implementations/StoreRegistrationSystem.sol
```diff
@@ -44,8 +45,8 @@ contract StoreRegistrationSystem is System, IWorldErrors, LimitedCallContext {
     FieldLayout fieldLayout,
     Schema keySchema,
     Schema valueSchema,
-    string[] calldata keyNames,
-    string[] calldata fieldNames
```
vdrg, line 48: what if we instead change it at the system lib level so we can keep the existing systems as they are?

vdrg (commented):
lgtm in general but maybe I would keep calldata args and fix at the system lib gen

packages/world/src/modules/init/implementations/StoreRegistrationSystem.sol
holic: as in turn `calldata` into `memory` for just system lib methods?

holic: fwiw I saw no change in gas for this change, what's the concern here?

vdrg: yeah, gas improvements from using calldata will only happen if you don't turn those args into memory later, which most of these systems might be doing anyways. But imo it's better to have them in case a system would benefit from it in the future, also the api remains the same.

holic: done

.changeset/dry-camels-accept.md
```diff
@@ -0,0 +1,17 @@
+---
+"@latticexyz/world": patch
+---
+
+Added experimental system libraries for World systems for easier ergonomics when interacting with core systems.
```
vdrg, line 5:
```suggestion
Added experimental system libraries for World systems for better ergonomics when interacting with core systems.
```

not sure about this, english hard

packages/common/src/codegen/render-solidity/common.ts
```diff
@@ -22,8 +23,11 @@ export function renderList<T>(list: T[], renderItem: (item: T, index: number) =>
  * Renders a comma-separated list of arguments for solidity functions, ignoring empty and undefined ones
  */
 export function renderArguments(args: (string | undefined)[]): string {
-  const filteredArgs = args.filter((arg) => arg !== undefined && arg !== "") as string[];
-  return internalRenderList(",", filteredArgs, (arg) => arg);
+  return args
+    .filter(isDefined)
+    .filter((arg) => arg !== "")
+    .map((arg) => arg.replace(/ calldata /, " memory "))
```
vdrg, line 29: ah I actually meant doing it for the libraries only. Will this affect other codegen?

packages/world/src/modules/init/implementations/StoreRegistrationSystem.sol
```diff
@@ -25,7 +26,7 @@ import { WorldRegistrationSystem } from "./WorldRegistrationSystem.sol";
  * @author MUD (https://mud.dev) by Lattice (https://lattice.xyz)
  * @dev This contract provides functionality for the registration of store-related resources within the World framework.
  */
-contract StoreRegistrationSystem is System, IWorldErrors, LimitedCallContext {
+abstract contract StoreRegistrationSystem is System, IWorldErrors, IStoreRegistration, LimitedCallContext {
```
vdrg, line 29: wdyt of always inheriting from interfaces first? not really an issue in this case but feels cleaner to me in the sense that interfaces go before implementations (that might implement some of the interface functions).

vdrg (commented):
leaving an initial review with a few questions but will look in more detail later

packages/common/src/codegen/render-solidity/common.ts
holic: doesn't seem to, I think because we already use memory everywhere in tables

packages/world/src/modules/init/implementations/StoreRegistrationSystem.sol
holic: I don't feel strongly. Does it impact anything on the compiler side?

packages/common/src/codegen/render-solidity/common.ts
holic: ah, it did have an impact on other interfaces, so I updated to just do this in lib output

packages/world/src/modules/init/implementations/StoreRegistrationSystem.sol
vdrg: in this case it doesn't, but if one of the interfaces declares one of the functions that is present in the other contracts it would not compile unless you put the interface first, so I feel following the convention I mentioned makes sense:

Screenshot 2025-02-10 at 9 30 17 AM

holic: looks like we're fairly inconsistent with it so maybe we can do a clean up pass separately?

image

PR #3605, opened 2025-02-26
feat(entrykit): add react 19 and missing peer dep

frolic:


packages/entrykit/package.json
```diff
@@ -61,6 +61,7 @@
     "react-merge-refs": "^2.1.1",
     "tailwind-merge": "^1.12.0",
     "usehooks-ts": "^3.1.0",
+    "webauthn-p256": "0.0.10",
```
frolic, line 64: this is declared as a peer dep in permissionless so we add it here to avoid "missing peer dep" warnings

this is changed to an optional dep (and moved to ox) in https://github.com/pimlicolabs/permissionless.js/pull/373 so we can remove this next time we upgrade

PR #3611, opened 2025-02-28
feat(store): use custom error for deleteCount overflow

dk1a:
Calling `_spliceDynamicData` with an overflowing `deleteCount`, it triggers a default arithmetic error

If it instead triggered `Store_InvalidSplice`, it would've saved me a couple hours of debugging
I also add tests, for this new behaviour and for the previous way `Store_InvalidSplice` was used

My used of `unchecked` isn't important and I can remove it if you don't like it

I assume this is also blocked by audit, so I'm stockpiling PRs for whenever that happens


dk1a:
The arithemetic error I replace with a custom error comes from the definition of `previousFieldLength`, which is unchanged in  https://github.com/latticexyz/mud/pull/3521 
Essentially `Store_InvalidSplice` would be used for 2 errors
The second error alvarius changes is unrelated (except semantically), although I add a test for it here (as old behaviour that my PR shouldn't affect) so the test would need to be adjusted

PR #3613, opened 2025-03-02
docs: fix broken links

hunterfarq:
Hi! While reviewing the documentation, I noticed that two links to Foundry's scripting tutorials were broken. These links are important for developers who rely on Foundry's documentation to understand scripting and module installation.


frolic:
thank you!

PR #3619, opened 2025-03-05
feat(explorer): add pyrope testnet support

karooolis:


packages/common/src/chains/pyrope.ts
```diff
@@ -0,0 +1,48 @@
+import { chainConfig } from "viem/op-stack";
+import { MUDChain } from "./types";
+import { Chain } from "viem";
+
+const sourceId = 17000;
```
frolic, line 5:
```suggestion
const sourceId = 11155111;
```

.changeset/angry-otters-joke.md
```diff
@@ -0,0 +1,6 @@
+---
+"@latticexyz/common": patch
+"@latticexyz/explorer": patch
+---
+
+Added Pyrope Testnet chain configuration and support for Worlds Explorer.
```
frolic, line 6:
```suggestion
Added Pyrope chain.
```

this will show up contextually in each changeset

alternatively, could split them into separate changesets

karooolis: yeah makes sense, will split adding pyrope chain to common package as a separate PR

alvrs (dismissed):

PR #3621, opened 2025-03-07
docs: fixed typo in tutorial text

zkpepe:
Corrected a typo in the tutorial where the phrase "optimistically render user actions using overridable tables" should have been "optimistically **rendering** user actions using overridable tables." The form of the verb was missing.


frolic:
thank you!

PR #3632, opened 2025-03-18
feat(entrykit): validate signature

frolic:


packages/entrykit/src/validateSigner.ts
```diff
@@ -0,0 +1,59 @@
...
+}) {
+  const ownerAddress = await readContract(client, {
+    address: sessionAddress,
+    abi: simpleAccountAbi,
+    functionName: "owner",
+  });
+
+  if (ownerAddress.toLowerCase() !== signerAddress.toLowerCase()) {
+    throw new Error(`Session account owner (${ownerAddress}) does not match message signer (${signerAddress}).`);
+  }
```
alvrs, line 29: would it be useful to return something like `{ valid: boolean, message: string }` instead of throwing to make it easier to build UI based on this downstream? (still possible with this, but requires `try/catch`)

alvrs (approved):

packages/entrykit/src/validateSigner.ts
frolic: This is all internal still so we can change it in the future, but I'm kinda following Viem's approach of typed errors with nice data + messages, except skipping the "typed error" step and just throwing a message for now.

PR #3634, opened 2025-03-21
fix: remove wiresaw from pyrope chain definition

alvrs:
There is no wiresaw deployed to Pyrope yet, but the wiresaw rpc url in the chain config caused the sync stack to wait for wiresaw events.


frolic:
whoops! copy pasta mistake

PR #3642, opened 2025-03-25
feat(store-sync): system/world ABI from metadata

karooolis:


packages/store-sync/src/world/getSystemAbi.ts
```diff
@@ -0,0 +1,35 @@
...
+    chainId,
+    indexerUrl,
+    client,
+    fromBlock,
+    toBlock,
+  });
+
+  const abi = records
+    .filter(({ tag }) => hexToString(tag).replace(/\0+$/, "") === "abi")
+    .flatMap(({ value }) => (value === "0x" ? [] : hexToString(value).split("\n").map(parseAbiItem)));
```
frolic, line 32: according to docs, `parseAbiItem` accepts an array so this could be 
```suggestion
    .flatMap(({ value }) => (value === "0x" ? [] : parseAbiItem(hexToString(value).split("\n"))));
```

packages/store-sync/src/world/getSystemAbi.ts
```diff
@@ -0,0 +1,35 @@
...
+  toBlock,
+  indexerUrl,
+  chainId,
+}: {
+  readonly client: Client;
+  readonly worldAddress: Address;
+  readonly fromBlock?: bigint;
+  readonly toBlock?: bigint;
+  readonly indexerUrl?: string;
+  readonly chainId?: number;
```
frolic, line 18: do we want system ID as a param?

karooolis: what do you mean by "abi + system ID for each system ABI"? Kind of get it but not sure how this would look exactly. Anyhow, I'd be more inclined towards this returning the full system ABI so you don't have to fetch each one individually.

frolic: you're generally only interacting with one or a handful of systems, not the entire world's systems, esp as worlds get bigger, so I am more inclined to do individual ones (i.e. system ID param)

you always need a system ID in addition to the ABI + func name to call the system

karooolis: Yeah actually makes sense when put that way. Wonder if we should have another helper for querying the list of systems in that case?

karooolis: and I'll add `getSystemAbis()` as well since querying multiple systems could be PITA.

frolic: > Yeah actually makes sense when put that way. Wonder if we should have another helper for querying the list of systems in that case?

You already can by reading the world's system tables

packages/store-sync/src/world/getSystemAbi.ts
```diff
@@ -0,0 +1,37 @@
...
+    worldAddress,
+    chainId,
+    indexerUrl,
+    client,
+    fromBlock,
+    toBlock,
+  });
+
+  const abi = records
+    .filter(({ resource, tag }) => hexToString(tag).replace(/\0+$/, "") === "abi" && resource === systemId)
```
frolic, line 33: we have the whole key here now, so we don't need to fetch all records 

```ts
getRecord({
  table: metadataConfig.tables.metadata__ResourceTag,
  ...
  key: { resource: systemId, tag: stringToHexH("abi", { size: 32 }),
});
```

packages/store-sync/src/world/getSystemAbis.ts
```diff
@@ -0,0 +1,40 @@
...
+    fromBlock,
+    toBlock,
+  });
+
+  const abi = Object.fromEntries([
+    ...systemIds.map((id) => [id, []]),
+    ...records
+      .filter(({ resource, tag }) => hexToString(tag).replace(/\0+$/, "") === "abi" && systemIds.includes(resource))
+      .map(({ resource, value }) => [resource, value === "0x" ? [] : hexToString(value).split("\n").map(parseAbiItem)]),
+  ]) as Record<string, Abi>;
```
frolic, line 37:
```suggestion
  ]) as Record<Hex, Abi>;
```

packages/store-sync/src/world/getSystemAbis.ts
```diff
@@ -0,0 +1,40 @@
...
+  chainId,
+}: {
+  readonly client: Client;
+  readonly worldAddress: Address;
+  readonly systemIds: Hex[];
+  readonly fromBlock?: bigint;
+  readonly toBlock?: bigint;
+  readonly indexerUrl?: string;
+  readonly chainId?: number;
+}): Promise<Record<string, Abi>> {
```
frolic, line 21:
```suggestion
}): Promise<Record<Hex, Abi>> {
```
or better yet
```suggestion
}): Promise<{ readonly [systemId: Hex]: Abi }> {
```

packages/store-sync/src/world/getSystemAbis.ts
frolic, line 37:
```suggestion
      .map(({ resource, value }) => [resource, value === "0x" ? [] : hexToString(value).split("\n").map(parseAbiItem)] as const),
  ]);
```

packages/store-sync/src/world/getSystemAbi.ts
karooolis: yeah makes sense, although `getRecord()` does not exist in `store-sync` the same way `getRecords()` does. Wonder if I should create a new helper function for `getRecord()`, or what do you reckon?

karooolis: ah, looks like one from `@latticexyz/store` might work

packages/store-sync/src/world/getSystemAbis.ts
karooolis: still need to cast, will probe further into it

karooolis: that, and the empty array needs to be casted to `Abi` as well, now works nicely.

packages/store-sync/src/world/getSystemAbi.ts
karooolis: worked perfectly, great suggestion

packages/store-sync/src/world/getSystemAbi.ts
karooolis: Looks like there is a bug as it does not work as described in the docs, will file an issue.

packages/store-sync/src/world/index.ts
frolic, line 1: while we're here, can we move this to `packages/store-sync/src/exports/world.ts

karooolis: oh actually no, need to update imports now, on it

karooolis: Alright, done. And yeah, best to avoid barrel files definitely, makes sense.

packages/store-sync/src/world/getSystemAbi.ts
karooolis: Oh just realized what it actually does. It only parses one ABI item but it accepts an array in case the function signature contains a struct. In such case the parser will know how to interpret the struct as well. There might possible be more use cases as well.

frolic: ahh okay! I misunderstood the docs then 

looks like `parseAbi` does what I was hoping for

karooolis: ah that's right, adjusted

PR #3646, opened 2025-03-26
feat(explorer): use systems abis in decode form

karooolis:


packages/explorer/src/app/(explorer)/api/system-abis/route.ts
```diff
@@ -0,0 +1,42 @@
+import { Abi, Hex } from "viem";
+import { getSystemAbis } from "@latticexyz/store-sync/world";
```
frolic, line 2:
```suggestion
import { getSystemAbis } from "@latticexyz/store-sync/internal";
```

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/decode/DecodeForm.tsx
```diff
@@ -1,8 +1,9 @@
 "use client";
 
-import { AbiEvent, AbiFunction, toFunctionSelector } from "viem";
+import { AbiFunction, AbiItem, toFunctionSelector } from "viem";
 import { formatAbiItem } from "viem/utils";
 import * as z from "zod";
+import { AbiError } from "abitype";
```
Comment (frolic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/decode/DecodeForm.tsx:6]: does viem not export this?

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/decode/DecodeForm.tsx
```diff
@@ -22,36 +23,48 @@ import {
...
 
 const formSchema = z.object({
   selector: z.string().min(1).optional(),
 });
 
+function isFunction(item: AbiItem): item is AbiFunction {
+  return item.type === "function";
+}
+
+function isError(item: AbiItem): item is AbiError {
```
Comment (frolic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/decode/DecodeForm.tsx:38]: if you wanted to avoid the `AbiError` import, you could also narrow 
```suggestion
function isError(item: AbiItem): item is AbiItem & { type: "error" } {
```

Comment (frolic) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/decode/DecodeForm.tsx:38]: maybe worth clarifying the names of these `isAbiError` and `isAbiFunction`? since JS has errors/functions and is a bit confusing

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/decode/DecodeForm.tsx
Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/decode/DecodeForm.tsx:6]: looks like no weirdly enough, cannot import it unlike other types.

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/decode/DecodeForm.tsx
Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/decode/DecodeForm.tsx]: yeah makes sense, renamed to `isAbiError` and `isAbiFunction`. And also got rid of `AbiError` import and declared it locally, actually prefer that to the whole import, good suggestion

packages/explorer/src/app/(explorer)/api/system-abis/route.ts
```diff
@@ -0,0 +1,42 @@
...
+  }
+  validateChainId(chainId);
+
+  try {
+    const client = await getClient(chainId);
+    const indexerUrl = getIndexerUrl(chainId);
+    const abis = await getSystemAbis({
+      client,
+      worldAddress,
+      systemIds,
```
frolic, line 32: instead of asking the indexer for system IDs, then passing them in here, then fetching all system resource tags, then filtering based on passed system IDs
why don't we make this param optional and return all system ABIs that are found in the resource tags table?
that way we don't have to query this table twice in two different ways

frolic: alternatively, since these are internal APIs, we could return all ABIs in one response instead of two juggling separate API calls and hooks?

frolic (approved):

PR #3647, opened 2025-03-27
build: enable DTS map

frolic:


frolic:
don't know if this helps with our build output as that still runs through tsup

but if we switched over to using tsc for builds, we might be able to get rid of tsconfig.paths.json

PR #3649, opened 2025-04-01
feat(world): find systems based on inheritance

vdrg:
TODO: add tests


packages/world/ts/node/getSystemContracts.ts
```diff
@@ -22,10 +22,22 @@ export async function getSystemContracts({
     pattern: path.join(config.sourceDirectory, "**"),
   });
 
+  // Get systems from the top-level systems object
+  const topLevelSystems = Object.keys(config.systems);
+
+  // Get systems from namespaces
+  const namespaceSystems = Object.values(config.namespaces || {}).flatMap((namespace) =>
+    Object.keys(namespace.systems || {}),
```
frolic, line 30: ideally we map these to namespace.label rather than key to keep tooling consistent

frolic: sorry, the namespaces flatMap is the "standard" path for this kinda thing

packages/world/ts/node/getSystemContracts.ts
```diff
@@ -22,10 +22,16 @@ export async function getSystemContracts({
...
+  const configSystemLabels = Object.values(config.namespaces || {}).flatMap((namespace) =>
+    Object.values(namespace.systems).map((system) => system.label),
+  );
+
   return solidityFiles
     .filter(
       (file) =>
-        file.basename.endsWith("System") &&
+        // Include files with the System suffix and files defined in config
+        (file.basename.endsWith("System") || configSystemLabels.includes(file.basename)) &&
```
frolic, line 34: My only concern with this approach is if a system name provided by the MUD config conflicts with any other contract name (including codegen). I think I avoided filtering out codegen dirs here because how those dirs are determined is a bit complex (and somewhat dependent on the source dir structure).

Some other ideas on my mind:
- config option for `isSystemSource: () => bool` where we'd provide this behavior by default but it could be added to or overridden
- config option for `systemSuffix: RegExp` where the default is something like `/System$/` but could be overridden to something like `/(System|Program)$/`

In an ideal world, we'd be able to parse the Solidity source and check if anything inherits from the base `System` contract and use that as an indicator, but I recall running into chicken-and-egg problems when I moved source parsing earlier in the stack, where a system may depend on table libs or world/system interfaces that don't exist yet.

frolic: Curious to hear @alvrs's thoughts

alvrs: > In an ideal world, we'd be able to parse the Solidity source and check if anything inherits from the base System contract and use that as an indicator, but I recall running into chicken-and-egg problems when I moved source parsing earlier in the stack, where a system may depend on table libs or world/system interfaces that don't exist yet.

I think this should be possible with the solidity parser package, the file doesn't have to compile to be parsed (not 100% sure if base contracts can be extracted from what the parser returns but would expect so)

packages/cli/src/build.ts
```diff
@@ -16,7 +16,8 @@ type BuildOptions = {
 };
 
 export async function build({ rootDir, config, foundryProfile }: BuildOptions): Promise<void> {
-  await Promise.all([tablegen({ rootDir, config }), worldgen({ rootDir, config })]);
+  await tablegen({ rootDir, config });
+  await worldgen({ rootDir, config });
```
frolic, line 20: this has to be done serially now that we're parsing all `*.sol` source files in worldgen to determine if a source file is a system, otherwise we get a bunch of "file does not exist" errors from tablegen deleting/creating files

alvrs (approved):
Looks good! Should we add a config option to enable/disable this new behavior? Thinking about codebases that organized in some way that didn't expect all contracts that extend `System` to be automatically deployed and registered, or codebases that may have too many (irrelevant) files for which it would be a slowdown to have to parse all of them every time? (I think it's fine to enable this new behavior by default but would be nice to be able to opt out)

frolic:
Happy to put this behind a default-on config option.

> If the opt out flag would be set, we could just early return after if (contractName.endsWith("System") && contractName !== baseSystemName) return true;)

Note that this would still read+parse each file (the expensive bit). This is mostly to avoid deploying abstract contracts. I don't think the perf cost is huge here, so can leave this as is and the improve this if we find opted-out folks with perf issues.

frolic:
added an issue to follow up on this later: https://github.com/latticexyz/mud/issues/3665

PR #3650, opened 2025-04-02
feat(world-module-metadata): remove resource existence check

frolic:
if I wanted to bring my own resource ID type, this check prevents registering tags on arbitrary resources

an alternative to this is exposing a way to registering resource IDs to the `ResourceIds` table but not sure we want to muck with that internal-ish table or even require that resource IDs be instantiated onchain before they can be used

the namespace ownership check should be enough enforcement (because I can only create arbitrary resources, like tables and systems, if I own the namespace)


alvrs (dismissed):

frolic:
from @alvrs in discord:
> agree with allowing the namespace owner to set metadata for any resource including custom types

PR #3661, opened 2025-04-10
fix(world): fix static array arguments in system libraries

dk1a:
Currently static arrays break system libs, I really like them otherwise

The TODO is related to https://github.com/latticexyz/mud/pull/3660 (which is a separate PR because it affects more than system libs)

I use `stringToHex` to account for future possibilities like `uint[1 + 2 + THREE]` - escaping all those operands and whitespaces is troublesome otherwise


vdrg (approved):
nice! ty

vdrg:
@alvrs I think it still requires a review from you as I'm not a code owner

PR #3671, opened 2025-04-12
fix(world): support functions with missing argument names in system libraries

vdrg:
This is needed so that systems can have functions with arguments that don't have a name (useful if you are making the system inherit from an interface and are not using all args). In the current implementation you would get a compile error due to the encoding of the system calls.

### DIFF
--- a/packages/world/ts/node/render-solidity/renderSystemLibrary.ts
+++ b/packages/world/ts/node/render-solidity/renderSystemLibrary.ts
@@ -22,9 +22,11 @@ export function renderSystemLibrary(options: RenderSystemLibraryOptions) {
     storeImportPath,
   } = options;
 
-  // Remove `payable` from stateMutability for library functions
   const functions = functionsInput.map((func) => ({
     ...func,
+    // Format parameters (add auxiliary argument names, replace calldata location)
+    parameters: formatParams(func.parameters),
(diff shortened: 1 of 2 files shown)


packages/world/ts/node/render-solidity/renderSystemLibrary.ts
```diff
@@ -312,3 +314,16 @@ function renderReturnParameters(returnParameters: string[]) {
 
   return `returns (${renderArguments(returnParameters)})`;
 }
+
+function formatParams(params: string[]) {
+  // Use auxiliary argument names for arguments without names
```
frolic, line 319: ooc do they need names?

vdrg: yes, you need them to encode the system call. The other option was to use the default value for each type, imo that is not worth it (we would need to parse and figure out the underlying type of user types, define empty structs, etc)

vdrg: added more context in the PR description:
> This is needed so that systems can have functions with arguments that don't have a name (useful if you are making the system inherit from an interface and are not using all args). In the current implementation you would get a compile error due to the encoding of the system calls.

alvrs (approved):

PR #3672, opened 2025-04-15
fix(store-sync): handle pending deletion in stash storage adapter

alvrs:
If there is a pending deletion, `value` is undefined. If there is another `Store_SpliceStaticData` update, we previously passed in a partial record (consisting of only the key) to `encodeValueArgs`, which would throw an error when trying to encode the missing fields. Now we're handling this case by setting the `previousValue` to `undefined`, just like it would be if the pending deletion was already applied.

### DIFF
--- a/.changeset/hot-bananas-invite.md
+++ b/.changeset/hot-bananas-invite.md
@@ -1,0 +1,5 @@
+---
+"@latticexyz/store-sync": patch
+---
+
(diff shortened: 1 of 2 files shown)


frolic (approved):
is there a test to add to capture this regression?

alvrs:
> is there a test to add to capture this regression?

i don't think our current setup makes it easy to add a test for this specific case, added an issue to track: #3673

PR #3676, opened 2025-04-16
feat(store-indexer): add experimental/local SQL API endpoint

karooolis:
The eventual goal is to remove `/api/sqlite-indexer` from the Explorer, and move the Dozer-like API endpoint to local SQLite indexer instead.

CleanShot 2025-04-16 at 23 40 47@2x


packages/store-indexer/src/sqlite/apiRoutes.ts
```diff
@@ -44,5 +45,44 @@ export function apiRoutes(database: BaseSQLiteDatabase<"sync", any>): Middleware
     }
   });
 
+  router.post("/api/sqlite-indexer", async (ctx) => {
```
frolic, line 48: should we name this `/q` or whatever the dozer one is?

(this is the sqlite indexer so we don't need to name the route that too)

packages/store-indexer/src/bin/sqlite-indexer.ts
```diff
@@ -121,6 +122,7 @@ if (env.SENTRY_DSN) {
 }
 
 server.use(cors());
+server.use(bodyParser());
```
frolic, line 125: does dozer do it via POST?

packages/store-indexer/src/sqlite/apiRoutes.ts
frolic: also I wonder if we should put this behind some sort of flag so folks can opt-in because this is exposing the entire DB to any arbitrary SQL, and not sure the indexer API has done this before

packages/store-indexer/src/bin/sqlite-indexer.ts
karooolis: yep, it's a `POST`

packages/store-indexer/src/sqlite/apiRoutes.ts
karooolis: It would be convenient if it was `/q` although not sure if it's the best naming tbh, wdyt? Thinking of either `/q`, `/api/q` or `/api/data`.

As for arbitrary SQL, yeah, it's safe for development but if the endpoint is exposed in anyway, not ideal. What about if we limit it to only `SELECT` statements? And still can add a flag.

frolic: I'm not confident we can reliably prevent all kinds of attacks, so I'd rather this endpoint be an opt-in flag

if we're aiming to create something dozer-like I'd prob just follow dozer's naming for now (can't recall if it's `/q` or `/api/q`)

packages/store-indexer/src/sqlite/apiRoutes.ts
```diff
@@ -44,5 +45,51 @@ export function apiRoutes(database: BaseSQLiteDatabase<"sync", any>): Middleware
...
+      }
+
+      const result = [];
+      for (const { query } of queries) {
+        const normalizedQuery = query.trim().toLowerCase();
+        if (!normalizedQuery.startsWith("select")) {
+          ctx.status = 400;
+          ctx.body = JSON.stringify({ error: "Only SELECT queries are allowed" });
+          return;
+        }
```
frolic, line 64: I am guessing I can already break this with `select 1; drop table if exists "some_table"`

karooolis: yeah, I guess there is not much point to it. Looking at dozer as an example, it parses the query as AST, and then selectively checks parts of the tree to make sure it's a valid query but that's quite involved, and still not sure if it couldn't be gamed. I've put it under flag so it's opt-in at least which is perfect for dev environment.

packages/store-indexer/src/bin/sqlite-indexer.ts
```diff
@@ -28,6 +29,11 @@ const env = parseEnv(
     z.object({
       SQLITE_FILENAME: z.string().default("indexer.db"),
       SENTRY_DSN: z.string().optional(),
+      ENABLE_QUERY_ENDPOINT: z
```
frolic, line 32: I would give this a bit stronger name
```suggestion
      ENABLE_UNSAFE_QUERY_API: z
```

packages/store-indexer/src/sqlite/apiRoutes.ts
```diff
@@ -7,10 +9,9 @@ import { debug } from "../debug";
 import { createBenchmark } from "@latticexyz/common";
 import { compress } from "../koa-middleware/compress";
 import { getTablesWithRecords } from "./getTablesWithRecords";
-import { BaseSQLiteDatabase } from "drizzle-orm/sqlite-core";
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-export function apiRoutes(database: BaseSQLiteDatabase<"sync", any>): Middleware {
+export function apiRoutes(database: BaseSQLiteDatabase<"sync", any>, enableQueryEndpoint = false): Middleware {
```
frolic, line 14: small nit but would turn this into an object now that we have more than one thing
```suggestion
export function apiRoutes({ database, enableUnsafeQueryApi = false }: { database: BaseSQLiteDatabase<"sync", any>, enableUnsafeQueryApi?: boolean }): Middleware {
```

packages/store-indexer/src/sqlite/apiRoutes.ts
```diff
@@ -44,5 +45,50 @@ export function apiRoutes(database: BaseSQLiteDatabase<"sync", any>): Middleware
...
 
+  router.post("/q", async (ctx) => {
+    if (!enableQueryEndpoint) {
+      ctx.status = 404;
+      ctx.body = JSON.stringify({ error: "Query endpoint is not enabled" });
+      return;
+    }
+
+    try {
+      const queries = Array.isArray(ctx.request.body) ? ctx.request.body : [];
```
frolic, line 56: does dozer allow for more than one query at a time? are they executed in the same db tx?

packages/store-indexer/src/bin/sqlite-indexer.ts
karooolis: nice suggestion!

.changeset/yellow-otters-swim.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/store-indexer": patch
+---
+
+Added Dozer-like query API endpoint `/q` to the SQLite indexer.
```
alvrs, line 5:
```suggestion
Added experimental SQL API endpoint `/q` to the SQLite indexer. This is only intended for local development purposes and should not be used in production.
```

packages/store-indexer/src/bin/sqlite-indexer.ts
```diff
@@ -136,7 +143,7 @@ server.use(
   }),
 );
 server.use(helloWorld());
-server.use(apiRoutes(database));
+server.use(apiRoutes({ database, enableUnsafeQueryApi: env.ENABLE_UNSAFE_QUERY_API }));
```
alvrs, line 146: should we print a fat `WARNING: unsafe query API is enabled, DO NOT USE IN PRODUCTION as it will lead to public database access` or something similar to the console if this is enabled? i know some people are using the sqlite indexer in prod, don't want them to enable this because they're like "oh cool sql api is available for sqlite indexer now"

karooolis: yeah very valid point, added the warning message (attached screenshot in PR description)

alvrs: :chefkiss:

alvrs (approved):

PR #3678, opened 2025-04-17
chore(explorer): remove indexer endpoint

karooolis:
Remove `/api/sqlite-indexer` API endpoint, and use SQLite indexer's `/q` directly instead which is functionally the same. The added benefit is that the `/q` endpoint used by Explorer can be queried by outside applications.


alvrs (approved):

alvrs (dismissed):

alvrs (approved):

packages/explorer/src/app/(explorer)/api/sqlite-indexer/worlds/route.ts
```diff
@@ -1,24 +1,41 @@
...
+      },
+      body: JSON.stringify([
+        {
+          query: "SELECT DISTINCT address FROM __mudStoreTables",
+        },
+      ]),
+    });
+
+    if (!response.ok) {
+      throw new Error(`HTTP error! Status: ${response.status}`);
```
frolic, line 22: it would be great if these errors were a little more descriptive, like "Failed to fetch from indexer"

frolic: also any reason we have to do this fetch in an API route rather than directly from the page that needs to fetch worlds?

PR #3687, opened 2025-04-23
feat(store-indexer): format sql queries

karooolis:
Currently, local SQLite indexer stores column as kebab case. In order to conform to SQL API queries format, camelCase column names are converted to snake_case automatically.

### DIFF
--- a/.changeset/happy-melons-own.md
+++ b/.changeset/happy-melons-own.md
@@ -2,4 +2,4 @@
 "@latticexyz/store-indexer": patch
 ---
 
-The local SQLite indexer now automatically converts camelCase column names to kebab-case to support Dozer-style SQL queries.
+The local SQLite indexer now automatically converts camelCase column names to snake_case to comply with the SQL API.
--- a/.changeset/happy-melons-own.md
+++ b/.changeset/happy-melons-own.md
@@ -1,0 +1,5 @@
+---
+"@latticexyz/store-indexer": patch
+---
+
+The local SQLite indexer now automatically converts camelCase column names to kebab-case to support Dozer-style SQL queries.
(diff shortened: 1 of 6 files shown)


.changeset/happy-melons-own.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/store-indexer": patch
+---
+
+The local SQLite indexer now automatically converts camelCase column names to kebab-case to support Dozer-style SQL queries.
```
alvrs, line 5:
```suggestion
The local SQLite indexer now automatically converts camelCase column names to snake_case to comply with the API SQL.
```

alvrs (approved):

packages/store-indexer/src/sqlite/formatSqlQuery.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  if (node && typeof node === "object") {
+    const astNode = node as AstNode;
+
+    if (
+      astNode.type === "column_ref" &&
+      typeof astNode.column === "string" &&
+      /^[a-z]+[A-Z][a-z]*$/.test(astNode.column)
+    ) {
+      astNode.column = camelToSnakeCase(astNode.column);
```
frolic, line 37: thoughts on using `change-case` lib for this instead of our own implementation? we already use this in store-sync for sqlite table names

frolic: also don't think it's necessary to test if the column has something that looks like camel case, we can just change it all to snake case if that's the expected format

packages/store-indexer/src/sqlite/formatSqlQuery.ts
```diff
@@ -0,0 +1,50 @@
...
+  [key: string]: unknown;
+};
+
+/**
+ * Transforms camelCase identifiers to snake_case in SQL queries
+ *
+ * @param sqlQuery The SQL query to transform
+ * @returns The transformed SQL query
+ */
+export function formatSqlQuery(sqlQuery: string): string {
```
frolic, line 14: I feel like it might be clearer if we name this after what it does, like `normalizeColumnCase`

PR #3689, opened 2025-04-30
feat(entrykit): loosen min gas balance

frolic:
min gas balance wasn't a requirement for returning a complete session account but was required for getting through onboarding steps in entrykit

for now, we'll remove the min gas balance and just require anything >0

### DIFF
--- a/packages/entrykit/src/onboarding/GasBalance.tsx
+++ b/packages/entrykit/src/onboarding/GasBalance.tsx
@@ -55,7 +55,7 @@ export function GasBalance({ isActive, isExpanded, sessionAddress }: Props) {
         ) : relayChain != null ? (
           // TODO: convert this to a <ButtonLink>
           <a
-            href={`${relayChain.bridgeUrl}?${new URLSearchParams({ toAddress: sessionAddress })}`}
+            href={`${relayChain.bridgeUrl}?${new URLSearchParams({ toAddress: sessionAddress, amount: "0.01" })}`}
             target="_blank"
             rel="noopener noreferrer"
           >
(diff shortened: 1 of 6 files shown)


packages/entrykit/src/onboarding/GasBalance.tsx
```diff
@@ -46,7 +46,7 @@ export function GasBalance({ isActive, isExpanded, sessionAddress }: Props) {
             onClick={() =>
               setBalance.mutate({
                 address: sessionAddress,
-                value: minGasBalance + (balance.data?.value ?? 0n),
+                value: parseEther("0.01") + (balance.data?.value ?? 0n),
```
frolic, line 49: this is only for topping up locally

packages/entrykit/src/onboarding/quarry/Allowance.tsx
```diff
@@ -30,7 +29,7 @@ export function Allowance({ isActive, isExpanded, userAddress }: Props) {
         claimGasPass.status === "idle" &&
         allowance.isSuccess &&
         allowance.data != null &&
-        allowance.data < minGasBalance
+        allowance.data < parseEther("0.01")
```
frolic, line 32: this is only used to trigger an automatic claim of a gas pass in the quarry paymaster when balance dips below some amount

packages/entrykit/src/onboarding/GasBalance.tsx
```diff
@@ -55,7 +55,7 @@ export function GasBalance({ isActive, isExpanded, sessionAddress }: Props) {
         ) : relayChain != null ? (
           // TODO: convert this to a <ButtonLink>
           <a
-            href={`${relayChain.bridgeUrl}?${new URLSearchParams({ toAddress: sessionAddress })}`}
+            href={`${relayChain.bridgeUrl}?${new URLSearchParams({ toAddress: sessionAddress, amount: "0.01" })}`}
```
frolic, line 58: this is just a "suggested" amount to save some typing/clicks when trying to quickly top up via relay

karooolis (approved):

PR #3697, opened 2025-05-09
fix(entrykit): sign session wallet after top-up

karooolis:


alvrs (dismissed):

packages/entrykit/src/onboarding/Session.tsx
```diff
@@ -1,30 +1,52 @@
...
     // seems to stay pending. This is sorta resolved by triggering this after a timeout.
     const timer = setTimeout(() => {
-      if (isActive && setup.status === "idle" && sessionClient.data && !hasSession) {
+      if (
+        isActive &&
+        setup.status === "idle" &&
+        sessionClient.data &&
+        !hasSession &&
+        balance.data?.value != null &&
+        balance.data.value > 0n
```
alvrs, line 48: In the "gas tank" variant, the session account may not have a balance, because we first need to register it as a spender for the main account. If i understand this correctly it would only trigger `registerSpender` after the session wallet has an own balance?

karooolis: Exactly, once it has own balance, it attempts to `registerSpender()`. Without this condition, it tries to register it even before it has own balance which is what causes the "Enable" button pending state prematurely.

alvrs (approved):

PR #3703, opened 2025-05-13
feat(entrykit): add wiresaw transport

alvrs:


packages/common/src/transports/wiresaw.ts
```diff
@@ -0,0 +1,163 @@
...
+type WiresawSendUserOperationResult = {
+  txHash: Hex;
+  userOpHash: Hex;
+};
+
+type WiresawOptions<transport extends Transport> = {
+  wiresaw: transport;
+  fallbackBundler?: Transport;
+  fallbackEth?: Transport;
+};
```
frolic, line 14: just sketching out other name ideas

```ts
type WiresawOptions<transport extends Transport> = {
  /** wiresaw-compatible transport */
  transport: transport;
  /** fallback transport for bundler RPC methods not supported by wiresaw */
  bundlerTransport?: Transport;
  /** fallback transport for all other RPC methods not supported by wiresaw */
  fallbackTransport?: Transport;
};
```

```ts
type WiresawOptions<wiresawTransport extends Transport> = {
  /** wiresaw-compatible transport */
  wiresawTransport: wiresawTransport;
  /** fallback transport for bundler RPC methods not supported by wiresaw */
  bundlerTransport?: Transport;
  /** fallback transport for all other RPC methods not supported by wiresaw */
  defaultTransport?: Transport;
};
```

packages/entrykit/src/quarry/transports/userOpExecutor.ts
```diff
@@ -18,8 +18,14 @@ import { setBalance } from "viem/actions";
...
 // TODO: move to common package?
 
-export function userOpExecutor({ executor }: { executor: ConnectedClient }): Transport {
-  return () => {
+export function userOpExecutor({
+  executor,
+  fallbackEth,
+}: {
+  executor: ConnectedClient;
+  fallbackEth: Transport;
```
frolic, line 26: same thoughts as above re naming (`fallbackTransport` or `defaultTransport`)

packages/common/src/transports/wiresaw.ts
alvrs: how about:
```ts
type WiresawOptions<transport extends Transport> = {
  /** wiresaw-compatible transport */
  wiresawTransport: transport;
  /** fallback transport for bundler RPC methods */
  fallbackBundlerTransport?: Transport;
  /** fallback transport for all other RPC methods */
  fallbackDefaultTransport?: Transport;
};
```

i'd like to make it clear that the bundler and eth transports are just used for fallback, otherwise it may confuse someone into passing in the wiresaw url into all three since it technically supports all the methods.

packages/common/src/transports/wiresaw.ts
```diff
@@ -0,0 +1,163 @@
...
+            if (knownTransactionHash) {
+              const transactionReceipt = await getTransactionReceipt(knownTransactionHash);
+              if (transactionReceipt) {
+                return getUserOperationReceipt(userOpHash, transactionReceipt);
+              }
+            }
+            if (transport.fallbackBundler) {
+              const { request: fallbackRequest } = transport.fallbackBundler(opts);
+              return await fallbackRequest(req);
+            }
```
frolic, line 82: we don't ask wiresaw because it's not supported? should we throw here and let the `catch` fallback below handle this?

packages/common/src/transports/wiresaw.ts
frolic: what does "fallback" mean in this case? maybe that's worth clarifying here in comments 

in viem, fallback means "when transport fails, fall back the request to the next transport" but here we mean a few different things
- for chain ID, block number, block, and fees requests, we prioritize the default transport for request otherwise we use wiresaw
- only for user op receipt, use the bundler transport (again, not really a fallback)
- estimate user op gas _actually_ does a fallback, first wiresaw then bundler
- a catch with actual fallbacks failed requests, falling back to appropriate transport based on req method

I guess "fallback" captures the intent, but the cases where it isn't used as a fallback is a bit weird

packages/common/src/transports/wiresaw.ts
```diff
@@ -0,0 +1,168 @@
...
+  /** fallback transport for all other RPC methods */
+  fallbackDefaultTransport?: Transport;
+};
+
+export function wiresaw<const wiresawTransport extends Transport>({
+  wiresawTransport,
+  fallbackBundlerTransport,
+  fallbackDefaultTransport,
+}: WiresawOptions<wiresawTransport>): wiresawTransport {
+  return ((opts) => {
```
frolic, line 24: iirc we have access to the chain here, so couldn't we do all the transport handling here internally?

packages/common/src/transports/wiresaw.ts
alvrs: yeah this method is not supported on wiresaw, but we can still do it via `getTransactionReceipt`, but only for known transaction hashes (those where we previously sent the user operation through this transport). And yep, can let the catch handle it.

packages/common/src/transports/wiresaw.ts
frolic: wonder what happens if we throw in that case 🤣

frolic: feels fine if we say "can only use wiresaw transport with a client that defines a chain"

or take in a chain param so wagmi transports look like

```
[garnet.id]: wiresaw({ chain: garnet })
```

alvrs: updated to allow `wiresaw()` as default or overriding the transports explicitly (similar to how other transports allow an explicit url to be passed in)

packages/common/src/transports/wiresaw.ts
frolic: also internal, so can change later, but still wouldn't mind a rename like you proposed while we're here

.changeset/tender-kangaroos-build.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/entrykit": patch
+---
+
+Added experimental support for fast user operations on wiresaw-enabled chains.
```
frolic, line 5:
```suggestion
Added experimental support for fast user operations on Wiresaw-enabled chains.
```

packages/common/src/transports/wiresaw.ts
```diff
@@ -0,0 +1,188 @@
...
+              return receipt;
+            }
+          }
+        }
+      },
+    };
+  }) as wiresawTransport;
+}
+
+function getWiresawBaseTransport(chain: Chain): Transport | undefined {
```
frolic, line 157: not blocking, but we could probably generalize this to something like `chainTransport` given a chain's RPC URLs object like `chainTransport(chain.rpcUrls.wiresaw)`, since there are a bunch of areas we use this pattern where we want websocket + http if one or both exist

test/mock-game-contracts/foundry.toml
```diff
@@ -19,3 +19,12 @@ extra_output_files = [
...
 fs_permissions = [{ access = "read", path = "./"}]
+
+[profile.garnet]
+eth_rpc_url = "https://rpc.garnetchain.com"
+
+[profile.pyrope]
+eth_rpc_url = "https://rpc.pyropechain.com"
+
+[profile.redstone]
+eth_rpc_url = "https://rpc.redstonechain.com"
```
frolic, line 30: prob should normalize this across all our foundry.toml

frolic (approved):

PR #3714, opened 2025-05-26
fix(store-sync): handle pending deleted dynamic value in stash storage adapter

alvrs:


packages/store-sync/src/stash/createStorageAdapter.ts
```diff
@@ -66,10 +69,7 @@ export function createStorageAdapter({ stash }: CreateStorageAdapter): StorageAd
 
         updates.push((pendingRecords[id] = { table, key, value }));
       } else if (log.eventName === "Store_SpliceDynamicData") {
-        const previousValue = pendingRecords[id]
-          ? ({ ...pendingRecords[id].key, ...pendingRecords[id].value } as TableRecord)
-          : getRecord({ stash, table, key });
```
alvrs, line 71: we were missing the logic for checking that `pendingRecords[id].value` is defined in the `SpliceDynamicData` branch

PR #3718, opened 2025-05-28
feat(explorer): interact system calls

karooolis:
CleanShot 2025-05-28 at 17 20 52@2x


alvrs (dismissed):
very nice! not blocking, but we should do a performance pass on the filter function, ie on the DUST world there is a noticeable lag after each character i type and it sometimes even misses key inputs. Maybe there are some things that don't depend on the input field value, ie constructing the search key and other formatting, so we could do it once at the beginning and then only filter the pre-formatted array of functions without doing additional work on each key input?

PR #3720, opened 2025-05-28
fix(store-sync): ignore store logs with invalid keys

alvrs:


packages/store-sync/src/zustand/createStorageAdapter.ts
```diff
@@ -139,7 +138,21 @@ export function createStorageAdapter<tables extends Tables>({
...
+            const keySchemaLength = Object.keys(keySchema).length;
+            if (keySchemaLength !== keyTupleLength) {
+              debug(
+                `key tuple length ${keyTupleLength} does not match key schema length ${keySchemaLength}, skipping record`,
+                {
+                  table,
+                  rawRecord,
+                },
+              );
+              return undefined;
```
frolic, line 152: is returning undefined required or can we just return?

```suggestion
              return;
```

PR #3721, opened 2025-05-29
feat(entrykit): move to connectkit

frolic:
- moves to ConnectKit (RainbowKit had issues with Vite and `Buffer`)
- removes light mode variants for components in deposit flow (EntryKit doesn't have a light mode and these variants were artifacts of the former AccountKit)
- throws when trying to sign calls with ERC-6492 until we can add support (see #3722)
- bumps wagmi/viem everywhere


packages/entrykit/src/ui/Shadow.tsx
```diff
@@ -52,7 +52,9 @@ export const Shadow = forwardRef<HTMLIFrameElement, Props>(function Shadow({ mod
           inset: "0",
           width: "100%",
           height: "100%",
-          zIndex: "2147483646",
+          // one less than ConnectKit's modal z-index
+          // so that ConnectKit can overlap properly
+          zIndex: "2147483645",
```
alvrs, line 57: this is funny

alvrs (approved):

PR #3724, opened 2025-05-29
feat(entrykit): gas balance withdraw

karooolis:
https://github.com/user-attachments/assets/011823e6-d0f2-4ea9-a7d4-2d7c2978ba1c

### DIFF
--- a/packages/entrykit/src/onboarding/ConnectedSteps.tsx
+++ b/packages/entrykit/src/onboarding/ConnectedSteps.tsx
@@ -129,7 +129,7 @@ export function ConnectedSteps({ userClient, initialUserAddress }: Props) {
     <div
       className={twMerge(
         "px-8 flex flex-col",
-        !focusedId && "divide-y divide-neutral-800",
+        "divide-y divide-neutral-800",
         "animate-in animate-duration-300 fade-in slide-in-from-bottom-8",
       )}
     >
--- a/packages/entrykit/src/onboarding/ConnectedSteps.tsx
+++ b/packages/entrykit/src/onboarding/ConnectedSteps.tsx
@@ -129,7 +129,7 @@ export function ConnectedSteps({ userClient, initialUserAddress }: Props) {
     <div
       className={twMerge(
         "px-8 flex flex-col",
-        focusedId && "divide-y divide-neutral-800",
+        !focusedId && "divide-y divide-neutral-800",
         "animate-in animate-duration-300 fade-in slide-in-from-bottom-8",
       )}
     >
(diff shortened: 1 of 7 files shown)


packages/entrykit/src/onboarding/ConnectedSteps.tsx
```diff
@@ -129,7 +129,7 @@ export function ConnectedSteps({ userClient, initialUserAddress }: Props) {
     <div
       className={twMerge(
         "px-8 flex flex-col",
-        focusedId && "divide-y divide-neutral-800",
+        !focusedId && "divide-y divide-neutral-800",
```
frolic, line 132: intentional?

packages/entrykit/src/onboarding/deposit/SubmitButton.tsx
```diff
@@ -2,20 +2,28 @@ import { useAccount, useBalance, useSwitchChain } from "wagmi";
 import { Button, type Props as ButtonProps } from "../../ui/Button";
 import { twMerge } from "tailwind-merge";
 import { parseEther } from "viem";
+import { useEffect } from "react";
 
 export type Props = Omit<ButtonProps, "type"> & {
   amount?: bigint | undefined;
   chainId?: number | undefined;
+  onChainSwitched?: () => void;
```
frolic, line 10: looks like this isn't being used?

packages/entrykit/src/onboarding/quarry/GasBalance.tsx
```diff
@@ -10,16 +10,53 @@ import { StepContentProps } from "../common";
...
   const balance = useShowQueryError(useBalance(userAddress));
   const prevBalance = usePrevious(balance.data || 0n);
 
+  const { chainId: userChainId } = useAccount();
+  const shouldSwitchChain = chainId != null && chainId !== userChainId;
+  const { switchChain } = useSwitchChain();
+
+  const handleWithdraw = async () => {
+    if (!paymaster) throw new Error("Paymaster not found");
+    if (!balance.data) throw new Error("Balance is 0");
```
frolic, line 36:
```suggestion
    if (!balance.data) throw new Error("No balance to withdraw.");
```

packages/entrykit/src/onboarding/deposit/SubmitButton.tsx
karooolis: was still wip, removed

packages/entrykit/src/onboarding/ConnectedSteps.tsx
karooolis: There is now a generalized logic for setting a focused "screen", like in the case of "Deposit" form. So we want to show the divider when there is no screen focused i.e. we are seeing the initial "steps" page. But if we go inside a "Deposit" form page, we don't want to have the divider while keeping the rest of the styling the same.

frolic (dismissed):

packages/entrykit/src/onboarding/ConnectedSteps.tsx
karooolis: right, actually did this another way, and put the whole "Deposit" form under one container. It was trying to divide the top "Go back / Close button" part from the actual form part. So removed the conditional styling check.

PR #3735, opened 2025-06-11
feat(explorer): resolve paths using chain ID or name

karooolis:


packages/explorer/src/app/(explorer)/[chainName]/layout.tsx
```diff
@@ -11,8 +11,21 @@ type Props = {
...
-  if (!isValidChainName(chainName)) {
+export default function ChainLayout({ params: { chainName: chainIdOrName }, children }: Props) {
+  const pathname = usePathname();
+  const searchParams = useSearchParams();
+
+  if (!isValidChainName(chainIdOrName)) {
+    const chainId = Number(chainIdOrName);
+    if (isValidChainId(chainId)) {
+      const chainName = chainIdToName[chainId];
+      const newPath = pathname.replace(chainIdOrName, chainName);
```
Comment (frolic) [inline packages/explorer/src/app/(explorer)/[chainName]/layout.tsx:22]: I am worried this might replace too eagerly if the chain ID is also in the address, like `/690/worlds/0xff690ff...`

since we know the chain name/ID param is first, we could do something like

```ts
const newPath = pathname.replace(new RegExp(`^/${chainIdOrName}/`), `/${chainName}/`);
```

PR #3736, opened 2025-06-11
fix(explorer): crashing interact tab

karooolis:


packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/interact/content/useEnsAddress.ts
```diff
@@ -9,8 +9,8 @@ const mainnetClient = createPublicClient({
   transport: http(),
 });
 
-function isValidEnsName(name: string): boolean {
-  return name.includes(".");
+function isValidEnsName(name: string | undefined): name is string {
+  return !!name && name.includes(".");
 }
```
Comment (frolic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/interact/content/useEnsAddress.ts:14]: erm does this mean `useEnsAddress` should also accept `undefined`?

seems like there's an inaccurate type floating around that we haven't fully plumbed through

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/interact/content/useEnsAddress.ts
```diff
@@ -2,34 +2,42 @@ import { createPublicClient, http } from "viem";
...
 });
 
 function isValidEnsName(name: string): boolean {
-  return name.includes(".");
+  if (!name.includes(".")) {
+    return false;
+  }
+
+  const parts = name.split(".");
+  return parts.length >= 2 && !!parts[1] && parts[1].length > 0;
```
Comment (frolic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/interact/content/useEnsAddress.ts:18]: easier with regex imo
```ts
function isValidEnsName(name: string): boolean {
  return /\w\.\w/.test(name);
}
```

Comment (frolic) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/interact/content/useEnsAddress.ts:18]: I still think this should be inlined though, because this isn't actually telling you if it's a valid ENS, just that the string is possibly an ENS

Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/interact/content/useEnsAddress.ts:18]: oh amazing, yeah much better

PR #3737, opened 2025-06-12
fix(explorer): display remotely updated table values

karooolis:


packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/EditableTableCell.tsx
```diff
@@ -37,102 +39,143 @@ export function EditableTableCell({ name, table, keyTuple, value: defaultValue }
...
+          address: worldAddress as Hex,
+          functionName: "setField",
+          args: [table.tableId, keyTuple, fieldIndex, encodedFieldValue],
+          chainId,
+        } as const;
+
+        await simulateContract(wagmiConfig, {
+          ...writeArgs,
+          account: account.address,
+        });
```
Comment (frolic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/EditableTableCell.tsx:61]: why? this is done already by gas calc

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/EditableTableCell.tsx
```diff
@@ -37,102 +39,143 @@ export function EditableTableCell({ name, table, keyTuple, value: defaultValue }
...
+        ref={formRef}
+        onSubmit={(event) => {
+          event.preventDefault();
+
+          if (!account.isConnected) {
+            return openConnectModal?.();
+          }
+
+          if (fieldType === "bool") {
+            write.mutate({ value: write.status === "success" ? !write.variables.value : !value });
```
Comment (frolic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/EditableTableCell.tsx:119]: this logic makes no sense to me, why submit the opposite of the checked val?

Comment (frolic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/EditableTableCell.tsx:119]: we're ignoring the checkbox value and assuming we always want to toggle? feels wrong and like we're sidestepping what the DOM gives us

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/EditableTableCell.tsx
```diff
@@ -37,102 +39,143 @@ export function EditableTableCell({ name, table, keyTuple, value: defaultValue }
...
-      {!isPending && (
-        <form
-          onSubmit={(evt) => {
-            evt.preventDefault();
-            handleSubmit(value);
-          }}
-        >
+      <form
+        ref={formRef}
+        onSubmit={(event) => {
```
Comment (frolic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/EditableTableCell.tsx:111]: I know I said in my PR that an improvement would be to wrap this entire thing in a form but I feel like this has made it harder to reason about because we're not using the form as a form but rather just a way to trigger the mutation

what I meant was that we could use the form like a real HTML form with field values, then use those values in the submit handler to perform the write (like a real form would), then we get all the benefits of the DOM like accessibility, browser APIs, etc.

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/EditableTableCell.tsx
```diff
@@ -37,102 +39,143 @@ export function EditableTableCell({ name, table, keyTuple, value: defaultValue }
...
+        const receipt = await waitForTransactionReceipt(wagmiConfig, { hash: txHash });
+        if (receipt.status !== "success") {
+          throw new Error("Transaction reverted. Please try again.");
+        }
+
+        toast.success(
+          <a href={blockExplorerTransactionUrl({ hash: txHash, chainId })} target="_blank" rel="noopener noreferrer">
+            Transaction successful: {txHash} <ExternalLinkIcon className="inline-block h-3 w-3" />
+          </a>,
+        );
```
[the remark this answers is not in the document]
Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/EditableTableCell.tsx:73]: yeah, now showing a spinning loader instead which feels a little nicer overall

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/EditableTableCell.tsx
Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/EditableTableCell.tsx:119]: this covers a case where we have a non-live query where we update a value but do not refetch it. In those cases we haven't yet cleared the `write` which holds the latest value which we then toggle.

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TableColumn/BooleanField.tsx
```diff
@@ -0,0 +1,55 @@
...
+    />
+  );
+}
+
+export function BooleanField(props: Props) {
+  if (props.isReadOnly) {
+    return <ReadonlyBooleanField {...props} />;
+  }
+  return <EditableBooleanField {...props} />;
+}
```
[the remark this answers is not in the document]
Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TableColumn/BooleanField.tsx:55]: If a component is read-only, it could be so due to various reasons e.g. wallet not connected, cell being part of a key schema, etc. Thus, in read-only case would still need to sprinkle the input component with if statements since some props would become optional rather than required (thanks to tagged union). Not sure if it's a good explanation but overall having editable, and non-editable, components separated this way cuts a clear line between the two, and makes it easier to work with imo.

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TableColumn/useSetFieldMutation.tsx
```diff
@@ -0,0 +1,115 @@
...
+        if (reset) {
+          reset();
+        }
+      }
+    },
+  });
+
+  // When the indexer has picked up the successful write, we can clear the write result
+  useEffect(() => {
+    if (write.status === "success" && BigInt(blockHeight) >= write.data.receipt.blockNumber) {
```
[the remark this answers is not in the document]
Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TableColumn/useSetFieldMutation.tsx]: good point, created `useTrackPendingValue` hook that does that

packages/explorer/src/app/(explorer)/queries/useTableDataQuery.ts
```diff
@@ -1,32 +1,33 @@
...
 import { DozerResponse } from "../types";
 
 type Props = {
   table: Table | undefined;
-  query: string | undefined;
   isLiveQuery: boolean;
 };
 
-export type TDataRow = Record<string, unknown>;
+export type TDataRow = Record<string, string | boolean | undefined>;
```
frolic, line 15: what about numbers? I imagine dozer can return numbers too?

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TableColumn/useSetField.tsx
```diff
@@ -0,0 +1,78 @@
...
+        { id: context.toastId },
+      );
+    },
+    onError(error, variables, context) {
+      console.error("Error:", error);
+      toast.error(error.message || "Something went wrong. Please try again.", { id: context?.toastId });
+    },
+    onSettled() {
+      queryClient.invalidateQueries({ queryKey: ["balance"] });
+    },
```
Comment (frolic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TableColumn/useSetField.tsx:76]: FYI I moved back to hooks for side effects to keep the body of the mutation easier to understand, since it's now a highly reusable component.

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TableColumn/useSetField.tsx
```diff
@@ -0,0 +1,78 @@
...
+  fieldName: fieldName;
+};
+
+export function useSetField<
+  table extends Table,
+  fieldName extends keyof getValueSchema<table> & string,
+  value extends getSchemaPrimitives<getValueSchema<table>>[fieldName] = getSchemaPrimitives<
+    getValueSchema<table>
+  >[fieldName],
+>({ table, keyTuple, fieldName }: Props<table, fieldName>) {
```
Comment (frolic) [inline packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TableColumn/useSetField.tsx:32]: turns out all the type fun that I had didn't matter because we don't have specific types in this context (just a generic `Table` and `string` field name

but leaving this in because it might be useful to extract out into a place where we do have stronger input types

packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TableColumn/useSetField.tsx
Comment (karooolis) [reply packages/explorer/src/app/(explorer)/[chainName]/worlds/[worldAddress]/explore/TableColumn/useSetField.tsx:76]: got it, makes sense

PR #3739, opened 2025-06-15
fix: assert -> with

### LINKED ISSUE #3389: mud cli do not work with node 22

roninjin10:
fixes #3389


roninjin10:
Ah yea that is likely a good point. Just sharing what I know about this

- `assert` got completely deprecated (and it was never standardized) so for sure don't want to use it
- `with` has a specification that says the entire way the module is loaded is allowed to be changed from how it would be loaded without the `with`
- Your tsup config should be defaulting to compiling to node16 support but I am guessing that it doesn't apply to mdoules? I'm not sure if tsup has module specifications like tsconfig has

Might be worth just removing the import attributes completely until it gets officially standardized

PR #3743, opened 2025-06-17
feat(store-sync): make pending logs sync more resilient

alvrs:
- previously if there was an error in the pending logs stream the sync stack fell back to the latest block logs stream (from indexer or rpc), but it never tried to reconnect to the pending logs stream
- previously the sync stack wasn't detecting if the pending logs api was inactive and wasn't falling back to a latest block stream in that case


packages/store-sync/src/watchLogs.ts
```diff
@@ -30,56 +30,19 @@ export function watchLogs({ url, address, fromBlock }: WatchLogsInput): WatchLog
...
-      client.socket.addEventListener("close", async () => {
-        debug("socket close, setting up new client...");
-        clearTimeout(pingTimer);
-        setupClient().catch((error) => {
-          debug("error trying to setup new client", error);
-          subscriber.error(error);
-        });
+      client = await getWebSocketRpcClient(getUncachedUrl(url), {
+        keepAlive: true,
+        reconnect: { attempts: 100, delay: 1_000 },
```
frolic, line 45: is it intentional that attempts is really high? won't this mean retrying ~forever because of how the delay scales/backs off?

frolic:
I wonder if it's time to move store sync to effect or something. It's getting really hard to understand.

packages/store-sync/src/createStoreSync.ts
```diff
@@ -41,10 +44,33 @@ import { fromEventSource } from "./fromEventSource";
...
+ * If pending logs are available:
+ * - Initialize from snapshot
+ * - Open a pending log stream
+ *   - On error recreate the stream
+ * - Open a fallback log stream (indexer or RPC)
+ * - Catch up logs between snapshot and latest block
+ * - Cache processed log indices from pending logs stream
+ * - On every new block from the fallback logs stream
+ *   - Verify that all logs have already been processed in the pending logs stream
+ *   - If missing logs are found, release the missing logs to subscribers and recreate the pending logs stream
```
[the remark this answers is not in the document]
alvrs: Passing the logs that were missing to the subscribers. Realizing that that's not fully accurate though, we're passing the whole block to subscribers to ensure ensure the block order. We're recreating the pending logs stream in that case because we shouldn't be missing logs from it unless there is a problem with it, in which case we try to reconnect to fix it (if the problem persists we keep trying to recreate until wiresaw is available again).

alvrs, line 72:
```suggestion
 *   - If missing logs are found, pass all logs from the affected block to subscribers (to ensure ordering) and recreate the pending logs stream
```
```suggestion
 *   - If missing logs are found, release the missing logs to subscribers and recreate the pending logs stream
```

packages/store-sync/src/createPendingBlockStream.ts
```diff
@@ -0,0 +1,232 @@
...
+  pendingLogsUrl: string;
+  indexerUrl?: string;
+  chainId: number;
+  address?: Hex;
+  filters: SyncFilter[];
+  latestBlockNumber$: Observable<bigint>;
+  maxBlockRange?: bigint;
+};
+
+export function createPendingBlockStream(opts: PendingBlockStreamOptions): Observable<StorageAdapterBlock> {
```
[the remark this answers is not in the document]
alvrs: changing everywhere to preconfirmed

packages/store-sync/src/watchLogs.ts
alvrs: yeah that's intentional, the idea is that we're just using the latest block stream while this is disconnected or not returning valid logs (ie because wiresaw is down), but we want to keep retrying to automatically recover once it becomes available again

alvrs:
> I wonder if it's time to move store sync to effect or something

Let's do that in a followup PR 🙈

frolic (approved):

packages/store-sync/src/watchLogs.ts
```diff
@@ -189,3 +151,9 @@ async function fetchInitialLogs({
   const formattedLogs = rawInitialLogs.map((log) => formatLog(log));
   return { blockNumber: BigInt(latestBlockNumber), logs: parseEventLogs({ abi: storeEventsAbi, logs: formattedLogs }) };
 }
+
+function getUncachedUrl(url: string): string {
+  const uncachedUrl = new URL(url);
+  uncachedUrl.searchParams.append(`__${Date.now()}`, `${Date.now()}`);
```
frolic, line 157:
```suggestion
  uncachedUrl.searchParams.append(`__${Date.now()}`, "");
```

PR #3744, opened 2025-06-17
feat(store-sync): add support for upper bound in fetchRecords

alvrs:


alvrs:
why no changeset comment 😢

PR #3745, opened 2025-06-17
feat(store-sync): add flag to define chunking behavior during initial hydration

alvrs:


packages/store-sync/src/createStoreSync.ts
```diff
@@ -155,26 +156,31 @@ export async function createStoreSync({
...
-          latestBlockNumber: 0n,
-          lastBlockNumberProcessed: blockNumber,
-          message: "Hydrating from snapshot",
-        });
-
-        // RECS is a synchronous API so hydrating in a loop like this blocks downstream render cycles
-        // that would display the percentage climbing up to 100.
-        // We wait for idle callback here to give rendering a chance to complete.
-        await waitForIdle();
+      if (!enableHydrationChunking) {
```
frolic, line 159:
```suggestion
      if (enableHydrationChunking) {
```

frolic (dismissed):
need to flip the bool but otherwise lgtm

PR #3746, opened 2025-06-17
feat(explorer): textarea string fields

karooolis:


packages/explorer/src/components/ui/Textarea.tsx
```diff
@@ -0,0 +1,26 @@
...
+    return (
+      <textarea
+        className={cn(
+          "flex min-h-[80px] w-full",
+          "px-3 py-2",
+          "rounded-md border border-input ring-offset-background focus-visible:ring-2 focus-visible:ring-ring focus-visible:ring-offset-2",
+          "bg-background text-base",
+          "text-base md:text-sm",
+          "placeholder:text-muted-foreground disabled:cursor-not-allowed disabled:opacity-50",
+          "focus-visible:outline-none",
```
frolic, line 15: not sure if your tailwind version has this but would be nice to add `field-sizing: content` to allow this to expand as you type
https://developer.mozilla.org/en-US/docs/Web/CSS/field-sizing
https://tailwindcss.com/docs/field-sizing

frolic (dismissed):

packages/explorer/src/components/ui/Textarea.tsx
karooolis: existing version doesn't have it as it looks like it came out in v4 but will do once we upgrade to tailwind 4

PR #3752, opened 2025-06-20
fix(store-sync): backoff on failed connection, close websockets

alvrs:
* attempt to prevent dangling http connections we're seeing in wiresaw infra metrics
* wait before reconnecting if precious connection attempt failed

### DIFF
--- a/packages/store-sync/src/createPreconfirmedBlockStream.ts
+++ b/packages/store-sync/src/createPreconfirmedBlockStream.ts
@@ -14,6 +14,7 @@ import {
   merge,
   filter,
   startWith,
+  delay,
 } from "rxjs";
 import { StorageAdapterBlock, StoreEventsLog, SyncFilter } from "./common";
 import { watchLogs } from "./watchLogs";
@@ -56,7 +57,7 @@ export function createPreconfirmedBlockStream(opts: PreconfirmedBlockStreamOptio
     switchMap(() =>
       createLatestBlockStream({ ...opts, fromBlock: restartBlockNumber }).pipe(
         catchError((e) => {
-          debug("Error in latest block stream, recreating", e);
+          debug("Error in latest block stream", e);
           recreateLatestStream$.next();
           return throwError(() => e);
         }),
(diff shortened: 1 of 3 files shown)


packages/store-sync/src/watchLogs.ts
```diff
@@ -109,8 +108,9 @@ export function watchLogs({ url, address, fromBlock }: WatchLogsInput): WatchLog
       debug("logs$ subscription closed, closing client");
       try {
         client?.close();
+        client?.socket?.close();
```
alvrs, line 111: i assumed the socket should automatically be closed if the client is closed, but adding this to make sure, since we're seeing many open connections in the infra metrics that abruptly drop when the indexer restarts

packages/store-sync/src/watchLogs.ts
```diff
@@ -42,7 +42,6 @@ export function watchLogs({ url, address, fromBlock }: WatchLogsInput): WatchLog
 
       client = await getWebSocketRpcClient(getUncachedUrl(url), {
         keepAlive: true,
-        reconnect: { attempts: 100, delay: 1_000 },
```
alvrs, line 45: Removing the aggressive retries here and just using the default reconnect setting, since the sync logic already recreates the stream if something goes wrong.

packages/store-sync/src/createPreconfirmedBlockStream.ts
```diff
@@ -81,13 +85,15 @@ export function createPreconfirmedBlockStream(opts: PreconfirmedBlockStreamOptio
         catchError((e) => {
           debug("Error in preconfirmed logs stream, recreating", e);
           recreatePreconfirmedStream$.next();
-          return throwError(() => e);
+          return of(null);
```
alvrs, line 88: Throwing an error here in combination with the delay in recreating caused the whole `createPreconfirmedBlockStream` to be closed and to fall back to the regular sync in `createStoreSync`. It's also not necessary to throw the error here since the expectation is we're reconnecting the stream.

PR #3762, opened 2025-06-24
refactor(store-sync): migrate to isomorphic-ws

alvrs:
* Replaced `viem`'s websocket client with `isomorphic-ws` in `watchLogs` for more low level control
* Better error handling for the `watchLogs` stream
* Fixed time backoff when recreating `watchLogs` stream
* More robust logic to detect mismatching logs from preconfirmed/latest stream
* Better logging
* Fixed potential duplicate/unordered emission of logs when restarting the `watchLogs` stream

### DIFF
--- a/packages/store-sync/src/watchLogs.ts
+++ b/packages/store-sync/src/watchLogs.ts
@@ -199,11 +199,11 @@ async function request<T>({ ws, method, params, wsId }: RequestAsyncArgs): Promi
 
   const timeout = setTimeout(() => reject(new Error("timeout waiting for response for " + requestId)), 10_000);
 
-  const listener = (message: WebSocket.MessageEvent): void => {
+  ws.addEventListener("message", function onMessage(message) {
     const data = JSON.parse(message.data.toString());
     if (data.id === requestId) {
       debug(`ws${wsId} request response for`, method, requestId);
-      ws.removeEventListener("message", listener);
+      ws.removeEventListener("message", onMessage);
       clearTimeout(timeout);
 
       if ("error" in data) {
@@ -214,8 +214,7 @@ async function request<T>({ ws, method, params, wsId }: RequestAsyncArgs): Promi
 
       resolve(data.result as T);
     }
-  };
-  ws.addEventListener("message", listener);
+  });
   return promise;
 }
 


packages/store-sync/src/createPreconfirmedBlockStream.ts
```diff
@@ -90,36 +99,85 @@ export function createPreconfirmedBlockStream(opts: PreconfirmedBlockStreamOptio
...
+
+      const mismatchingTransactions: string[] = [];
+      if (preconfirmedLogsState === "initialized") {
+        const logsByTransaction = block.logs.reduce<Record<string, Partial<StoreEventsLog>[]>>((acc, log) => {
+          const txHash = log.transactionHash;
+          if (txHash == null) return acc;
+          acc[txHash] ??= [];
+          acc[txHash].push(log);
+          return acc;
+        }, {});
```
frolic, line 144: could also do
```ts
const transactionLogs = groupBy(block.logs.filter((log) => log.transactionHash), (log) => log.transactionHash)
```
(`groupBy` from common package)

packages/store-sync/src/createPreconfirmedBlockStream.ts
```diff
@@ -43,19 +45,23 @@ export function createPreconfirmedBlockStream(opts: PreconfirmedBlockStreamOptio
   const recreatePreconfirmedStream$ = new Subject<void>();
   const recreateLatestStream$ = new Subject<void>();
 
-  let restartBlockNumber = opts.fromBlock;
+  let processedLatestBlockNumber = opts.fromBlock - 1n;
```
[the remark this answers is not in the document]
alvrs: We have to do `-1` or `+1` in some places - the `fromBlock` param we get is the first block we want to include, we then update the variable whenever we processed a `latestBlock` to know what's the latest block we've confirmed is correct, so we want to start from one after that when creating the `preconfirmedStream` and also want to skip any blocks smaller or equal to this in future updates to subscribers.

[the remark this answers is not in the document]
alvrs: we could change it to "first unconfirmed block" but i feel like that makes it less clear what it actually is

packages/store-sync/src/watchLogs.ts
```diff
@@ -23,137 +26,227 @@ type WatchLogsEvent = {
...
+
+  debug(`ws${wsId} sending request`, method, requestId);
+  ws.send(JSON.stringify({ jsonrpc: "2.0", id: requestId, method, params }), (error) => {
+    if (error) {
+      debugError(`ws${wsId} request error`, error);
+      reject(error);
+    }
+  });
+
+  const timeout = setTimeout(() => reject(new Error("timeout waiting for response for " + requestId)), 10_000);
```
frolic, line 200: should this remove the message listener?

packages/store-sync/src/watchLogs.ts
```diff
@@ -23,137 +26,227 @@ type WatchLogsEvent = {
...
+    );
+
     return () => {
-      debug("logs$ subscription closed, closing client");
+      debug(`ws${wsId} logs$ subscription closed, closing client`);
+      clearInterval(ping);
       try {
-        client?.close();
-        client?.socket?.close();
+        ws?.close();
```
frolic, line 157: doesn't look like there's anything here to auto-close after an error? intentional?

packages/store-sync/src/watchLogs.ts
```diff
@@ -23,137 +26,227 @@ type WatchLogsEvent = {
...
+      if ("error" in data) {
+        debugError(`ws${wsId} request json-rpc error`, data.error);
+        reject(new Error(data.error.message));
+        return;
+      }
+
+      resolve(data.result as T);
+    }
+  };
+  ws.addEventListener("message", listener);
```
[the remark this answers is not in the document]
alvrs: that's cool! updated

packages/store-sync/src/watchLogs.ts
alvrs: this will cause the connection to be closed and recreated so don't necessarily have to remove the msg listener on the old ws

packages/store-sync/src/watchLogs.ts
alvrs: yeah the idea is errors are passed to the subscriber to handle it, and stopping the subscription closes the websocket

frolic (approved):

PR #3765, opened 2025-06-25
fix(store-sync): setup message listener before setting up subscription

alvrs:


.changeset/sharp-shrimps-dream.md
```diff
@@ -0,0 +1,5 @@
+---
+"@latticexyz/store-sync": patch
+---
+
+Fixed a race condition in the preconfirmed logs stream by setting up the message listener before setting up the subscription.
```
frolic, line 5: we haven't released so this changeset isn't strictly necessary

PR #3770, opened 2025-06-26
feat(common): add paymaster for testnet chains, remove rhodolite

frolic:


packages/common/src/chains/pyrope.ts
```diff
@@ -27,6 +27,9 @@ export const pyrope = {
         address: "0xC24932c31D9621aE9e792576152B7ef010cFC2F8",
       },
     },
+    quarryPaymaster: {
+      address: "0xD40C9cFc97b855B2183D7e1c8925edF77C309b85",
```
frolic, line 31: https://github.com/latticexyz/quarry-paymaster/blob/40f99890bbda36a997c1b8435fae4c9d18da794c/packages/contracts/worlds.json#L18

PR #3776, opened 2025-06-27
feat(store-sync): include registered system functions in ABI

frolic:
fixes explorer not showing metadata system functions


frolic:
this is an internal export and I wrote tests, so gonna land

PR #3778, opened 2025-06-27
chore(explorer): remove URL copy on function click

karooolis:


frolic (approved):
why not just delete the code?

karooolis:
> why not just delete the code?

Fair enough, given copying is not used/required anywhere atm removed entirely

frolic (approved):

PR #3784, opened 2025-07-10
feat(entrykit): non-blocking fee estimation

alvrs:


packages/entrykit/src/createBundlerClient.ts
```diff
@@ -59,7 +58,6 @@ function createFeeEstimator(client: Client): undefined | (() => Promise<Estimate
   // because viem sets fees way too high by default
   // https://github.com/wevm/viem/blob/253b1072ad9fe36a0e0491e173c85a6d69209ada/src/account-abstraction/actions/bundler/prepareUserOperation.ts#L436-L457
   if ([690, 17069, 695569].includes(client.chain.id)) {
-    // TODO: move to fee ref or similar approach
-    return () => getAction(client, estimateFeesPerGas, "estimateFeesPerGas")({ chain: client.chain });
+    return cachedFeesPerGas(client);
```
alvrs, line 61: viem prioritizes this method over the chain's `estimateFeesPerGas` config: https://github.com/wevm/viem/blob/3aa882692d2c4af3f5e9cc152099e07cde28e551/src/account-abstraction/actions/bundler/prepareUserOperation.ts#L447. If we override only the `estimateFeesPerGas` config, we still have one unnecessary call for `getBlock`: https://github.com/wevm/viem/blob/3aa882692d2c4af3f5e9cc152099e07cde28e551/src/actions/public/estimateFeesPerGas.ts#L132

packages/entrykit/src/utils/withFeeCache.ts
alvrs, line 1: we don't strictly need this file anymore (for our use case the fee estimation happens via `userOperation.bundlerClient.estimateFeesPerGas`), could remove it from this PR

packages/entrykit/src/utils/withFeeCache.ts
```diff
@@ -0,0 +1,25 @@
+import { Chain, createClient, http } from "viem";
+import { cachedFeesPerGas } from "../actions/cachedFeesPerGas";
+
+type WithFeeCacheOptions = {
+  refreshInterval?: number;
+};
+
+export function withFeeCache(chain: Chain, options: WithFeeCacheOptions = { refreshInterval: 10_000 }): Chain {
```
frolic, line 8:
```suggestion
export function withFeeCache<const chain extends Chain>(chain: chain, options: WithFeeCacheOptions = { refreshInterval: 10_000 }): chain {
```

just to keep strong types

frolic (approved):

PR #3794, opened 2025-07-29
feat: add custom paymaster client override

dk1a:
Currently it's impossible to use entrykit with a custom paymaster client

This override is a simple solution, it may be better to move all the configuration there from `Chain` and making it just `paymaster`, rather than `paymasterOverride`. Related to the TODO for configuring `createBundlerClient` in general.

Since it's probably not relevant for quarry stuff, I ignore the override there

### DIFF
--- a/packages/entrykit/src/config/input.ts
+++ b/packages/entrykit/src/config/input.ts
@@ -1,3 +1,4 @@
+import { BundlerClientConfig } from "viem/account-abstraction";
 import { Address } from "viem/accounts";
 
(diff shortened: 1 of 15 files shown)


frolic:
can you share more about your use case? what do you need a custom client for that the chain config options doesn't help with?

dk1a:
The main reason I need to fork entrykit is that it does `registerDelegation`, which I also want to be aware of the custom paymaster, otherwise I could've probably just added it to the session client within the app itself

PR #3813, opened 2025-09-09
add BatchStoreModule

frolic:


packages/world-module-batchstore/src/BatchStoreSystem.sol
```diff
@@ -0,0 +1,60 @@
...
+        staticData: staticData,
+        encodedLengths: encodedLengths,
+        dynamicData: dynamicData
+      });
+    }
+  }
+
+  function setTableRecords(ResourceId tableId, TableRecord[] memory records) external {
+    AccessControl._requireAccess(tableId, _msgSender());
+
```
frolic, line 39: need to get field layout here to save gas

packages/world-module-batchstore/src/BatchStoreSystem.sol
```diff
@@ -0,0 +1,60 @@
...
+        records[i].keyTuple,
+        records[i].staticData,
+        records[i].encodedLengths,
+        records[i].dynamicData
+      );
+    }
+  }
+
+  function deleteTableRecords(ResourceId tableId, bytes32[][] memory keyTuples) external {
+    AccessControl._requireAccess(tableId, _msgSender());
```
frolic, line 52: should these be `requireAccess` rather than `_requireAccess`?

PR #3816, opened 2025-09-26
feat(cli): save deployed contracts info

dhvanipa:


packages/cli/src/runDeploy.ts
```diff
@@ -219,9 +220,9 @@ export async function runDeploy(opts: DeployOptions): Promise<WorldDeploy> {
     );
   }
 
-  console.log(deploymentInfo);
+  console.log({ ...deploymentInfo, contracts: undefined });
```
alvrs, line 223: Nit: i assume this will also print `contracts: undefined`, could we instead explicitly print `{ worldAddress, blockNumber }` ?

alvrs (dismissed):

alvrs (approved):

